PR #19, opened 2017-04-10
Tooling improvements

zakx:
I have started to improve the basic tooling by replacing homegrown helper files with commonly used patterns. There's also some preparation for the introduction of a proper setup.py, which will need to happen anyway sometime in the feature to enable users to install InvenTree via package managers like pip.


zakx:
Note that this will of course not yet result in a passing build; I created #20 to solve that.

SchrodingersGat:
@zakx so this should be merged before #20 ?

What advantage does flake8 offer over the pep8 utility I was using? I am fine to change it, just wondering what the rationale is.

zakx:
Yeah, this should be merged first.

flakes8 ist basically a merger of pep8 (style checking only) and pyflakes (error checking only). See https://dancallahan.info/journal/python-flake8/ for more information and examples.

In #20 I have also removed (not fixed) two errors that a simple style check would not have found.

SchrodingersGat:
Ok, I am happy with this, it looks a lot more robust and forward-thinking. Thanks :)

PR #20, opened 2017-04-10
flake8 fixes

zakx:
I've removed the whitespace exception and cleaned up all occurrences.
I also "fixed" other errors in favor of a clean flake8 run. Please have a look.

This is based on #19.

### DIFF
--- a/InvenTree/InvenTree/models.py
+++ b/InvenTree/InvenTree/models.py
@@ -8,10 +8,10 @@ from django.contrib.contenttypes.models import ContentType
 class Company(models.Model):
     """ Abstract model representing an external company
     """
-    
+
     class Meta:
         abstract = True
-        
+
     name = models.CharField(max_length=100)
     URL = models.URLField(blank=True)
     address = models.CharField(max_length=200,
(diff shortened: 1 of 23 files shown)


InvenTree/InvenTree/models.py
```diff
@@ -73,19 +72,16 @@ def getUniqueChildren(self, unique=None):
 
         unique.add(self.id)
 
+        # commenting this out as it throws errors.
```
SchrodingersGat: What errors are you getting here? This code is required to ensure that a category cannot be set as a parent or child of itself.

zakx: Well, for flake8 in the current setting, unused code segments are considered an error. As you did not make use of `children` yet, it is considered unused. When I comment it out, `contents` will become unused too. You are very welcome to re-include the code, but it would be preferable if you'd then make use of `children` so it isn't just dead code.

The actual `make style` output was:
`./InvenTree/InvenTree/models.py:77:9: F841 local variable 'children' is assigned to but never used`

SchrodingersGat: Ah, I see the problem. I have added the code that was missing

SchrodingersGat:
Closing as these are addressed in #23

PR #53, opened 2018-04-16
Stock track

SchrodingersGat:
Moved part tracking information into the StockItem model

- Does not make sense to keep double records
- StockItems can be in a location, or "in" another stock item (e.g. serialized part within another serialized part)


SchrodingersGat:
Thanks!

Issue #127, opened 2019-04-16
Code cleanup


SchrodingersGat:
There are a lot of modules used for experimentation that are no longer used.

Also dead code / templates / views / etc.

Needs a good ol' pruning

SchrodingersGat:
Fixed some dead scripts in https://github.com/inventree/InvenTree/pull/130

SchrodingersGat:
Ongoing, but closed for now

PR #137, opened 2019-04-18
Delete old forms

SchrodingersGat:
Remove a lot of dead templates / views. They were leftover from initial attempts and are no longer valid.

Everything is done with modal forms now.


SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/127

Issue #122, opened 2019-04-14
Mark parts as 'inactive'


SchrodingersGat:
Instead of deleting parts, they should be retained but marked as 'inactive'

This will prevent them from showing up in searches, etc.

But it means that part tracking history can be retained.

PR #185, opened 2019-04-28
Inactive part

SchrodingersGat:
Add ability to mark part as 'inactive'

- Prevent part deletion (can still delete from Admin interface)
- Hide inactive parts from main search
- Display **INACTIVE** warning on Part page
- Use JSON API to toggle part active status


SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/122

Issue #88, opened 2018-05-04
"Did you mean"?


SchrodingersGat:
Prevent user from making a new part which is really the same as an old part

- Look for 'similar named' parts
- Suggest these parts before part creation
- Part of Modal Wizard Form?
  * Step 1 - Create part
  * Step 2 - Check for dupes
  * Step 3 - ???
  * Step 4 - PART!

PR #309, opened 2019-05-11
Fuzzy search

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/88

Fuzzy-search lookup on part names when creating a new part. If there are any parts with a significantly close match, they are displayed and part creation is prevented unless the user selects a confirmation checkbox.

Note: The search algorithm isn't *fantastic* yet, could certainly do with some work.


SchrodingersGat:
image

PR #368, opened 2019-05-24
BOM Upload

SchrodingersGat:
First pass at uploading BOM file is working

- Upload a file to the server and load into memory
- Parse the file with tablib (verify it is a valid tabulated data file)


SchrodingersGat:
@jnewlands here's how you can upload a file to the server using the existing modal forms interface. The file is memory in the server and can be parsed using tablib...

Issue #373, opened 2019-05-27
Collapse (many) duplicate stock items in the same location for readability


SchrodingersGat:
Most likely a front end only implementation

"_2300 total pcs in 3 stock items_"

SchrodingersGat:
This allows nice display of multiple single-quantity stock items (e.g. individually serialized parts)

SchrodingersGat:
Only group stock items together if the other parameters are equal.

e.g.

* Status
* Packaging types

Issue #354, opened 2019-05-20
Further speed improvements


SchrodingersGat:
References:

https://docs.djangoproject.com/en/2.2/ref/models/querysets/#prefetch-related

(Read the entry about __str__ of pizza objects.

https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.prefetch_related_objects

Useful if you already have a list of objects

PR #375, opened 2019-05-28
Group stock items

SchrodingersGat:
Display stock items as "groups" relating to the same part

- This allows us to show total stock for many singleton items
- Items which have serial numbers *must* be singletons

Fixes https://github.com/inventree/InvenTree/issues/373
Fixes https://github.com/inventree/InvenTree/issues/354


SchrodingersGat:
Was actually a pretty hard problem to solve. Ended up doing all the grouping / sorting at the front end, but some changes were needed to the bootstrap-table code base.

PR #414, opened 2019-07-02
doc: python3 instead of python

diegoherranz:
Otherwise, python2 would be used in most systems.

Thanks!


SchrodingersGat:
Good call. I've been trying to merge everything over to Python3 :)

PR #430, opened 2019-08-02
Makefile: add .PHONY targets

diegoherranz:
Given all the targets are actions instead of files to be generated,
.PHONY should be used, mainly to avoid a conflict with a file of the same name.

See https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html


SchrodingersGat:
Thanks @diegoherranz I wasn't aware of this.

Ideally I would like to move away from reliance on a makefile - so that python is the only requirement... Something I am looking into.

PR #447, opened 2019-08-28
Feature/add method to delete token

rrakso:
I've added an DELETE method, to remove token from DB (I'm using this in my app)


SchrodingersGat:
Hey @rrakso thanks for adding this, looks good :)

PR #484, opened 2019-09-07
Install package for fast string matching

SchrodingersGat:
- But really, mostly to supress a warning!


SchrodingersGat:
Also made some other changes I guess...

PR #526, opened 2019-09-14
Fixes problem with 'New Purchase Order' button from Suppliers page.

linucks:
This fixes an issue where the 'New Purchase Order' button doesn't work from the 'Purchase Order' tab of a Supplier page.

It seems to be an issue with the id of the button and the second hyphen as renaming the button id from 'company-order-2' to 'company-order2' resolves the issue.

The issue has been confirmed on Mac OSX 10.14.6 with Brave Version 0.67.125 Chromium: 76.0.3809.100 and Safari Version 12.1.2 (14607.3.9)

### DIFF
--- a/InvenTree/company/templates/company/detail_purchase_orders.html
+++ b/InvenTree/company/templates/company/detail_purchase_orders.html
@@ -9,7 +9,7 @@
 
 <div id='button-bar'>
     <div class='btn-group'>
-        <button class='btn btn-primary' type='button' id='company-order-2' title='Create new purchase order'>New Purchase Order</button>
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
Thanks @linucks good spot. Not sure what is going on there but your change seems to have fixed it :)

linucks:
It is a bit weird as no error message is generated and it's not the length (I tried with a string of the same length),  so it's either the actual name or the additional hyphen that causes the problem. Go figure :)

Thanks for all your work on InvenTree - I'm find it really useful.

SchrodingersGat:
Yeah I'm putting it down to a weird quirk of jQuery, or something... Into the bucket of mysteries it goes.

Really glad you're finding it useful - please keep contributing back any fixes or improvements as you come across them :)

PR #577, opened 2019-11-09
Updated pip call for compatibility.

ionisingTuna:


SchrodingersGat:
Hey @ionisingTuna this change is breaking the CI tests - https://travis-ci.org/inventree/InvenTree/jobs/609633528

I am not familiar with calling pip in this way. What are the requirements for running the `py` command?

ionisingTuna:
Hi @SchrodingersGat , The py command is a python launcher for windows which helps to determine the correct version of python to run. It was introduced in python 3.3 to introduce functionality already present on Unix.

Currently on Windows, a strategy of "last installed Python wins" has been used and while not ideal, has generally been workable due to the conservative changes in Python 2.x releases. As Python 3.x scripts are often syntactically incompatible with Python 2.x scripts, a different strategy must be used to allow files with a '.py' extension to use a different executable based on the Python version the script targets. This will be done by borrowing the existing practices of another operating system - scripts will be able to nominate the version of Python they need by way of a "shebang" line.

The py command provides an implementation of a launcher which can offer the same benefits for Python on the Windows platform and therefore allows the launcher to be the executable associated with '.py' files to support multiple Python versions concurrently.

Details for the above can be found at the following links:
https://www.python.org/dev/peps/pep-0397/

py can be invoked to select a desired python version.
such as py -2 program.py, py -2.7 program.py, py -3.5 program.py, py -3.6 program.py.
If you want to get the REPL, just use py or py -version_specifier.


I understand that Unix style OS do not have this launcher- So perhaps there should be a check for the host operating system & if python has been installed past 3.3, to determine whether to use the launcher or not

Thanks for checking this out - Is this something that could be useful?

SchrodingersGat:
If the tool is unique to windows then this PR cannot be merged in - the default OS is assumed to be Linux (but supports windows).

Does the default call of `pip3` not call the correct python install on your system?

ionisingTuna:
Hi @SchrodingersGat , this python launcher is indeed unique to Windows. ``pip3`` will work for both Windows and Linux. If cross-platform compatibility is an aim without additional changes, I would stay with the current project approach. Thanks for checking out this pull request.

Issue #596, opened 2019-12-20
Allow public access


Zeigren:
I'm looking at allowing visitors to view my InvenTree instance, I don't currently see a way to do this but I could be missing something.

Any ideas or preferences on how I might go about implementing this feature?

SchrodingersGat:
Django provides for "user permissions" which can apply to individual users or entire groups. These can provide fine-grained control of what users can access / read / write / delete (etc).

I think this would be a good way to approach the problem, and create a "guest" account which can only view certain parts of the data.

However this functionality needs to be used to actually restrict access to the particular data. I would also like to see it coded in way that is not particular to a single InvenTree instance (e.g. at the very least it covers InvenTree in a generic way, or it can be extended to meet particular needs).

Docs:

- https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Authentication
- https://docs.djangoproject.com/en/3.0/topics/auth/

Zeigren:
Oh I never shared my thoughts on the actual implementation...

My idea is to focus on group/role based permissions although user based permissions would of course still work. I plan on sticking with the default Django permissions as I think they should be good enough (view, change, add, delete).

You'd create a group in the admin panel and set whatever permissions you would like that group to have, then add users to that group. These would be the broad restrictions that would apply to the API and top level UI elements, based on the model. These can be implemented with just Django.

In order to use object based permissions we need a per object authentication backend like django-guardian, since that isn't natively supported in Django. On creation (or while being edited) each object should have a form that can be used to pick which groups have access to that object. What object permissions they have should be based on that groups model permissions, to keep it simple.

You can use the admin panel to set user permissions per object or to fine tune the group permissions per object, but I think that's a bit much to add to the front end UI.

I wound up with this approach since it was easier to implement an anonymous user and would add more to the project.

SchrodingersGat:
Hi @Zeigren check out the implementation in https://github.com/inventree/InvenTree/pull/1018

Does this meet your requirements here?

eeintech:
@SchrodingersGat I think we should close this ticket, the role and permission system definitely allows for visitor/public/view-only profile.

SchrodingersGat:
Agreed, thanks for bringing this to my attention @eeintech

PR #603, opened 2020-01-04
Permissions and Public Access

Zeigren:
In order to allow anonymous viewing (see #596), I've been implementing a permissions system using [django-guardian](https://github.com/django-guardian/django-guardian), which handles per object permissions and has a built-in AnonymousUser (that can be assigned permissions).

It's a WIP but here's what I have so far and my plans.

Done:
- Various fixes for AnonymousUser
- Removed Middleware that checked for authenticated users (no longer needed)
- settings.py
- Permissions checking in views using a [PermissionRequiredMixin](https://django-guardian.readthedocs.io/en/stable/api/guardian.mixins.html), if the user doesn't have the correct permissions for the object they're redirected to the login page

In Progress:
- Improve the user experience by using [template tags](https://django-guardian.readthedocs.io/en/stable/api/guardian.templatetags.guardian_tags.html) to hide any UI elements the user doesn't have permission to use
- Developer Documentation

Planned:
- Add UI elements for choosing what permissions to assign to an object
- Improve the object permission admin interface
- Use [django-rest-framework-guardian](https://github.com/rpkilby/django-rest-framework-guardian) for improved REST integration.  It has a filter so only objects the user has access to are returned and a mixin for assigning permissions
- Modify current tests to work with permissions and create new ones for specifically testing various permissions
- User Documentation


SchrodingersGat:
Interesting approach, I hadn't heard of these plugins before. I do not have enough time to look at this right now but I would like to soon.

SchrodingersGat:
@Zeigren are you able to fix the conflicts with master here so I can have a look at what you've done?

Cheers

Zeigren:
Yep can do!

Zeigren:
It's been a minute since I've worked on it but I want to say object permissions can only be assigned in the admin panel currently. I think permission checks work for most of the UI and in views. It's worth checking out [DjangoObjectPermissions](https://www.django-rest-framework.org/api-guide/permissions/#djangoobjectpermissions) for the rest framework as well.

Zeigren:
Still lots to do but I've made some minor changes/fixes. Most of the tables should actually be filtered now.

I've been trying to figure out the "best" way to actually assign the permissions since there's a number of ways to do it and I'd like to keep it simple. One way would be to handle it in the view or it could be done inside a [signal](https://django-guardian.readthedocs.io/en/stable/userguide/assign.html#assigning-permissions-inside-signals).

To Do:
- Add/Fix permissions checks in views/templates for new UI elements and where object permissions make sense
- Add/Fix django-rest-framework-guardian filters and add any querysets needed for them to work properly
- Don't redirect users to the login page if they don't have the correct permissions
- Use [direct foreign keys](https://django-guardian.readthedocs.io/en/stable/userguide/performance.html#direct-foreign-keys) for performance and to avoid orphaned permissions
- Add UI elements for choosing what permissions to assign to an object
- Improve the object permission admin interface
- Modify current tests to work with permissions and create new ones for specifically testing various permissions

SchrodingersGat:
Hey @Zeigren I'm trying to assess what is going on here - but as i haven't used the "permissions" system in django before I'm a little lost. I've pulled your branch, but the default (non-staff) account can still perform every available action. I had assumed that most actions would be prevented by default? Is there something I am missing here? Cheers

Zeigren:
That would make sense, the per object permissions weren't being assigned so it should have defaulted to checking the regular Django model permissions.

This commit adds in a very simple automatic object permissions assignment that should be a better representation. I've added a bunch of post_save signals to `part>models.py` that will automatically assign viewing permissions to the group `view` for any objects created under those models. You'll need to create a group called `view` first and give it viewing permissions in the admin panel, then add whatever users to that group.

If you want to test the anonymous user add `anonymous_user_name: AnonymousUser` to your `config.yaml` and add it to the `view` group.

If it still isn't working correctly for normal users try changing `rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly` to `rest_framework.permissions.DjangoObjectPermissions` in `settings.py`.

eeintech:
@Zeigren This draft is little out of date now, any chance you can bring it up-to-date? I'm interested to have a look, this could be handy for our company too.

SchrodingersGat:
@Zeigren I would also be quite keen to see this updated to the latest InvenTree codebase. I would like to start looking seriously to integrate a permissions architecture. And your approach looks very sensible. Do you have the time to look at fixing the conflicts?

eeintech:
@SchrodingersGat For the permission architecture, I would like to introduce fine control at the Part, Stock, Build, etc. level to decide which role and users have write access on those. So that for instance, EEs should have full control over Parts/Build/Stock, PMs on Stock/Sales/Purchase orders, manufacturing people on Stock/Build, etc. Does that make sense?

I think this implementation is slowly becoming higher priority at our company so I'm happy to take a look and send out a proposal, what's your take on this?

SchrodingersGat:
@eeintech I agree completely with your approach - this is how I want to use it at my company also.

Fine-grained control is great, then the admin user can create their own groups which contain the correct permissions.

For example you could have a group that can view part pages but not edit them.

SchrodingersGat:
Started to work on the framework for this here - https://github.com/inventree/InvenTree/pull/1010

eeintech:
@SchrodingersGat Great, I did some reading too and am going to start implementing the group/roles in the admin interface so that we can add, edit, remove roles and have "fine" control what role has write access to Parts, Stocks, Build, Companies, etc. Then we can have a one-to-one user to role relationship and the user's role would drive the write access to InvenTree.

I'm thinking to bring [django-role-permissions](https://github.com/vintasoftware/django-role-permissions) to the mix, but I wonder how it handles dynamically created roles, maybe building on top of the [Django Authentication System](https://docs.djangoproject.com/en/3.1/topics/auth/default/) would be better.

SchrodingersGat:
The django-role-permissions integration looks really nice, I like the idea of "roles" rather than "groups".

### Advantages

- Not just limited to the view / add / edit / delete functions of groups
- Can define default roles that make sense for InvenTree, such as "sales_officer" who can create / edit sales orders, assign stock to orders, etc

### Disadvantages

- Not sure (yet) if it can handle dynamically created roles, or if users who with to have different role implementations will need to edit the code in `roles.py`. This would be a large disincentive to me. If they could edit roles in the admin interface, sure...

eeintech:
@SchrodingersGat I agree that InvenTree should support dynamic role management in the admin interface and not at the `roles.py` level. I will give it a shot and see where it leads us 👍

eeintech:
I was not very familiar with the Django users and permissions system but here are my today's findings (mostly notes for myself but feel free to comment/react).

## [Django Authentication System](https://docs.djangoproject.com/en/3.1/topics/auth/default/#using-the-django-authentication-system)

When I started to look into it, I discovered that `Groups` seemed like a candidate for the `Role` management mission. I was like "sweet, it looks much simpler than I thought and this is the Holy Grail of "fine-control" for permissions" but then I realize how painful it was to attach every single permission for every single model to a group, definitely not user friendly (can be very confusing if not familiar with InvenTree's models), with a huge amount of work to update views and templates, prone to permission errors, etc.

Conclusion: we definitely need a role-based system!

## [django-role-permissions](https://github.com/vintasoftware/django-role-permissions)

I was not able to figure out how to define `AbstractUserRole` dynamically. Also I don't like the fact that the permissions related to the `auth.User` model instead of being tied to the specific model it relates to.
For instance:
```
class Engineer(AbstractUserRole):
    available_permissions = {
        'add_part: True,
    }
```
will create a new `auth.User.create_part` permission instead of using the default `part.add_part` permission automatically generated for the Part model. I don't think there is a way to connect the two, or I didn't find it.

Conclusion: we need to extend the User model to define custom roles.

## Extending User Model with One-to-One relationship

I started naively looking into the Django docs and the implementation from [this section](https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#extending-the-existing-user-model).

I started "playing" with one-to-one user-to-role relationships but when I added a model called `RoleOptions` to define the level of permissions for each `Role` model, I realized 1:1 isn't enough, I don't want to have to create X number of `Engineer` roles for X number of users. And then have to attach the same permissions each time, no no too much work and not user friendly...

Conclusion: we need a custom User model to define roles.

## Extending User Model using AbstractUser

My last attempt of the day which quickly concluded when I saw how many errors pointing to a new `InvenTreeUser` model got me:
```
SystemCheckError: System check identified some issues:

ERRORS:
build.Build.completed_by: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'.
order.PurchaseOrder.created_by: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'.
order.PurchaseOrder.received_by: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'.
order.PurchaseOrderAttachment.user: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'.
order.SalesOrder.created_by: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
...
stock.StockItem.stocktake_user: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'.
stock.StockItemAttachment.user: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'.
stock.StockItemTestResult.user: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'.
stock.StockItemTracking.user: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'.
```

Conclusion: Adding a Custom User model pushes for a new database structure, I'm not sure to which extent...

SchrodingersGat:
I actually think the fine-grained controls of the existing user / group permissions are what we want here. Any higher-level permission roles that we decide to create here might not match a particular use case.

For example, let's say we create a role "engineer" which can edit parts etc but cannot edit or create sales orders (or see sales pricing).

This role might work for your company but not someone else's?

I would suggest that we allow very fine-grained control of the permissions, as this allows the system to be quite flexible.

What I would also suggest is that we create a set of "default" or "example" groups, which have a set of model permissions that make general sense. Then end-users can edit these groups, or duplicate them, to match their particular requirements.

But the available groups would be pretty generic and cover most use cases...

eeintech:
@SchrodingersGat I had this brief moment of thinking InvenTree could have "template groups", I guess this is what you had in mind.

Maybe this is me being over-the-top here, I envisioned some easy to customize table that will be defined for each role.
Here are some quick mockups:

image

image

image

image

Each check will add/remove group of permissions for related objects. And this is without knowing what `BomItem`, `PartParameter`, `PartAttachment`, etc. InvenTree objects are. Also this could live in the InvenTree settings instead of the Django Admin interface.

It's possible the [django-guardian](https://github.com/django-guardian/django-guardian) library can help with this implementation (eg. grouping view/add/change/delete of multiple objects and to add/remove group of permissions with each checkmark).

If some users wants more "fine-control", in this case they would have to edit the Django Groups manually.

Is that too much you think?

SchrodingersGat:
If we can make a simplified table like you mocked up, with names like "build" rather than actual table names, I'm all for it.

Each column represents a "proxy" for a set of permissions.

If an admin wants a more fine-grained control of the groups then that is possible too!

Not sure how to:

a) Create the "table" view in the admin interface (or maybe we put it in the "settings" tab rather than the admin interface?)
b) Create this meta-concept of "roles" which are a superset of table permissions

eeintech:
> a) Create the "table" view in the admin interface (or maybe we put it in the "settings" tab rather than the admin interface?)

I'm thinking it would be inherent to InvenTree so maybe in the "settings" tab would make more sense (reserved only for superusers and staff). I will investigate both to see what makes more sense and is more convenient.

> b) Create this meta-concept of "roles" which are a superset of table permissions

I will take a look as well. I am hoping to get something as simple as (or similar) `{% if perms.user_role.add_part %}` in templates, but I don't know yet what it would take to get there.

SchrodingersGat:
> I will take a look as well. I am hoping to get something as simple as (or similar) {% if perms.user_role.add_part %} in templates, but I don't know yet what it would take to get there.

Already you can use the following code in templates: `{% if perms.part.add_part %}` or `{% if perms.stock.view_stockitem `%}

eeintech:
> Already you can use the following code in templates: `{% if perms.part.add_part %}` or `{% if perms.stock.view_stockitem %}`

Yes you're right, my example was a little silly... I was trying to show a "user role-based" check vs "model-based" check. Now I realize `perms` would carry the permissions for the role: instead of seeking permissions from the `User`, it will seek permissions from the `UserRole`. I guess this what I was trying to do "extending" the `User` model and connecting a `Role` object to it. The new `User` model could extend the `has_module_perms()` method and go check permissions into the associated `Role` object instead of the default. The new `Role` model could be defined in the `users` app and be a ForeignKey of `User`, it would be storing a "bundle" of permissions and be used as the "proxy" for the permissions check.

I'm still uncovering ground at this point, I may be leading myself in wrong directions... I'll do some more research and reading!

eeintech:
My thoughts depicted above are put into action through the draft PR #1016

SchrodingersGat:
@Zeigren check out the changes made in https://github.com/inventree/InvenTree/pull/1018

If you have any further comments or requirements feel free to re-open this issue or start another.

PR #630, opened 2020-02-02
Update German translation

chschlue:


SchrodingersGat:
Thanks for the translations @chschlue 

If you see any files / strings which are not currently translated, please feel free to add those too. If you want some pointers on how to add translatable strings I'd be happy to help.

SchrodingersGat:
Oh also @chschlue please make sure you run `make translate` before pushing the updates, so that the translation files get re-compiled! Cheers.

chschlue:
Didn't I?
Why else would the mo file have changed?

SchrodingersGat:
Hmmm I think actually I might have had some confusion between branches on my end - it looks good now.. Sorry!

chschlue:
> If you see any files / strings which are not currently translated, please feel free to add those too. If you want some pointers on how to add translatable strings I'd be happy to help.

There are quite a lot of fixed strings in the code.
Do you have a preference about how to structure a PR so it doesn't become overly convoluted? Like one commit per file?

SchrodingersGat:
I'm happy for multiple files to be updated in a single commit - as long as the changes made are all related. e.g. "update translations for Part application" or similar.

Issue #617, opened 2020-01-17
Geewhiz metrics


a7hybnj2:
This is just for fluff but could possible be useful somehow.

Add an area similar to `Part Categories ##` and `Stock Locations ##` to give you total # of parts in inventory. Maybe add some other metrics in there like least/most expensive part, most used part, most purchased part, ect.

My primary interest was in total parts in inventory so I could say I stock ## parts.

SchrodingersGat:
What if it were an option in the main drop-down menu in the top-right corner which shows database stats. It is only called when required but can include a lot of varied stats as required.

a7hybnj2:
Sounds perfect!
Jan 19, 2020, 05:25 by notifications@github.com:

SchrodingersGat:
![image](https://user-images.githubusercontent.com/10080325/73607425-fd819780-4609-11ea-83bf-ac4f810c3e0b.png)

![image](https://user-images.githubusercontent.com/10080325/73607431-070aff80-460a-11ea-8c2e-ff030ddea07c.png)

a7hybnj2:
Perfect. If I think of any additional metrics that could be useful I will let you know.

PR #632, opened 2020-02-02
Database stats

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/617

Implements a new dialog view which shows some simple (for now) database stats.

### DIFF
--- a/InvenTree/InvenTree/static/script/inventree/inventree.js
+++ b/InvenTree/InvenTree/static/script/inventree/inventree.js
@@ -41,6 +41,12 @@ function inventreeDocReady() {
 
         modal.modal('show');
     });
+
+    // Callback to launch the 'Database Stats' window
+    $('#launch-stats').click(function() {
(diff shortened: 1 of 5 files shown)


mtrazakhan:
Hi, I want a access management app in this inventory project separately for role based permissions(read, write, update and delete) permissions for users

SchrodingersGat:
Have you looked into the Admin interface? This provides user and group management?

image

mtrazakhan:
okay, how can I show group in navbar only when admin login and then at the time of adding user i can add this user in a particular group.

mtrazakhan:
from outside of my admin panel

SchrodingersGat:
Sorry for my difficulty in understanding - are you trying to replicate the admin functionality here? I do not see the point of duplicating this when the "admin" users can log into the admin interface directly.

mtrazakhan:
Want something like this:-
createRole
EditRole
RoleListing

PR #651, opened 2020-02-17
Use black and isort for automatic formatting

Zeigren:
I know this is a big change that can be a bit polarizing but I think automatically enforcing a style would be helpful, especially for those of us that want to contribute. I would propose using [isort](https://github.com/timothycrosley/isort) for sorting imports and [black](https://github.com/psf/black) for the actual python formatting.

I've added both and used the recommended configurations for using isort, black, and flake8 together. I had to bump up the versions of flake8, python-coveralls, and coverage to make them compatible. I also added isort and black to the Makefile so they're ran whenever `make style` is used.

Instead of making the case for it myself here's why [Django](https://github.com/django/deps/blob/master/accepted/0008-black.rst) started using black, they also use isort.


Zeigren:
Here's a [comparison ](https://github.com/Zeigren/InvenTree/tree/black-isort-example)with the current master branch. What I should have done in the first place instead of adding it to the pull request...

eeintech:
@SchrodingersGat Did you decide in favor of using black and isort? The current formatting looks fine to me, I was just wondering what's your take on this proposal.

SchrodingersGat:
Thanks for bumping this @eeintech I had forgotten about this one. I think this will take focus away from actually doing core dev on the product which is not appropriate at this time. I'm going to close it out @Zeigren - sorry that it took me so long to get back to you!

PR #676, opened 2020-03-22
use rapidfuzz instead of fuzzywuzzy

maxbachmann:
FuzzyWuzzy is GPLv2 licensed which would force you to licence the whole project under GPLv2. I had the same problem on one of my projects and so I wrote [rapidfuzz](https://github.com/rhasspy/rapidfuzz) which is implementing the same algorithm but is based on a version of fuzzywuzzy that was MIT Licensed and is therefor MIT Licensed aswell, so it can be used in here without forcing a License change. As a nice bonus it is fully implemented in C++ and comes with a few Algorithmic improvements making it between 5 and 100 times faster than FuzzyWuzzy.

### DIFF
--- a/InvenTree/part/bom.py
+++ b/InvenTree/part/bom.py
@@ -3,7 +3,7 @@ Functionality for Bill of Material (BOM) management.
 Primarily BOM upload tools.
 """
 
-from fuzzywuzzy import fuzz
+from rapidfuzz import fuzz
 import tablib
 import os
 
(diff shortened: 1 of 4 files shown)


SchrodingersGat:
The CI toolchain is failing, can you please add "wheel" as a pre-requisite in the requirements.txt file?

maxbachmann:
I did now build the wheels for a lot of platform: https://pypi.org/project/rapidfuzz/#files so it is now not longer required to compile it on most platforms

SchrodingersGat:
Ok, great! Thanks for your contribution

Issue #674, opened 2020-03-22
[Enhancement] Add "On Order" flag to Part Status 


maholli:
Similar to the "No Stock" flag for Part Status:
![image](https://user-images.githubusercontent.com/29153441/77239496-be4ade00-6b98-11ea-923f-7c6811c55006.png)

There should be an "On Order" flag for Parts that have an On Order quantity >0:
![image](https://user-images.githubusercontent.com/29153441/77239511-e4707e00-6b98-11ea-9cbe-018cb47d2dba.png)

I'd make the flag blue-ish to visually correlate to PO "Placed" status:
![image](https://user-images.githubusercontent.com/29153441/77239528-08cc5a80-6b99-11ea-8daa-b1e07bfa8ad9.png)

SchrodingersGat:
Hi @maholli can you please mock-up a "screenshot" of where this on-order flag might be placed?

maholli:
@SchrodingersGat I'm envisioning it something like this:
(from that Parts tab)
![image](https://user-images.githubusercontent.com/29153441/77575584-fd2aad80-6e90-11ea-945f-529a2cc41e71.png)

Obviously not critical, but would nice to have :)

a7hybnj2:
I like it

PR #678, opened 2020-03-26
Add "On Order" badge

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/674

image

If a part has no stock, but is on order, display a special "on order" badge


SchrodingersGat:
This implementation actually has a bug - the "on_order" sum also counts orders which are "closed" (e.g. received etc)

PR #715, opened 2020-04-13
Manufacturer

SchrodingersGat:
The "manufacturer" field for the SupplierPart model now points to a Company reference, rather than just being a simple text field (e.g. Manufacturer name)

This now provides a "Company" view for each "manufacturer" allowing the user to show all parts from a given manufacturer, even if they come from multiple suppliers, etc.

This requires a pretty complex custom migration to help the user migrate from manufacturer "names" to actual manufacturer objects.


SchrodingersGat:
Gak, stupid rebase.

Issue #713, opened 2020-04-13
Barcode Scanning API


SchrodingersGat:
Following on from the discussion in https://github.com/inventree/InvenTree/issues/677

Provide an API endpoint for sending scanned data for validation.

### Workflow

1. Client application (web / app / third-party) scans barcode and extracts json data object
1. JSON data is sent to server
1. Data are parsed by the server, and "sanitized" into a standard barcode-metadata object
1. Sanitized (and error-checked) data object passed back to the client for further actions

### Barcode Formats

**InvenTree**
Barcodes can be provided in different "formats". The default format is the QR code generated by InvenTree itself for a StockItem or StockLocation. This barcode data contains some very simple ID information about an element - other data is provided via API lookup.

**Manufacturer**
Manufacturer / supplier barcode support can be added via a simple translation layer. These barcodes are actually much more useful than the InvenTree ones, as they can be used for:

- Unique lookup of StockItem by scanning original manufacturer tag
- Scanning of incoming goods - backend compares metadata to match against a purchaseorder

### Plugins

Each "manufacturer" barcode style will be supported via a "Plugin" which is loaded by the server. The plugin must provide a certain set of functions and be able to validate a certain style of barcode.

maholli:
Hooray! Awesome work @SchrodingersGat 

# RE: Plugins

Rather than having each manufacturer as a separate plugin, I took a different approach with my [getparts()](https://github.com/maholli/getparts) plugin (new name). We could break it up as you suggest, but then we'll need some preliminary script that determines the appropriate supplier from the barcode information. Getparts() does supplier determination and part query in one step.


# RE: workflow

> 3. Data are parsed by the server, and "sanitized" into a standard barcode-metadata object

Do we know exactly what we want in the standard barcode-metadata object? If you can list them for me I'll modify getparts() to output only those fields.

Currently:

```python
>> api = getparts.API(app_credentials)
>> result = api.search(barcode) # result is JSON
>> print(result)
>> {
>>    "Capacitance": "10nF",
>>    "Description": "10nF ±10% 50V X7R 0603 Multilayer Ceramic Capacitors MLCC - SMD/SMT RoHS",
>>    "Datasheet": "Walsin Tech Corp 0603B103K500CT",
>>    "Manufacturer": "Walsin Tech Corp",
>>    "Package": "0603",
>>    "Voltage - Rated": "50V"
>> }
```

So we need to decide on...

1. What the standard metadata should be, 
[...]

PR #722, opened 2020-04-13
Create simple endpoint for barcode decode

SchrodingersGat:
API endpoint for checking barcode data

Fixes https://github.com/inventree/InvenTree/issues/713

This PR provides a "plugin" architecture (finally!) - the first implemented plugin is a barcode interface.
Third-party barcode plugins can then be implemented wholly separate from the database code!


SchrodingersGat:
@maholli there is now an endpoint /api/barcode/ to throw some data at. It expects a simple data format:

```
{
  barcode: "barcode_data_goes_here"
}
```

Currently it works to decode the native InvenTree style barcodes, with a skeleton wrapper for DigiKey codes. 

The StockItem model now also has a "UID" field which can be mapped to a barcode hash

maholli:
> 
> 
> @maholli there is now an endpoint /api/barcode/ to throw some data at. It expects a simple data format:
> 
> ```
> {
>   barcode: "barcode_data_goes_here"
> }
> ```

@SchrodingersGat I think there's some confusion of terminology (at least in my head).

I've been calling:

   - BARCODE data = dict returned from a barcode decoding tool (like zxing)
   - PART data = dict returned from my getparts() "plugin."

Since you mentioned you have zxing working for the mobile app, I've been under the assumption you have a method/plan for acquiring the barcode data. 

So is the endpoint /api/barcode/ expecting barcode data or part data?

SchrodingersGat:
The endpoint expects `BARCODE` data (i.e. data decode from a barcode).  It could be a dict or a raw string 

```
{
barcode: "[)>06PVSOP98260CT-ND1PVSOP98260K1K5415234510K6195019511K14LQ1011ZPICK12Z350468613Z15851020Z000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
}

**Comment by SchrodingersGat** [comment]
The server then tries to find a plugin that knows how to decode the particular barcode data, and then will return a response from that plugin.

The native "InvenTree style" barcode plugin simply performs a database lookup on the provided data which might be like:

image

```
{"instance": "Oliver's Test Instance", "stockitem": {"id": 407, "url": "/api/stock/407/"}, "tool": "InvenTree", "version": "0.0.11_pre"}
```

**Comment by SchrodingersGat** [comment]
So a digikey barcode plugin logic would be, (roughly):

1. Does this look like a digikey barcode?
1. Does the hash of this barcode map directly to a stockitem?
  YES -> Return that item
  NO -> API lookup required for further barcode data
1. Request barcode data from digikey API
1. Does the barcode map to a PurchaseOrder / SupplierPart / etc?
1. Work out what to do from here based on the point above:
  a) Create a new stockitem and mark as "received" against a certain purchase order
  b) Create a new stockitem without a PO
  c) Redirect to information about the part / supplier / etc

**Comment by SchrodingersGat** [comment]
In the case of the InvenTree barcode shown above, the server would respond with something like:

```
{stockitem: {id: 262}, success: Barcode successfully decoded, plugin: InvenTreeBarcodePlugin, hash: bffa382ed6114aa6d7ef385fed904d2f, barcode_data: {"instance": "Oliver's Test Instance", "stockitem": {"id": 262, "url": "/api/stock/262/"}, "tool": "InvenTree", "version": "0.0.11_pre"}}
```

(Note this is actually a scan result for a different stockitem, but you get the idea)

**Comment by maholli** [comment]
Looking great @SchrodingersGat!  👍

Did you get a chance to run one of your component-bag barcodes through [getparts()](https://github.com/maholli/getparts)? I can easily add InvenTree QR codes as an option, but I'll need your feedback on usability

**Comment by SchrodingersGat** [comment]
Not yet, although I would like to soon.

The inventree-style QR codes do not need any API lookup - they map directly to items in the database so all that info is directly available.

Your API code  would be useful when the parts are first scanned into the system - grab any data from the API and match it against the database. 

I need to have a think about how best to present the data to the user, though.

And now that the barcode API is working, I'd like to start thinking about supporting barcode scanning via the web interface - e.g. a webcam or connected USB camera

**Comment by SchrodingersGat** [comment]
@maholli here's where I see this progressing from here:

There's currently a skeleton for a digikey barcode plugin. You have already written code to recognize a digikey barcode and then perform API lookup. This code should be integrated into the barcode plugin:

## Recognize Barcode
First step is to parse the barcode data and say "Yes, this looks like a digikey barcode! 

Simply extend the class skeleton which is provided at:

`./InvenTree/plugins/barcode/digikey.py`

## Request API data
Send the barcode data to the digikey API and then return a JSON object. Here you will need to supply an API key. The InvenTree database already supports arbitrary key:value pairs (see `./InvenTree/common/models.py`

So you could call something like:

```python
from common.models import InvenTreeSetting

# See if there is a digikey api key available
# The call will return None if the key does not exist.
api_key = InvenTreeSetting.get_setting('digikey-api-key')
```

Then the API key can be added to the database:

image

## Work out what to do with the returned data

If you are willing to work on the two above points, then once we have data coming back from the digikey API is when things get really interesting!

Let me know if you're able to look at these points. I'm happy to provide pointers where you need them!

maholli:
@SchrodingersGat I'm happy to put together the plugin and demo the functionality, so I'll take a crack at this weekend.

That being said, I still think the supplier-specific plugin approach is setting us up for inefficiency. Would you be opposed to something like an `id_supplier.py` plugin that determines the appropriate supplier from the barcode, then dispatches the correct supplier-specific API. Should those be separate plugins?

SchrodingersGat:
I would strongly prefer the functionality to be split into separate plugins, the point of the plugin architecture is to separate the functionality out so code for one supplied does not "infect" the code for another. I can imagine the code for the digikey api growing and becoming pretty complex. Then as more manufacturers are added, the single plugin becomes very bloated.

With regards to inefficiency, code which is common between supplier plugins can be refactored to the parent BarcodePlugin class, there should be minimum code duplication.

rco3:
If I understand correctly, the /api/barcode/ endpoint looks for either a raw string OR a dict , and then attempts to match it to the appropriate decoding plugin by asking each available plugin to validate it?

So, could I pass in a list of barcode objects captured from a single bag?   I see that the Inventree barcode plugin looks first at the data type to see if it's a dict or a string - is there data validation that would reject a list, or some other reason I couldn't pass a list straight from pyzbar?? 

Extending maholli's webcam_example.py, I am decoding all the barcodes from a bag in one capture.  I have devised a strategy for uniquely identifying the vendor who produced the bag based on the pattern of barcode locations and types, so that I can then map the decoded barcodes to the appropriate fields (MPN, vendor, SKU, quantity, etc).  We can create a unique hash from the combination of these fields, too.

My intended strategy was to capture the array of barcodes, identify the vendor, and then call the appropriate mapping function to assign content to fields - all outside of InvenTree. 
With InvenTree's new Barcode Plugin architecture, the more appropriate path seems to be building a MultipleBarcode plugin class (to include a common Barcode_Sort method, e.g.) and a subclass for each vendor with the appropriate  Validate, Hash, Decode, etc methods.  

Seems like the main question (to me) is whether I can pass a list in. If so, I think the rest fits well into the architecture you've developed.  Thoughts?

SchrodingersGat:
That's a very interesting approach! There's no reason that you can't make a Barcode plugin which accepts multiple barcodes at once. The barcode data could be a list of barcode data for sure. The particular plugin just needs to be able to interpret the data. Please keep me posted if you actually develop this!

PR #735, opened 2020-04-18
Improvements for token authentication

SchrodingersGat:
- Do not POST username and password data
- This is already encoded in the Authorization: Basic headers


SchrodingersGat:
Note: This is a "breaking change" in the API!

Issue #746, opened 2020-04-21
Redirect from login view


SchrodingersGat:
When a user is not "logged in" and tries to access a page, they are redirected to the login page (as is desired). However after logging in they are sent to the home page. Instead, they should be sent to the page they were originally trying to access.

andyseracuse:
Hello again! I'm new to all this and I'm trying to find a good first issue.  I've completed an online course that gave me familiarity with java, html, and css and the MEN stack  and I'm working through the MIT online courseware for python right now.  would this be a good first issue?  Thanks so much for your response!

SchrodingersGat:
This would be a great issue to start with! 

To get you started, the 'login' page is rendered here: https://github.com/inventree/InvenTree/blob/master/InvenTree/templates/registration/login.html

And the code which is responsible for redirecting an unauthenticated user to the login page is under `AuthRequiredMiddleware` - https://github.com/inventree/InvenTree/blob/master/InvenTree/InvenTree/middleware.py

andyseracuse:
great!  I am gonna start working on this.  Can you assign me?  Also, to be clear, as far as I can tell you can't access any page without being logged first.  Is the user in question trying to access a part by it's URL and then being redirected to the login page?

SchrodingersGat:
> Also, to be clear, as far as I can tell you can't access any page without being logged first. Is the user in question trying to access a part by it's URL and then being redirected to the login page?

Yep, exactly. Sometimes if the browser logs you out, but the page is still "loaded" in the browser, you'll click on a link and instead be redirected to the login page and then have to find your way back.

andyseracuse:
i pushed a fix, but then I hit close and comment.  I think it made it through though?  let me know if I need to do anything else on my end.  Cheers!

SchrodingersGat:
You can hit "reopen" if you prematurely close it - I have now reopened it for you. But it appears there are still some kinks to work out

PR #782, opened 2020-05-03
after a login redirect and a successful login, the user is now redirected to the originally desired url

andyseracuse:
made changed middleware and made necessary changes to settings and urls files so that the user is redirected to their desired page after a login redirect unless the desired page was logout, in which case the user will be redirect

- Fixes https://github.com/inventree/InvenTree/issues/746


andyseracuse:
thanks for your patience.  It took me a while to figure out how to do this because of the workaround in the middleware for different verification methods from the mobile and web app.  I think there is a more concise way to do this with the login_required() decorator, but I don't think it would work with the current workaround in place.  In any case this seems to work just fine!

SchrodingersGat:
Hmm something isn't right here - looks like there is a recursive redirect happening:

image

I tried to access `127.0.0.1:8000/part/` when logged out.

SchrodingersGat:
Actually it seems to be working ok ... thanks for the fix! Please continue to contribute!

PR #837, opened 2020-05-26
Attachment serializer fix

SchrodingersGat:
Fix a "bug" relating to serialization of attachments,

if the server is hidden behind a reverse-proxy, attachments may not work proper good in the API.


SchrodingersGat:
Ah. phooey. Branch mixup.

PR #882, opened 2020-07-08
API: Delete Inactive Part

eeintech:
@SchrodingersGat This is related to https://github.com/inventree/inventree-python/issues/31
I wasn't so familiar with the Django REST Framework so I read some doc and came up with this implementation.
I have tested it with a few parts from my own InvenTree database and it seems to be functioning fine.


eeintech:
Sorry, forgot to switch back to master before creating this branch and another change sneaked in.

PR #883, opened 2020-07-08
API: Allow part deletion if inactive

eeintech:
@SchrodingersGat This is related to inventree/inventree-python#31
I wasn't so familiar with the Django REST Framework so I read some doc and came up with this implementation.
I have tested it with a few parts from my own InvenTree database and it seems to be functioning fine.

(Note: This is a fixed branch for previous pull request that you can ignore: https://github.com/inventree/InvenTree/pull/882)


SchrodingersGat:
Looks pretty good to me! Nice work

PR #886, opened 2020-07-20
Allow Part thumbnail update via API

eeintech:
Added `PartThumbsUpdate` API view to allow Parts thumbnails to be updated directly from the API


eeintech:
I merged too many commits, again...

PR #887, opened 2020-07-20
Allowing Part thumbnail (image) update via API

eeintech:
Added `PartThumbsUpdate` API view to allow Parts thumbnails to be updated directly from the API

### DIFF
--- a/InvenTree/part/api.py
+++ b/InvenTree/part/api.py
@@ -190,6 +190,21 @@ class PartThumbs(generics.ListAPIView):
         return Response(data)
 
 
+class PartThumbsUpdate(generics.RetrieveUpdateAPIView):
+    """ API endpoint for updating Part thumbnails"""
+
+    queryset = Part.objects.all()
+    serializer_class = part_serializers.PartThumbSerializerUpdate
+
+    permission_classes = [
+        permissions.IsAuthenticated,
+    ]
+
+    filter_backends = [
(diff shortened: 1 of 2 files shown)


eeintech:
I've also enabled this feature in `inventree-python`: https://github.com/inventree/inventree-python/pull/33

SchrodingersGat:
I assume you have tested this (via the Python API)?

eeintech:
@SchrodingersGat Correct, both manually (Django Rest Framework API page) and programmatically. I'm able to upload part images (jpeg format) directly downloaded from the internet, it works great 😄 
Now that I think about it, I'm don't think I have tried any other format like PNG or BMP, but I don't think that changes much to the code. I can give those other formats a shot, and maybe protect against someone uploading something other than a picture?

eeintech:
@SchrodingersGat So alright, you did well in making me wonder about other types of files and it was indeed triggering a Django error. To avoid this, I added a custom validation method to the thumbnail serializer (`PartThumbSerializerUpdate`) which allows the API view to raise an error and not crash Django, yay! :smiley: 
The validation method is based on the Django REST Framework API doc, specifically this section: https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation

SchrodingersGat:
Ok, this seems pretty nice! Thanks for this addition :)

PR #908, opened 2020-08-17
BoM Upload: Automatic Part match using name or IPN

eeintech:
@SchrodingersGat Starting to look into the BoM upload flow, I actually like what I see so far. This is a little improvement to attempt to 1-o-1 match the Part field using the `allowed_parts` set and part name. In my opinion, matching by IPN would be more efficient than using the part name so I also added it (in that case name matching isn't processed).


SchrodingersGat:
I like this, lookup by part IPN is a great addition.

I'm currently doing some work on improving the "template" file used to specify which columns names are used. 

I'll merge yours in first so we do not get out of sync. 

Cheers

SchrodingersGat:
See the updates here - https://github.com/inventree/InvenTree/pull/909

eeintech:
@SchrodingersGat Wan't me to hold off any more changes related to BoM management right now? I saw you also merged multi-level BoM update. I was interested into potentially leveraging it for showing supplier parts under each "Component" part that has one or more supplier parts. Also would like to improve the export supplier data, not just part data.

SchrodingersGat:
I'm happy for you to keep working on this, the multi-level BOM was something I needed urgently for my work, but I am not planning on any new development right away.

Cheers!

eeintech:
Ok sounds good!

Issue #841, opened 2020-05-27
Alternate Part


lookme2:
Do we have a way to handle alternate parts on a BOM?

SchrodingersGat:
@lookme2 not currently, although there are some ways to work around. What exactly are you imagining here?

lookme2:
I'm not really sure.  May you could add the Alt part# as a Qty of 0.  Just a thought.

eeintech:
@lookme2 @SchrodingersGat 
In my experience with BoMs, alternate parts are extra columns with manufacturer data, for instance:
| Name | IPN | Designator | Quantity |  Manufacturer A |  MPN A |  Manufacturer B | MPN B |
|---|---|---|---|---|---|---|---|
| RES 10K OHM 5% 1/10W 0603 | RES-000039-00 | R1, R2  | 2 | KOA Speer Electronics, Inc. | RK73B1JTTD103J | Panasonic Electronic Components | ERJ-3GEYJ103V |

CMs usually understand the liberty to use whichever manufacturer part is easier for them to source.
What do you guys think?

PR #911, opened 2020-08-18
Added manufacturer data to BoM export

eeintech:
@SchrodingersGat I took at a stab at adding manufacturer data to the BoM export, never mind if you have something better.
I think we can close [**this issue**](https://github.com/inventree/InvenTree/issues/841) then.

### DIFF
--- a/InvenTree/part/bom.py
+++ b/InvenTree/part/bom.py
@@ -14,6 +14,7 @@ from InvenTree.helpers import DownloadFile, GetExportFormats
 
 from .admin import BomItemResource
 from .models import BomItem
+from company.models import SupplierPart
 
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
This is great! I did find a small bug, which is that the code crashes if there is a SupplierPart which does not have an associated Manufacturer reference.

I have made a new PR here - https://github.com/inventree/InvenTree/pull/913 - which makes a fix for this.

I have also added an option on the BOM export form, allowing users to select whether supplier information is included in the exported BOM file:

image

SchrodingersGat:
I'm going to close this out as my slightly updated PR is merged. Thanks for making this happen!

Issue #793, opened 2020-05-07
Printing BOM with location list


lookme2:
Is there a way to print the BOM with a location.  I like to print a list when I start the build.

lookme2:
@SchrodingersGat  and @eeintech the export from BOM with location works perfect!  Thanks!!!!

eeintech:
Awesome, good to know 😄

Issue #126, opened 2019-04-16
Add custom fields to BOM export


SchrodingersGat:
- Parts can have 'custom fields'
- Add these fields to the BOM export

PR #919, opened 2020-08-20
BoM export: added option to export part parameters (#126) and stocks (#793)

eeintech:
Intention to fix both #126 and #793 
I have to admit I'm not quite sure the exported stock data is what @lookme2 was expecting, but at least the option is there and can be built on


eeintech:
@SchrodingersGat Just to be clear, is this only gated by @lookme2 approval? What minor change were you referring to?

SchrodingersGat:
The minor change is to alter the column heading from `Location` to `Default Location`

eeintech:
Ok, I just did. Sorry I couldn't see your review, not sure where to find it at :thinking:

SchrodingersGat:
Great, thanks for the fix @eeintech :)

PR #921, opened 2020-08-20
Show 'available_stock' in Part string representation

### LINKED ISSUE #760: Show quantity when adding to BOM
I am building a BOM and I have a few parts like 1µF caps where there are multiple entries for different manufactures even thought they would be considered interchangeable. It would be nice to have the quantity displayed next to the add item to bom sub-part so that you can pick which one actually has parts in stock.

I think this touches on an issue of is there a way to have a generic heading like 1µF 0805 Caps and have multiple children fill it?

eeintech:
I think this fixes #760 
At least in my opinion, gives a good idea of parts stocks during BoM item creation/upload process:

image


SchrodingersGat:
@eeintech I like this idea, however overriding the __str__ function for the model means that the stock calculation will be performed wherever this representation is used.

The `available_stock` function is reasonably expensive, so I think it should only be used where necessary.

I personally think a "cleaner" implementation would be to make a custom FormField which will be used wherever we want to display the "Stock available" for a given Part.

https://docs.djangoproject.com/en/3.1/ref/forms/fields/#django.forms.ModelChoiceField

Have a look at the "label_from_instance" function which could then be used to include these custom strings:

image

eeintech:
@SchrodingersGat Fair enough, it was a little crude implementation, see updated one.

I'm no JS expert so I can't quite figure out where can I manipulate the select field colors in the template to render something like:

image

SchrodingersGat:
Thanks, I think this looks pretty good.

Not sure how to change the color in the drop-down options. I think that there might be a way to override the "widget" for the FormField, and apply a custom css class based on whether there is stock available or not.

Then in `inventree.css` you could apply the CSS styling to those extra class types.

Something to investigate!

eeintech:
Ha yes, this could be a way. I'll see if I can take a look soon!

Issue #625, opened 2020-01-30
Enhancement, SSO log-in 


sihawken:
I work for a local startup that was/is in need of an inventory/BOM solution for our products. We found Inventree, and so far its been fantastic. 
 
We have Inventree securely hidden behind an SSO sign-in reverse proxy called [Pomerium](https://github.com/pomerium/pomerium), which is essentially our solution to not having a conventional intranet (and for working remotely).

This means we have to log in twice, once through google SSO, and a second time with InvenTree.

Is there a way to 1) unify the accounts (inventree SSO login) and/or 2) or have some other sort of automated log in? 

Thanks

SchrodingersGat:
Hey @sihawken I'm so excited that you're finding Inventree useful :)

I do not have any experience with such systems. Do you have any way to work out how to connect these two systems - and potentially contribute these enhancements back to the project?

sihawken:
I do not either, I set up the SSO with Pomerium after a night of research. I would absolutely love to attempt this, but it's very daunting. I'm a mechanical engineer by trade with only a tad bit of python experience. Pardon my general lack of understanding if I get something wrong.

Regardless, here are a few details that could help with finding a way to accomplish this:

**Pomerium allows setting request headers, which seems like it could be used to authenticate users.** 

> "This can be useful if you want to pass along additional information to downstream applications as headers, or set authentication header to the request."
Copy + pasted from: https://www.pomerium.io/configuration/#set-request-headers

**And here are a few details regarding SSO. SSO services are often used to "stitch" together services with a single user account.**.

The most common protocol for SSO is OAuth2.0, which Twitter, Facebook, **Microsoft, and Google** support: https://oauth.net/2/. Office 365/Sharepoint and GSuite users would definitely see the most benefit from this. Second to that is SAML, which is generally only used in large enterprises.

There seems to be a great [Django OAuth Toolkit](https://github.com/jazzband/django-oauth-toolkit) (FreeBSD License). Perhaps that could make something like this much easier, I am unsure.

Theoretically, it would be possible to pass a header to the downstream application (InvenTree) to log someone in with an already authenticated SSO session. Signing them in automagically.

Zeigren:
This is closely related to #634, it seems relatively straight forward to add providers to django-allauth. I also would like to use InvenTree with SSO so I'm willing to work on this at some point.

SchrodingersGat:
@sihawken - check out the change made in https://github.com/inventree/InvenTree/pull/930

[...]

PR #930, opened 2020-08-24
Allow custom authentication/middleware in config.yaml

bcc:
It is useful to be able to customise the middleware / authentication_backends without editing settings.py. This PR optionally loads them from config.yaml.

My use case is to be able to add 'django.contrib.auth.middleware.RemoteUserMiddleware' to MIDDLEWARE  and set AUTHENTICATION_BACKENDS to 'django.contrib.auth.backends.RemoteUserBackend' to make inventree use the REMOTE_USER variable alongside client certificate TLS auth. 

I'm happy to maintain this change in a local branch if you're not keen, but I thought it might potentially be a useful option for others.

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -163,7 +163,7 @@ LOGGING = {
     },
 }
 
-MIDDLEWARE = [
+MIDDLEWARE = CONFIG.get('middleware', [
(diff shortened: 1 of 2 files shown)


bcc:
This also potentially helps with #625 if the SSO system / webserver supplies REMOTE_USER.

SchrodingersGat:
Sure, this seems like a reasonable addition to me!

PR #931, opened 2020-08-24
Fix internal server error when serializing stock with no location

bcc:
If you attempt to serialize stock which has no location, the following error occurs:

  File "inventree/InvenTree/stock/views.py", line 1123, in get_initial
    initials['destination'] = item.location.pk
AttributeError: 'NoneType' object has no attribute 'pk'

This is a trivial fix, and the following form offers to set a location, so there's no sense in making that a requirement.


SchrodingersGat:
Great catch, thanks

Issue #932, opened 2020-08-25
Dynamic field updates in modal forms


SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/920

Some form fields are dependent on other fields within the same form. i.e. changing the `Part` field for a StockItem should change the available options for the `SupplierPart` field.

## Desired Functionality

Add the ability to mark certain fields as "dynamic" - when they are changed, a callback function is performed, which can do "whatever" (e.g. update the selections for another field)

Probably the best place to put this functionality is in the `launchModalForm` function (`/InvenTree/static/script/inventree/modals.js`).

Follow the pattern used to attach secondary dialogs to fields.

Issue #920, opened 2020-08-20
Create new Stock item form: selecting "Base Part" won't refresh "Supplier Part" list


eeintech:
Base part selected:
![image](https://user-images.githubusercontent.com/4020546/90819501-ae752880-e2f5-11ea-8225-024f62850849.png)

"Submit" button clicked:
![image](https://user-images.githubusercontent.com/4020546/90819768-00b64980-e2f6-11ea-953d-d032f6dcdad0.png)

Maybe a 2-step form?
1. Select base part, stock location, etc.
2. Select supplier part

SchrodingersGat:
This is potentially a difficult one to solve. Currently these modal forms are served "once", and then re-computed every time the form is submitted.

What would be nice is to have some sort of AJAX layer in the middle, where certain fields can be marked as "dynamic" (i.e. updating these fields causes other fields to be updated also). Ideally this would be reasonably abstract so it does not have to be "hacked" into each case where it is useful.

There some documentation on how the forms work [here](https://inventree.readthedocs.io/en/latest/forms.html) - this would be a good starting point. 

It feels like this would be a job for some front-end javascript code which is called whenever a tagged field is updated

eeintech:
Thanks for getting back on this, sounds like a good challenge. I'm not so familiar with JS so I hope someone can pick it up!

SchrodingersGat:
Someone did!

eeintech:
Haha who better than @SchrodingersGat!

PR #934, opened 2020-08-26
Reactive forms

SchrodingersGat:
Adds functionality to make modal forms more "reactive" to user input

## Callbacks

Adds the ability to attach a callback function to any named field. When the value of that field is changed, the callback is called (with the new value).

## Update Field Options

Adds helper functions to change the provided options in a given modal form field. 

--------

With these two new features, the following sequence can be allowed:

1. User changes selected value in a field
1. The new value is used to perform an AJAX query, resulting in a list of values being returned
1. The list of values is formatted and creates new `<option>` elements for a different field

------

- Fixes https://github.com/inventree/InvenTree/issues/932
- Fixes https://github.com/inventree/InvenTree/issues/920

### DIFF
--- a/InvenTree/InvenTree/static/script/inventree/modals.js
+++ b/InvenTree/InvenTree/static/script/inventree/modals.js
@@ -87,6 +87,53 @@ function setFieldOptions(fieldName, optionList, options={}) {
 }
 
 
+function reloadFieldOptions(fieldName, options) {
+    /* Reload the options for a given field,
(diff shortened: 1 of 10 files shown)


SchrodingersGat:
@eeintech this implementation is super simple, and very extensible. If you can find any other forms which would benefit from inter-linking fields, please point them out.

FYI, once the javascript code was setup, this is all it takes to link the particular form fields together:

image

PR #941, opened 2020-08-28
Support non-integer serial numbers

### LINKED ISSUE #860: Serial numbers: integer only?
Many of the parts I'll be using in assemblies include a factory-assigned serial number.  For a variety of obvious reasons I'd like to use that SN for tracking stock if present.  Currently the web interface, at least, rejects serial numbers that are non-integer.  I assume that this is to ease auto-generation of SNs; perhaps either option can be supported?

bcc:
This resolves #860 in a manner that shouldn't break the numeric serial number support.

I've done a lot of testing with random serials over the last few days (as we've been populating stock in our "production" InvenTree) and that all seems to be fine. Integer ones seem OK from the basic testing I've done, but more eyes would be good at this point.


SchrodingersGat:
@bcc  looks great! I've added a couple of other fixes and re-submitted this over at https://github.com/inventree/InvenTree/pull/948

SchrodingersGat:
Thanks for getting this 99% of the way 👍

PR #942, opened 2020-08-28
Fix invoke pty error when running 'invoke superuser'

bcc:
`invoke superuser` requires a proper TTY if asking for a password. Following the setup instructions for a new install results in the following error:

> Superuser creation skipped due to not running in a TTY. You can run `manage.py createsuperuser` in your project to create one manually.

This PR fixes this.


SchrodingersGat:
Ah, great catch @bcc thanks :0

PR #961, opened 2020-09-04
Implemented part parameters copy from template/original part

### LINKED ISSUE #955: Variant Part Parameters: should they be copied over from Template part?
This is first a question before being an issue. I realized parameters were not copied over from the Template part, shouldn't they be?
My preference would be that they are copied at the creation of the part variant, but their presence and value post-creation are not tracked with reference to the Template part. That way one can setup the Template part with placeholder parameters and they can easily be edited in the variants of that part.

@SchrodingersGat What do you think?
If you are on-board, happy to take a look on the implementation side.

EDIT: reformulation to match InvenTree's language

eeintech:
Closes #955

### DIFF
--- a/InvenTree/part/forms.py
+++ b/InvenTree/part/forms.py
@@ -129,10 +129,17 @@ class EditPartForm(HelperForm):
         'IPN': 'fa-hashtag',
     }
 
-    deep_copy = forms.BooleanField(required=False,
-                                   initial=True,
-                                   help_text=_("Perform 'deep copy' which will duplicate all BOM data for this part"),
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@eeintech one thing confuses me here - it seems you are creating the form with "hidden input" configured for bom_copy and parameter_copy fields, and then later adding a checkbox input. Is there a particular reason for this? If these fields were not initially created as "hidden input" they would work just fine...

SchrodingersGat:
This looks great! Thanks for making this contribution :)

PR #962, opened 2020-09-04
Added PartDetailFromIPN view (subclass of PartDetail) and url pattern

eeintech:
Now one can load Part detail page from `/part/<part_ipn>/`


eeintech:
Umm I'll have to dig into the part tests before it can be merged

eeintech:
@SchrodingersGat I'm happy with the current implementation, I tried to keep it non-intrusive so that all other views prevail. I just wanted to be able to access the part details using the IPN, it can be quite handy instead of guessing the pk.

SchrodingersGat:
I can see the appeal for this. However what happens when multiple parts share the same IPN? This is allowed and intended behavior - to allow multiple parts with the same IPN.

I would be happy if the new URL works if (*and only if*) there only exists one part with the matching IPN. Otherwise just redirect to the part index.

If you can ensure this behavior is observed then I'm happy to merge :)

eeintech:
@SchrodingersGat You made a very good point, I assumed IPNs were unique. I added a check for multiple objects returned in the `get_object` queryset. I also added unit tests for this new view, to ensure only a unique match returns the part detail page.

SchrodingersGat:
Thanks for the changes, and the contribution! :)

PR #963, opened 2020-09-05
Allow selection of bom_copy and parameter_copy when making a new variant

SchrodingersGat:


eeintech:
Thanks, I hesitated adding this as new Variants were copying everything by default, I did not want to alter the behavior.

PR #970, opened 2020-09-07
Color theme selection added to settings

### LINKED ISSUE #969: Is it possible to change main css?
I've created little more darker CSS. Is it possible to somehow use it via config or some settings so it will not conflict on upgrade?
If anyone is interested:
[inventree-darker.zip](https://github.com/inventree/InvenTree/files/5182314/inventree-darker.zip)

eeintech:
If approved by @SchrodingersGat and @Kedarius: closes #969 

An attempt to make InvenTree more colorful! (or less colorful and easier on the eye? Not sure :thinking:)

* Added `ColorTheme` (attibutes: name and user) model to **common** app
* Added `ColorThemeSelectForm` (uses crispy form helper), `ColorThemeSelectView` (manages user's color theme preference page in InvenTree settings) and corresponding url to **InvenTree** app
* Added `theme.html` template to `templates/InvenTree/settings/` folder
* Added `get_theme_css` tag to `part/templatetags/inventree_extras.py` which return the CSS sheet based on the user color theme selection
* Added `inventree-darker.css` CSS sheet provided by @Kedarius in the #969 issue, this is what it looks like in Chrome:

image


eeintech:
After reflection, I thought it would be better if the custom color theme CSS sheets were completely separated from main InvenTree CSS and had their own folder (`css/color-themes/`).
I've improved the color theme form and template and also added a Dark CSS sheet generated from the "Dark-Reader" chrome extension, looks pretty sweet:

image

SchrodingersGat:
@eeintech this is amazing! I love the dark theme too. I think the approach you have taken is very sensible.

I have a couple of suggestions which I'd like to run by you.

## Remove Theme "Hardcoding"
 
Currently the theme options are hard-coded in, which means to add a new theme the database model has to be changed.
What if instead, when the user selects a theme, it simply lists all the .css files which are in the `color-themes` directory, and presents these to the user in a list. Then adding a new theme is as simple as dropping a new .css file into that directory.

This would also require a few other subtle changes:

- If the user's currently selected theme file does not exist, the 'default' theme is used as a fallback
- If the user's currently selected theme file does not exist, display a warning next to the the theme selection box

## Add Unique Constraint

In the `ColorTheme` model, the `user` field should have the keyword argument `unique=True` added - this ensures that there can be only one row per user.

## Documentation

This is a cool new feature, it would be great to add a PR documenting this over at the [docs repo](https://github.com/inventree/inventree.github.io). It does not have to be exhaustive, just a couple of quick points:

- Describe the "theming" feature
- How to select your theme
- How to add a custom theme

Looking really good :)

eeintech:
@SchrodingersGat Thanks! Your suggestions look very good to me, I'll work on them soon

eeintech:
@SchrodingersGat Made the changes you've proposed, except for the documentation (my next step)

SchrodingersGat:
Nice work. Looks good on my end, thanks for this effort!

Issue #980, opened 2020-09-17
Expanding allocated items within a build fails if an allocated item is missing a location


wolffam:
When I go to expand "Allocated Parts" within a build, it works fine if the stock items within that build have "locations" associated.  This correct behavior is show below:
<img width="772" alt="with_location" src="https://user-images.githubusercontent.com/11801197/93428244-f5622980-f873-11ea-86ef-6e1a2fc98345.png">

However, if I try the same set of actions when a stock item inside a build does NOT have a location, the console reports that the pathstring is null.  That behavior is shown here:
<img width="1432" alt="without_location" src="https://user-images.githubusercontent.com/11801197/93428250-f72bed00-f873-11ea-921c-bd54183e9734.png">

I have a fix for this on my local branch and can get a PR out for it soon.

My fix is a simple `if` statement but would we prefer that all stock items that get added to a build adopt the builds location?  I'm not sure if that is logic that is wanted/needed.

wolffam:
Fixed with https://github.com/inventree/InvenTree/pull/981

PR #981, opened 2020-09-17
Handle situation where allocated items within a build do not have a location set

wolffam:
A fix for [#980](https://github.com/inventree/InvenTree/issues/980)

This is my first time modifying a django template so please let me know if there are any conventions that I should be aware of.


SchrodingersGat:
Fantastic, looks like a good solution here. And thanks for fixing the translation bug too.

PR #982, opened 2020-09-18
New homepage layout

eeintech:
Closes #403 Closes #140 

image

Happy to make any changes :wink:

PR #984, opened 2020-09-19
Tweaks for new index page

SchrodingersGat:
Follows on from the work done in https://github.com/inventree/InvenTree/pull/982


InvenTree/part/api.py
```diff
@@ -405,13 +405,6 @@ def filter_queryset(self, queryset):
             except (ValueError, Part.DoesNotExist):
                 pass
 
-        # Filter by latest part creation date
-        latest_parts = params.get('latest_parts', None)
-
-        if latest_parts is not None:
-            # Get the last 5 created parts
-            queryset = queryset.order_by('-creation_date')[:5]
```
SchrodingersGat, line 413: @eeintech the [django DRF framework](https://www.django-rest-framework.org/) already provides queryset ordering functionality, you just need to add 'creation_date' to the fields available for ordering

InvenTree/part/api.py
```diff
@@ -499,6 +503,9 @@ def filter_queryset(self, queryset):
         # Filter by "parts which need stock to complete build"
         stock_to_build = params.get('stock_to_build', None)
 
+        # TODO: This is super expensive, database query wise...
+        # TODO: Need to figure out a cheaper way of making this filter query
+
         if stock_to_build is not None:
```
SchrodingersGat, line 509: @eeintech this approach is pretty "expensive" in terms of database hits. It runs into the [N+1 query problem](https://scoutapm.com/blog/django-and-the-n1-queries-problem) as the query cannot be performed in a single run.

This will need some further investigation to determine if it can be sped up

InvenTree/part/api.py
eeintech: @SchrodingersGat Ha right, this looks much more elegant. Thanks for the improvements!

InvenTree/part/api.py
eeintech: @SchrodingersGat Right, I wasn't in love with this implementation, I'll have to think about this more...

eeintech: I just realized `django-debug-toolbar` can be enabled in the config, I'll check it out!

SchrodingersGat: Yeah it is a good way to see how many SQL queries are actually being executed (among other things too)

Issue #977, opened 2020-09-14
Add ability to modify a completed build 


wolffam:
Hi there!  I am hoping to use Inventree for keeping an inventory of hardware components but there is a feature that I need which is not currently implemented.

An example of the use case:
- 2 components (Part A and Part B) make an assembly (Assembly A)
- A "build" is performed given the BOM items above and serial numbers 1 and 2 are used for Part A and Part B respectively
- The build is completed and that assembly becomes a stock item (subsequently removing stock for parts A and B)
- Some time passes and the Part A (serial number 1) installed in that completed build needs to be replaced by Part A (serial number 11)
- Part A (serial 1) would be unallocated and the status can be changed to reflect its new state.  Part A (serial 11) would need to be allocated to the build and the build would need to be re-completed again.

From messing around the code base a bit, it seems that every time the build is moved from "Pending" to "Completed", it makes a new build output. I would imagine some business logic needs to be added to handle this scenario.

Is this a use-case Inventree purposely does not want to allow or does it just need to be coded up?  I'm new to django but wouldn't mind giving this a try if I'm aware of the common pitfalls when adding a feature like this.

SchrodingersGat:
This is a great idea, and I have been thinking about how to implement something like this for quite some time.

> Is this a use-case Inventree purposely does not want to allow or does it just need to be coded up?

It "simply" needs to be implemented!

**Background Reading:** - https://github.com/inventree/InvenTree/issues/918

For this to work, the first step is to implement the concept of "installing" a StockItem inside another StockItem (e.g. track a single serialized item inside another).

Currently this *does not work* - a "build" captures much more generic information about stock items that go into a particular build. But it does not currently record which particular unique items were assigned to which particular build outputs.



- 20 screws from this packet were consumed to make this build

wolffam:
Starting to try and make some progress on this for single output builds.  Here are some initial screenshots for how it would look to move a build into "MODIFYING" state.
<img width="500" alt="Screen Shot 2020-09-20 at 11 59 53 PM" src="https://user-images.githubusercontent.com/11801197/93739821-070e3e80-fb9e-11ea-8813-349b1c793858.png">

<img width="500" alt="Screen Shot 2020-09-21 at 12 00 03 AM" src="https://user-images.githubusercontent.com/11801197/93739811-02e22100-fb9e-11ea-9126-876b095144b9.png">

[...]

Issue #918, opened 2020-08-20
Add ability to link items (i.e. specific pcb with a specific daughter board)


Kedarius:
It would be great if the user would be able to easily link two or more items together. Something like a build but "weaker". 
For example I have a PCB with connector for Raspberry Pi. I have the PCB and the RPi as stock items. Now I need to say that this RPi has been put into that PCB....

SchrodingersGat:
@Kedarius this is a great idea, and one that I have been trying to work out how to do "simply".

Currently the `StockItem` model has a field `installed_in` which is set-aside for this exact purpose. However the "business logic" side of things has not been worked out yet.

What features / implementation details would be of particular interest to you here?

wolffam:
@Kedarius , if this functionality was added around the build object, would that work for your use case?  
I am envisioning a workflow where you stock components, make a build with the assembly level part, and the components that are allocated then "belong_to" that build. Then we can include functionality to "modify a completed build" which will let the user uninstall and install new items into that same build and the "belongs_to" identifier will change appropriately.
I'm still scoping out what this would look like and would like to make any changes as useful to others as I can so any input would be appreciated.

SchrodingersGat:
@wolffam there is an additional consideration which may make this *very* complex

Let's say you have a "build" which (as its output) makes 15 serialized widgets. Each serialized widget is tracked individually in the system. Each of these widgets has installed in them 1 serialized sub-widget (so 15 sub-widgets in total).

So, each serialized sub-widget must be "installed_in" the serialized widget. We cannot simply assign the sub-widgets to the build, as that does not provide tracking of each individual item.

I think a user-interface on the build which forces the user to assign any serialized stock to a particular serialized output is required.

wolffam:
Can you describe the workflow within Inventree to replicate that scenario?  I'm still trying to grok the project as a whole and want to make sure I understand the intended use of builds.

LarryJoy:
[...]

PR #991, opened 2020-09-20
Add the concept of "belongs to"/"installed in" to single output trackable builds

wolffam:
Multiple issues related to this PR but the most recent are #977 and #918 .
This PR only adds the concept of "belongs_to" for builds that meet the following criteria:

1. There is a single build output
2. The build part is trackable

It handles the case where the subparts are trackable as well as non-trackable.
There are follow-on features of "modifying" a build as well as supporting multiple build outputs (this would require a good deal of UI and a specific workflow) can be added later.

This is my first time working with a Django model so any feedback regarding conventions (the right way to query data for instance) is much appreciated!

### DIFF
--- a/InvenTree/build/models.py
+++ b/InvenTree/build/models.py
@@ -329,7 +329,7 @@ class Build(MPTTModel):
             item.save()
 
         # Install stock items into the build output(s)
-        self.install_subitems(subitems_used)
+        self.install_subitems(subitems_used, user)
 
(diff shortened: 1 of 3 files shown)


wolffam:
What the completed reference looks like:
Screen Shot 2020-09-20 at 12 39 02 AM

wolffam:
Superseded by #1000

PR #993, opened 2020-09-21
[WIP] Modify a completed build

wolffam:
Follows issues #977 , #918 and is downstream of PR https://github.com/inventree/InvenTree/pull/991
The goal of this PR is to put in place functionality that allows a user to "modify" a single output build by unallocating items and re-allocating new items before completing the build again.


wolffam:
## Starting Point

- Part: mod_assembly, SubPart: mod_component
- Stock: 2 stock items of mod_component serials 201 and 202
- Build of part mod_assembly completed with mod_component 201

## Workflow 

1. On a completed build, hit the blue "modify" button to change the build status to MODIFYING
Screen Shot 2020-09-25 at 4 17 08 PM

2. Build now looks like this.  I think we should have the stock name included in the dropdown to show the serial numbers but that stock item is serial 201.
Screen Shot 2020-09-25 at 4 17 21 PM

3. From here, it's the same as unallocating stock
Screen Shot 2020-09-25 at 4 17 37 PM

4. Serial 201 shows it has been installed and subsequently uninstalled:
Screen Shot 2020-09-25 at 4 18 03 PM

5. Serial 202 can now be installed in the build
Screen Shot 2020-09-25 at 4 18 38 PM
Screen Shot 2020-09-25 at 4 18 53 PM

6. Snapshot of the build output.  When a build is moved from MODIFYING to COMPLETED, instead of creating a new build output, it gets the current one and saves it with the new information.  I think I should also add transaction notes on the build output to show that the build was modified
Screen Shot 2020-09-25 at 4 19 13 PM

Any thoughts on what I've got so far?  Things you like or things you'd like to see changed/improved?  I'm not handling the situation where there is more than 1 build output, I am punting that one down the road.  Not sure if that's an extremely popular use case or not (it is not for me).

wolffam:
Closed due to taking a different approach for install/uninstall of stock items into builds

PR #995, opened 2020-09-22
Reduced SQL load for 'Require Stock to Complete Build' widget on homepage

eeintech:
Follow-up of #984 (2 SQL queries vs N+1 for build related items)

### DIFF
--- a/InvenTree/part/api.py
+++ b/InvenTree/part/api.py
@@ -7,7 +7,7 @@ from __future__ import unicode_literals
 
 from django_filters.rest_framework import DjangoFilterBackend
 from django.http import JsonResponse
-from django.db.models import Q, F, Count
+from django.db.models import Q, F, Count, Prefetch, Sum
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
Have you profiled this using the debug toolbar?

eeintech:
Oh yeah I should have mentioned this!

I have 21 parts in my database. I tested with following URL: `/api/part/?stock_to_build=true`

With the original code, I had `20 similar queries` regarding build items fetching.
With this code, those `20 similar queries` are gone and the number of SQL queries went down by 19.

However, the potentially redundant stock item queries are not eliminated. It's a bit more complex as the `prefetch_related` is embedded in the `Part.can_build` property method.
But the subset of parts is greatly reduce by the time it reaches this step, so I don't see any major hiccup here.

SchrodingersGat:
Nice, thanks for implementing this and testing it :)

PR #1001, opened 2020-09-28
Change layout for the index page

SchrodingersGat:
- Expanding one table doesn't force the neighbouring one to be expanded also

### DIFF
--- a/InvenTree/templates/InvenTree/index.html
+++ b/InvenTree/templates/InvenTree/index.html
@@ -8,41 +8,19 @@ InvenTree | Index
 <h3>InvenTree</h3>
 <hr>
 
-<div class="row">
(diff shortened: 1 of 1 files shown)


eeintech:
@SchrodingersGat I can still expand one table at the time:

image

It's a fine behavior for me, not sure why would someone want the neighbor table to expand to? :thinking:

SchrodingersGat:
Yes, this is the new and desired behaviour.

With the old layout, if you expanded (for example) the second row in the left column, then the second row in the right column would expand to. It was due to the ordering or rows / columns in the layout.

eeintech:
Umm ok, it did not have this effect when I tested it. But good that you fixed it then!

PR #1008, opened 2020-09-29
Add parametric part tables to category detail page

eeintech:
* Added tabs to category detail view: `Parts` and `Parametric Table`
* Display number of parts in given category next to the `Parts` tab name
* Fetching all parameters for all parts in given category (methods added to `PartCategory` class) to load bootstrap table

Current result is as-follow:

image

## TODO
- [X] Hide part specific toolbar

image

- [X] Improve table part information: `Part` and `Description`

- [X] Fix sorting for `Part` column

- [X] Add input filtering options for each parameter
https://bootstrap-table.com/docs/extensions/filter-control/
https://github.com/askalione/bootstrap-table-filter

### DIFF
--- a/InvenTree/InvenTree/static/css/bootstrap-table-filter-control.css
+++ b/InvenTree/InvenTree/static/css/bootstrap-table-filter-control.css
@@ -1,0 +1,13 @@
+@charset "UTF-8";
+/**
+ * @author: Dennis Hernández
+ * @webSite: http://djhvscf.github.io/Blog
+ * @version: v2.1.1
+ */
+.no-filter-control {
+  height: 34px;
+}
+
+.filter-control {
+  margin: 0 2px 2px 2px;
+}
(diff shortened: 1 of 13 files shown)


eeintech:
@SchrodingersGat I'm pretty happy with where the table stands right now, at least for the first version. Let me know if there is anything you'd like me to improve or change :wink:

SchrodingersGat:
This looks fantastic! I'm happy to merge this now, here's what I'd love to see in the future:

## Value comparison

For example, filter "Length" parameter by "< 10" and it matches anything less than 10.

eeintech:
> ## Part Index
> The parameter tab is visible on the "Part Category" page, but not on the "Part Index" page. It would be useful to be able to view parameters for every part...

I actually avoided the part index on purpose because what I consider "parameters" are different for each category, at least in my setup. For instance, I didn't see the point of mixing resistors and capacitors parameters. However I'm happy to add it, it wouldn't be that much work I think.

> ## Value comparison
> For example, filter "Length" parameter by "< 10" and it matches anything less than 10.

I'll need to play more with the bootstrap-table `filter-control` extension, this would be very handy indeed.
I tried the "select" mode for that extension and it threw a bunch of JS errors... May have to open an issue to figure this out.

Btw, would you know why icons can't be displayed on your implementation of the bootstrap-table? I couldn't figure out why in the search box the "cross/times" icon isn't showing to delete the whole query.

SchrodingersGat:
> Btw, would you know why icons can't be displayed on your implementation of the bootstrap-table? I couldn't figure out why in the search box the "cross/times" icon isn't showing to delete the whole query.

No idea..

PR #1016, opened 2020-10-02
Roles and permissions management system

eeintech:
This is an attempt to introduce `Roles` and `RuleSets` into InvenTree's user authorization flow.

## Roles and RuleSets
image

### What is a Role
A `Role` is mirroring the Django `Group` model instead of extending it or creating a custom model.
A Role has currently 5 `RuleSet` instances automatically created and managed from the role perspective.

### What is a RuleSet
It is the standard set of Django permissions for model (eg. view, add, change, delete) except that each applies to a superset of models. Basically, the `RuleSet` tries to simplify the language for admin users and automatically matches an InvenTree "application" (eg. Parts, Stock, Build, etc.) to its corresponding set of models so that the user does not have to know which model to set permissions on.

### What is implemented so far
`Role` and `RuleSet` models, along with a set of starter methods.
Those models are located in the `users` app.

The roles can be created and managed inside the Admin interface, under `Users` (very bottom of the index):

image

To create a Role:
* Click on "+ Add" button next to "Roles"
* Give it an awesome name
* Click on "Save and continue editing"
* Set the values for the rulesets:

image

* Click on "Save"

### What about the Group model in all this?

Well, after the `Role` instance is saved, all permissions related to all models from the rulesets are updated inside a `Group` instance of the same name as the `Role` instance (open the "Home › Authentication and Authorization › Groups" page and click on the role name you just created):

image

### What about users?

To add a role to a user, it's as simple as adding the corresponding group:

image

## TODOs

- [ ] A nice way to integrate Roles inside templates? Not sure if needed (we can use Django permission tags), requires research and brainstorming
- [ ] Move the Role management form to InvenTree's settings instead of admin interface? Currently feels fine in admin interface
- [ ] Update the hard-coded set of `RuleSet` to better represent InvenTree applications (need your insight here @SchrodingersGat)
- [ ] Test cases for `Role` and `RuleSet` models
- [ ] If approved, need to update documentation: https://inventree.readthedocs.io/en/latest/admin/permissions/ 

> Note: I'm not counting the work to update all views and templates with this new permission-based system (or the standard Django permission system if we decide to stick to it), that would come in a separate PR!


SchrodingersGat:
Looks like you've been doing a lot of research here! I've spent some time playing around with your code, and experimenting myself.

I think that this is becoming a *lot* more complicated than I first expected, which is fine, but means we'll need to consider this very carefully.

## Complex Permission Rules

It is not sufficient to match each "RuleSet" with an "app" e.g. Part. There will be interlinked permissions (e.g. the Build permissions will require part.bomitem permissions or stock.stocklocation) etc.

Also what about models such as "PartStar" - this links a "starred" part to the given user. We should not require that a user be given permissions to create a PartStar object in the part_part_star table when they click the "star" button (for example).

## Role Model

I do not think that the one-to-one addition of the `Role` model adds much in this case. It should be sufficient to add the rule_set model and link it to the (existing) `Group` model. Refer to the code in my branch (linked below) for another way of approaching this.

Adding a `Role` model in the middle simply increases complexity and I cannot really see the benefit of it.

## Overall Thoughts

Currently the web interface and API do not implement the permissions system at all - it is only implemented in the admin interface. We do not necessarily have to add permissions requirements to every element of the system - we could add it on an "as-needed" basis to simplify the implementation.

For example, the "part star" example as listed above.

Secondly, there are probably some permissions that every user needs to have.

Also, there will be many linked permissions as I suggested above.

Do we "need" the fine-grained control that the default django permission system allows? Or can we hide this entirely, start with some coarse permissions, and increase fidelity as needed?

I've also added some other thoughts inline with the code.

Let me know what you think.. I feel we are on the right track, and I really like the simple grid interface, but not sure we're 100% there yet.

SchrodingersGat:
For reference: https://github.com/inventree/InvenTree/pull/1018

SchrodingersGat:
## Further Thoughts

### Permissions Fidelity

As the complexity of the "roles" increases, we get closer and closer to just using the default django permission fields. Not sure what my point is here, just thought it was worth noting.

Even if all the "roles" implementation is doing is increasing readability so the user does not have to wade through entries like `part_edit_partcategory` then I think it is a net improvement

### Permissions Conflicts

Let's say some roles have overlapping permission on a given table, e.g. `part_part`. One role is marked as "can read", one role is marked as "cannot read". I think that "can" will override "cannot" in this circumstance - but we need to ensure that the code to update group permissions can handle these cases.

eeintech:
@SchrodingersGat Thanks for looking into this and creating that new PR, I will try to look into it soon and respond/comment on your thoughts.

SchrodingersGat:
Following on from the fantastic work you've done here, I think #1018 provides a solution that I'm happy with. Let me know what you think! 

This PR does not actually *do* anything with the permissions yet - that will take some doing as we have to guard a *lot* of parts of the interface. But that can be an ongoing work in progress!

For example, see https://github.com/inventree/InvenTree/pull/1011

eeintech:
> ## Complex Permission Rules
> It is not sufficient to match each "RuleSet" with an "app" e.g. Part. There will be interlinked permissions (e.g. the Build permissions will require part.bomitem permissions or stock.stocklocation) etc.
> 
> Also what about models such as "PartStar" - this links a "starred" part to the given user. We should not require that a user be given permissions to create a PartStar object in the part_part_star table when they click the "star" button (for example).

I anticipated it would the case, I had left room for improving this part (as suggested in the TODO list, the models weren't finalized). Looks like you had it covered in #1018 

> ## Role Model
> I do not think that the one-to-one addition of the `Role` model adds much in this case. It should be sufficient to add the rule_set model and link it to the (existing) `Group` model. Refer to the code in my branch (linked below) for another way of approaching this.
> 
> Adding a `Role` model in the middle simply increases complexity and I cannot really see the benefit of it.

`Role` was an extra layer which I started with to avoid messing with the `Group` model directly. My plan was actually to use it to control templates, and leave the `User` and `Group` models from Django as untouched as possible, to avoid future compatibility problems.

I agree it adds a bit of complexity and you're more familiar with Django than I am so I'll follow your call on this one 😉 

> ## Overall Thoughts
> Currently the web interface and API do not implement the permissions system at all - it is only implemented in the admin interface. We do not necessarily have to add permissions requirements to every element of the system - we could add it on an "as-needed" basis to simplify the implementation.

True, I was going to tackle this next, after defining and stabilizing the logic for the permissions.

> Do we "need" the fine-grained control that the default django permission system allows? Or can we hide this entirely, start with some coarse permissions, and increase fidelity as needed?

I see that you've removed it in #1018, maybe we wanted to keep it for the time being, especially when developing templates and would be easier for testing (eg. quickly add and remove a model permission).

> ### Permissions Fidelity
> As the complexity of the "roles" increases, we get closer and closer to just using the default django permission fields. Not sure what my point is here, just thought it was worth noting.
> 
> Even if all the "roles" implementation is doing is increasing readability so the user does not have to wade through entries like `part_edit_partcategory` then I think it is a net improvement.

I agree with the net improvement, I think it's fair to consider users don't need to look at the source code to define permissions. Any great software cannot be without a user-friendly interface, InvenTree is on the path of joining that category so I think it's worth the effort!

> ### Permissions Conflicts
> Let's say some roles have overlapping permission on a given table, e.g. `part_part`. One role is marked as "can read", one role is marked as "cannot read". I think that "can" will override "cannot" in this circumstance - but we need to ensure that the code to update group permissions can handle these cases.

That's a very good point, we should have the ability to drive this case-by-case and take the "safer" path but always good to keep in mind.

SchrodingersGat:
Thanks for the great work here @eeintech 👍

PR #1018, opened 2020-10-03
Roles and Permissions

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/pull/1018#issue-497274664

This PR augments the great work done by @eeintech in https://github.com/inventree/InvenTree/pull/1016

The default django permission management system is pretty complex to use, and while the fine-grained permissions control can be useful, we really want a "role" or "ruleset" approach.

This PR adds the RuleSet concept which maps to a particular group. The following table is provided in the admin page for each group:

image

Each row covers a particular set of database models, and each column determines the permission type for that ruleset.

Future work will be required to actually make use of the per-model permissions

- Add particular permission requirements to the DRF API
- Add particular permission requirements to various web pages
- Add particular permission requirements to the modal forms

(All this is pretty straight forward really)


SchrodingersGat:
@eeintech please let me know what you think of this implementation. I think it's pretty clean, but I might have missed something! 👍

InvenTree/users/models.py
```diff
@@ -1 +1,314 @@
...
+            'part_partcategory',
+            'part_bomitem',
+            'build_build',
+            'build_builditem',
+            'stock_stockitem',
+            'stock_stocklocation',
+        ],
+        'purchase_order': [
+            'company_company',
+            'part_supplierpart',
```
eeintech, line 75: `SupplierPart` and `SupplierPartBreak` are models of the `Company` app, not the `Part` app.

InvenTree/users/models.py
```diff
@@ -1 +1,314 @@
...
+            permission_string - a simplified permission_string e.g. 'part.view_partcategory'
+
+        Returns the permission object in the database associated with the permission string
+        """
+
+        (app, perm) = permission_string.split('.')
+
+        (permission_name, model) = perm.split('_')
+
+        content_type = ContentType.objects.get(app_label=app, model=model)
```
eeintech, line 278: To avoid Django to error for any mislabeled app and model (human mistakes can happen), I suggest replacing with:
```
        try:
            content_type = ContentType.objects.get(app_label=app, model=model)
            permission = Permission.objects.get(content_type=content_type, codename=perm)
        except ContentType.DoesNotExist:
            permission = None
```

`group.permissions.add(permission)` does not error out if None object is passed.

Better yet would be to report this to the user in a "nice" way.

eeintech (commented):
So far what I see looks nice, however I don't see the option to dive down to the Model level for managing permissions inside the Admin interface, would be handy to still have the choice to do so, no?

SchrodingersGat:
> So far what I see looks nice, however I don't see the option to dive down to the Model level for managing permissions inside the Admin interface, would be handy to still have the choice to do so, no?

Originally I had thought this was a nice idea, but there is an issue. The new RuleSet approach is pretty "heavy handed" with the permission control - every time you save a group, the permisions all get updated on the group. I couldn't think of a clean way to still allow fine-grained control of the permissions (and this is the antithesis of what we are trying to achieve by simplifying the permissions in the first place).

I think that the new ruleset approach is sufficient. Permissions can also be set individually for users if need be. 

What do you think?

InvenTree/users/models.py
SchrodingersGat: They were *originally* in the Part app, so the database tables retain the old names:

image

SchrodingersGat: And the unit testing has been setup to detect any of the following cases

- A ruleset references a permission that does not exist
- An existing permission is not referenced by any ruleset

eeintech: @SchrodingersGat I was not able to create a new `Group` after pulling your branch, I hit the `ContentType.DoesNotExist` error...

eeintech:
> Originally I had thought this was a nice idea, but there is an issue. The new RuleSet approach is pretty "heavy handed" with the permission control - every time you save a group, the permisions all get updated on the group. I couldn't think of a clean way to still allow fine-grained control of the permissions (and this is the antithesis of what we are trying to achieve by simplifying the permissions in the first place).
> 
> I think that the new ruleset approach is sufficient. Permissions can also be set individually for users if need be.
> 
> What do you think?

Umm I see, this would require a way to log individual users overwrites on each permission and add extra logic and complexity... I guess you are right, better abstract this layer completely and move forward with the rulesets.

InvenTree/users/models.py
SchrodingersGat: You're right, it does not actually use the table name, it *does* use the app name.

So `company_supplierpart` and `company_supplierpricebreak` are correct here.

I'm just writing some better unit testing to check for this. I'll push it momentarily

SchrodingersGat:
> Umm I see, this would require a way to log individual users overwrites on each permission and add extra logic and complexity... I guess you are right, better abstract this layer completely and move forward with the rulesets.

It's a good start - if in the future we find this approach is not sufficient for some nominal use case, we can always re-address it then...

SchrodingersGat:
@eeintech please pull down the latest commit and see if it works for you now

InvenTree/users/models.py
```diff
@@ -1 +1,317 @@
...
+
+        app, model = model.split('_')
+
+        return "{app}.{perm}_{model}".format(
+            app=app,
+            perm=permission,
+            model=model
+        )
+
+    def __str__(self):
```
eeintech, line 158: Replace with something like?
```
    def __str__(self):
        """ Ruleset string representation """

        return f'{self.name} | v: {self.can_view} | a: {self.can_add} | ' \
               f'c: {self.can_change} | d: {self.can_delete}'
```

eeintech:
@SchrodingersGat I noticed that `update_group_roles` is called 3 times upon `RoleGroupAdmin` is saved... Not quite sure why :thinking: 

I was testing your latest changes (with `part_partsupplierbreak` instead of `company`) and this is the output when printing the `RuleSet` (custom) string value and added/deleted permissions:
```
admin | v: False | a: False | c: False | d: False
part | v: False | a: False | c: False | d: False
stock | v: False | a: False | c: False | d: False
build | v: False | a: False | c: False | d: False
purchase_order | v: False | a: True | c: False | d: False
sales_order | v: False | a: False | c: False | d: False
Adding permission part.add_supplierpricebreak
Error: Could not find permission matching 'part.add_supplierpricebreak'
Removing permission company.add_company
admin | v: False | a: False | c: False | d: False
part | v: False | a: False | c: False | d: False
stock | v: False | a: False | c: False | d: False
build | v: False | a: False | c: False | d: False
purchase_order | v: False | a: True | c: False | d: False
sales_order | v: False | a: False | c: False | d: False
Adding permission company.add_company
Adding permission part.add_supplierpricebreak
Error: Could not find permission matching 'part.add_supplierpricebreak'
admin | v: False | a: False | c: False | d: False
part | v: False | a: False | c: False | d: False
stock | v: False | a: False | c: False | d: False
build | v: False | a: False | c: False | d: False
purchase_order | v: False | a: True | c: False | d: False
sales_order | v: False | a: False | c: False | d: False
Adding permission part.add_supplierpricebreak
Error: Could not find permission matching 'part.add_supplierpricebreak'
Removing permission company.add_company
```

InvenTree/users/models.py
eeintech: Gasp, this actually is captured in the admin form... Can put this on backburner / for debug only

SchrodingersGat:
Does it work better now that `part_supplierpricebreak` has been changed to `company_supplierpricebreak`?

eeintech:
> Does it work better now that `part_supplierpricebreak` has been changed to `company_supplierpricebreak`?

Yes it does 👍

SchrodingersGat:
Great! Now we'll need to add some documentation about this great new feature :)

PR #1019, opened 2020-10-04
Improvements for the "Installed Items" tab for StockItem display

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/977


SchrodingersGat:
@wolffam here's how the "rough mockup" from #977 ended up:

image

PR #1024, opened 2020-10-05
Permissions

SchrodingersGat:
Change what is visible on the various views, based on user permisisons


SchrodingersGat:
@eeintech I do not have time at the moment to change *all* the views / pages / etc, but here is a start on what is required to change what the user sees (and what actions they can perform) based on their assigned permission roles.

## Views

To prevent a user from looking at a particular page, you'll need to add the `PermissionRequiredMixin` to the view class.

And also define the `permission_required` value for that class.
e.g.

```python
class CategoryDetail(PermissionRequiredMixin, DetailView):
    """ Detail view for PartCategory """

    model = PartCategory
    context_object_name = 'category'
    queryset = PartCategory.objects.all().prefetch_related('children')
    template_name = 'part/category_partlist.html'

    permission_required = 'part.view_partcategory'
```

## Templates

In each template, you can use code such as `{% if perms.part.add_part %}` to determine if the user can add new parts (etc). Then these sort of checks can be used to hide buttons, etc.

## Modal Forms

Modal forms can be set to require permissions simply by setting the `permission_required` value. This is because of some code hooks I've added to the `AjaxView` mixin class.

e.g.

```python
class PartParameterDelete(AjaxDeleteView):
    """ View for deleting a PartParameter """

    permission_required = 'part.delete_partparameter'

    model = PartParameter
    ajax_template_name = 'part/param_delete.html'
    ajax_form_title = _('Delete Part Parameter')
```

## API

The API uses DRF which allows you to set the permission requirements to `rest_framework.permissions.DjangoModelPermissions`

This basically takes care of everything!

----------------------------------

So this PR is a great start to the process. But there's a lot more to do. Please use this as a reference and if you are able to look at any of the other pages / views / etc that would be great.

----------------------------------

Does this approach line up with what you wanted for your particular requirements?

eeintech:
@SchrodingersGat I started updating the stock views and templates but few things bugged me along the way:
- Rulesets were saved 3 times so I added logic to speed it up #1025 
- Users could be in multiple groups and it was confusing which permissions were enabled, I don't think we want to let this happen (it does not make sense in our implementation), see #1026 
- I fixed a 'View has no queryset method' DRF error in my current branch: https://github.com/eeintech/InvenTree/tree/permissions

And I have questions/comments:
1. Why the "Admin" RuleSet?
```
        'admin': [
            'auth_group',
            'auth_user',
            'auth_permission',
            'authtoken_token',
            'users_ruleset',
        ],
```
I would expect only staff and superusers to access the admin interface.

2. Now that we have bundled permissions together, I would have expected to take benefit from this implementation and instead of:
```
class PartIndex(PermissionRequiredMixin, ListView):

    permission_required = ('part.view_part', 'part.view_partcategory')
```

extend the `PermissionRequiredMixin` with:
```
class InvenTreePermissionRequiredMixin(PermissionRequiredMixin):
    model_permission = None

    def get_permission_required(self):
        permission_required = []

        if self.model_permission:
            (permission_model, permission_level) = self.model_permission.split('.')

        for model in RuleSet.RULESET_MODELS.get(permission_model, None):
            permission_required.append(
                RuleSet.get_model_permission_string(model=model,
                                                    permission=permission_level)
            )
        
        return permission_required
```
and switch from `permission_required` to `model_permission` in the views:
```
class PartIndex(InvenTreePermissionRequiredMixin, ListView):
...
    model_permission = 'part.view'
```

The custom Mixin could re-used across views in Part app as now `permission_required` is auto-populated with the given RuleSet and permission level, in the case of the `PartIndex` this is what we would get:
```
permission_required = ['part.view_part', 'part.view_bomitem', 'part.view_partcategory', 'part.view_partattachment', 'part.view_partsellpricebreak', 'part.view_parttesttemplate', 'part.view_partparametertemplate', 'part.view_partparameter']
```
I see where this could hit road bumps in the case one wants to customize who can create/update/delete for categories vs parts, but then it goes **against** our "Role" model which is there to enable/disable permissions altogether.

A potential other benefit is if the `RULESET_MODELS` dictionary changes, the views permissions will be automatically updated (not sure how often this would happen).

I would be happy to implement this if you think the same in term of benefits (I also hate to remember which models belongs where and the exact name).

I haven't thought about templates yet but I am leaning towards using `simple_tags` to mirror this behavior.

SchrodingersGat:
>     1. Why the "Admin" RuleSet?

The permissions system also governs what you can access in the admin interface. The **superuser** account can see everything, but any other **staff** user can only see what their permissions allows.

So, to be able to view / edit users, group, etc in the admin interface, the staff user must have the admin role assigned.

SchrodingersGat:
I like your suggestions regarding extending the "PermissionRequiredMixin" class, as it makes our life easier. 

In terms of the simple_tag approach in templates, we could mirror the behaviour of `{% if perms.part.add_part %}` with a check on various roles.

e.g. `{% if roles.part.add %}` would check if the user has "add" functionality in the "part" role.

We could also make this tag throw an error if you pass an invalid role name, rather than just failing slightly.

So views and templates would be simplified to just use the role permissions.

API endpoints will probably still use model-based permissions but that's OK because we don't have to code that manually, we get it for free as part of DRF.

eeintech:
> > ```
> > 1. Why the "Admin" RuleSet?
> > ```
> 
> The permissions system also governs what you can access in the admin interface. The **superuser** account can see everything, but any other **staff** user can only see what their permissions allows.
> 
> So, to be able to view / edit users, group, etc in the admin interface, the staff user must have the admin role assigned.

Oh okay, I didn't think of the staff option, that makes sense then.

> I like your suggestions regarding extending the "PermissionRequiredMixin" class, as it makes our life easier.

I'm currently fixing up the tests and will commit a improved version to what is shown above.

> In terms of the simple_tag approach in templates, we could mirror the behaviour of `{% if perms.part.add_part %}` with a check on various roles.
> 
> e.g. `{% if roles.part.add %}` would check if the user has "add" functionality in the "part" role.
> 
> We could also make this tag throw an error if you pass an invalid role name, rather than just failing slightly.
> 
> So views and templates would be simplified to just use the role permissions.

Exactly, that was the idea I was trying to convey.

> API endpoints will probably still use model-based permissions but that's OK because we don't have to code that manually, we get it for free as part of DRF.

And who doesn't like free goodies? :smile:

eeintech:
@SchrodingersGat About those tests failures... I couldn't figure it out so I opened an issue on SO hoping for insight:
https://stackoverflow.com/questions/64216938/django-limit-on-permission-required-size-for-testcase-permissionrequiredmixin

Maybe you got an idea why?

SchrodingersGat:
Have you added the required permissions to the user model in the test case?

e.g.

image

eeintech:
> Have you added the required permissions to the user model in the test case?
> 
> e.g.
> 
> image

This is really embarrassing and will blame it on my busy day... you are right.
I couldn't but tell myself that if at least one of the permission from the `permission_required` group was present in the user's permission, it should be allowed... Gasp!

InvenTree/part/templates/part/build.html
```diff
@@ -1,15 +1,18 @@
...
 {% include 'part/tabs.html' with tab='build' %}
 
-<h3>Part Builds</h3>
+<h3>{% trans "Part Builds" %}</h3>
 
 <div id='button-toolbar'>
     <div class='button-toolbar container-flui' style='float: right';>
         {% if part.active %}
-        <button class="btn btn-success" id='start-build'>Start New Build</button>
+        {% if roles.build.add %}
```
SchrodingersGat, line 13: @eeintech checking roles in templates is now done this way as per your suggestion

InvenTree/part/views.py
```diff
@@ -38,17 +38,21 @@
...
-class PartIndex(ListView):
+class PartIndex(InvenTreeRoleMixin, ListView):
     """ View for displaying list of Part objects
     """
+
     model = Part
     template_name = 'part/category.html'
     context_object_name = 'parts'
 
+    role_required = 'part.view'
```
SchrodingersGat, line 54: @eeintech -  Views can simply inherit `InvenTreeRoleMixin` and set the `role_required` attribute

eeintech:
I'm pleased to see the "Role" concept brought back in. When I initially introduced a model I had this sort of implementation in mind, but maybe an abstract "model" as you did here was better in the end.

So do you need help with updating templates and views? Which ones are on the TODO list?

InvenTree/part/templates/part/build.html
eeintech: Is the `user_roles` evaluated on each request? Or how is the context updated when user makes a ruleset change?

SchrodingersGat:
> So do you need help with updating templates and views? Which ones are on the TODO list?

I'm pretty sure I've covered off these views:

- Part
- Build
- Purchase Order
- Sales Order

But none of the Stock views / templates have been updated yet. Feel free to implement!

PR #1025, opened 2020-10-05
Fixed Group model permissions redundant saves

eeintech:
Now only updating all permissions on Group creation and after Rulesets are saved.


SchrodingersGat:
Thanks @eeintech 👍

PR #1026, opened 2020-10-05
Added check for multiple groups assigned to user

eeintech:
Added `clean` method to `InvenTreeGroupAdminForm` form which validates that added Users don't belong to any group other than the current one (else permissions from a second group would overwrite the first one)

Edit: I initially committed too much so I did a force push to revert tracking to master


SchrodingersGat:
I don't think I agree with the logic here - I can certainly see the benefit of assigning users to multiple groups. Certainly I would like to retain the flexibility of "mixing and matching" the groups.

eeintech:
Ok so let's say "John" is part of _Group A_ with ruleset **Part: change=False** and also member of _Group B_ with **Part: change=True**, which permissions do you follow? What Group is "right"?

I personally want to know if this happens, I don't want to _think_ someone is given certain limited rights but didn't realize another Group would give this user more rights.

SchrodingersGat:
I see the purpose of group roles as being "additive". let's say you have have the following groups:

**Readers** - This group can view everything in the database, but not *edit* anything
**Part Editors** - This group can view / edit / create / delete parts, but not see anything else

You could then add a user to both groups and they would get the permissions from both roles.

The alternative approach would be to create separate groups such as:

- Readers
- PartEditors
- ReadersAndPartEditors

For every combination you now need a separate group, rather than stacking them together.

The role permissions are always additive - a negative won't cancel out a positive, if that makes sense.

I definitely think there is a use case for this

eeintech:
Alright, then I will find a way to warn when a user is added to multiple groups instead, without this I will get lost in who belongs where and I want the option to know my users are only in a certain group.

Why is that? Because for me "Group" is not the group from Django but the "Role" model I wanted to introduce.
A fellow coworker, even though they usually take multiple hats, will own only one role (basically their work title).
I believe that with this rule in place, user management's load will be lower.

eeintech:
@SchrodingersGat Changed from validation error to warning message when Group instance is saved:

image

eeintech:
@SchrodingersGat I'd be keen to have this PR merged (it does not change behavior of Group admin management)

eeintech:
@SchrodingersGat Any luck having a glimpse at this PR?

SchrodingersGat:
Sorry, I missed the updates here.

I'm happy to merge this as it doesn't change the behaviour, but I can appreciate that the warning messages are very useful to have. 

I did not know that you could do such things in the admin interface - very handy to know 👍

eeintech:
Nice thanks! I've also discovered about the admin's messaging system, it's quite handy.

PR #1030, opened 2020-10-06
Stock permissions

eeintech:
Added role permissions on Stock views and templates, tested with few roles (view only, change, delete/staff).


SchrodingersGat:
This looks pretty good @eeintech thanks for the quick addition 👍 We are now well on the way to support good user segregation!

PR #1031, opened 2020-10-06
Fixed context permissions for superuser with no group assigned

eeintech:
The `roles` context dictionary was returned empty because of absence of group assigned


SchrodingersGat:
Good catch :)

PR #1038, opened 2020-10-09
fix error when submitting some non-dict barcodes

mpdgraev:
An assumption was made that json.loads() will always return an object with a .keys() function. However this is only true for a dict.

I noticed the issue by trying to search or add a 'barcode' that only contained numbers, which was parsed to int by json.loads(), and this then gave an error in the logs (line 54). All this happened silently for the user, without any indication of why this was not allowed on the UI side.

As I don't see why arbitrary strings like 'foobar' would be correct 'barcodes', while '123456789' or '["foo", "bar"]' wouldn't be, I think just handling all these strings in the same way (hashing them and using them as UID, unless another plugin picks them up) seems sensible.

### DIFF
--- a/InvenTree/barcode/plugins/inventree_barcode.py
+++ b/InvenTree/barcode/plugins/inventree_barcode.py
@@ -42,6 +42,8 @@ class InvenTreeBarcodePlugin(BarcodePlugin):
         elif type(self.data) is str:
             try:
                 self.data = json.loads(self.data)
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
Hey @mpdgraev I think your solution here is appropriate. I'm just excited that someone is making use of the barcode features! Please LMK if there are any other issues you come across.

PR #1043, opened 2020-10-12
[Company] Allow duplicate names - Unique name/email pair

### LINKED ISSUE #1040: Customer company name uniqueness issue for B2C application
Due to the reuse of company for customers, customer 'names' must be unique.  This is not a reasonable expectation for business selling direct to consumers as customers.   The unique key for a company should be relaxed or changed.  The only unique part for a 'customer' should be the email address.

eeintech:
Fixes #1040.

* Company `name` field _unique_ constraint was **removed**
* Company `email` field _null_ constraint was set to `True` (convert empty string to `Null` value)
* ~~Company `email` field _unique_ constraint was set to `True` (check for uniqueness)~~
* Added custom method to migration file to convert `email` field empty string value to `Null` value
* Added `UniqueConstraint` on name/email pair (instead of unique constraint on email only)

image


SchrodingersGat:
Now that the "name" field does not have to be unique, but the email does, there still could be a situation where multiple company objects exist with the same name and a blank email.

So, I would suggest that we add a requirement that if the email is empty, then the name must be unique.

It may be as simple as adding:

```python
class Meta:
    unique_together = [
        ('name', 'email'),
    ]
```

(but this requires testing)

eeintech:
Thanks @SchrodingersGat I'll take another shot with this in mind. Seems like the better way is to create a `UniqueConstraint` as `unique_together` could be [deprecated in the future](https://docs.djangoproject.com/en/3.1/ref/models/options/#unique-together).

SchrodingersGat:
@eeintech I'm pretty happy with this on my side, I don't see any issues with this as it stands. Thanks for addressing the issue 👍

PR #1047, opened 2020-10-15
Added IPN to Supplier Part string representation (global)

### LINKED ISSUE #616: Add IPN as a sort option when building a purchase order
When I was adding items to my PO I noticed you can only search by the 'Supplier SKU' since all my parts have either the MFG# or my IPN on them it would be awesome if you could add those part numbers in the search when adding to PO.

Some suppliers SKUs don't always make sense and it would be easier if you could search for it by IPN.

eeintech:
Also improved Supplier Part templates and fixed 'Order Part' button in supplier_part_orders

Fixes #616

### DIFF
--- a/InvenTree/company/models.py
+++ b/InvenTree/company/models.py
@@ -427,8 +427,9 @@ class SupplierPart(models.Model):
 
         if self.part.IPN:
             s += f'{self.part.IPN}'
+            s += ' | '
 
-        s += f' | {self.supplier.name} | {self.SKU}'
+        s += f'{self.supplier.name} | {self.SKU}'
 
         if self.manufacturer_string:
             s = s + ' | ' + self.manufacturer_string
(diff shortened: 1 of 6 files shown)


InvenTree/company/models.py
```diff
@@ -423,12 +423,15 @@ def pretty_name(self):
...
     def __str__(self):
-        s = "{supplier} ({sku})".format(
-            sku=self.SKU,
-            supplier=self.supplier.name)
+        s = ''
+
+        if self.part.IPN:
+            s += f'{self.part.IPN}'
+
+        s += f' | {self.supplier.name} | {self.SKU}'
```
SchrodingersGat, line 431: In the case where the part does *not* have an IPN, this will result in a leading space and pipe character.

I think the following approach would be cleaner:

```python
if self.part.IPN:
    s += f'{self.part.IPN}'
    s += ' | '

s += f'{self.supplier.name} | {self.SKU}`
```

eeintech: Yes right, much better indeed! Just updated the PR! :zap:

SchrodingersGat:
Thanks for the fixes :)

PR #1048, opened 2020-10-16
Related Parts

### LINKED ISSUE #652: New Feature Request
I have been adding parts to InvenTree.  I use Molex connectors often and I think it would be nice to be able to link two or more parts together.  Like for example the Molex connector with the crimp pin.

eeintech:
This added feature, if approved, fixes #652 

### What is Related Parts

Denotes a relationship between two parts, when users want to show their usage is "related" to another part or simply emphasize a link between two parts.

### Implementation

* New `PartRelated` model/table to store relationships between parts (requires migration)
* New `Related` tab shown in Part detail page
* Ability to add relationships between parts through both main and admin interfaces
* Can only manage relationship if user has `change` permission on `Part` ruleset

### Template View

image


eeintech:
@SchrodingersGat Feedback is positive (see #652), wanna take a look and merge?

SchrodingersGat:
@eeintech I really like the implementation here. When I was originally thinking of fixing #652 there were a number of potential issues I thought of, but you've addressed them all! Very impressive instinct into how the code base works 👍 

Thanks for the contribution!

PR #1056, opened 2020-10-19
History tracking

### LINKED ISSUE #377: Keep a complete history of records
Use django-simple-history

https://django-simple-history.readthedocs.io/en/2.7.2/

### LINKED ISSUE #537: User History
Please provide a user history or global history that shows things such as added items, stock adjustments, Modified locations and who made the changes.

### LINKED ISSUE #547: Show transaction history for user
Add a page that displays all the historical actions performed by the current user (or for any user).

- Stock transaction history
- Part history
- etc etc

eeintech:
If approved, fixes #547 fixes #537 fixes #377 

Proposal is to use the [django-simple-history](https://github.com/jazzband/django-simple-history) library (actively maintained and great documentation).

For testing, I added historical data on the `Part` model. Here are my observation in a pros/cons list.

## Pros
- Non-intrusive: **django-simple-history** creates a separate table for each model for storing historical data
- Built-in admin access: The history of an instance can be added to the admin interface and be seen via a "History" button
- Transparency: for each model tracked, all changes are recorded and automatically linked to the corresponding `User`
- Version control for models: Any instance can be reverted to any previous state and [diff can be executed](https://django-simple-history.readthedocs.io/en/latest/history_diffing.html) to understand what changed

## Cons
- Extra database load:
  - Each model would have it's own `historical` table (that's a lot of tables :open_mouth: )
  - Long list of parameters added each time a change happen on an instance
- Hard-coded and enabled by default for all InvenTree users: that's mostly because it uses the admin interface (unless there is a way to subclass a configurable admin model?)
- Admin interface unfortunately does not display diffs (would have to be ran through command line or added as InvenTree feature) and each import-export based model would have to use [the mixin](https://django-import-export.readthedocs.io/en/latest/api_admin.html#import_export.admin.ImportExportMixin) instead.

## Final words
I personally think the pros outweighs the cons. But I could understand not everybody want to have it enabled.
In that case, tracking would need to be enabled/disabled via the `config.yaml` (or even better would be the Admin interface). 

Edit: added more related open issues


eeintech:
@SchrodingersGat What's your take, should I keep pushing for this?

eeintech:
This would also fix: #537 #377

SchrodingersGat:
I haven't had time to ponder fully on this one yet, sorry!

eeintech:
@SchrodingersGat I'm thinking of closing this PR, I don't think it is really useful at this time to track history of instances (our team do not really need it and I did not see any other recent inquiries) and I would prefer to focus on other development efforts.

I also stumbled upon [django-reversion](https://github.com/etianen/django-reversion) and I like it more than the proposed [django-simple-history](https://github.com/jazzband/django-simple-history) package, because it full-fledged revision control system for instances, not only history, which means one can revert the state of an instance to previous state. In other words, Git for InvenTree :smile:

SchrodingersGat:
django-revision looks interesting! But implementing any revision system is going to be a **huge** amount of work. I'm happy for you to close this and we can re-assess this at a later date :)

eeintech:
Ok, that's what I was thinking too :+1:

PR #1021, opened 2020-10-05
Updates to build system

wolffam:
1. Remove ability to modify quantity of build outputs once the build has been created

Issue #572, opened 2019-10-31
Partial build completion


SchrodingersGat:
Add the option to partially complete a build - e.g. make 10 out of 25 units, leaving 15 still to be made...

Issue #292, opened 2019-05-10
Build allocation ideas


SchrodingersGat:
Show list of 'remaining' allocations - this quickly shows what you need to buy or build or whatever.

Filter 'allowable locations' to take stock from. e.g. only allow stock items from /work/office/* or from /* for all stock items

jnewlands:
More fine grained process of what a 'build allocation' is - currently you put parts in (and can overallocate if you want), but then the parts disappear. Need a way of managing:

- Getting a full reel from stock, putting it in a box. Can't allocate all of it to the build or it will disappear. Manually moving it from one location to the temporary location currently doesn't have any structure (e.g. a page prefilled against the BOM).
- Filtering stock items in the build allocation by an existing location (so that e.g. an existing parts kit can be pulled from).
- Receiving parts back from a supplier - some mixture of surplus parts and the output of the build. May want to explore approximate quantities - a reel having some rough minimum quantity is often all the info that is needed.
- A build might create parts that then need to be serialised, but briefly exist in a state without one.

This may look like a more generic 'location to location' transfer that can be prefilled against certain models (e.g. a BOM or a salesorder).

SchrodingersGat:
https://github.com/inventree/InvenTree/compare/master...SchrodingersGat:take-from?expand=1 adds the ability to specify where parts are taken from for a build.

PR #1063, opened 2020-10-20
Build system updates

SchrodingersGat:
This PR represents a major change to the way that builds operate:

Stock gets allocated against a build output (which is a StockItem). There can be multiple build outputs per build, allowing for partial build completion, and also allowing for tracked stock item to be installed inside a particular "output" of the build process.

- Fixes https://github.com/inventree/InvenTree/issues/977
- Fixes https://github.com/inventree/InvenTree/issues/572
- Fixes https://github.com/inventree/InvenTree/issues/292
- Closes https://github.com/inventree/InvenTree/pull/1021


eeintech:
@SchrodingersGat I haven't tested this yet but I've noticed that you had removed the `post_save` method in ajax views, which broke #1048 after it was merged as it relied on this method to save the model... What was the reason for removing it? `pre_save` was also removed it seems.

SchrodingersGat:
See comments in #1099

Issue #1068, opened 2020-10-23
Code formatting style guide


SchrodingersGat:
Look at options for checking (and enforcing) coding style guide

- UPPERCASE_FOR_SOME_STUFF
- e.g. camel case (or other options) for variables / functions / classes
- tools for automated checking of code
- Add "contributing.md" file

jnewlands:
Popular formatters

- [black](https://github.com/psf/black)
- [yapf](https://github.com/google/yapf)

eeintech:
https://github.com/inventree/InvenTree/pull/651

I'm not a big fan of black, too opinionated IMO.
yapf seems more flexible, have to give it a shot!

eeintech:
@SchrodingersGat @jnewlands I played a bit with `yapf` over on my fork: https://github.com/eeintech/InvenTree/tree/yapf_test

I've ran the yapf formatter with standard settings on all Python files in the `part/` folder and `flake8` only reported 3 issues.
The concern is that among those 3 issues, 2 of them are conflicts between `flake8` rules and `yapf` formatting (the last one was a  `flake8` internal conflict). But I guess this will be the case with any code formatter out there, unless there is a formatter out there is developed by the same team as `flake8`, not sure...

I do like the [flexible settings](https://github.com/google/yapf#knobs) of `yapf` and would prefer it over `black`.

I think with with a couple hours, InvenTree's entire code base could be formatted using `yapf` and passing CI.
What do you think, worth the push into this direction?

PR #1091, opened 2020-10-29
Yapf

eeintech:
[Yapf formatter](https://github.com/google/yapf) tests!

### DIFF
--- a/InvenTree/part/admin.py
+++ b/InvenTree/part/admin.py
@@ -25,28 +25,42 @@ class PartResource(ModelResource):
     """ Class for managing Part data import/export """
 
     # ForeignKey fields
-    category = Field(attribute='category', widget=widgets.ForeignKeyWidget(PartCategory))
-    
-    default_location = Field(attribute='default_location', widget=widgets.ForeignKeyWidget(StockLocation))
+    category = Field(attribute='category',
+                     widget=widgets.ForeignKeyWidget(PartCategory))
 
-    default_supplier = Field(attribute='default_supplier', widget=widgets.ForeignKeyWidget(SupplierPart))
+    default_location = Field(attribute='default_location',
+                             widget=widgets.ForeignKeyWidget(StockLocation))
+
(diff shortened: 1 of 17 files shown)


SchrodingersGat:
It is intriguing that the formatter seems to be doing opposite things in different instances:

e.g. list unrolling

image

vs

image

@eeintech can you comment on this?

eeintech:
Ummm the changes looked "clean" for the ones I saw, but this one is bizarre...
I'll check more into yapf tomorrow to understand why would this happen.

eeintech:
@SchrodingersGat Looks like the termination comma in the list matters:
* The first list is comma-terminated -> list is unrolled
* The second list is not -> list is rolled back

This behavior can be changed with the following option:
image

Disabling it (`disable_ending_comma_heuristic = true`) would re-roll all the lists, independently if they are comma-terminated or not.

I do prefer them **unrolled**, it creates much cleaner and visually obvious diffs, which means we would need to add a termination comma on all lists and keep that option disabled.

SchrodingersGat:
Ah, at least there was a reason!

Is the idea here that yapf will just point out errors, or it will actually format the code on a check-in?

Also, I assume this is going on top of the flake8 checking which is already in place - that checks a lot of things over and above style issues.

eeintech:
Yes yapf formats the code when called on a file, it won't point out errors unless maybe when it fails to run.

There is a bit of conflict between flake8 and yapf as I mentioned here: https://github.com/inventree/InvenTree/issues/1068#issuecomment-719064834

Maybe a code formatter tool was not the real idea behind that #1068 issue, however it was proposed and I wanted to get a look at what yapf is capable of.

eeintech:
@SchrodingersGat Let's put this on the back burner for the time being, I'm actually pretty happy with the format of the code, you did a great job keeping it clean and I'm trying too. Unless more users start contributing to InvenTree more frequently I would say `yapf` or any other code formatter tools is not really needed right now.

PR #1092, opened 2020-10-29
Copy bom

SchrodingersGat:
Add function to copy a Bill of Materials from a template part


SchrodingersGat:
@jnewlands your suggestion is my implementation

PR #1096, opened 2020-10-30
Categories parameter templates

### LINKED ISSUE #1083: Assign default Parameter Templates by Category
For example, Category: Capacitors might be assigned (e.g.) Capacitance in Farads, Voltage in Volts, etc. When a new Part is created in that Category, it would be lovely to either 

a) automatically add a new Parameter for each Default Template assigned to Parts in this Category, with value = NULL unless otherwise assigned, or

b) at Part creation, offer to add aforementioned Parameters but only do so for those Parameters for which the user supplies a value. (probably the most flexible method)

From the API, a method might be added to report all of the default Parameter Templates assigned to a Category, so that they may be added programmatically, and another to assign a default Parameter Template to a new or existing Category.  When creating a Part and assigning a Category, the remote system can then retrieve all of the expected Parameters for that Category and assign values (or not) as appropriate.

eeintech:
If approved, fixes #1083 

## Features
* User can now setup a list of parameter templates for each (or all) part category(ies) in InvenTree settings
* During part creation, part parameters are automatically created using the list of parameter templates from the parent category it belongs to (if option is enabled)

## Screenshots

* Select category

image

* Add parameter template

image

* Control default behavior within part settings

image

* Control instance behavior during part creation

image


SchrodingersGat:
I like the look of what you have implemented here - putting it into the settings view makes a lot of sense!

Can you please address conflicts here, and I'll pull it down to have a look.

eeintech:
@SchrodingersGat Solved the conflict 😃

InvenTree/part/views.py
```diff
@@ -688,6 +693,18 @@ def post(self, request, *args, **kwargs):
             except AttributeError:
                 pass
 
+            # Create part parameters
```
SchrodingersGat, line 696: I think it would be a cleaner approach to use the save method of the Part model to add default parameter values, rather than manually performing the action here.

That way, the parameters will also be added if a Part is created via the API or the admin interface, or wherever.

In the save() method for the Part object, you can check if the object is being created (first call to save) or updated (subsequent calls)

SchrodingersGat:
How does this approach handle multi-level categories?

Let's say you have the following category structure

`Electronics -> Passives -> Resistors`

with the following category parameter templates:

**Electronics**
- Voltage
- Package

**Passives**
- Tolerance

**Resistors**
- Resistance

If you create a new part under *Resistors*, then I believe the part should be created with default parameters for:

- Voltage
- Package
- Tolerance
- Resistance

That is to say, when a new part is created, it traverses up the entire category tree and applies the category parameter templates from each category.

What are your thoughts on this?

eeintech:
@SchrodingersGat Yes I did work on this today and updated with the behavior you mentioned, I added an additional checkbox for traversing parent categories (eg. ancestors), see updated screenshot in original post.

InvenTree/part/views.py
eeintech: @SchrodingersGat This is a smart idea, I like it.
I don't have time to do this now but I'll work on it asap.

Thanks for taking a look at this :smile:

SchrodingersGat:
This looks like it is working pretty well, although I do not think I agree with the concept of copying a parameter template to other categories on the same level or "all" categories. 

image

What is the justification here?

eeintech:
@SchrodingersGat See https://github.com/inventree/InvenTree/issues/1083#issuecomment-721876577

#### Same-level

There are users (like me) who like to share parameters across multiple categories, all my top-level categories have at least `symbol`, `footprint` and `value` parameters. I wouldn't want to have to repeatedly had them one-by-one to each top level category.

#### All categories

Maybe this one is a stretch, I can see the appeal for people to link a template to all categories, for instance when you have a reference field that is needed across all. Of course it could be inherited from your top-level categories (if set with the "same-level" option) but some people may prefer to have it at the subcategory level too.

eeintech:
@SchrodingersGat I can remove the "Add to all categories" option if you really wish so but I would definitely have great use for the "same-level" one. But I personally think it's okay to leave that door open to users, let me know.

SchrodingersGat:
I think that entertaining the concept of "copy to all" runs against the idea of having hierarchical category parameters. If you have a use case where you want every category to have a copy of the same parameter, then you should probably just add the parameter to a higher-level category and it will flow down.

If you remove the "add to all categories" option I'm happy to merge this though :)

eeintech:
> I think that entertaining the concept of "copy to all" runs against the idea of having hierarchical category parameters. If you have a use case where you want every category to have a copy of the same parameter, then you should probably just add the parameter to a higher-level category and it will flow down.

I agree with you on the general concept, and also disagree because of user preferences. Mostly because there is the option when adding a part to use or not the hierarchical structure:

image

This leave it up to the user to decide to:
* use hierarchical parameters
* or use a strict "per-category" list of parameters.

The preferred method would be the first option and should be the one recommended.
However, users could also want to get a global parameter into their non-hierarchical setup, if they want to.

I'm happy to remove it, but I personally don't think it hurts.

SchrodingersGat:
I suppose if it is disabled by default, and well-documented, it's ok. I personally don't think I would use it but some users might...

Are you happy for this to be merged in?

eeintech:
Yup ready to merge. I did bunch of testing myself and did not encounter any issue with the current implementation.

SchrodingersGat:
Ok, let's go then. Thanks for all your work on this :)

rco3:
Thank you both very much.

PR #1099, opened 2020-11-03
Fix PartRelated (based on updated AjaxCreateView class)

eeintech:
I think we should keep it, this method can be handy and PartRelated view relies on it.

### DIFF
--- a/InvenTree/InvenTree/views.py
+++ b/InvenTree/InvenTree/views.py
@@ -333,6 +333,12 @@ class AjaxCreateView(AjaxMixin, CreateView):
     - Handles form validation via AJAX POST requests
     """
 
+    def post_save(self, **kwargs):
+        """
+        Hook for doing something with the created object after it is saved
+        """
+        pass
+
     def get(self, request, *args, **kwargs):
         """ Creates form with initial data, and renders JSON response """
 
@@ -386,6 +392,7 @@ class AjaxCreateView(AjaxMixin, CreateView):
 
             # Save the object to the database
             self.object = self.save(self.form)
+            self.post_save()
 
             # Return the PK of the newly-created object
             data['pk'] = self.object.pk
--- a/InvenTree/InvenTree/views.py
+++ b/InvenTree/InvenTree/views.py
@@ -333,12 +333,6 @@ class AjaxCreateView(AjaxMixin, CreateView):
     - Handles form validation via AJAX POST requests
     """
 
-    def post_save(self, **kwargs):
-        """
-        Hook for doing something with the created object after it is saved
-        """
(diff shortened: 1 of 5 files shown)


SchrodingersGat:
#1063 refactored the way that forms are handled in a major way. One part of this was "simplifying" the form saving hooks.

Prior to this, the form would call (after validation):

- pre_save()
- save()
- post_save()

Now, this has been simplified to just `save()` - which can easily be overridden by a custom subclass.

The default implementation of save() is pretty simple:

```python
    def save(self, form):
        """
        Method for actually saving the form to the database.
        Default implementation is very simple,
        but can be overridden if required.
        """

        self.object = form.save()

        return self.object
```
I should have spotted that you had a `post_save` call in your PR, missed that one sorry.

Here's what it is currently:

```python
    def post_save(self):
        """ Save PartRelated model (POST method does not) """

        form = self.get_form()

        if form.is_valid():
            part_1 = form.cleaned_data['part_1']
            part_2 = form.cleaned_data['part_2']

            PartRelated.create(part_1, part_2)
```

With the new system, it should be:

```python
    def save(self, form, **kwargs):
        """ Save PartRelated model """

        # Note: form.is_valid() is already true by the time save() is called!
        # if form.is_valid():
  
        # Also, you could run some custom form validation before getting to save(),
        # by overriding the new 'validate' method
        part_1 = form.cleaned_data['part_1']
        part_2 = form.cleaned_data['part_2']

        PartRelated.create(part_1, part_2)
```

SchrodingersGat:
The refactoring is much simpler, I think. And importantly, it adds the ability to do custom form validation, by overriding the `validate` method and using the `form.add_error` call.

Previously, extra form validation had to be done in a custom `post` method which then required re-implementation of all the other stuff that happened in the top-level `post` call.

If you can adjust this PR to use the new `save` method I'll merge it in 👍

SchrodingersGat:
Actually, looking at this a bit closer, is there a need to even implement a custom save call here? The default `save` call just takes the form data and creates a database object, and the only thing you are doing is creating a new `PartRelated` object with the fields `part_1` and `part_2` (which the form validation would do anyway)...

I think perhaps that the custom `save` method could be deleted entirely here?

eeintech:
Oh I see, I missed the fact that the `save` method was now the way to go. I'll check if it is actually needed (like you mentioned) and refactor this PR as necessary.

eeintech:
@SchrodingersGat Alright, it's fixed the right way :smile:

SchrodingersGat:
Yeah, that looks a lot cleaner. Thanks for the fix 👍

PR #1102, opened 2020-11-03
Updated to latest version of coverage and coveralls

SchrodingersGat:
Checking if this fixes the coveralls issues


SchrodingersGat:
Yay!

Thanks for the tip on using `coveralls` rather than `python-coveralls` @eeintech

eeintech:
@SchrodingersGat Glad to know it's fixed now!

PR #1111, opened 2020-11-05
Disable crispy form errors

eeintech:
## Enabled (old)

image

## Disabled (new)

image

### DIFF
--- a/InvenTree/InvenTree/forms.py
+++ b/InvenTree/InvenTree/forms.py
@@ -27,6 +27,7 @@ class HelperForm(forms.ModelForm):
         self.helper = FormHelper()
 
         self.helper.form_tag = False
+        self.helper.form_show_errors = False
 
         """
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@eeintech this PR (and #1115) have actually introduced a new bug...

If a form has a field which is *not* in the referenced model, and that field has an error, then the error message is not rendered.

Commenting out the following line fixes the issue on my end:

```python
# self.helper.form_show_errors = False
```

I have provided a further fix in https://github.com/inventree/InvenTree/pull/1147

eeintech:
Ok thanks, I'll take a look because there were multiple instances where those errors would show up as duplicate...

SchrodingersGat:
I *think* I have addressed this by altering the form template itself.. But let me know if you can see any further issues.

eeintech:
@SchrodingersGat Looks good to me now, thanks!

Issue #1125, opened 2020-11-12
Add unit test for "global settings" forms / views


SchrodingersGat:
Add some unit testing for the global settings view, which is a generic view for editing global settings.

Ref: https://github.com/inventree/InvenTree/pull/1123 (some issues were found with this form)

- Check that boolean values work well (both True and False)
- Check that multiple-choice options work
- Check that string values work

PR #1127, opened 2020-11-13
Extra unit testing for settings forms / views

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/1125

### DIFF
--- a/InvenTree/InvenTree/urls.py
+++ b/InvenTree/InvenTree/urls.py
@@ -82,7 +82,7 @@ settings_urls = [
     url(r'^purchase-order/?', SettingsView.as_view(template_name='InvenTree/settings/po.html'), name='settings-po'),
     url(r'^sales-order/?', SettingsView.as_view(template_name='InvenTree/settings/so.html'), name='settings-so'),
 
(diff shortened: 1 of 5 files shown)


eeintech:
You may want to try `TransactionTestCase` for the `SettingsViewTest` class. I think it has to do with `TestCase` wrapping all tests into atomic blocks and ignoring method exceptions/errors.

I had to do this for `TestCategoryTemplates` in [part/test_param.py](https://github.com/inventree/InvenTree/blob/0bb8c0a1e34218149ff0f035d7976684233ecf22/InvenTree/part/test_param.py#L50).
The Part `save` method had this [`IntegrityError` exception](https://github.com/inventree/InvenTree/blob/0bb8c0a1e34218149ff0f035d7976684233ecf22/InvenTree/part/models.py#L383) when parameter already exists, which works fine when run live but was causing the same test case headache.

References:
https://docs.djangoproject.com/en/3.0/topics/testing/tools/#transactiontestcase
https://stackoverflow.com/questions/44450533/difference-between-testcase-and-transactiontestcase-classes-in-django-test

Issue #1134, opened 2020-11-18
Updating to latest: fails due to new currency updates?


rco3:
Running migrations:
  Applying company.0026_auto_20201110_1011...Updating currency references for SupplierPriceBreak model...
Traceback (most recent call last):
  File "/opt/invenv/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql)
psycopg2.errors.UndefinedColumn: column "USD" does not exist
LINE 1: ...PDATE part_supplierpricebreak set price_currency= "USD" wher...

I don't know how to recover from this.  When I updated my install at home, I had to run invoke update and tell it I wanted USD - twice.  This time I wasn't asked, it just failed.

rco3:
Upon further examination, I find that the Inventree install is still at least somewhat operational despite the failed upgrade.  I determined that I did not have any currencies set at all, and promptly set USD as the base currency.  invoke update fails again with the same error.

eeintech:
Weird, I had no problem migrating on both SQLite and PostgreSQL...
`psycopg2` line indicates that you are using PostgreSQL database and Python connector, is that correct?

Try this:
```
$ cd path/to/InvenTree
$ ./manage.py migrate company 0025_auto_20201110_1001
```

and if succeeds, re-run `invoke update`.

rco3:
Hi, @eeintech, thanks for looking at this.

Sadly, that didn't fix it. Executing " ./manage.py migrate company 0025_auto_20201110_1001 " gave me the same error.

Running migrations:
  Applying company.0026_auto_20201110_1011...Updating currency references for SupplierPriceBreak model...
Traceback (most recent call last):
  File "/opt/invenv/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql)
psycopg2.errors.UndefinedColumn: column "USD" does not exist
LINE 1: ...PDATE part_supplierpricebreak set price_currency= "USD" wher...

I also tried running the migration that failed (0026), directly from the command line, and failed that way as well.

eeintech:
@rco3 Maybe try to reset migrations for the company app?
https://simpleisbetterthancomplex.com/tutorial/2016/07/26/how-to-reset-migrations.html

rco3:
That tutorial says to start after all migrations are applied:
"If you see the message:

No changes detected
You are good to go."

[...]

PR #1135, opened 2020-11-18
Fixes for custom SQL queries

SchrodingersGat:
- Don't use double quotes!
- NO NO NO!
- Single quotes only

Fixes https://github.com/inventree/InvenTree/issues/1134


eeintech:
Haha!

[For reference](https://stackoverflow.com/a/41396319/12794913):
"Double quotes are for names of tables or fields. Sometimes You can omit them. The single quotes are for string constants. This is the SQL standard. In the verbose form your query look like this:

``` sql
select * from "employee" where "employee_name"='elina';
```

PR #1136, opened 2020-11-19
Add framework for "health checks"

SchrodingersGat:


amishHammer:
There is a bug with this:  it will fire multiple times for the same request (not sure why but other projects work around this).

This commit takes an approach similar to others to make sure we don't do the checks more than once:  https://github.com/inventree/InvenTree/pull/1113/commits/37fc320457cbb6b033058b4121f9604a5d003ab3

SchrodingersGat:
Similar patch applied here - https://github.com/inventree/InvenTree/pull/1140

Issue #656, opened 2020-02-18
part ownership and lending


Chrescht:
One way to use InvenTree is to share parts among a group of people (friends, ..).
This means a part 
- has to have ownership,
- can be lent to a user, with a start and end date

Is somebody else interested in this?

a7hybnj2:
I don’t currently need this but could foresee it being useful in the future.

Would the “lent to a user” need an inventree account or would they just be added as a “client/supplier/contact”?

Chrescht:
> 
> 
> I don’t currently need this but could foresee it being useful in the future.
> 
> Would the “lent to a user” need an inventree account or would they just be added as a “client/supplier/contact”?

They need an InvenTree account since they can also add they own stuff, be parts owner, and lend parts.

chschlue:
I don't really get the "lend" part.
Isn't lending parts akin to lending firewood?

a7hybnj2:
@chschlue I think its more for if you track things like your multimeters and dso and whatnots. As an asset management system.

SchrodingersGat:
You could pretty easily hack this in by creating a StockLocation tree like:

``Stock/Friends/Bob`` which then contains all the things that Bob has selfishly never returned to you. This also allows automated history tracking of when something was transferred to Bob. The ``Stock/Friends`` location would then show all the items you have lent out.

Chrescht:
> You could pretty easily hack this in by creating a StockLocation tree like:
> 
> `Stock/Friends/Bob` which then contains all the things that Bob has selfishly never returned to you. This also allows automated history tracking of when something was transferred to Bob. The `Stock/Friends` location would then show all the items you have lent out.
> 
> i.e. if you instead encode a "who" as a "where" you should get all the features you are looking for, without scope-creeping the software :)

This is one way it could work, but I was more thinking about a shared stock management, where some stuff is owned by me, some by my friend Bob. So Stock/Friends/Bob doesn't really work unless we have the parts tagged with an owner.

Then there are _parts_ like firewood as @chschlue pointed out, these are _commonly owned_ or have _no owner_, just a location, an can be consumed. Nails, screws, and paint are more examples.

SchrodingersGat:
Ok, what if in addition to having a "virtual" location like "lent to friends", each Part can have an (optional) "owner" field. In a factory or business context this would be the person responsible for that part, for example.

[...]

PR #1155, opened 2020-12-01
Stock Locations and Items Ownership

### LINKED ISSUE #1129: Stock location linked to user group
Hi!

At [fablab bcn](https://fablabbcn.org/) we are running an instance of [Inventree](https://inventory.fablabbcn.org/) - so far it's working great and we really appreciate the fantastic work.
Since we have various projects with different inventory locations, we wonder if there's the possibility (already implemented or not) to have a stock location linked to an user group. This facilitates the case in which an organization like ours has various projects in parallel with different teams, but you don't want them to have edit access to other stock locations, but yes view access. 

Would this be a potential enhancement? If not, we can always run different instances. 
Thanks!

eeintech:
If approved, fixes #1129 (and maybe #656)

## Global
* Added `owner` fields to `StockLocation` (ForeignKey to Django `Group`) and `StockItem` (ForeignKey to Django `User`)
* Added global setting for ownership control of stock locations and items (under "Stock" tab): enabling this setting enforces ownership for add/edit operation on stock locations/items

image

## Web Interface/Templates
* Only the stock item owner (user) can see "add/edit" item
* If user is member of stock location ownership (group), user can "add/edit" location

## Forms/Views
* The `Owner` field is shown in both stock location and item forms only **when the global setting is enabled** (otherwise hidden)
* Adding a stock location requires:
  * the owner field (group) to be set
  * if the location has a parent, the owner is automatically set to the parent's owner and can't be changed
* When editing a stock location's owner, all children locations are updated with the same owner
* Adding a stock item requires:
  * the owner field (user) to be set
  * If the location is set, only a user associated to location ownership (group) can be selected as owner


SchrodingersGat:
Hi @eeintech I've finally had a chance to look at your implementation here. In general, it looks pretty nice!

A couple of fine points to be clarified:

### Group vs User

I think it makes sense to allow some flexibility in determining "who" owns or controls a StockLocation or StockItem. I would like to see the ability to assign either a StockLocation or StockItem to a user *or* a group.

- Assign StockItem to group or single user
- Assign StockLocation to group or single user
- Cannot assign to both a user and a group

### Anonymous Items

In your current implementation, if "ownership" is enabled, can locations and/or items exist without an owner? Your description above seems to indicate that if ownership is enabled, then stock items and locations *must* have an owner assigned. Is this the case? I think that un-owned items and locations must still be supported in this case.

eeintech:
@SchrodingersGat Thanks for having a look :+1:

Please check out the following table:

image

* `UC` stands for "Use-Case" (only for referencing each change)
* Red text is when a change was made by user
* Green text is when a location/item owner is automatically updated in database

#### Worth-to-mention notes
* Changing the location's owner forces changes of all items' owners at that location (UC#2, 4 and 6)
  * Unless the new location's owner is a **group** and the item's owner is part of that group (UC#10)
* Changing the item's location forces change of owner (UC#7)
* There is no memory from previous items' owners in the case stock location's owner changes (UC#8)

Does it cover all your suggestions?
If/when you approve, I will update this PR to cover all those use-cases.

SchrodingersGat:
Thanks for putting this table together. I believe this covers the use cases I can think of.

How do you plan to implement the dual group/user ownership rules?

eeintech:
My plan was to have two separate fields in the models (one for group owner, one for user owner) and only one of them can be set at a given time. I was thinking to use the `UniqueConstraint` ([ref](https://docs.djangoproject.com/en/3.1/ref/models/constraints/#uniqueconstraint)) along with a conditional check that either one if set, but not both at the same time (XOR-type of logic). That way, whichever model owner field (group or user) is set is the actual owner.

That would also mean logic in the form to gather both users and groups in the same "owner" field and in the model/view to handle the switch of owner.

What do you think?

SchrodingersGat:
That sounds like a very sensible approach

eeintech:
@SchrodingersGat I think this pull request is ready for review (round 2).

I have changed the default behavior for parts without owners when stock ownership is enabled, both locations and items are now only editable by an **Admin** user.
The reason for this is to enforce that each location/item has its owner set, and not leave the door open to malicious users wanting to play around because the admin/InvenTree manager hasn't set the owner yet.
Does it make sense?

Here is the updated table of use-cases (I did traverse most of it manually through the web interface, and the test cases also try to repoduce those use-cases):

image

Also, I don't know what's happening with the Travis hook but it seems way behind.
For reference, the result of the last build should show up here: https://travis-ci.org/github/inventree/InvenTree/builds/754194770

eeintech:
@SchrodingersGat I realized that I forgot to mentioned how I implemented the "owner" field.
Instead of my original thoughts of using two separate database fields, I went with a [`GenericForeignKey` implementation](https://docs.djangoproject.com/en/3.1/ref/contrib/contenttypes/#generic-relations).

I created a "proxy" class conveniently called `Owner` in the `users` app.
An "owner" is a link to either a `User` or `Group` instance.
Owners creation and deletion is linked closely to users and groups, so that there is no discrepancy there (eg. owner pointing to deleted user or group, user or group missing an owner entry).

Now what's great about this implementation of ownership is that it can be re-use in other apps other than the stock management :smiley: 
It's also quite easy to add to existing models and forms as it is just one extra field!
I found it quite easy to retrieve authorized owners in templates too.

Take a look into `InvenTree/users/models.py` for more details, I hope you like it too!

SchrodingersGat:
@eeintech thanks for your fantastic effort here. I should have some time to look over this in the next couple of days 👍

SchrodingersGat:
Hey @eeintech sorry some recent merges have caused some conflicts here. Can you please fix and I'll review :)

eeintech:
@SchrodingersGat Here you go! :wink:

SchrodingersGat:
Hi @eeintech I like what you have implemented here. I do not see anything which would prevent me from merging this. Are you 100% happy for this to be merged now or are there any things you still need to work on?

eeintech:
@SchrodingersGat Yes this is ready to be merged, I don't intend to add more for the time being.

SchrodingersGat:
Fantastic, thanks for the great work on this. This (and the documentation in https://github.com/inventree/inventree-docs/issues/47) are all that I'm waiting on for the next major release :)

PR #1158, opened 2020-12-08
Expiry for Stock Items

nwns:
Considers any item with an expiry that is past today will have an `EXPIRED` status and it to be not available for sale.

Any item whose expiry is within 30 days will have a status of `STALE` and but is otherwise not changed.

Created in collaboration with @Simage, that is, he wrote it months ago and I have code janitored it into submission :smile:


SchrodingersGat:
Hey @nwns this is a great idea! I've got a few additional thoughts below:

## Status Field

The existing status field should not encompass the "staleness" state of a StockItem - it is really for the physical state of the stock item. Storing the expiry date of the stock item is already enough data to determine if the item is "stale" or "expired".

As you have the date stored, it can always be evaluated at run-time to determine if the particular stock item is stale (or not). In the API we can add a new boolean field called 'stale' (for e.g.) which is computed at run-time for each object in the query.

## Default Part Lifetime

If a stock item can "expire" I think it would make sense for the `Part` model to have a corresponding field "stock expiry duration" (or similar). This could be specified in units of days, for example. Whenever a new stock item is created, if the corresponding part has a default expiry duration, that is added to the current date and used as the default value.

For example, stock of `ACME Widget` expires 30 days after creation by default.

## Filtering

The stock table can be updated to add a filter for "stale" stock - allowing the user to quickly determine what stock is out of date.

Let me know your thoughts on the points above. Cheers.

Simage:
I believe I added the "staleness" to the status for a couple reasons.
- First I have to disagree with you, I believe 'staleness' is a physical state, akin to destroyed, a moldy loaf of bread for instance is expired, If I'm looking at my inventory and see a "stale" loaf that I know is about to expire, I may (should?) prioritize its use.
- Semantically it makes sense to display a product as "Stale" and/or "Expired" rather than "Available"

That having been said, My largest concerns with the way the status is being handled are two-fold
- The state is generally set by the user, The current implementation silently overwrites and discards the user set state 
- The state is set during object instantiation by the ORM, This just feels really hacky

Default part lifetime and filtering both make sense to me,

nwns:
After discussion with @Simage, I think the best way to get this merged would be to:

- [ ]  Add the stock item api to return a set of `statuses`, which may include `Expired`/`Stale`
- [ ]  Display only `statuses` when viewing stock items
- [ ]  Per Part default stale time (in days)
- [ ]  Per Part default expired time (in days)
- [ ]  Add to the Global Settings framework, Do not allow sale of expired Items.
- [ ]  If Global Setting set and if an expiry is defined and the StockItem is expired, do not allow the StockItem to be sold
- [ ]  Add an index on expiry field
- [ ]  Enable the stock table to filter on expiry

@SchrodingersGat would this work?

SchrodingersGat:
The list of points above make sense. I would also suggest the following items:

- [ ] List "expired" stock on the index page
- [ ] Calculate "default" expiry date for StockItem objects based on the "expired time" of the Part

In terms of how the expiry data are stored - in general it is bad practice to store calculated values in a database - especially when the calculations are very simple.

Here, the calculated value is "has this stock item expired". As the expiry date is a single static datefield, it can be inexpensively compared to the current date. All the data we need to determine if an item has expired (or is about to expire) is already there.

Also, the "expired" state is orthogonal to the "physical" state. For example, a StockItem could be damaged, but not out of date, or destroyed, and also out of date. Overloading the existing `status` field means that these combinations cannot be encoded.

To accept, this is what I would like to see:

- [ ] Add an "is_expired()" function to a StockItem which simply compares the expiry date to the current date
- [ ] Use this function to display expiry status on StockItem detail page
- [ ] Remove extra status codes from StockStatus class
- [ ] Remove `after_init_stock_item` hook which updates the database model

If you can address these four items above, I'm happy to merge and then work on the other improvements (such as default Part expiry, table filtering, etc). 

Thanks!

SchrodingersGat:
Hey @nwns please check out the implementation made in #1202  :)

nwns:
@SchrodingersGat Thank you.  The implementation works well.

PR #1174, opened 2020-12-17
Fixed category parameter templates settings view

eeintech:
Previous to this fix, the Category parameter templates settings view resulted in error with URL reverse when a category was selected


SchrodingersGat:
Thanks!

PR #1175, opened 2020-12-17
Keep 'stock actions' button enabled when on children tab

eeintech:
Bug was in the Stock item detail view when the "Children" tab is selected (ex: http://127.0.0.1:8000/stock/item/1/children/).

## Before
image

## With Fix
image

## Root Cause
The `stock_table.html` file already contains a `<button id='stock-options' ...>` button group with the same id (`stock-options`) as the one present on the Stock item detail page (in `item_base.html`). The button group in the `stock_table.html` is referenced in multiple places, whereas the button group on the Stock item detail page is not. Therefore I renamed the ID for the later to `stock-actions`.


SchrodingersGat:
Simple fix, nice work :)

PR #1217, opened 2021-01-12
Allow user with part.change permission to delete BOM items, part attachments and parameters

eeintech:
I think they deserve it :smile: 

And more seriously, it should be allowed IMO.
Any object connected to a Part instance should be able to be deleted if `part.change` permission is granted.
However, user should **NOT** be able to delete the part itself if missing the `part.delete` permision, does it make sense?


SchrodingersGat:
Fair enough! I think this makes sense. Thanks for the contribution :)

PR #1221, opened 2021-01-13
Separated category from part permissions and location from stock item permissions

### LINKED ISSUE #1176: Separate "container" from "content" in permission rule sets?
@SchrodingersGat Starting a discussion here before going heads down into the implementation.

For our team, we need more flexibility into the permission system.
I start seeing some users creating **Categories** and **Stock Locations** ("containers") instead of **Parts** and **Stock Items** ("content").
I would like to prevent this to happen as I'm the one cleaning-up the mess afterwards...

The problem is that both "containers" and "content" models permissions are bundled together:
- **Category** permissions are tied to the **Part** permission rule set
- **Stock Location** permissions are tied to the **Stock** permission rule set

Would you be on-board for splitting those two rule sets, giving **Category** and **Stock Location** their own rule sets?
Any other ideas?

eeintech:
If approved, fixes #1176


SchrodingersGat:
Great work, thanks 👍 

As an aside, I'm adding major changes (and crucial bug fixes) to the release notes file which is now on the documentation repo:

https://github.com/inventree/inventree-docs/blob/master/docs/releases/0.1.5.md

eeintech:
@SchrodingersGat Oh I see, I was going to add this but it seems like you pushed it to master already, fast! Thanks!

PR #1237, opened 2021-01-16
Add dockerfile

rcludwick:
This builds the mariadb and psycopg2 drivers, as well as install all the software needed to run inventree in a container.

One can build from a forked repo and a custom branch as well.  

/usr/src/app is the root directory of the Inventree git repo.  This can be mounted as a volume with your local git repo for development.


rcludwick:
This should be an example of how I use docker-compose.  

Running ```docker-compose build inventree``` will build the docker image from my repo (see the build section in inventree) on the branch "add-dockerfile".

The only thing that doesn't happen automatically is creating the inventree database.  

``` bash
$ docker-compose up -d db

$ docker-compose run inventree bash
Creating dev_inventree_run ... done

bash-5.0# pgcli -h db -p 5432 -u pguser
> create database inventree;

(ctrl-D)(ctrl-D)

$ docker-compose up -d
$ docker-compose logs --follow inventree
``` 

``` docker-compose.yml
version: "3.8"

services:
  db:
    image: postgres
    container_name: db
    ports:
      - 5432/tcp
    environment:
      - PGDATA=/var/lib/postgresql/data/pgdb
      - POSTGRES_USER=pguser
      - POSTGRES_PASSWORD_FILE=pgpassword
    volumes:
      - /path/to/pg/data:/var/lib/postgresql/data
    restart: unless-stopped

  inventree:
    build:
      context: ./InvenTree
      args:
        VERSION: add-dockerfile
        INVENTREE_REPO: "https://github.com/rcludwick/InvenTree.git"
    image: inventree:latest
    container_name: inventree
    ports:
      - 9767/tcp
    depends_on:
      - db
    volumes:
      - /path/to/static:/usr/src/static
      - /path/to/media:/usr/src/media
    environment:
      # RUN MIGRATE_STATIC WHENEVER YOU UPDATE THE INVENTREE CONTAINER TO A NEW VERSION
      - MIGRATE_STATIC=True
      - DATABASE_NAME=inventree
      # CREATE SUPERUSER ONCE THEN DELETE THESE
      - CREATE_SUPERUSER=False
      - DJANGO_SUPERUSER_USERNAME=admin
      - DJANGO_SUPERUSER_EMAIL=admin@admin.com
      - DJANGO_SUPERUSER_PASSWORD=admin
      # SECRETS
      - DATABASE_USER=pguser
      - DATABASE_PASSWORD=pgpassword
      - INVENTREE_SECRET_KEY=abcdefghijklm
      - DATABASE_PORT=5432
      - DATABASE_HOST=db
    restart: unless-stopped
```

fredrikkarlsson-pervanovo:
I used this to create a Docker image of InvenTree, but I made some changes:
* Moved the DEBUG variable initiation out of the start script
* Wrapped the manage.py migrations to check the MIGRATE_STATIC variable
* Added exit after the migration block and the create superuser command, ie. not starting server afterwards
* Made the port (9767) configurable
* Moved secrets to .env file

But as a starting point this worked out great! Thanks!

rcludwick:
> I used this to create a Docker image of InvenTree, but I made some changes:
> 

Good, I'm glad you found it useful.  I'm hoping this can be the basis of a standard docker container you can just pull from docker hub upon every new release of inventree.

SchrodingersGat:
Hi @rcludwick I've been able to give this some more attention now.

Background: I've never used docker before, so I've had to do some reading into how it all fits together. I did not realize dockerhub was a "thing" - but having official InvenTree releases pushed there looks like a cool option!

I have some docker related questions which I'm hoping you can answer for me:

### image

In your example compose file above, you have the line:

```
    image: inventree:latest
```

Does this try to grab an inventree image from dockerhub? And then "extend" upon it?

### Customization

If the user wishes to use a different database backend, I'm assuming they can simply override the `db` service?

### Variables

I note in your compose file you have to put the database username / password in twice. Is it possible to define some variables in the file which can then be re-used?

rcludwick:
So currently this builds the docker image from the local inventree git repo.  It then saves it in my local docker container cache.  

So inventree:latest pulls from there, and it always pulls the latest image.

In deploying inventree to docker hub, the CICD system would build the docker container using the Dockerfile, then publish it to dockerhub.  This would probably happen on a release tag pushed to github.  And the Dockerfile can either use the git repo or just pip install from a published package on PyPI say, depending upon how you want to do releases.  

Then I as a user can just deploy using inventree/inventree:latest say, when I feel like upgrading it by doing a "docker pull"  I have about 20 docker containers running on an old laptop, and when I did a OS upgrade, it was the easiest migration I ever did, because docker-compose downloaded all the images and started them up.

SchrodingersGat:
That makes sense, thanks Rob. 

In your example dockerfile you have it setup to run a gunicorn server. I'm assuming it would be possible to configure the dockerfile to use a different webserver if the users require?

rcludwick:
Yes.  I would just use gunicorn or uwsgi with nginx in reverse proxy, and it's ready to be deployed to production... 

Or you scale up to two and then run your own load balancer in front of them.    

In my case I have 20 docker containers using traefik to proxy them.

miniPCR-Jesse:
Thank you both rcludwick and SchrodingersGat for looking into providing a Docker image for Inventree.  I recently joined a start-up and Inventree looks nearly perfect for our inventory needs, but I'm trying to stick to Docker apps due to their ease in maintenance and app-isolation.  I'd definitely throw my vote towards adding InvenTree to the docker hub!

rcludwick, two questions if you don't mind:
- The dockerfile you provided looks like it pulls Inventree's latest release, but when I spun up the docker file it's running version "0.1.5 pre", which looks to be the last version of your personal branch.  So safe to assume that what you provided as an example implementation is dependent on the repo you split?
- The InvenTree logo on the login page is broken until you log in and then log back out, so it looks to be a permission issue.  Is this normal to InvenTree or an artifact of the docker implementation?

SchrodingersGat:
I am actually focusing on improving deployment methods for the next major release. This will involve a complete overhaul of the docs, with a focus on robust deployment, rather than simple "getting started" instructions. Inventree is a popular enough project now to warrant this. 

Also I'll be looking at providing a docker image (or a set of images) with examples of compose files.  

Regarding the logo bug - this has been fixed in latest master.

rcludwick:
> Thank you both rcludwick and SchrodingersGat for looking into providing a Docker image for Inventree. 

To be fair it's just a Dockerfile that needs to be built to build a local docker image (which serves my purpose)

> * The dockerfile you provided looks like it pulls Inventree's latest release, but when I spun up the docker file it's running version "0.1.5 pre", which looks to be the last version of your personal branch.  So safe to assume that what you provided as an example implementation is dependent on the repo you split?

It requires a git clone to a local directory.  Then you can docker build it, so you have to maintain that version yourself.

> * The InvenTree logo on the login page is broken until you log in and then log back out, so it looks to be a permission issue.  Is this normal to InvenTree or an artifact of the docker implementation?

Yes, I believe that's true.  I believe the reason for that is there needs to be an nginx frontend to serve the media directories without auth.   Currently I'm using the django static files server and that in order for the media to show up on the login page, you have to auth first (because of the auth middleware, don't you know)

The next step is to add nginx to the install for the Dockerfile as well as in a supervisord.conf file so that the gunicorn workers, nginx.conf, and the new django q workers all get started at once. 

@SchrodingersGat has started on supervisord.conf.  I can probably find an nginx.conf template to use and modify the Dockerfile.

SchrodingersGat:
@rcludwick I think I'm done now :)

The official InvenTree docker image is here - https://hub.docker.com/r/inventree/inventree

And instructions on how to use it with a compose file: https://inventree.readthedocs.io/en/latest/start/docker/

I've got the docker containers working seamlessly now!

image

Thank you **so much** for all your help getting this up and running. I'm very happy that InvenTree has a robust dockerized setup process now!

Let me know if you see any issues

rcludwick:
Congratulations.   

So one thing you can do now, is it should be pretty easy to set up development for everyone.  You can mount your source code in a volume in the docker container where gunicorn runs it, and you can test it out directly.

I'm going to switch my "production" server over tomorrow.  I'll let you know what I think then.

miniPCR-Jesse:
Just installed Inventree using the docker documentation and it worked perfectly.

One additional question.  [Watchtower](https://containrrr.dev/watchtower/) is commonly used to update docker images.  I'm assuming that to update Inventree you can let watchtower run, and then (depending on whether it's required for the specific release) run the following command afterwards:

`docker-compose run inventree invoke update`

Is that correct?

rcludwick:
I've never used watchtower, but it shouldn't be any different than any other docker container.

SchrodingersGat:
@miniPCR-Jesse sounds about right. You should only have to run `invoke update` once you've pulled down the new image.

PR #1247, opened 2021-01-19
Fix misspelled query filter method name

danielpervan:
Opening a sales order will break because query.filter was misspelled.


SchrodingersGat:
Thanks @danielpervan

Issue #1256, opened 2021-01-22
Critical: PDF report generation is now broken


SchrodingersGat:
![image](https://user-images.githubusercontent.com/10080325/105430764-62655000-5ca8-11eb-866d-5aeef2402c05.png)

- Report generation recently moved to the API
- Now is BAD
- Fix!!

SchrodingersGat:
Seems like it might only be an issue when generating LaTeX reports?

neco1982:
Hello I made some PDF report 
![Ekran Resmi 2021-01-28 21 20 37](https://user-images.githubusercontent.com/31635566/106181803-16863f80-61af-11eb-8d49-34d73f10960c.png)

SchrodingersGat:
@neco1982 that looks great - could you share the .html template you used to generate this report?

neco1982:
Here is codes
1.urls.py
url(r'^print/', views.BuildPrint, name='build-print'),
2. views.py
from django.http import HttpResponse
from django.template.loader import render_to_string
from weasyprint import HTML
import tempfile

def BuildPrint(request, **kwargs):     

    items = Build.objects.filter(**kwargs).values('reference','quantity','notes','creation_date','target_date', 'part__name', 'part__description','part__variant_of__name', 'title','allocated_stock__stock_item__part__name', 'allocated_stock__quantity').distinct()
    

    html_string = render_to_string('build/build_print.html', {'items': items, 'time':datetime.now()})
    html = HTML(string=html_string)
    result = html.write_pdf()
        
    response = HttpResponse(content_type='application/pdf;')    
    response['Content-Dispositon'] = 'inline; attachment; filename=İşEmri.pdf' 
    response['Content-Transfer-Encoding'] = 'binary'    
    with tempfile.NamedTemporaryFile(delete=True) as output:
            output.write(result)
            output.flush()
            output = open(output.name, 'rb')
            response.write(output.read())
     
    
    return response

3. build_base.html for link pdf
<a href="{% url 'build-print' build.id %}" type='button' class='btn btn-default btn-glyph' id='build-print' title='{% trans "Yazdır" %}'>
            <span class='fas fa-clipboard-list'/>
        </a>
4. build_print.html
<!doctype html>
<html lang="en">
  <head>
    
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">

[...]

PR #1270, opened 2021-02-03
Adds "report snippet" class allowing re-usable report snippets

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/1256

- [x] Automatically add a "generic" test report when migrating
- [x] Remove all LaTeX support
- [x] Update documentation (including removing LaTeX references) - https://github.com/inventree/inventree-docs/pull/53
- [x] Add user-configurable setting for enabling / disabling test reporting interface


SchrodingersGat:
Need to work out how to ensure that the report "snippet" template files are always uploaded with the same file name. Otherwise, they cannot be included using the template syntax!

Does this go "against" the django way of doing filefield uploads? Need to work out a solution here...

PR #1279, opened 2021-02-06
Default page size

SchrodingersGat:
Adds some more reporting features

## Default Page Size

Sets the page size for default InvenTree reports (and any reports derived from the base templates)

## Debug Mode

- When set, reports are generated to HTML rather than PDF
- Lets users look at actual generated HTML if there is a problem


SchrodingersGat:
Docs: https://github.com/inventree/inventree-docs/pull/54

Issue #310, opened 2019-05-12
Server-size pagination for tables


SchrodingersGat:
- Fetch and filter lag becoming noticeable with larger number of parts
- bootstrap-table has option for server-side pagination and ordering
- Need to implement this!

PR #1373, opened 2021-02-28
Server pagination

SchrodingersGat:
Adds server-side pagination for (some) data tables which **significantly** speeds up the UI - especially when there are a large number of records.

Fixes https://github.com/inventree/InvenTree/issues/310

The query is limited on the database side.

For a dataset of ~1000 records, this reduces query time from 6,000ms to ~100ms.

- [x] Check that the "limit" functionality still works for the home page
- [x] Unit testing for the stock list API
- [x] Unit testing for the part list API
- [x] Refactor duplicated "queryParams" function for inventreeTable

### DIFF
--- a/InvenTree/InvenTree/api.py
+++ b/InvenTree/InvenTree/api.py
@@ -18,7 +18,7 @@ from rest_framework.response import Response
 from rest_framework.views import APIView
 
 from .views import AjaxView
-from .version import inventreeVersion, inventreeInstanceName
+from .version import inventreeVersion, inventreeApiVersion, inventreeInstanceName
 
 from plugins import plugins as inventree_plugins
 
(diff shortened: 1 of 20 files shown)


eeintech:
@SchrodingersGat Wow very impressive! Nice work!

Seems like large BOM loading is still slow, any idea why?

SchrodingersGat:
That's pretty slow indeed. With only 67 items I'd say there is something more serious going on that needs to be looked at.

For reference, before I added server-side pagination for the "parts" list, loading *all* parts on my test database (~1000) took around 6 seconds.

eeintech:
You're right, we have 200+ parts in our database and the part list is loading almost instantaneously. I'll have to investigate why our BOM doesn't follow the same logic...

SchrodingersGat:
Using the django debug toolbar, I was actually able to replicate this. The issue is *not* the data loading into the table

The server request to generate the page takes about 18s

image

But the subsequent database request is reasonably fast:

SchrodingersGat:
On further inspection it appears that *any* of the part detail views are very slow (~15 seconds just to load the "part detail" page).

PR #1363, opened 2021-02-26
Api permissions

SchrodingersGat:
Adds role checks to API endpoints:

- Fixes https://github.com/inventree/InvenTree/issues/1362
- Fixes https://github.com/inventree/InvenTree/issues/1358
- Fixes https://github.com/inventree/InvenTree/issues/1293
- Fixes https://github.com/inventree/InvenTree/issues/1365
- 
Add role requirements (and unit testing!) for

- [x] Build API
- [x] Company API
- [x] Stock API
- [x] Stock Location
- [x] Part API
- [x] Part Category
- [x] Report API
- [x] Sales Order
- [x] Purchase Order
- [x] Label API

PR #1376, opened 2021-03-01
Fix for SupplierPart API permissions

eeintech:
Missing `part_supplierpart` table permissions

### DIFF
--- a/InvenTree/users/models.py
+++ b/InvenTree/users/models.py
@@ -68,6 +68,7 @@ class RuleSet(models.Model):
             'part_partparameter',
             'part_partrelated',
             'part_partstar',
+            'part_supplierpart',
         ],
         'stock_location': [
             'stock_stocklocation',


eeintech:
@SchrodingersGat Unfortunately this is was just a hacky fix...

`check_table_permission` method does not know how to fetch the `SupplierPart` permission because of the remaining link with the `part` app (in `SupplierPart` model `db_table = 'part_supplierpart'`) but in reality its permissions should be tied to the `company` app.

Do you have an idea how to fix this?

SchrodingersGat:
@eeintech what exactly is the issue here? 

I see that the table name `company_supplierpart` is actually not valid, but what is the problem that you are fixing that led you here?

And is the `check_table_permission` function not performing correctly?

eeintech:
@SchrodingersGat A user with edit `Part` and `Purchased Order` permissions (add/change) is not able to create a supplier item anymore... I think it has to do with #1363

eeintech:
I have to admit this is quite an annoying issue as we are forced to use the Admin/superuser to create supplier items at the moment... Maybe I'll have some time tomorrow to do more debug but if you have any idea let me know @SchrodingersGat

eeintech:
Was fixed by #1381

PR #1393, opened 2021-03-08
updated german translations

matmair:
Updated a few translations for German, did not change anything in code.
There are still **414 fuzzy** (sadly mostly to be newly translated) and **136 missing** translations. I will try to do those also, but I am not sure about a lot of them and the timeline.

### Language background
I am a native German-speaker from Austria so there might be subtle differences to how Germans would phrase it although I tried keeping it conforming to normal engineering vocabulary.

Also thank you very much for this awesome app and chosing MIT as license


SchrodingersGat:
Thanks @matmair I welcome any further contributions you can make! And I appreciate your support :)

PR #1398, opened 2021-03-11
Django q

SchrodingersGat:
Support for background tasks.

- Closes https://github.com/inventree/InvenTree/issues/1097
- Closes https://github.com/inventree/InvenTree/pull/1237
- Closes https://github.com/inventree/InvenTree/pull/1113

Docs: https://github.com/inventree/inventree-docs/pull/92


InvenTree/InvenTree/version.py
```diff
@@ -23,6 +25,53 @@ def inventreeVersion():
...
+
+    if version is None:
+        version = INVENTREE_SW_VERSION
+
+    match = re.match(r"^.*(\d+)\.(\d+)\.(\d+).*$", str(version))
+
+    return [int(g) for g in match.groups()]
+
+
+def versionTupleToInt(version):
```
rcludwick, line 39: You can compare version tuples directly.  

(2,4,1) < (2,4,2):  True
(3) > (2,4,2): True

InvenTree/InvenTree/status.py
```diff
@@ -1,15 +1,49 @@
...
+    """
+    Sometimes Stat.get_all() returns [].
+    In this case we have the 'heartbeat' task running every five minutes.
+    Check to see if we have a result within the last ten minutes
+    """
+
+    now = datetime.now()
+    past = now - timedelta(minutes=10)
+
+    results = Success.objects.filter(
```
rcludwick, line 38: I would just delete any heart beat other than the last one while I'm in this code.

InvenTree/InvenTree/status.py
```diff
@@ -1,15 +1,49 @@
...
 
 
+def is_q_cluster_running(**kwargs):
+    """
+    Return True if at least one cluster worker is running
+    """
+
+    clusters = Stat.get_all()
+
+    if len(clusters) > 0:
```
rcludwick, line 26: if clusters: 
  return True

Since it's more pythonic.

InvenTree/InvenTree/version.py
SchrodingersGat: Huh, you learn something new every day. I did not know Python could do that!

InvenTree/InvenTree/context.py
```diff
@@ -30,10 +30,22 @@ def health_status(request):
...
 
-    return {
-        "system_healthy": InvenTree.status.check_system_health(),
+    status = {
+        'django_q_running': InvenTree.status.is_q_cluster_running(),
     }
 
+    all_healthy = True
+
+    for k in status.keys():
```
rcludwick, line 39: for k in status:

deploy/inventree.conf
```diff
@@ -0,0 +1,39 @@
...
+[program:inventree-server]
+user=inventree
+directory=/home/inventree/src/InvenTree
+command=/home/inventree/env/bin/gunicorn -c gunicorn.conf.py InvenTree.wsgi
+startsecs=10
+autostart=true
+autorestart=true
+startretries=3
+; Change these paths if log files are stored elsewhere
+stderr_logfile=/home/inventree/log/server.err.log
```
rcludwick, line 27: Inside docker should not redirect to files.  Outside of docker they should be files.

docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+# Install required PIP packages
+RUN python -m venv $INVENTREE_VENV && pip install --upgrade pip setuptools wheel
+RUN python -m venv $INVENTREE_VENV && pip install --no-cache-dir -U invoke
+RUN python -m venv $INVENTREE_VENV && pip install --no-cache-dir -U psycopg2 mysqlclient pgcli mariadb
+
+# Install InvenTree packages
+RUN python -m venv $INVENTREE_VENV && pip install --no-cache-dir -U -r $INVENTREE_SRC_DIR/requirements.txt
+
+# Install supervisor
+RUN apk add --no-cache supervisor
```
rcludwick, line 58: I believe you can pip install this to get the latest version of supervisord.

InvenTree/InvenTree/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+
+def delete_successful_tasks():
+    """
+    Delete successful task logs
+    which are more than a month old.
+    """
+
+    try:
+        from django_q.models import Success
+        logger.warning("Could not perform 'delete_successful_tasks' - App registry not ready")
```
rcludwick, line 85: logger statement needs to be after the except.

InvenTree/InvenTree/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+    repeats = kwargs.pop('repeats', -1)
+
+    try:
+        from django_q.models import Schedule
+    except (AppRegistryNotReady):
+        logger.warning("Could not start background tasks - App registry not ready")
+        return
+
+    try:
+        if Schedule.objects.filter(func=taskname).exists():
```
rcludwick, line 34: So what happens when the task parameters change?   Like you want to shift the period from 5 minutes to 30 minutes say?

docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+      org.label-schema.url="https://hub.docker.com/r/inventree/inventree-docker" \
+      org.label-schema.version=$INVENTREE_VERSION \
+      org.label-schema.vcs-url=$INVENTREE_REPO \
+      org.label-schema.vcs-branch=$BRANCH \
+      org.label-schema.vcs-ref=$COMMIT
+
+RUN echo "Installing InvenTree '${INVENTREE_VERSION}' from ${INVENTREE_REPO}"
+
+# Create user account
+RUN addgroup -S inventreegroup && adduser -S inventree -G inventreegroup
```
rcludwick, line 31: If I run a docker container as myself (1000, 1000) and have a volume for the config files.  And then if the docker container has inventree user/group set to 500,500 (e.g.).  Then my home directory outside of docker will contain files with owner 500,500.   

The linuxserver.io series of docker containers support PUID, PGID as integers for process uid and process groupid.  

Usually I set these to be 1000,1000 because that's my user id from which I run docker containers.

ARG INVENTREE_GID=1000
ARG INVENTREE_UID=1000

RUN addgroup -S inventree --gid ${INVENTREE_GID} && adduser -S inventree -G inventree --uid ${INVENTREE_UID}

docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+RUN apk add --no-cache fontconfig ttf-droid ttf-liberation ttf-dejavu ttf-opensans ttf-ubuntu-font-family font-croscore font-noto
+RUN apk add --no-cache python3
+RUN apk add --no-cache postgresql-contrib postgresql-dev libpq
+RUN apk add --no-cache mariadb-connector-c mariadb-dev
+
+# Clone source code
+RUN git clone --branch $INVENTREE_VERSION --depth 1 ${INVENTREE_REPO} ${INVENTREE_SRC_DIR}
+
+# Install required PIP packages
+RUN python -m venv $INVENTREE_VENV && pip install --upgrade pip setuptools wheel
```
rcludwick, line 50: So playing with the Dockerfile tonight, this doesn't install anything in to $INVENTREE_VENV... my fault here.

```
+RUN python -m venv $INVENTREE_VENV 
+RUN $INVENTREE_VENV/bin/python3 -m pip install --upgrade pip setuptools wheel
+RUN $INVENTREE_VENV/bin/pip install --no-cache-dir -U invoke
+RUN $INVENTREE_VENV/bin/pip install --no-cache-dir -U psycopg2 mysqlclient pgcli mariadb gunicorn
```

docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+RUN python -m venv $INVENTREE_VENV && pip install --no-cache-dir -U invoke
+RUN python -m venv $INVENTREE_VENV && pip install --no-cache-dir -U psycopg2 mysqlclient pgcli mariadb
```
rcludwick, line 52: gunicorn needs to be installed as well.

docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+# Install InvenTree packages
+RUN python -m venv $INVENTREE_VENV && pip install --no-cache-dir -U -r $INVENTREE_SRC_DIR/requirements.txt
+
+# Install supervisor
+RUN apk add --no-cache supervisor
+
+# Create required directories
+RUN mkdir /home/inventree/media /home/inventree/static /home/inventree/log /home/inventree/backup
```
rcludwick, line 61: I would use the variables you have defined above.

docker/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+# Copy supervisor file
+COPY deploy/inventree.conf /etc/supervisor/conf.d/inventree.conf
+
+CMD ["/usr/bin/supervisord"]
```
rcludwick, line 66: Supervisord whines when running as root.  so I did this:

```
ENV INVENTREE_HOME="$INVENTREE_HOME"  # <-- ARG is not automatically an ENV
COPY deploy/inventree.conf $INVENTREE_HOME/supervisord.conf
CMD ["sh", "-c", "/usr/bin/supervisord -c ${INVENTREE_HOME}/supervisord.conf"]

File: deploy/inventree.conf [hunk 14]
```diff
@@ -0,0 +1,39 @@
...
+; The recommended way to handle the web server is to use gunicorn
+;
+; ## Background Tasks
+; A background task manager processes long-running and periodic tasks
+; InvenTree uses django-q for this purpose
+
+[supervisord]
+; Change this path if log files are stored elsewhere
+logfile=/home/inventree/log/supervisor.log
+nodaemon=true
```
rcludwick, line 16: Add:

```
user=inventree
```

deploy/inventree.conf
```diff
@@ -0,0 +1,39 @@
...
+autorestart=true
+startretries=3
+; Change these paths if log files are stored elsewhere
+stderr_logfile=/home/inventree/log/server.err.log
+stdout_logfile=/home/inventree/log/server.out.log
+
+[program:inventree-cluster]
+user=inventree
+directory=/home/inventree/src/InvenTree
+command=/home/inventree/env/bin/python manage.py qcluster
```
rcludwick, line 33: So this needs to be something like this.

```
command=sh -c 'DJANGO_SETTINGS_FILE=InvenTree.settings /home/inventree/env/bin/python /home/inventree/src/InvenTree/manage.py qcluster'
```
But this fails immediately because config.yaml isn't found.  So generation of an initial config.yaml file needs to be done.

rcludwick:
Hi,  

So I pulled your branch tonight and tried to get the docker container built.  

So the following things were broken: 

* gunicorn isn't being installed.
* config.yaml doesn't exist when qcluster tries to run, so supervisord kills it after three attempts.
* supervisord whines about running in root mode
* python -m venv $INVENTREE_VENV && pip install ...   doesn't work (though that one is my fault)

Hope this helps.  

--Rob

deploy/inventree.conf
SchrodingersGat: Thanks! I'd just read about piping logs to stdout in docker. This has now been addressed :)

InvenTree/InvenTree/tasks.py
SchrodingersGat: Thanks!

InvenTree/InvenTree/tasks.py
SchrodingersGat: Good point. Code has been updated now.

docker/Dockerfile
SchrodingersGat: Fixed in latest

SchrodingersGat:
@rcludwick thanks for your helpful advice. I'm getting there, still a couple of kinks to work out. But I'm finally getting my head around the docker ecosystem!

One question - I want to setup nginx in the "default" InvenTree docker / compose files. Would it be better to:

a) Run nginx in a separate container, with a custom Dockerfile for configuration
b) Run nginx as a separate (third) managed process in the `inventree` dockerfile?

rcludwick:
I tend to say yes for simplicity with dealing with the media/statics, but you could also just make life easier for yourself and just provide a configuration for a docker-compose with nginx sitting in front of it.

SchrodingersGat:
@rcludwick I think this is ready for a test drive now. Please let me know if you can get it working on your end, and if there's anything else I should add, before it goes live on dockerhub

rcludwick:
Apologies, I didn't get a chance this weekend to look at it this weekend.  I can't wait to try it out on my "production" server at home.  

If I see anything wrong, I'll just file issues/PR's.

:tada:

PR #1400, opened 2021-03-12
Add setting to disable IPN editing after part is created (web only)

eeintech:
I wanted a way to prevent users to edit an auto-generated IPN number after part is created.

The title says "web only" because I think the API request should still be able to edit it.
That's because an automated API app may want to grab the part's primary key after creation and generate the IPN based on the PK (like in Ki-nTree :smile: ).

To keep the current behavior (IPN edit is allowed) I set this setting default to True (eg. Editing IPN is allowed).


eeintech:
@SchrodingersGat Looks like Travis got disconnected?

eeintech:
Looking good now!

SchrodingersGat:
@eeintech I'm relatively happy with this, although I don't think I would use the feature personally. Can you please add a PR to the docs explaining this new feature?

eeintech:
Added the PR https://github.com/inventree/inventree-docs/issues/82, I will get on the doc soon 👍

eeintech:
For the doc, see https://github.com/inventree/inventree-docs/pull/83

SchrodingersGat:
Thanks @eeintech

eeintech:
No problem!

PR #1401, opened 2021-03-12
Stock filter for parts assemblies

eeintech:
Would be useful at our company :smiley:


SchrodingersGat:
This seems more than reasonable!

PR #1403, opened 2021-03-12
unified translation scheme

matmair:
I tried to coordinate with @mosenturm to make a few translations a bit more concise. There are still 68 strings missing, those will follow shortly but I need to discuss those.

To all German-speaking folks: I would love to discuss a few terms to find better strings. Hit me up-everything is linked on my profile.


SchrodingersGat:
Thanks :)

Issue #1404, opened 2021-03-15
Recommended workflow for importing KiCad BOMs


cdwilson:
Is there a recommended workflow for exporting a BOM from KiCad and importing it into InvenTree?  

I just went through the process (for the first time) of exporting a CSV from KiCad using [KiBoM](https://github.com/SchrodingersGat/KiBoM) and [importing](https://inventree.readthedocs.io/en/latest/build/bom/) it into InvenTree.  

I ran into a couple issues along the way, and I'm curious if anyone has developed a more "seamless" workflow for importing the BOM from KiCad into InvenTree.

1. I had to manually remove some of the metadata that KiBoM adds to the end of the CSV (InvenTree initially tried to import those extra lines as actual BOM lines).  Is there a way to tell InvenTree to ignore those extra lines?
2. I currently have the `Part_IPN` stored in my KiCad library symbols, but I don't currently have the `Part_Name` in KiCad.  I'm curious why the `Part_Name` is required if the `Part_IPN` is provided in the CSV, because I would assume the `Part_IPN` can be used to match the imported BOM line to a part in the InvenTree database.  @SchrodingersGat can you help me understand why both the `Part_Name` and `Part_IPN` are required on import?

Also, if there is a better/preferred place to post questions like this instead of creating a GitHub issue, please let me know.

SchrodingersGat:
Hi @cdwilson 

>  Also, if there is a better/preferred place to post questions like this instead of creating a GitHub issue, please let me know.

This is the right forum for posting such questions.

> I had to manually remove some of the metadata that KiBoM adds to the end of the CSV (InvenTree initially tried to import those extra lines as actual BOM lines). Is there a way to tell InvenTree to ignore those extra lines?

Not at this stage. The library InvenTree uses for data importing does not like having extra lines at the end of the file.

> I currently have the Part_IPN stored in my KiCad library symbols, but I don't currently have the Part_Name in KiCad. I'm curious why the Part_Name is required if the Part_IPN is provided in the CSV, because I would assume the Part_IPN can be used to match the imported BOM line to a part in the InvenTree database. @SchrodingersGat can you help me understand why both the Part_Name and Part_IPN are required on import?

Good point. Originally the `Part_Name` field was required because the import process did not support lookup by IPN. However when IPN was added, the Name was still required.

Honestly, the BOM import process is in need of some polishing - it is a very difficult process to streamline!

If you have some suggestions as to how it could be improved, please let me know.

[...]

PR #1407, opened 2021-03-15
Split required and part match headers for BOM import

eeintech:
As mentioned in #1404 only one of the following field should be required for part matching: part ID, part IPN or part name


eeintech:
Parts with no BOMs were showing in "BOM Waiting Validation" section of the index page. A part's BOM shouldn't be in that list if the BOM does not exist yet. So I think it is fair assumption that `is_bom_valid` should return **True** if part has no BOM items

SchrodingersGat:
Thanks for the quick fix @eeintech - looks pretty good to me. I agree with your comment about blank BOMs being "valid", also.

PR #1423, opened 2021-03-28
german translations

matmair:
Finally got around to finish the German translations, they are at 100% at time of commit. I also updated the po-references for en and es.


SchrodingersGat:
Awesome work, thanks @matmair

PR #1445, opened 2021-04-08
Simplified stock table view in Part and SupplierPart detail pages

### LINKED ISSUE #1442: Remove duplicate columns from part tables
(Adding this as a separate issue per request in https://github.com/inventree/InvenTree/pull/1417#issuecomment-815024403 from @eeintech)

There are a couple places (e.g. part stock pages) where the tables include info that is duplicated from the page header.  This info is identical for each line in the table, and it causes the table to be crowded on smaller screens.  Would it be OK to remove these duplicate columns?

image

image

eeintech:
Fixes #1442

### DIFF
--- a/InvenTree/company/templates/company/supplier_part_stock.html
+++ b/InvenTree/company/templates/company/supplier_part_stock.html
@@ -22,7 +22,7 @@
         params: {
             supplier_part: {{ part.id }},
             location_detail: true,
-            part_detail: true,
+            part_detail: false,
         },
         groupByField: 'location',
         buttons: ['#stock-options'],
--- a/InvenTree/part/templates/part/stock.html
+++ b/InvenTree/part/templates/part/stock.html
@@ -40,7 +40,7 @@
         params: {
             part: {{ part.id }},
             location_detail: true,
-            part_detail: true,
+            part_detail: false,
         },
         groupByField: 'location',
         buttons: [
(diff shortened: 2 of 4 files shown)


SchrodingersGat:
Neat fix!

Have you checked that **all** calls to `loadStockTable` now explicitly set the `part_detail` option (either to True or False). Now that it is not *assumed* to be true, it must be specified.

eeintech:
@SchrodingersGat I checked some of them, but haven't done a lookup on the code base. I will do that and confirm

SchrodingersGat:
Thanks. Happy to merge in once you've checked.

eeintech:
@SchrodingersGat Scanned through all the calls of `loadStockTable` method and only in one place `part_detail` was not specified, it is corrected now. I've also removed the commented line that was hard coding `part_detail` in the `loadStockTable` method because I don't think it makes sense to keep it. However, it could be that in the future one forgets about specifying `part_detail` parameter, it throws JS errors in this case. I don't know a good way to translate this python line to javascript, do you?

``` python
part_detail = options.get('part_detail', False)
```

SchrodingersGat:
You could write 

```javascript
part_detail = options.part_detail || false;
```

Or

```
var part_detail = false;
if ('part_detail' in options) {
    part_detail = options.part_detail;
}
```

SchrodingersGat:
Thanks for fixing this :)

eeintech:
Oh it is that simple too, nice! :+1:

PR #1446, opened 2021-04-08
Replace normalize with integer wrapper for quantity field

### LINKED ISSUE #1443: Error exporting BOM in YAML format
The error is:

Error

eeintech:
In a attempt to fix #1443 

I don't see when can the BOM Item quantity be a decimal number and it was causing the error in the BOM export because the PyYAML library doesn't have a `Decimal` object representation.

### DIFF
--- a/InvenTree/part/admin.py
+++ b/InvenTree/part/admin.py
@@ -180,7 +180,7 @@ class BomItemResource(ModelResource):
 
         Ref: https://django-import-export.readthedocs.io/en/latest/getting_started.html#advanced-data-manipulation-on-export
         """
-        return normalize(item.quantity)
+        return int(item.quantity)
 
     def before_export(self, queryset, *args, **kwargs):
 
--- a/InvenTree/part/admin.py
+++ b/InvenTree/part/admin.py
@@ -16,8 +16,6 @@ from .models import PartCategoryParameterTemplate
 from .models import PartTestTemplate
 from .models import PartSellPriceBreak
 
-from InvenTree.helpers import normalize
-
 from stock.models import StockLocation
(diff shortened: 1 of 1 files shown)


eeintech:
Looks like float would work for YAML:

``` python
>>> import yaml
>>> dec = 1.23456
>>> print(dec)
1.23456
>>> type(dec)
<class 'float'>
>>> yaml.safe_dump(dec)
'1.23456\n...\n'
>>> dec = 1.2345600000
>>> yaml.safe_dump(dec)
'1.23456\n...\n'
```

But I'm not sure all the other formats would remove the trailing zeros...

SchrodingersGat:
Can we check the output format and choose 'float' if it is a yaml file?

eeintech:
I don't really find a good way to check the export format and pass it to the `dehydrate_quantity` method...

SchrodingersGat:
Ok, what does the output look like for a .csv file? Does it suppress the trailing zeros correctly?

eeintech:
@SchrodingersGat I tested all formats with `float` and CSV, TSV and XLS gets no decimal digits for integer quantities and the right amount of digits for decimal numbers, while XLSX, JSON and YAML get one digit for integer quantities and also the right amount of digits for decimal numbers.

So if we are okay that quantity of `10` becomes `10.0` for XLSX, JSON and YAML then `float` is a great option :smiley:

SchrodingersGat:
Seems like a reasonable compromise.

SchrodingersGat:
Thanks for the fix :)

eeintech:
No problem, thanks for merging 👍

PR #1449, opened 2021-04-09
Fixed transfer stock action in template

### LINKED ISSUE #1448: Serialised Stock Items can't be transferred between locations
Non serialised stock item:
working_transfer

Serialised stock item:
not_working_transfer

eeintech:
Fixes #1448 

Thanks for the catch @jnewlands


SchrodingersGat:
Thanks @eeintech

Issue #1436, opened 2021-04-02
Why is company description a required field?


cdwilson:
Just curious if there is a reason why company description is a required field.  I'm usually not really sure what to put in the description field, so I typically just put the company name again.  As a result, all my suppliers and manufacturers show up with an awkward name where it's duplicated twice...

1. Is it possible to make the description field optional?
2. I'm curious how other people use this field.  Can you give some examples of what you typically put in the description for companies like Digi-Key, Mouser, etc?

![image](https://user-images.githubusercontent.com/14631/113456145-52ad6a80-93c1-11eb-8a54-b8c15f745fc6.png)

SchrodingersGat:
Fair question, it could be made an optional field without detriment. 

I do find it useful as while the engineering dept knows all about suppliers like Digi-Key etc, other depts might need some context info e.g. "supplier of electronic components."

But sure, can make this field optional

SchrodingersGat:
I'm away on holiday for a week but feel free to submit a PR to make this field optional :)

PR #1417, opened 2021-03-23
Manufacturer Parts

eeintech:
When merged, fixes #1389 also fixes #1436

PR #1450, opened 2021-04-10
Company description is no longer a required field

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/1436


eeintech:
@SchrodingersGat I had done this change part of #1417 because it affected the migrations of the company app.

Would be nice if we could merge #1417 soon so that I don't have to maintain the branch each time there is a merge to master.
Mind having a look?

PR #1458, opened 2021-04-13
updated translation references

matmair:
@SchrodingersGat just an update to the .po files for the few new strings. And the german translations.


SchrodingersGat:
Keep it coming @matmair , this is great :) Many thanks.

PR #1462, opened 2021-04-14
Added revision and stock item QR code URL for label creation

eeintech:
@SchrodingersGat I wanted to code the stock item absolute URL directly into the QR code, so a phone can open the link directly.

In the label template, this will generate the QR code URL:
``` html
<img class='qr' src='{% qrcode qr_url %}'>
```

I also added the part revision in the context data so I could create some sweet labels like:

image


SchrodingersGat:
Fantastic! Please also add a short note to the revision info in the docs :)

eeintech:
@SchrodingersGat See https://github.com/inventree/inventree-docs/pull/100!

Issue #1463, opened 2021-04-15
use instance name in title and views?


matmair:
@SchrodingersGat this is just a suggestion - close this if you disagree, no hard feelings.

There are a few views and (meta) titles where 'InvenTree' is used where it could make sense to use the instance name. I currently have a few instances deployed in places where English is not widely used or known. A few users, especially in manufacturing, are confused by the 'InvenTree' as they are used to only use the internal system names (for exampel 'TeilDB').

I would make a PR replacing 'InvenTree' with the instance name in all end-user facing parts. About, version and setting-views should keep 'InvenTree' in my opinion as to show what system is used.

SchrodingersGat:
I'm not opposed to this, feel free to submit a PR showing how you would like to see it implemented

matmair:
@SchrodingersGat I threw together how I would do it.
It am missing:
- unittests
- UI-tests
- documentation

the feature branch is in [my github](https://github.com/matmair/InvenTree/tree/instance-name-in-title)

PR #1464, opened 2021-04-15
Translation improvments

matmair:
this PR translates / makes translatable
- the root-elements of stock and part-trees in the API (which are directly used for building the sidebar)
- boolean values in table-filters (in creation and display)

-> from my testing it seems like all parts of the UI should be translatable now, maybe except for #1463


SchrodingersGat:
Nice work again, thanks @matmair

Issue #1457, opened 2021-04-13
Docker: allow user to use the official nginx server


rcludwick:
I would recommend in the docker_compose example, to rather than deploy a custom nginx container, just specify a config file in the directions and allow the user to mount it inside the docker container.

This will require less infrastructure overhead for you to maintain your own custom docker container, when all the users really need is the nginx config.

rcludwick:
```
server {

    listen 80;

    location / {
        proxy_pass http://inventree;
```

As a docker user I might need to change the port as well as the name of the server (since something else may be running on 80 in the same box/network).  Right now I have mine named "inventree-web", but this nginx file wouldn't work.

SchrodingersGat:
Excellent suggestions. I was not sure if that was possible with the nginx image but if so, that's much better. 

Can the conf file support environment variables?

rcludwick:
Yes.... through an env file.  Or though secrets.

rcludwick:
Okay ... this is what I got working tonight.  @SchrodingersGat Tell me what you think:

nginx.conf:  (Note: proxy_pass needs to use whatever the container name and port is in docker, in this case I use "inventree-web")
```
server {
    listen 80;
    location / {
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_buffering off;
        proxy_request_buffering off;
        proxy_pass http://inventree-web:8080;
    }

    location /static/ {
        root /var/inventree/www;
        autoindex on;
    }

}
```

Then my docker-compose.yml looks like this: 

```
  inventree-web:
    container_name: inventree-web
    image: inventree/inventree:latest
    networks:
      - proxy
      - db
    expose:
      - 8080
    depends_on:
      - db
    volumes:
      - ./data/inventree/data:/home/inventree/data
      - ./data/inventree/static:/home/inventree/static
    environment:
      - INVENTREE_DB_ENGINE=postgresql
      - INVENTREE_DB_NAME=inventree
      - INVENTREE_DB_USER=pguser
      - INVENTREE_DB_PASSWORD=pgpasswd
      - INVENTREE_DB_PORT=5432
      - INVENTREE_DB_HOST=db
    restart: unless-stopped

  inventree-worker:
    container_name: inventree-worker
    image: inventree/inventree:latest
[...]

PR #1476, opened 2021-04-18
Cleanup docker files

SchrodingersGat:
Simplification of docker files, conf files, etc.

Closes https://github.com/inventree/InvenTree/issues/1457

Docs PR - https://github.com/inventree/inventree-docs/pull/102


docker/docker-compose.yml
```diff
@@ -50,21 +50,26 @@ services:
...
         restart: unless-stopped
 
     # nginx acts as a reverse proxy
     # static files are served by nginx
     # web requests are redirected to gunicorn
+    # NOTE: You will need to provide a working nginx.conf file!
     nginx:
         container_name: nginx
-        image: inventree/nginx:latest
+        image: nxinx
```
rcludwick, line 79: Typo.  :)

rcludwick:
So with my default config I'm getting something that looks like this: 

```
inventree-web       | Could not connect to database: FATAL:  database "inventree" does not exist
db                  | 2021-04-18 08:50:10.614 UTC [113] FATAL:  database "inventree" does not exist
inventree-worker    | Could not connect to database: FATAL:  database "inventree" does not exist
```

It's easy to fix for me, but it would be nice if it were configured.

SchrodingersGat:
@rcludwick currently (as per the [docker setup guide](https://inventree.readthedocs.io/en/latest/start/docker/#create-database) ) the database is created as a manual step in the setup process.

Are you suggesting that the database should be created automatically? Is this "safe"?

docker/docker-compose.yml
SchrodingersGat: Thanks, fixed now.

rcludwick:
I pretty sure you just need to run django migrations to create the db for you.  And it's safe.  So django migrations won't destroy any existing DB that exists.

rcludwick:
Looked over your docker changes and they look good to me.

SchrodingersGat:
I do not think that you can auto-create a db (except if you're using the sqlite backend):

https://docs.djangoproject.com/en/3.2/intro/tutorial02/#database-setup

rcludwick:
It looks like you're right.   I use postgres at work for my development database, and I swear I didn't have to, but it's also 3:30am here so I can't really remember right now.  :smile: 

What's weird is that unit tests have no issues creating databases on postgres.... So ...

Issue #1477, opened 2021-04-18
inventree data leak


rcludwick:
Inventree leaks the DB password in the logs.

Probably best to remove that line or replace it with a sha-256 hash.

SchrodingersGat:
Great catch @rcludwick - can you please check the referenced PR to ensure that I've fixed the issue at the right spot.

PR #1478, opened 2021-04-18
Remove log output which was leaking password into the logs

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/1477


rcludwick:
Pulled the latest docker image today and validated it.  Thanks!

PR #1479, opened 2021-04-18
Feature/translations

matmair:
@SchrodingersGat turns out I still missed a few things, I am too used to English it seems.

In this PR there are translations for:
- the yes/no partial
- a new formatter in the BOM-table for the field *optional* - with translations
- a new template-tag that returns a highlighted link for an object - currently used in a few alerts. This makes translation-strings cleaner while also allowing for more flexible grammar-use
- changing of a lot of translations from *trans* to *blocktrans* to enable translation of whole sentences. You had to bend the German grammar quite hard to translate those before. And the segments tended to shift in the po-files, which makes translations quite hard
- enabled translations for a few new functions from @eeintech 
- added translations for German - a few are still missing

### DIFF
--- a/InvenTree/build/templates/build/build_base.html
+++ b/InvenTree/build/templates/build/build_base.html
@@ -108,7 +108,7 @@ src="{% static 'img/blank_image.png' %}"
         <td>
             {{ build.target_date }}
             {% if build.is_overdue %}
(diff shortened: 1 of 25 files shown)


SchrodingersGat:
Nice work again :)

PR #1481, opened 2021-04-18
Instance name in title

### LINKED ISSUE #1463: use instance name in title and views?

matmair:
closes #1463 

- adds a setting for choosing if 'InvenTree' or the instance name should be used in views and title -> defaults to 'InvenTree'
- adds a function to easily query the for that string
- adds a template tag for easy use in templates
- implements new function in current views

### DIFF
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -58,6 +58,13 @@ class InvenTreeSetting(models.Model):
             'description': _('String descriptor for the server instance'),
         },
 
+        'INVENTREE_INSTANCE_TITLE': {
+            'name': _('Use Instance Name'),
+            'description': _('Use the instance name in the Titel-Bar'),
+            'validator': bool,
+            'default': False,
+        },
+
         'INVENTREE_COMPANY_NAME': {
             'name': _('Company name'),
             'description': _('Internal company name'),
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -59,8 +59,8 @@ class InvenTreeSetting(models.Model):
         },
 
         'INVENTREE_INSTANCE_TITLE': {
-            'name': _('Use Instance Name'),
-            'description': _('Use the instance name in the Titel-Bar'),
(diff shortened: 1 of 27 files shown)


matmair:
@SchrodingersGat are there dev-docs for template tags / templates? I didn't find any so there is no docs for the new tag at this point

InvenTree/common/models.py
```diff
@@ -58,6 +58,13 @@ class InvenTreeSetting(models.Model):
             'description': _('String descriptor for the server instance'),
         },
 
+        'INVENTREE_INSTANCE_TITLE': {
+            'name': _('Use Instance Name'),
+            'description': _('Use the instance name in the Titel-Bar'),
```
SchrodingersGat, line 63: "Titel" should be "Title"

SchrodingersGat:
Nice work @matmair a very clean implementation.

> are there dev-docs for template tags / templates? I didn't find any so there is no docs for the new tag at this point

Currently there are no separate developer docs...

InvenTree/common/models.py
matmair: oh sorry, my german is showing. Will fix ASAP

matmair:
> 
> 
> Nice work @matmair a very clean implementation.
>
I love Django for how easy KISS is with it :-)
Thank you, I am trying to use all the interfaces you laid out @SchrodingersGat

SchrodingersGat:
Nice work, thanks

PR #1482, opened 2021-04-18
Stock price history

matmair:
It seems like price-history is not used currently. This is WIP as i want to check it the tests are running and coverals seems to not work on my fork for some reason.


SchrodingersGat:
Coveralls won't work on your fork because the coveralls repo is linked to the InvenTree github account : )

matmair:
@SchrodingersGat is there a reason why there are no graphs in InvenTree? In this WIP - PR I implement [chart js](https://github.com/chartjs/Chart.js). It is MIT licensed and only 172 KB in size, but I am flexible in what to use.

SchrodingersGat:
> @SchrodingersGat is there a reason why there are no graphs in InvenTree? 

Yes - to date there has been no historical data to "graph". However I would love to support charts.

If you are going to add graph support here (and chart.js looks great) then I would like to implement a 'stock level history' graph - and I'm sure there are other areas it could be used too!

Excited to see what you come up with here :)

matmair:
@SchrodingersGat that would be all for the first implementation

matmair:
@SchrodingersGat ready for merge if you are, stock history will be done on a later date

InvenTree/order/models.py
```diff
@@ -254,7 +254,11 @@ def add_line_item(self, supplier_part, quantity, group=True, reference=''):
             if matches.count() > 0:
                 line = matches.first()
 
-                line.quantity += quantity
+                # update quantity and price
+                quantity_new = line.quantity + quantity
+                line.quantity = quantity_new
+                if line.purchase_price:
+                    line.purchase_price = supplier_part.get_price(quantity_new) / quantity_new
```
eeintech, line 261: Is there any case for which `quantity_new` could be equal to 0? (eg. both `line.quantity` and `quantity` both equal to 0 or cancelling each other)

eeintech: Also maybe handle the case where `get_price` returns `None`?

InvenTree/order/views.py
```diff
@@ -995,6 +996,7 @@ def post(self, request, *args, **kwargs):
 
                 part.order_supplier = supplier_part.id if supplier_part else None
                 part.order_quantity = quantity
+                part.purchase_price = supplier_part.get_price(quantity) / quantity if supplier_part else None
```
eeintech, line 999: Same here, `quantity` can be 0 and `get_price` can return `None`

InvenTree/order/views.py
```diff
@@ -1095,7 +1097,10 @@ def order_items(self):
                         sp=item.order_supplier))
                     continue
 
-                order.add_line_item(supplier_part, quantity)
+                # get purchase price
+                purchase_price = item.purchase_price if item.purchase_price else None
```
eeintech, line 1101: Wouldn't `purchase_price` be `None` already if it is not set?

eeintech:
@matmair 

> stock history will be done on a later date

What data would the current graph show right now? I haven't pulled your branch to check :smile:

matmair:
@eeintech the purchase prices and current stock numbers at that price (on separate scales). The data gets read from the StockItems.

InvenTree/order/models.py
matmair: > 
> 
> Is there any case for which `quantity_new` could be equal to 0? (eg. both `line.quantity` and `quantity` both equal to 0 or cancelling each other)

No, that section is only called if there is already a line for that supplier part - which must be 0 or higher and you can not add a part with a quantity 0 or smaller (see line 241).

> 
> 
> Also maybe handle the case where `get_price` returns `None`?

Is now included from line 261 following

InvenTree/order/views.py
matmair: quantity can not be 0 (line 992), price is handeld now

InvenTree/order/views.py
matmair: removed that bit

matmair:
@SchrodingersGat anything you would like to see changed in this PR? I would need some stuff of this for implementing #1484.

SchrodingersGat:
Wow, this looks amazing!! I had not had a chance to check it out yet, but I am very impressed by how it looks.

A couple of things popped up immediately.

### Currency Conversion

If different currencies are used, and no conversion exists, the following error occurs:

image

Have you taken currency conversion into account? When I used [django-money](https://github.com/django-money/django-money) for currency management, my intent was also to provide a conversion rate system. However this has not been implemented yet! Are you ignoring this for now, or managing currency conversion somehow on your end?

### Display

I think that this new information warrants a new "tab" rather than displaying in the modal form dialog.

matmair:
> A couple of things popped up immediately.
> 
> ### Currency Conversion

Seems like you did the hard part already, will add a check and error message into the PR
> 
> ### Display
> 
> I think that this new information warrants a new "tab" rather than displaying in the modal form dialog. 

I thought it would be fitting there as a single min-max seldom shows all information you might need when thinking about how to price something. Especially with how the chip-shortage is running the prices up rn.
Should I leave the rest of the pricing in the modal or move everything over into the tab?

SchrodingersGat:
Probably makes sense to move all the pricing information into the separate tab.

Another thought that has occurred to me - it looks like you are simply recording stock price history from StockItem entries? What happens when those entries are removed from the database? 

Also, should we be considering pricing from received purchase orders?

matmair:
@SchrodingersGat I started implementing and found a problem:  the modal is not only used in the part itself but also in the sales order to show prices for each LineItem. That feature is very handy so my proposal is to make a pricing tab with the additional Infos and keeping the modal (maybe adding a link to the tab there). Thoughts?

SchrodingersGat:
That sounds like a good compromise. A quick dialog for showing pricing info in various contexts sounds like it will still be useful

matmair:
@SchrodingersGat I moved the new stuff into a separate pricing tab. I have added a pie-chart showing from which BOM items the prices come from but without something like #1606 I think that makes not a lot of sense.
My stuff is finished but I would wait for #1600 to be merged before pulling this.

SchrodingersGat:
@matmair is this ready for review?

matmair:
@SchrodingersGat yes, is ready now that #1600 is merged.

SchrodingersGat:
Run into a "missing rate" error again

image

image

However I definitely have exchange rates available.

image

Any idea why this isn't working?

SchrodingersGat:
The issue is here: @eeintech 

https://github.com/inventree/InvenTree/blob/3a36e800c0f2338c55424a76b997a0b02e71363c/InvenTree/InvenTree/exchange.py#L48

Adding the currency suffix means that the django-money library cannot find the exchange rate objects associated with the backend:

https://github.com/django-money/django-money/blob/9b1e6ad7e93a3ea150c1cc7f97d00b2d0ff06b29/djmoney/contrib/exchange/models.py#L44

And thus is not finding the correct exchange rates.

Removing the offending line (which adds the suffix) fixes the problem here.

This raises a few potential problems / complexities:

Firstly, django-money library expects the `EXCHANGE_BACKEND` variable to be supplied in `settings.py` and cannot be altered at run time. This means that if the user switches between "custom" (manual) exchange rates and the automatic ones (via exchangerate.host) then money conversion will be an issue!

I would suggest the following (which I'm happy to implement if you're both on board):

We simplify the exchange "backend" approach by making a single back-end (e.g. `InvenTreeExchangeBackend`) which allows both manual and automatic updates

- If in "auto" mode the background worker task pulls exchange rate data from exchangerate.host automatically
- Data can also be pulled on request
- If in "manual" mode then the worker does *not* do this, and the user can manually edit the exchange rate values

This way there is only ever a single backend (which is hard-coded into settings.py)

eeintech:
@SchrodingersGat The reason behind adding the suffix was to store rates for multiple base currencies.
Let's say you decide to use `USD` as base currency, then switch the base currency to `EUR` and fetch rates, it would delete all the the initial `USD` rates... When using the API, you are just a click away from refreshing the rates but with manual control it is more tedious so wanted to take that pain away.

However, I agree with you the implementation should stay as close as possible to the `djmoney` config, which we deviated greatly adding the flexible option in #1600 of changing backend on the fly and removing `EXCHANGE_BACKEND` from the settings. It would take more effort to support this implementation.

I'm looking into #1581 and same thing the rates are not found because the backend is not loaded properly.

I'm fine with having `auto`/`manual` options for backend in `config.yaml` and `settings.py` loading the backend accordingly, sounds simple, leaves enough customization and would be compatible with `djmoney`.

matmair:
@SchrodingersGat @eeintech one backend seems like the best option. Especially if the InvenTree admins are not server admins and do not have the privileges to change files on the system.

eeintech:
I guess we all agree on the direction then. On a side note, I probably won't have time to push a PR for the rest of the week... sorry :disappointed:

SchrodingersGat:
Ok, I'll have a look at implementing something today so @matmair can merge it into this branch

SchrodingersGat:
@matmair please merge latest master in here and then I'll have another final look. I think it should be good then.

I'd like to get this merged in for 0.2.2 release :)

matmair:
@SchrodingersGat would be a go from my side

SchrodingersGat:
Great work, thanks again @matmair

PR #1483, opened 2021-04-18
german translations

matmair:
- translated new strings
- fixes for a few translations


matmair:
There are merge-conflicts that I need to investigate, will open again when fixed

Issue #1485, opened 2021-04-18
performance optimisation


matmair:
## How I noticed
I am currently transitioning from testing InvenTree in small academic / teaching environments (max. 10 concurrent users) to a larger pool before deployment to est. 100 concurrent users.

With that came a switch to quite powerful production webservers and bare-metal configs instead of docker. Out of curiosity I measured the performance of the new docker-image and the performance is more than 2x slower than an apache2 + mod_wsgi / postgres bare-metal config on the same server.

I found that no kind of expiry or cache control is set for the static content in docker. Also, the reloading of all the 'dynamic' JS-files seems to cause quite a lot of blocking-time.

# Ideas - RFC
@SchrodingersGat @eeintech
My understanding of the layout of InvenTree is limited so I would ask you to weigh in on these ideas to make InvenTree quicker:
- [x] enabling [manifest storage](https://docs.djangoproject.com/en/3.2/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage) and enable long-time caching in the docker config (I already do that on my bear-metal config and it makes everything considerable faster)
- [x] only load needed JS-files per view (not really needed with caching enabled)
- [x] rendering the 'dynamic' JS stuff once while updating the app with the translations backed and adding it to the static content

SchrodingersGat:
@matmair thanks for starting this conversation! Performance has been a core design criterion for me, so I'm very happy to support improvements in this area.

> only load needed JS-files per view (not really needed with caching enabled)

This is a good idea, and a pretty low barrier to implementation. Currently all the JS files are loaded via the "base.html" template as it's simpler 

> enabling manifest storage and enable long-time caching in the docker config (I already do that on my bear-metal config and it makes everything considerable faster)

I have not looked into this before, but it looks like a great idea. Approximately how much of a speed improvement would you typically expect?

> rendering the 'dynamic' JS stuff once while updating the app with the translations backed and adding it to the static content

This I think is a key improvement. Currently the server has to render each javascript file each time it is requested! So somehow we'll need to split these out into separate locale files, something like:

- part.en.js
- part.de.js

and then serve these files statically (which means Gunicorn does not even need to touch them).

The "conversion" process could be handled in the current `invoke static` script - translate the files before copying them over to the static directory.

----------------------

In general terms, where are you seeing the main performance hits currently? And what are the major points of difference between a docker setup and a bare-metal setup?

matmair:
[...]

PR #1495, opened 2021-04-20
performance improvments docker

matmair:
the rest of the suggestions from #1485 
just so docker uses the new methods and gains a bit performance


SchrodingersGat:
Thanks @matmair

Issue #1459, opened 2021-04-13
advanced translation pipeline


matmair:
Hi @SchrodingersGat I thought a bit about how one could streamline the translation pipeline - especially for people who are not familiar with Djangos language system. An OS-project I love and use ([Snipe-It](https://github.com/snipe/snipe-it)) uses crowdin. I setup a test [check it out here](https://crwd.in/inventree) which watches the django.po file in my [master](https://github.com/matmair/InvenTree/tree/master) and pushes all changed translation into [a translation branch](https://github.com/matmair/InvenTree/tree/l10n_master). Coupled with an action that regenerates the .po with each commit that could greatly automate the whole translation system and maybe enable more languages.
Crowdin is free with advanced features for [OS](https://crowdin.com/page/open-source-project-setup-request).

matmair:
I found a flow diagram I made for another thing, shows how the process / dataflow could work.
![flow_diagramm_crowdin](https://user-images.githubusercontent.com/66015116/114789127-e51b1a00-9d82-11eb-9565-ae72f727d99e.png)

SchrodingersGat:
Interesting, I had not heard of this before. I've been wondering for a while how to make the translation process easier for anyone to contribute.

Are there any examples of using crowdin with a django project in particular? I'm trying to find some django specific examples without much luck so far.

matmair:
I do not know of any examples with Django but I configured the shown configuration to try it out in my fork.

matmair:
@SchrodingersGat the crowdin project with CI enabled is hosted [here](https://crowdin.com/project/inventree1).
The [github action](.github/workflows/translations.yml) creates the current po files and pushes them to the [base branch](https://github.com/matmair/InvenTree/blob/l10_base/InvenTree/locale/en/LC_MESSAGES/django.po).
The crowdin translations are pushed every 10 minutes to [this branch](https://github.com/matmair/InvenTree/tree/l10n_trans/InvenTree/locale)

SchrodingersGat:
@matmair I've had a look at the crowdin page, it looks **really** good - I'd like to make this the "official" way to do translations :)

How do we proceed from here?

FYI the github action link you posted doesn't work

SchrodingersGat:
https://github.com/inventree/InvenTree/pull/1498 adds some more locale entrypoints too

SchrodingersGat:
And looks like crowdin supports translation status badges we can include in README.md

https://store.crowdin.com/products/bds

[...]

PR #1500, opened 2021-04-21
translations action

matmair:
- requires a base-repo for the translation files - in this case named 'l10_base'
- action should only be run on master to preserve the translation files
- see #1459 for background


SchrodingersGat:
By "base repo" do you actually mean branch?

matmair:
> 
> 
> By "base repo" do you actually mean branch?

branch or repo - the translations build up quite a lot of commits. Or maybe squash all the commits on merge to master. You get one commit per language, every time a file is updated.

SchrodingersGat:
Alright, I've made a separate repo - https://github.com/inventree/translation

SchrodingersGat:
And I've claimed the crowdin project space again - https://crowdin.com/project/inventree

Also I've submitted a request for an open source license for the project

matmair:
Ok now we need to add the following file to the root of the master-translation branch [crowdin](https://github.com/matmair/InvenTree/blob/9ee0d6982d9c8d40a61b0bed15367c82dc4c678a/crowdin.yml).

Then add the github-integration [link for this setting](https://crowdin.com/project/inventree/settings#integration-github). It should auto-detect the settings from the file.

Then the paths in the github-action need to be adapted (replace origin/l10_base or l10_base with the path to the translation-repo). I would set the action to run on every push to the master.
After the first action-run the new django.po should be pushed to the translation repo, where crowdin will pick it up and load all the strings, contexts etc. Every time a translation is made a few minutes afterwards the latest translation will be push from crowdin to the translation repo and before every release you can merge/squash the translations for all languages into the master.

SchrodingersGat:
I've set up the `origin/l10` branch to be the link to crowdin, I think that will be easier to work with than using a separate repo.

Can you adjust your workflow script to match?

.github/workflows/translations.yml
```diff
@@ -0,0 +1,62 @@
...
+        git fetch
+        git checkout origin/l10_base -- `git ls-tree origin/l10_base -r --name-only | grep ".po"`
+        git reset
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install Dependencies
+      run: |
+          sudo apt-get update
```
SchrodingersGat, line 36: As the translation (`*.po`) files are already committed, do we need to run `invoke translate` as a CI step?

matmair: - it first overrides the *.po-files to keep the changes in the structure of the po files as small as possible
- make translations to adapt the po
- push the updated po-file

matmair:
@SchrodingersGat do you mind canceling all the actions running? It will take forever to make those checks.

matmair:
@SchrodingersGat the action-bot is missing the permission to push changes. Could you set a secret for that?

SchrodingersGat:
Having some more time to think on this, I am wondering:

Is there any real disadvantage to letting crowdin work off the master branch? That way we do not need any special hooks to keep the translation branch up to date with the separate translation branch. And any PR's made by crowdin can be squash merged straight back into master.

If you think there is a major downside, please let me know. Otherwise, the much simpler path to me seems to be running from master.

matmair:
@SchrodingersGat django tends to need an action for translations because people normally forget to run the command which makes all contexts invalid -> lookup for translations is more difficult.
Also, if you do translations and merges manually you tend to get invalid lookup-files, which breaks the compiling-process (e.g. if you have two msgid with the same value everything errors out - you would need to run a full translation-cycle on each merge to make sure that is not happening).

With the pipeline always using the dedicated branch and never using submitted translations you prevent that.

SchrodingersGat:
Ok, yeah that makes sense then. So a separate branch sounds like the best way to go then :)

SchrodingersGat:
> @SchrodingersGat the action-bot is missing the permission to push changes. Could you set a secret for that?

What do I need to do on my end?

SchrodingersGat:
ok so I have done a bit of a "deep dive" reading about this, and subsequently learned a lot of new information...

In short, GITHUB_TOKEN has read-only access when an action is triggered from a fork (e.g. a pull request).

So this workflow is *only* going to be able to run once it is actually merged.

SchrodingersGat:
Ok, looks like it's up and running now!

SchrodingersGat:
@matmair I think something might not configured correctly. A few people have started translating, and proofreading, which is great! But there is no sign of crowdin pushing these back to the github repo.

Does it just take a while between sync events? 

image

Almost 24 hours between the two branches now..

matmair:
@SchrodingersGat that is odd, takes a few minutes normaly. Will investigate

SchrodingersGat:
Any ideas what's going wrong here?

matmair:
@SchrodingersGat I re-setup my environment and would like you to check the following:
- remove the integration for the `l10` branch,
- delete the `l10_translation` branch in GitHub
- reconnect the `l10`  branch in Crowdin

Crowdin should create the branch you name as a translated branch automatically if it does not there is our problem. I had issues with a PAT without the right privileges once.

If those steps do not work and you just authorized via OAuth try repeating the steps but use another branch name than `l10_translation`, if even that is not working, try to use the recommended name.

I am not sure what is causing the problem with pushing as pulling source-changes seems to work fine.

SchrodingersGat:
I've raised a ticket with the crowdin support team - we will see what they say

matmair:
@SchrodingersGat have you heard back? I would be interested what caused the issue.

SchrodingersGat:
No, nothing yet unfortunately.

SchrodingersGat:
Actually it looks like there's *something* working now - although the branches have got out of sync and there's merge conflicts:

https://github.com/inventree/InvenTree/pull/1534

matmair:
Interesting, it seems to work now. The last update is 15 hours ago, that is also the last time new strings were approved. Maybe your support-mail already worked ;-).

SchrodingersGat:
Hopefully!

PR #1505, opened 2021-04-21
Enabled default email sender setting

eeintech:
Was just testing the email reset form and found out that Microsoft requires the sender email address to match the username so I added this setting to the `config_template.yaml` and `settings.py`.

Also in `settings.py`, it looked like the default backend (`django.core.mail.backends.smtp.EmailBackend`) was not loaded if not found in the config, however it is commented in the `config_template.yaml` so I think it was a mistake and corrected it.


SchrodingersGat:
Thanks, this seems like a good fix.

Issue #1438, opened 2021-04-04
StockItemTracking actions are not mulitlingual


matmair:
The StockItemTracking adds static messages to the database; that poses be a problem with translations / multilingual installations.
Currently the action-note gets translated to the language of the current user at the time the action is executed. So when an user of a different language access the history for that stock item he gets the message in the language of the user who made the action.

I do not have a concrete idea how to solve that problem. Maye an action-message-id that gets saved alongside with all arguments needed for the string-format? The translation could be done on load. There might be performance implications with a lot of Tracking-Items but usually Django is quite fast with these kinds of actions.

What do you think @SchrodingersGat? That is quite a significant architectural decision.

Issue #1303, opened 2021-02-16
Stock status has no tracking history


jnewlands:
It would be nice to track when a stock item is destroyed and add a note (much like how a location transfer is tracked at the moment).

This is part of a broader issue (#1056) where one might wish to track _any_ changes to a stock item. @eeintech 
 and @SchrodingersGat mention [django-simple-history](https://github.com/jazzband/django-simple-history) as a potential solution.

eeintech:
@jnewlands There is also [django-reversion](https://django-reversion.readthedocs.io/en/stable/) as an alternative to django-simple-history, it gives the ability to "roll-back" any instance in a previous state.

I actually don't know that those packages are needed for stock, the current "home-made" history seems fine to me, as long as destroyed stocks can still be accessed through a filter on the Stock page (and filtered out by default).

matmair:
@SchrodingersGat the package(django-simple-history) itself is great but I have had problems when people **restore** data. Used as a fancy log it worked fine.
Some stuff to consider: 
- values in unique fields might be taken again - creating conflicts
- if you have a lot of signals in your model restores might start a lot of tasks - in one case that led to data-continuity problems in a multi-user environment, which are hard to debug if your dataflows were not designed to be rolled back in the first place
- foreign keys might have changed too, that can lead to very strange problems and tends to be hard to figure out afterwards

SchrodingersGat:
All very reasonable points. A good is example is if stock was moved from a location which has now been deleted - there's no way to "undo" that in a single action!

Maybe this is a case of "asking the wrong question". What I am looking to do is refactor the current "stock history" implementation to solve a number of outstanding issues:

- Stock "status" changes are not tracked - https://github.com/inventree/InvenTree/issues/1303
- Stock history items are not properly translated (you know this one)
- Stock history items are not "hyperlinked" - e.g. if you change location, I'd like to be able to click on that location (if that location still exists)

I *think* that the django-simple-history solution is a fine one here, but maybe it is overkill. Hooking into the signals might also be overbearing, we could potentially extend the current functionality of stock-tracking table to include location foreign keys, and status information...

matmair:
I mean if you do not plan to restore anything I do not see a problem, then it might be a great + fast solution.

Stock history items would be translated now 👼. How one would implement that with an automatically generated record/table I would have to check. But passing through custom change-messages will be necessary anyway because the defaults do not mean anything to users.

matmair:
1) this feature was implemented in #1572 
[...]

PR #1520, opened 2021-04-26
multilingual Stocktake actions

matmair:
PR for #1438

- [x] tests?
- [x] use right translations in UI
- [x] how to keep old translations around for backwards-compatibility


matmair:
Default fallback when no translation available: English! Gets quite difficult to maintain otherwise - with vastly decreasing returns.

SchrodingersGat:
@matmair I would suggest putting this one on hold for a little while. I am currently looking into how to remove the `StockItemTracking` model completely, in favor of a more robust "stock item history tracking" approach - see https://github.com/inventree/InvenTree/issues/1303

So this PR would conflict with that change a lot!

The same goes for - https://github.com/inventree/InvenTree/pull/1482

I'm keen to address both these issues but not until the stock history tracking framework is refactored.

matmair:
@SchrodingersGat the functionality is done already, it just needs a bit of polishing / testing.

I have had bad experiences with rollbacks in django-simple-history in other projects before but maybe I can learn from your implementation.

SchrodingersGat:
@matmair I have not settled on django-simple-history yet, if there's some issues you have come across I'd be interested to know!

matmair:
I have posted my thoughts to #1303 as it seems that that is the issue regarding the implementation.

matmair:
@SchrodingersGat made a PR with a better system in #1572.

Issue #1519, opened 2021-04-26
More syntax for adding serial numbers


matmair:
I will add the following two options to serial number-recognition:
\<start>+ :add expected number of serials after \<start>
\<start>+\<length> :add \<length> numbers starting at \<start>

I am thinking of also including a character for filling a pattern with all missing serials (to get the expected number of serial numbers).
What do you think @SchrodingersGat ?

SchrodingersGat:
I like it! Both suggestions I would support.

eeintech:
Sweet! Nice job!

PR #1521, opened 2021-04-26
More syntax for adding serial numbers #1519

matmair:
PR for #1519

- [x] documentation
- [x] maybe extra syntax


SchrodingersGat:
Simple clean change, thanks for the addition :

matmair:
@SchrodingersGat I left out a 'all remaining' symbol as it would see virtual no usage. Also in my testing / writing the docs it made the options more difficult to explain.

SchrodingersGat:
Sure, understood. I think the additions you have made here are sufficient!

PR #1522, opened 2021-04-27
Add owner model to imports ignore list

eeintech:
It was throwing a unique constraint error

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -248,6 +248,7 @@ def content_excludes():
         "django_q.schedule",
         "django_q.task",
         "django_q.ormq",
+        "users.owner",
     ]
 
     output = ""


SchrodingersGat:
Good catch :)

SchrodingersGat:
Actually, maybe I merged this too hastily. Is the users.owner model completely "calculated" when the database data is imported? Can we safely exclude this from the data migration files?

eeintech:
@SchrodingersGat Yes all owners are automatically created/updated when the user app is loaded: https://github.com/inventree/InvenTree/blob/9feed0dcdc4da0f6022133a7769263c54e853bd1/InvenTree/users/apps.py#L44

PR #1523, opened 2021-04-27
Updated "Required for Build Orders" API queryset

eeintech:
Seems like it was broken since the introduction of incremental build


SchrodingersGat:
Ah good catch @eeintech thanks

Issue #1527, opened 2021-04-28
Set user language per session


matmair:
Support setting the user language in the settings (like the theme). On a per-session basis using the included [Django tools](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#the-set-language-redirect-view).

PR #1529, opened 2021-04-28
per session user-language settings

matmair:
Using the standard cookie; see #1527


SchrodingersGat:
Thanks @matmair - was not aware of the set_language convenience functions! Great addition :)

matmair:
Django comes decked out. I am finding cool built-in stuff with every project.

PR #1530, opened 2021-04-28
Fixes to create build output form

eeintech:
Added validation on quantity field and removed image tag `hover` argument which was falsely creating the entire form being a single link or something like that.


matmair (approved):
I was wondering if that was an error 👼

SchrodingersGat:
Thanks @eeintech

Issue #1512, opened 2021-04-22
Part - get_price


matmair:
@SchrodingersGat  get_price currently just returns None when called with a quantity lower than the lowest pricebreak.
2 questions:
- intentional?
- should the automatic calculation when a part is added warn if there if that results in no price added?

SchrodingersGat:
Not "intentional", perhaps it should just return the lowest quantity price break in this situation?

matmair:
Changing the default behaviour creates a lot of problems, I will not poke this hornets-nest further.

eeintech:
@matmair What kind of problems? Would be nice to have it return the lowest quantity unit price and calculate for the requested quantity.

eeintech:
#1178 was related. I think we should find a solution to this problem

matmair:
@eeintech stuff broke in a lot of places and I could not find out why everywhere.
I think it makes sense that you can set a minimum order quantity just by starting the price-breaks at the minimum quantity. But correct me if you think otherwise.

eeintech:
@matmair Does #1535 implementation works for you?

matmair:
In the core functions (purchase-orders, pricing information) it does but I get a wired error when deleting stock - which hinders me from getting rid of stock. Will investigate further and submit my findings.

eeintech:
Weird indeed, I just checked and I don't have this kind of error. Have you tried the same thing with `master` on your setup?

matmair:
I have reset the deployment in which I got the error, seems to run fine now.

PR #1535, opened 2021-04-29
Added support for fractional price and lower price break

### LINKED ISSUE #1178: No Pricing information found when fractional amount is used
I have a part which uses a cable as a sub-part. The cable comes in rolls of 3 meters, so I created a supplier part where 1 unit costs a third of the manual prize. 

If I use a fractional amount of 0.25 (representing a 25 cm piece of ribbon cable in my case) I get a hint that says there is no pricing information available:  
2020-12-20 14_08_59-Window  

If I use anything bigger than 1 it works as expected, even with 1.5:  
2020-12-20 14_11_22-Window  
2020-12-20 14_12_29-Window  

I also tried to enter a Prize-break for an amount <1, but this is not allowed. Expected behaviour would be that fractional amounts <1 cost a fraction of the prize.

eeintech:
Allows Part and BOM pricing with fractional quantity.
Also selects the lowest price break found if part quantity is below any price break quantity.

Fixes #1178 and is related to #1512

### DIFF
--- a/InvenTree/company/models.py
+++ b/InvenTree/company/models.py
@@ -6,7 +6,7 @@ Company database model definitions
 from __future__ import unicode_literals
 
 import os
-
+import decimal
 import math
 
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
Thanks @eeintech this seems pretty reasonable

PR #1558, opened 2021-05-06
Bug/calender

### LINKED ISSUE #1557: [BUG] calender not rendering on view-switch
In all views with calendars the calendar is not rendered when switching to the calendar, only on the first refresh. At least on WSL with Firefox or Chrome.

matmair:
Fixes #1557


SchrodingersGat:
Awesome, thanks @matmair

PR #1560, opened 2021-05-07
enable live-serving of i18n_static tag in debug

matmair:
funny that I did not see the need for that earlier


SchrodingersGat:
Very good catch indeed! Please address style issue and I'll merge :)

PR #1564, opened 2021-05-09
inventree.css cleanup

matmair:
A few clean-ups in the main inventree.css. The problems kept popping up while I was developing a few new things, so I fixed them.

Does not change functionality.


SchrodingersGat:
Thanks :)

Issue #997, opened 2020-09-24
Add "copy to clipboard" button to table cells


SchrodingersGat:
Like DigiKey, for example, add a button (which is only visible when you hover over a table cell) which copies the cell contents into the clipboard.

cdplayer:
assistive info:
https://w3c.github.io/clipboard-apis/
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

https://clipboardjs.com/ 3kGZIPd library.

https://www.iconfinder.com/icons/326595/content_duplicate_copy_icon
https://thenounproject.com/term/copy-to-clipboard/3548032/
![image](https://user-images.githubusercontent.com/184022/96974287-5c05f180-155c-11eb-8938-cc0d23ef6e88.png)

matmair:
I hacked this feature in 2 pages by hand
- a table
- a detail-page

and I find it makes using the table quite distracting as the button to copy always fades in.
My solution would be to add the copy function - using the great js library @cdplayer suggested - only to the detail-pages (see mock-up below - would be adjusted to fit the style)
@SchrodingersGat What is your opinion on that? Is this thing still relevant?

![grafik](https://user-images.githubusercontent.com/66015116/117555002-b0397480-b05b-11eb-85f1-e4945ad21eb6.png)

matmair:
@SchrodingersGat should we close this now?

matmair:
@SchrodingersGat do you consider this done or should additional functionality be implemented?

SchrodingersGat:
Thanks for reminding me! Happy to close this one out now.

Issue #725, opened 2020-04-13
Add `copy version information` button


a7hybnj2:
Add a button to the `InvenTree Version Information` window that copies all important data to the paste buffer for easier supplying that info on bug reports/issues. Or, format that window where it can be copied and pasted because when I do it now it looks like this:
```

  | Instance Name | 1e8f50a4e38e49969ec510f2f5dc91c1
-- | -- | --
 

```

<img width="1151" alt="Screen Shot 2020-04-13 at 5 26 31 PM" src="https://user-images.githubusercontent.com/36831279/79162833-f0241e80-7dab-11ea-808d-56ebf23d4e28.png">

the screen shot is what was selected when I hit copy.

SchrodingersGat:
Nice idea. It would a perfect one to contribute if you were able to :)

a7hybnj2:
Maybe I will take a look. I poked around the code before and was completely lost.

SchrodingersGat:
The about dialog is handled here:

https://github.com/inventree/InvenTree/blob/master/InvenTree/templates/about.html

If there are any other contributions you want to make, and you want me to point you in the right direction, please let me know. I'd be more than happy to assist.

SchrodingersGat:
@a7hybnj2 did you want me to assign this to you?

eeintech:
May be related to https://github.com/inventree/InvenTree/issues/997

matmair:
@SchrodingersGat @eeintech I finished the work on this, should be maybe add something about the deployment method?

SchrodingersGat:
> should be maybe add something about the deployment method?

What exactly do you mean here?

matmair:
> 
> 
> > should be maybe add something about the deployment method?
> 
> What exactly do you mean here?

If the installation uses docker (and if it is in development or production) or is setup manually. Maybe also if MySQL is used - that sometimes causes problems with search / filtering.

SchrodingersGat:
Ah I see, yes that is a good idea. How can you determine at run-time if docker is being used?

I would like to see, as additions to the "System Information" dialog:

- Database backend
- Is the server running in debug mode?
- running in docker or bare-metal? (If this can be determined)

SchrodingersGat:
See https://github.com/inventree/InvenTree/pull/1583

I have added database engine and debug mode

matmair:
I will think about how do detect docker, maybe an enviroment-variable set in the docker-file and read by the settings.py?

PR #1565, opened 2021-05-09
Clipboard UI

matmair:
Implements the functions for #997 and - as an example - copy-buttons in the part-detail view (/part/pk/).

If you want to add copy-functionality to any element just add `{% include "clip.html"%}` in the that elements text. This will include html-code that shows a button that copies the text-content of this element. The text is fetched at event-time so dynamic texts are no problem. The btton is light-grey till hvered and floats right.


SchrodingersGat:
Looks pretty good! Two small requests:

- Please add a mouse-over text (translated) to the effect of "copy to clipboard"
- When you hover over the button, the css padding changes and causes elements on the screen to shift around a bit

SchrodingersGat:
May as well address this issue too - https://github.com/inventree/InvenTree/issues/725

SchrodingersGat:
Thanks @matmair

eeintech:
Nice stuff @matmair !

PR #1576, opened 2021-05-12
copy-btn in all the places

### LINKED ISSUE #725: Add `copy version information` button

### LINKED ISSUE #1574: Adding copy-button to interface
@SchrodingersGat are you happy with the clipboard in function, appearance, and usage in the templates? I thought I check before implementing it all over the place.
I would add it to the detail-pages for all models (eg. titles, references, descriptions) and implement #725.

matmair:
also adds templates for mailto: and tel: links - just in case we need that in more templates (maybe a user-popup to directly contact people in maker-spaces?)

Closes #1574
Closes #725

### DIFF
--- a/InvenTree/InvenTree/static/script/inventree/inventree.js
+++ b/InvenTree/InvenTree/static/script/inventree/inventree.js
@@ -1,37 +1,28 @@
-function attachClipboard(selector) {
+function attachClipboard(selector, containerselector, textElement) {
+    // set container
+    if (containerselector){
+        containerselector = document.getElementById(containerselector);
+    } else {
+        containerselector = document.body;
+    }
 
-    var cis = new ClipboardJS(selector, {
-        text: function(trigger) {
-            var content = trigger.parentElement.parentElement.textContent;
-
-            return content.trim();
+    // set text-function
+    if (textElement){
+        text = function() {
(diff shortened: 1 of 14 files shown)


matmair:
Closes #997
@SchrodingersGat I cleaned up the js, would be ready for merge

eeintech:
@matmair @SchrodingersGat Ok so this button has a JS error popping up, at least on Chrome:

image

Looks like it has to do with the `trigger` variable not being declared, indeed I don't see where is it declared.

Also I found there is some leftover console logging:
https://github.com/inventree/InvenTree/blob/a1fa3315b17ea15c9d0487ae34081a36793b8a0f/InvenTree/InvenTree/static/script/inventree/inventree.js#L25

Is that intentional? It gives this in the JS log:

image

The copy to clipboard function is not working in my case... not sure if it is browser related.

PR #1581, opened 2021-05-13
Added purchase price range and average to BOM items/view

eeintech:
Looks like this:

image


matmair:
@eeintech looks nice. Would it be easy to format the prices in the correct format for the currency? Also how are different currencies and their respective value handled? 0,85$ is currently less than 0,75€.

eeintech:
@matmair I always forget about currency... I wanted just a simple aggregator of purchase prices over the stock items. I guess different currency would mess that up as you mentioned. I'll take a look into the currency conversion strategy unless there is one you have in mind already.

matmair:
@eeintech there is
`from djmoney.contrib.exchange.models import convert_money`
in #1482 I converted all prices to the default currency.

eeintech:
@matmair Alright I just added a bit of conversion to the mix, it's a naive implementation but not sure how to be more efficient in the conversion using SQL query functions...

eeintech:
@matmair I have trouble with the annotating the API DB query for currencies as it returns duplicates (one BOM item can return multiple currencies). I did not find a clear way to associate a currency to a minimum, maximum or average `purchase_price` value. 

My workaround is to just match the first currency found for key pair (bom_item.pk, bom_item.sub_part) and use it for the conversion. However, I _think_ this could lead to erroneous conversion step... it is actually a more complicated problem than I thought! Grrr

eeintech:
Actually, thinking a bit more about it, the conversion is not the problem. The accuracy of the purchase price information could be an issue, as it will use whatever first currency found for a part stock item. So if we assume users set the currency for ALL stock items associated to a part to the same currency, the information will be correct. Else it could be you see the wrong currency but the price data is correct.

matmair:
@eeintech seems to be a good compromise, maybe the model should be changed to include helper-functions to make calculations to the default currency and additions of different currencies simpler.
I will check that possibility next week when I am finished with the pricing stuff, I need to get the price-history done first.

eeintech:
@matmair Seems like a great idea. Ok to wait for that too for this PR if we get this done in the next few weeks, not so urgent

matmair:
@eeintech I will finish my open PRs first but then look at this. After a few issues, conversations off-git and explaining I think currencies need a bit love in doc and code.

eeintech:
WIth #1611 merged, I can see that the currency conversion is happening!

## No exchange rates
image

## With exchange rates
image

Ready to merge :+1:

SchrodingersGat:
Thanks @eeintech this looks fantastic :)

Issue #551, opened 2019-09-23
Receiving order and allocating line items 


Marbles87:
When Receiving orders please add following feature in the purchase order feature page.

- Add an options to allocate different items to different locations.
- Add a field/column to show what the component was ordered for and where to place it.
- Be able to back track the Po to find where parts were sent in case of misplacement and back orders

Is it possible to have some type of filter system to to be able to set item locations easier?

Not sure how to word this buuuuttt!

If ordering items for multiple parts. Have it referenced showing what item it was ordered for. so for the person closing the Purchase order in InvenTree knows where to place them.

![Capture4](https://user-images.githubusercontent.com/48982658/65402520-403c9300-de12-11e9-9150-ff41596cd12a.PNG)

Marbles87:
please provide a new status in the "purchase order details" to allow a new status from pending to received then to "Allocation"? to provide a stop gap to allow a palm from one person to another. When item is received the purchase officer can then check for all line items then pass on packages to intended rooms where they can then be further allocated to final destination (BINS, build kits).

![Capture5](https://user-images.githubusercontent.com/48982658/65573935-5cc20200-dfaf-11e9-8e39-7addd8820ab4.jpg)

eeintech:
@SchrodingersGat What are your general thoughts for entering parts from purchase orders?
I think what @Marbles87 hinted at sounds like a fair request, just wondering what you think.

SchrodingersGat:
An "Intended location" column would certainly be a handy addition - and could default to the "default location" for the particular part.

I've opened a new issue for adding the ability to track down stock items received against a certain purchase order

eeintech:
Looks like you were quick to jump on this, nice. Should we close this issue then?

SchrodingersGat:
I think leave open until we have implemented a "receive location" for a purchase order line item. I have a few ideas how to implement this, might be able to look at it today.

Kedarius:
What is the status on this? I was wondering if it would be possible to assign parts from a purchase order to a location. Just simple select items and click set to different location...

SchrodingersGat:
@Kedarius no further development on this one yet...

PR #1587, opened 2021-05-14
feat(purchase orders): show the preferred location for each PO Line

### LINKED ISSUE #1467: See or select the default location when receiving stock
When receiving parts on a PO, unless you happen to have the correct location of the part memorised, you have no idea what the correct location of received part is.

There are a couple of ways that this could be solved satisfactorily:
1) When receiving a single line item, automatically select the default location as the location destination.  I looked at the code and as far as I can tell, this is not going to be easy to accomplish as the table of line items to receive is filtered separately from the selection of locations to put the line items; so short of creating a purpose specific modal this probably not a great way forward.
2) Add the default location as a column to the table of parts to receive.  This should be fairly straight forward as the table seems to be populated the same way as all other tables.  

Rejected ideas:
1) Receive the items to "Desk" or something and then transfer them to the correct location.  This doesn't work because it is more unnecessary work, and more crucially, the transfer dialog has the same problem, it also doesn't show the default location and additionally requires a note to be added.

I think 2 is probably the expedient way forward, but would the project accept a change like that?

nwns:
Adds the ability for the Purchaser to specify where the item is
intentended to go when received.  If the Purchaser does not set a
preferred location, then the default location for the part is displayed.
If the item is received them where it was actually placed is shown.

NOTE: if an item is split when received only one of the resulting
StockItem location is used.

Fixes #1467
Addresses some of the requests in #551


SchrodingersGat:
Hey @nwns can you please address the failing tests here

nwns:
@SchrodingersGat Tests should be good now.

SchrodingersGat:
Thanks @nwns this seems to work very cleanly, nice feature!

PR #1588, opened 2021-05-15
Part import

matmair:
Enables Import of parts building on the great additions of @eeintech in #1561.

Missing:

- [x] tests
~~updating of parts? Maybe out of scope~~
- [x] category
- [x] default location


matmair:
@eeintech would you mind taking a look at this? I tried making the html-implementation a bit simpler.

InvenTree/part/views.py
```diff
@@ -718,6 +722,151 @@ def get_initial(self):
...
+                    continue
+
+                # general fields
+                row[idx.lower()] = data
+
+    def done(self, form_list, **kwargs):
+        """ Create items """
+        items = {}
+
+        for form_key, form_value in self.get_all_cleaned_data().items():
```
eeintech, line 799: We can maybe move this for-loop in `FileManagementFormView` as I feel this will be reused a lot

eeintech: `items` could be accessed with `self.items` instead

InvenTree/part/views.py
```diff
@@ -718,6 +722,151 @@ def get_initial(self):
...
+        'default_expiry': 'default_expiry',
+        'minimum_stock': 'minimum_stock',
+        'units': 'units',
+        'notes': 'notes',
+        'category': 'category',
+        'default_location': 'default_location',
+        'default_supplier': 'default_supplier',
+    }
+
+    def get_field_selection(self):
```
eeintech, line 756: Very interesting implementation which makes it very generic!
Could think of moving most of this code to the `FileManagementFormView` view instead?

eeintech:
@matmair I pulled your branch and here are my thoughts:

- Part import can be a "dangerous" operation, meaning not all users should be able to do it as it is too easy to make a mistake. I would advise that for database entries like parts we create an import section in the settings. Also, why not improving the import process in the admin section instead? Also, what is the use case for quickly importing parts into the database?
- Some of the field keys are lower case while other are "titled"

image

- The category matching needs thought, one should be able to match the PK or by name, maybe separating tree element by comma or slash? Like "Capacitor, Ceramic, 0603" will parse the category tree and try to match it to what's in the database.

InvenTree/common/forms.py
```diff
@@ -145,12 +145,14 @@ def clean(number):
...
+                    # Set field name
+                    field_name = col_guess.lower() + '-' + str(row['index'])
 
                     # Create input for required headers
                     if col_guess in file_manager.REQUIRED_HEADERS:
-                        # Set field name
-                        field_name = col_guess.lower() + '-' + str(row['index'])
                         # Set field input box
+
+                        # TODO maybe not here but in an own function?
```
eeintech, line 155: Could be a good idea to break it down by the type of file ('order', 'part', etc)

InvenTree/common/files.py
```diff
@@ -109,6 +111,33 @@ def setup(self):
             # Update headers
             self.update_headers()
 
+        # TODO maybe not here but in an own function?
```
eeintech, line 114: How do you see this? Like `setup_order` and `setup_part` methods? I think I prefer to keep the `setup` method generic and re-using `self.name`.

InvenTree/common/views.py
```diff
@@ -269,6 +270,15 @@ def get_form_kwargs(self, step=None):
         
         return super().get_form_kwargs()
 
+    def get_form(self, step=None, data=None, files=None):
```
eeintech, line 273: Good stuff here!

InvenTree/part/templates/part/import_wizard/match_fields.html
```diff
@@ -0,0 +1,99 @@
...
+    </thead>
+    <tbody>
+        <tr>
+            <td>{% trans "Match Fields" %}</td>
+            <td></td>
+            {% for col in form %}
+            <td>
+                {{ col }}
+                {% for duplicate in duplicates %}
+                    {% if duplicate == col.name %}
```
eeintech, line 58: Should actually be `{% if duplicate == col.value %}`, see #1582 for the fixes regarding duplicate highlighting

InvenTree/part/templates/part/import_wizard/match_references.html
```diff
@@ -0,0 +1,91 @@
...
+                </td>
+                <td>
+                {% add row.index 1 %}
+                </td>
+                {% for item in row.data %}
+                <td>
+                    {% if item.column.guess %}
+                        {% with row_name=item.column.guess|lower %}
+                        {% for field in form.visible_fields %}
+                            {% if field.name == row|keyvalue:row_name %}
```
eeintech, line 60: Nice :+1:

eeintech (commented):
I really like those improvements, I'm wondering if we can get them into the more generic implementation of the `FileManagementFormView` with associated templates so we can re-use them in all places with import option?

matmair:
> Part import can be a "dangerous" operation, meaning not all users should be able to do it as it is too easy to make a mistake. I would advise that for database entries like parts we create an import section in the settings. Also, why not improving the import process in the admin section instead? Also, what is the use case for quickly importing parts into the database?
> 
The environment in which I plan to use InvenTree has multiple teams. I do not want these teams - which will add a lot of parts - to use the admin. There are a lot of options which will confuse the user. We could but a permission-flag on that function.
One of the use cases I have is in a mid-size commercial environment with a lot of R&D going on. The other is a uni-adjacent lab with a high turnover of students - which should all use the system. That is also why I invest a lot of time into translation and ease of use.

>  Some of the field keys are lower case while other are "titled"
> 
> 
> image
> 

That is something I am working on, is a result of how headers are configured and used.


> The category matching needs thought, one should be able to match the PK or by name, maybe separating tree element by comma or slash? Like "Capacitor, Ceramic, 0603" will parse the category tree and try to match it to what's in the database.

I would like to keep it simple and only human-ready at first.  manly see this feature used to import an extended BOM to quickly get stuff going and keep the barrier of entry low.

> 
>     * Should we try to make some generic templates for the different steps of `FileManagementFormView` which can be re-used in all places?

Absolutely, I just wanted you to check how I did it first and gauge if the way I went is something you approve of. I try to not mess with others features.

InvenTree/common/views.py
matmair: this is what I mean by inherited functions - reduce the changes to a minimum

InvenTree/part/templates/part/import_wizard/match_fields.html
matmair: maybe the whole section should be generalised.

InvenTree/part/views.py
matmair: thank you
yes 👍

eeintech:
> The environment in which I plan to use InvenTree has multiple teams. I do not want these teams - which will add a lot of parts - to use the admin. There are a lot of options which will confuse the user. We could but a permission-flag on that function.
> One of the use cases I have is in a mid-size commercial environment with a lot of R&D going on. The other is a uni-adjacent lab with a high turnover of students - which should all use the system. That is also why I invest a lot of time into translation and ease of use.

I understand your use-case. However keep in mind that we don't use InvenTree for the same thing. So for me this function should be hidden somewhere, not in clear view. Staff users have access to InvenTree global settings so I would suggest to move it there, in the `Parts` settings maybe. I personally don't want our team users to be able to easily import parts from a file, but all of our users have all permissions for parts. So I don't know that permission will solve that.
If you intend your use-case to expand to other sections, I would really like to see it in the settings instead. It could be a specific section too, accessible by all users deemed allowed for "database imports".

InvenTree/part/views.py
matmair: will try that tomorrow.
Keeping items a funtion-scope variable is in preperation for inherited functions - keeps the effects of wrong inherited functions to this step and therefore debugable

InvenTree/common/files.py
eeintech: Makes sense :+1:

matmair:
> 
> 
> > The environment in which I plan to use InvenTree has multiple teams. I do not want these teams - which will add a lot of parts - to use the admin. There are a lot of options which will confuse the user. We could but a permission-flag on that function.
> > One of the use cases I have is in a mid-size commercial environment with a lot of R&D going on. The other is a uni-adjacent lab with a high turnover of students - which should all use the system. That is also why I invest a lot of time into translation and ease of use.
> 
> I understand your use-case. However keep in mind that we don't use InvenTree for the same thing. So for me this function should be hidden somewhere, not in clear view. Staff users have access to InvenTree global settings so I would suggest to move it there, in the `Parts` settings maybe. I personally don't want our team users to be able to easily import parts from a file, but all of our users have all permissions for parts. So I don't know that permission will solve that.
> If you intend your use-case to expand to other sections, I would really like to see it in the settings instead. It could be a specific section too, accessible by all users deemed allowed for "database imports".

So a permission for parts-import would solve that? The normal users would not see the button but team leads for example would. Putting stuff into settings manly means people will be asking a lot where the function is in my experience. @SchrodingersGat do you prefer a position for this?

SchrodingersGat:
I think that it makes sense to have it in the "parts settings" page - a little bit separate to the everyday views

InvenTree/part/templates/part/import_wizard/match_references.html
matmair: I think it will be easier to keep the templates like they are , will not be touching existing stuff.

InvenTree/part/views.py
```diff
@@ -718,6 +723,157 @@ def get_initial(self):
         return initials
 
 
+class PartImport(FileManagementFormView):
+    ''' Part: Upload file, match to fields and import parts(using multi-Step form) '''
+    permission_required = 'part.add'
+
+    class MyFileManager(FileManager):
```
eeintech, line 730: Name it `PartFileManager`? "My" is not very explicit

InvenTree/order/views.py
```diff
@@ -572,7 +576,58 @@ def post(self, request, *args, **kwargs):
 class PurchaseOrderUpload(FileManagementFormView):
     ''' PurchaseOrder: Upload file, match to fields and parts (using multi-Step form) '''
 
+    # overriden classes
+    class MyMatch(cm_forms.MatchItem):
```
eeintech, line 580: I don't know if that's good Django practice, nesting a form class definition inside a form view..? What do you think @SchrodingersGat?

InvenTree/order/views.py
```diff
@@ -572,7 +576,58 @@ def post(self, request, *args, **kwargs):
...
+                return MoneyField(
+                    label=_(col_guess),
+                    default_currency=InvenTreeSetting.get_setting('INVENTREE_DEFAULT_CURRENCY'),
+                    decimal_places=5,
+                    max_digits=19,
+                    required=False,
+                    default_amount=self.clean_nbr(row.get('price', '')),
+                )
+
+    class MyFileManager(FileManager):
```
eeintech, line 611: Name it `OrderFileManager` instead?

InvenTree/order/views.py
eeintech: Maybe better declaring it in `order/forms.py`?

eeintech: Also the name should be `OrderMatchItem`, it should be explicit

eeintech:
I did a quick review and check of the feature for the purchase order, the step 3 validation fails to proceed when rows were removed from the view. Below I removed row 3 but when clicking on "Submit Selections" it did not validate the form and returned empty quantity field:
image

Also I see that you kept the "Part Import" view in the navigation bar of the "Part" section. Why not moving it to the settings?
I saw permission is set for staff users only so I guess that could work but I'm still not convinced this is the right place...

InvenTree/order/views.py
matmair: @eeintech I would not see this as a normal form - it just overrides one function. I could refactor it to look more like a override but that makes it more difficult to read.

matmair:
> 
> 
> @matmair I did a quick review and checked the features for the purchase order import process, the step 3 validation fails to proceed when rows were removed from the view. Below I removed row 3 but when clicking on "Submit Selections" it did not validate the form and returned empty quantity field:
> 
I cannot reproduce that, could you try again and check if it throws the same error with another PO? I am not sure why I cannot reproduce - seems to be a straight-forward problem.
> 
> Also I see that you kept the "Part Import" view in the navigation bar of the "Part" section. Why not moving it to the settings?
> I saw permission is set for staff users only so I guess that could work but I'm still not convinced this is the right place...
>
I am working on a method to change the position of buttons without forking the project or adding tons of settings. I respect your opinion where the button should be but I do not believe people will use it if it is buried.

eeintech:
> > @matmair I did a quick review and checked the features for the purchase order import process, the step 3 validation fails to proceed when rows were removed from the view. Below I removed row 3 but when clicking on "Submit Selections" it did not validate the form and returned empty quantity field:
> 
> I cannot reproduce that, could you try again and check if it throws the same error with another PO? I am not sure why I cannot reproduce - seems to be a straight-forward problem.

I have tried different purchase order files and always have the same result. I just checked on master and everything is working.
Try removing rows in step 3 and clicking on "Submit Selections", those rows should pop up again and show you an invalid form view.

eeintech:
@matmair Given this has a lot of multi-form implementation updates, do you think this will be ready for review soon? The test unit for that wizard is still on my TODO list and I would like to give it a shot soon :smiley:

InvenTree/common/forms.py
```diff
@@ -176,22 +170,14 @@ def __init__(self, *args, **kwargs):
...
-                            self.fields[field_name] = forms.CharField(
-                                required=False,
-                                initial=value,
-                            )
+                        self.fields[field_name] = forms.CharField(
+                            required=False,
+                            initial=value,
+                        )
+
+    def get_special_field(self, col_guess, row, file_manager):
```
eeintech, line 181: Shouldn't it return something by default, `None` maybe?
What would happen with `overriden_field` if it is not defined in subclass?

InvenTree/order/views.py
```diff
@@ -572,7 +576,58 @@ def post(self, request, *args, **kwargs):
...
+            elif 'price' in col_guess.lower():
+                return MoneyField(
+                    label=_(col_guess),
+                    default_currency=InvenTreeSetting.get_setting('INVENTREE_DEFAULT_CURRENCY'),
+                    decimal_places=5,
+                    max_digits=19,
+                    required=False,
+                    default_amount=self.clean_nbr(row.get('price', '')),
+                )
+
```
eeintech, line 610: Return `None` by default?

InvenTree/part/views.py
```diff
@@ -719,6 +724,168 @@ def get_initial(self):
         return initials
 
 
+class PartImport(FileManagementFormView):
+    ''' Part: Upload file, match to fields and import parts(using multi-Step form) '''
+    permission_required = 'part.add'
+
+    class PartFileManager(FileManager):
```
eeintech, line 731: Move to `InvenTree/common/files.py` for re-use?

InvenTree/part/views.py
eeintech: This is still relevant, hasn't been moved but marked "resolved"..?

InvenTree/common/views.py
```diff
@@ -362,20 +401,24 @@ def set_form_table_data(self, form=None):
...
                     'index': row_idx,
                     'data': data,
                     'errors': {},
                 }
-                self.rows.append(row)
+
+                # make sure that the row was submitted - solves https://github.com/inventree/InvenTree/pull/1588#issuecomment-847889353
+                row_exist_check = [a for a in self.request.POST.keys() if a.startswith(f'row_{int(row_idx) +1}_')]
+                if True in row_exist_check:
+                    self.rows.append(row)
```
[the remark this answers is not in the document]
eeintech: @matmair I don't know if this is a fix but right now nothing is showing on step 3 anymore... at least on purchase order upload...
There were a lot of addition to the form wizard so not sure where the hiccup is but will try to investigate.

eeintech:
@matmair I think I found what's happening here. In step 3 ("items") the `get_special_field` method fails to return the right fields and quantity is not set to `required=False`. So it will still try to validate rows that have been removed (eg. quantity is None).

The `get_special_field` should really be implemented in the subclass that should be located in `forms.py` of the corresponding app, as declaring the sub-classed form inside the view isn't working.

The "fix" that you added was removing all data from the step 3 so please revert to simply:
```python
self.rows.append(row)
```

InvenTree/common/views.py
```diff
@@ -188,11 +189,40 @@ class FileManagementFormView(MultiStepFormView):
     media_folder = 'file_upload/'
     extra_context_data = {}
 
-    def get_context_data(self, form, **kwargs):
+    def __init__(self, *args, **kwargs):
+        """ initialize the FormView """
+        # check if form_list should be overriden
+        if hasattr(self, 'form_list_override'):
```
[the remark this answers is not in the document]
matmair: I changed it but I am not sure if this is the more readable solution.
With the old method all the overridden/ custom stuff was in class attributes on top, now some stuff goes into init. We should document all these implementations carefully to enable other ppl. to also work with these functions. I am used to overrides in classes but maybe that is just how the libraries I encounter work.

InvenTree/order/views.py
```diff
@@ -572,7 +576,58 @@ def post(self, request, *args, **kwargs):
 class PurchaseOrderUpload(FileManagementFormView):
     ''' PurchaseOrder: Upload file, match to fields and parts (using multi-Step form) '''
 
+    # overriden classes
+    class OrderMatchItem(cm_forms.MatchItem):
```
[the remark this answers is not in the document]
matmair: I disagree. This form is just completely useless without the context of this FileManagementFormView and this specify FileManager and the correct `form_field_map` variable.
So while it is a form I would keep it here. Before moving I would rather move the `get_special_field` function into the view.

InvenTree/common/forms.py
matmair: pass just returns nothing so `bool(overriden_field) => False`

InvenTree/part/views.py
matmair: see previous comment about context

InvenTree/order/views.py
matmair: changed that

matmair:
> 
> 
> The `get_special_field` should really be implemented in the subclass that should be located in `forms.py` of the corresponding app, as declaring the sub-classed form inside the view isn't working.
> 
I commented how I think about that above. A lot of stuff can break when the context is wrong – the form is not standalone.
>
> The "fix" that you added was removing all data from the step 3 so please revert to simply:
> 
> ```python
> self.rows.append(row)
> ```
I cannot reproduce that. Were you already between steps when pulling? The submitted data lives through updates in the current implementation. Found that a bit odd but I do not want to rewrite everthing.

InvenTree/common/views.py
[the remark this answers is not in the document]
matmair: @eeintech have you tested your proposal? I cannot get that to work and it seems to not what the library [suggests](https://django-formtools.readthedocs.io/en/latest/wizard.html#formtools.wizard.views.WizardView.as_view).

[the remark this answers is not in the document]
eeintech: No I haven't tested it, it was only off-my-head proposal to make it cleaner.

I don't know why it wouldn't work, as long as `form_list` is created during init and immutable, shouldn't be an issue...

eeintech:
> > The `get_special_field` should really be implemented in the subclass that should be located in `forms.py` of the corresponding app, as declaring the sub-classed form inside the view isn't working.
> 
> I commented how I think about that above. A lot of stuff can break when the context is wrong – the form is not standalone.

It is my understanding that to conform to the Django structure, any form class should be kept in `forms.py` and not in `views.py`. I am not sure that it wouldn't work but I do think it makes sense to keep the InvenTree codebase conform to the Django rules.

> Were you already between steps when pulling? The submitted data lives through updates in the current implementation. Found that a bit odd but I do not want to rewrite everthing.

What do you mean? I am not pulling your branch in the middle of my test no, if this is what you mean. That'd be silly.

> Also reverting it would not solve the BUG with deleting elements in step 3?

Along with moving `get_special_field` into the `forms.py` associated to the app, everything worked fine on my side. It's quite easy issue to reproduce while uploading purchase order so I'm not sure why you are not able to.
If the `get_special_field` method of the purchase order "items" form is not called, then it will fail form validation.

matmair:
@eeintech I think we are talking about different things. Please submit the changes you want in a PR to my branch and I will merge them in.

InvenTree/common/views.py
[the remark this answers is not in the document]
matmair: It throws this [error](https://github.com/inventree/InvenTree/pull/1588/checks?check_run_id=2732179245#step:4:341)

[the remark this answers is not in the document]
matmair: Fixed it (I think)

eeintech:
@matmair Sure, will do when I get the chance :smiley:

eeintech:
@matmair See https://github.com/matmair/InvenTree/pull/18

matmair:
@eeintech everything seems to work now, I tested part import too. Thank you for your PR!

@SchrodingersGat I think we are ready for merge (and unit tests by @eeintech ).

SchrodingersGat:
@matmair this is my first opportunity to take a close look at this. I think I've run into a few issues that need addressing:

## Category Assignment

If I import parts with a *valid* category ID, they do not seem to get properly assigned to that category in step 3:

e.g. some sample rows:

image

and then, on step 3:

image

The "category" is not being properly selected here...

## Modal Dialog Issues

When launched in a "modal" window from the "Settings" view, there are some more serious issues:

### Form Errors Exist

The *Form Errors Exist* warning appears all the time, with no indication of that the errors (if any actually exist) may be

### Navigation Buttons Missing

The buttons to navigate between steps are not displayed:

image

Compared to 

image

### Close and Re-Open Dialog

- Run through the steps up to a certain point
- In my example, I got to "Step 3 of 3 - Match References"
- Close the form
- Open the form again (i.e. press *Import Parts* button)
- Instead of starting at step 1, the form is initialized at step 3
- Refresh the page
- Still stuck at Step 3!!
- Kill the server 
- Re-launch the server
- Press the *Import Parts* button
- **STILL** stuck on step 3 of 3

Is this intended behavior?

matmair:
@SchrodingersGat 

> ## Category Assignment

In the example screenshots the field for category_name seems to be not selected, I also can not reproduce that:
Step 2:
grafik
I selected the category field

Step 3:
grafik
the category that existed was selected by default, the category that does not was not.

> ## Modal Dialog Issues
> ### Form Errors Exist

The modal currently just supports successful -> closes and unsuccessful -> keeps open with warning. The pricing modal handles keeping the modal open the same way, I will adjust the modal-js-code.

> ### Navigation Buttons Missing

Form code inside a modal is not supported currently.

> ### Close and Re-Open Dialog

Yeah that is a big bug, will work on that and ping you when I am done. seems like the session form-data handeling @eeintech implemented is quite resilient. Could be usefull in other instances.

SchrodingersGat:
@matmair I'll try to have another look at this ASAP - I don't want you to think I'm ignoring this but very busy ATM !

SchrodingersGat:
@matmair I've played around with this again and found a few more issues. I can enumerate these if you like.

However after some more thinking I wonder if we are approaching this the wrong way.

We already rely on the [django-import-export](https://django-import-export.readthedocs.io/en/latest/index.html) library which does a *LOT* of the heavy lifting for us. It is very good at validation of fields, and the import workflow can be highly customized: https://django-import-export.readthedocs.io/en/latest/import_workflow.html

I think that we should look at leveraging `django-import-export` and see if we can create the specific features we are looking for. I'd extend this to the BOM import process too!

For example, the import screen in the admin interface is "most" of the way there already:

image

I would argue that the error / warning messages provided here are already very good:

- Indication of invalid data
- Indication of duplicate records

For the BOM importer example, we could potentially extend the functionality that exists in django-import-export to do things like:

- Create a new part if the referenced part does not exist
- If the part ID is not provided, lookup by part Name

The implementation here still seems rough around the edges, and I feel it still has quite a way to go in terms of data validation and edge cases (and I think this is where django-import-export excels).

What are your thoughts?

I am not sure yet if django-import-export can be operated from "outside" the admin context - this would be crucial I think?

Another option, which might actually not be too bad, is to expand general use of the admin interface itself. For operations which are seldom used (bulk imports for example) perhaps we are much better served polishing the tools that are already there, instead of recreating the wheel. This PR already feels like we are basically re-inventing what's already present in the admin interface...

Let me know what you think. @eeintech  too

I appreciate that this comes after a lot of work - sorry!

matmair:
@SchrodingersGat you are the maintainer - your call. I will respect that.
One caveat: I use ‘django-import-export’ in nearly all projects for fast backend import of existing data. As far as I am aware some of the core functions of @eeintech 's solution for PO import like easy row/column deletion and field-mapping in the wizard are not available in import-export. And having messed with it previously turns me really of from trying that any time soon.

import-export is really rigid about input-formatting and user-configurable imports could be a great value add - esp. in environments where you do not want / can't make users admins to use the admin-area import-functions or the user-base is not used to fixing a table in Excel.

Does your comment regard all wizards or just part-imports? In the latter case I would try to cherry-pick the general improvements in this PR an make a new one without the part-import-stuff.

eeintech:
@SchrodingersGat It's funny you brought that up, I did look into the `import-export` package for the purchase order importer and strongly considered it because it is already a dependency. But for all the reasons @matmair mentioned above, I thought this would take a lot more work to leverage into a "user-friendly" multi-step environment. This package is really good for admin level (where the person using it know what's happening) but the gap to being the mainstream importer was just too big. On the other hand, your custom BOM importer had all the "user-friendly" hooks in place, so I thought it made more sense to re-use what's already there.

Now for import operations altering directly the target table (import part directly alters the `part` table, import `build` directly alters the `build` table and so on) is basically considered as "database setup or maintenance" in my mind so not for the every user, at least in my opinion. So this is why I was pushing back not to have this feature in the main view.
That said, all the work you've done to improve the multi-step form framework is great, so that I would like to see it merged! :smile:

matmair:
@eeintech I would be happy to cherry-pick that stuff but will wait till there is a clear statment if the components even will be used

SchrodingersGat:
True that django-import-export has some limitations. I am reluctant to have "two solutions" to the same problem if we can help it, but maybe we can't help it.

I've just come across this project - https://github.com/wq/django-data-wizard - which *may* be another interesting option. I have not yet had time to fully look into it.

I am just concerned that we are boxing ourselves into maintaining a complex data importer tool when (potentially) there's something out there that does *most* of the work for us (and has other people using and validating it too!)

eeintech:
I have looked at many projects, packages, etc. before jumping into the purchase order importer wizard but have not came across `django-data-wizard`. I like that it is nicely documented, provides API and can combine well with background tasks. However, it has been more than a year it hasn't been updated and I'm always reluctant to add new dependencies if the project is not actively maintained... Once Django moves on, all the "abandoned" dependencies might create more issues...
If there is a similar package out there which is still pretty well alive I would definitely be on board!

SchrodingersGat:
@eeintech you're definitely right to be concerned about the maintenance issues of the library.

I think that perhaps "rolling our own" is the best approach here after all. Given the complex interactions of the various models I think it is going to be easier to write something that takes the particular requirements into account...

I've had a more thorough look around this evening and cannot find another library that looks "perfect"...

@matmair can you please fix the conflicts that have crept in and I'll have another look

matmair:
@SchrodingersGat should we just merge the whole thing or split it up into feature branches?

You mentioned several errors above so I think that could make merging and testing simpler.

SchrodingersGat:
I think merge the whole thing. It's pretty close to being good.

matmair:
@eeintech @SchrodingersGat mind checking if I missed an issue / Bug? The merge was pretty simple but better check twice before merging.

SchrodingersGat:
@matmair sorry looks like a conflict has snuck in - please address this and I'll have a final look

SchrodingersGat:
Thanks @matmair for your great work here. Happy to merge this in now. Please submit a documentation PR when you have a chance.

PR #1600, opened 2021-05-19
InvenTreeFixerExchangeBackend and manual pull of currency exchange rates

eeintech:


SchrodingersGat:
@eeintech what is the intent here?

eeintech:
@SchrodingersGat Create a "Currencies" view in the settings to nest all the global settings related to currencies (default currency, fixer API keys, etc) and show the last exchange rates pulled from Fixer.io. Also the user will be able to click on a button to update the exchange rates immediately, instead of relying on the scheduler.

eeintech:
@SchrodingersGat @matmair This is ready for review :)

Note that the exchange rate backend is defined in the `config.yaml` file:
``` yaml
# Define exchange backend
# Choices are:
# - InvenTreeManualExchangeBackend
# - InvenTreeFixerExchangeBackend
exchange_backend: InvenTreeFixerExchangeBackend
```

Below are the currency settings view based on the backend.
Let me know if you run into any issue!

## Fixer.io API
image

## InvenTree Manual
image

eeintech:
Btw we should add a note that Fixer.io free plan does not allow any base currency conversion other than Euro...

https://fixer.io/documentation#basecurrency

> Provided changing base currency is supported on your subscription plan, you can change the currency to which your output currency rates are relative to...

InvenTree/InvenTree/views.py
```diff
@@ -908,3 +911,87 @@ def get_context_data(self, **kwargs):
...
+        form = super().get_form()
+
+        # Get exchange rate backend
+        exchange_rate_backend = self.get_exchange_rate_backend()
+
+        # Get stored exchange rates
+        stored_rates = exchange_rate_backend.get_stored_rates()
+            
+        for field in form.fields:
+            if 'fixer' in exchange_rate_backend.name:
```
matmair, line 962: maybe this is a bit static

InvenTree/templates/InvenTree/settings/currencies.html
```diff
@@ -0,0 +1,54 @@
...
+        {% include "InvenTree/settings/setting.html" with key="INVENTREE_DEFAULT_CURRENCY" icon="fa-dollar-sign" %}
+        {% if 'fixer' in exchange_backend %}
+        {% include "InvenTree/settings/setting.html" with key="INVENTREE_FIXER_API_KEY" icon="fa-key" %}
+        {% endif %}
+    </tbody>
+</table>
+
+<div class='row'>
+    <div class='col-sm-6'>
+        <h4>{% trans "Exchange Rates - Convert to " %}{{ default_currency }}</h4>
```
matmair, line 27: not optimal translation, will submit PR

SchrodingersGat:
> 
> 
> Btw we should add a note that Fixer.io free plan does not allow any base currency conversion other than Euro...
> 
> https://fixer.io/documentation#basecurrency
> 
> > Provided changing base currency is supported on your subscription plan, you can change the currency to which your output currency rates are relative to...
> 
> Request with the free plan and different base currency returns:
> 
> ```json
> {"success":false,"error":{"code":105,"type":"base_currency_access_restricted"}}
> ```

Interesting, was not aware of this. Perhaps there is another online currency service we could use instead which allows this

matmair:
That is an inherit restriction caused by fixers use of the European central banks API as a datasource.
BTW should we maybe implement this as a third source? It is just a [xml](https://www.ecb.europa.eu/stats/eurofxref/eurofxref-daily.xml)

matmair:
@eeintech nice code, found 3 small things I would change, submitted a PR https://github.com/eeintech/InvenTree/pull/3.

SchrodingersGat:
https://exchangerate.host/#/

Looks like a potential alternative for fixer.io which supports different base currency

SchrodingersGat:
https://www.exchangerate-api.com/docs/standard-requests has a free plan which looks like it supports custom base currency

SchrodingersGat:
If I get some time today I'll have a look at adding support for these, and we can potentially remove fixer.io support

SchrodingersGat:
https://github.com/inventree/InvenTree/pull/1605 adds support for https://exchangerate.host/#/docs which does not require an API key

For now I think the simplest approach would be to just use that

SchrodingersGat:
I like what you've done so far - but while we are addressing currency support I'd like to see a couple of other changes.

The currency options (base currency / supported currencies) are currently specified in the settings file, but I think we should move them to the "settings" architecture which is written to the database.

The settings file (or environment variables) are nominally reserved for system admin type settings (database backend, etc) but I think these currency settings should be configurable without having to restart the server.

Does this all make sense?

SchrodingersGat:
See the changes in https://github.com/inventree/InvenTree/pull/1608 also

- exchangerate.host is now used in the background task for currency updates
- the INVENTREE_DEFAULT_CURRENCY setting is now used as the base currency

eeintech:
> The currency options (base currency / supported currencies) are currently specified in the settings file, but I think we should move them to the "settings" architecture which is written to the database.
> 
> The settings file (or environment variables) are nominally reserved for system admin type settings (database backend, etc) but I think these currency settings should be configurable without having to restart the server. 

Totally agreed with your comment. I moved it to a boolean setting allowing user to choose between fetching exchange rates from the exchangerate.host API or using custom rates set manually. The exchange rate backend is picked accordingly to that global setting (`CUSTOM_EXCHANGE_RATES`).

image

image

Does it make sense to you?

This PR is ready for review and merging :smiley:

SchrodingersGat:
Thanks @eeintech seems to work really well. Thanks again for the great work on this!

PR #1611, opened 2021-05-27
Exchange simplification

SchrodingersGat:
Simplification of currency exchange rates.

Ref: https://github.com/inventree/InvenTree/pull/1482

### DIFF
--- a/InvenTree/InvenTree/exchange.py
+++ b/InvenTree/InvenTree/exchange.py
@@ -1,120 +1,29 @@
 from django.conf import settings as inventree_settings
 
-from djmoney.contrib.exchange.backends.base import BaseExchangeBackend
-from djmoney.contrib.exchange.models import Rate
-
-from common.models import InvenTreeSetting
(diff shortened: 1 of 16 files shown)


SchrodingersGat:
@matmair @eeintech this PR represents somewhat of a regression in terms of functionality, but the way that currency exchange works now has been greatly simplified.

The currency "settings" view now looks like this:

image

Notes:

- There is no facility to manually edit exchange rate data (this will be added back in later after more of a case study)
- Currency settings are now reverted to being configured in the settings file rather than database settings (to work more cleanly with django-money)
- Simple button to "update now"

eeintech:
Thanks for the quick fix @SchrodingersGat 

Minor things but worth mentioning:
- Changing the base currency in the `config.yaml` does not reset the exchange rates (which could lead to erroneous conversions)
- There is a leftover print here:
https://github.com/inventree/InvenTree/blob/59aca2791ac6dc4d0e291d0609f25c531d34cb4f/InvenTree/InvenTree/views.py#L786

PR #1627, opened 2021-06-01
Fixed price break row actions after sorting

eeintech:
To reproduce the issue:
* Navigate to supplier page view
* Sort the price break table by quantity or price
* Click on "edit" or "delete" button on a row, it wouldn't do anything

Seems like `onPostBody` event handler handles sorting better than `onLoadSuccess` would do.
Reference: https://bootstrap-table.com/docs/api/events/#onpostbody

Note: `onLoadSuccess` is also used in `bom.js`, `build.js` and `stock.js` files for respective bootstrap tables but have not looked those up to see if `onPostBody` would fit better there too.


SchrodingersGat:
Thanks @eeintech - I agree that onPostBody is more appropriate here.

PR #1633, opened 2021-06-05
Money format

matmair:
For some reason the formatting of prices seems to be broken, this PR fixes it by using the dj-money formatting tag.
Also the default currency will be used in all forms now.


SchrodingersGat:
Thanks!

Issue #1484, opened 2021-04-18
stock level graph


matmair:
> @SchrodingersGat is there a reason why there are no graphs in InvenTree? 

Yes - to date there has been no historical data to "graph". However I would love to support charts.

If you are going to add graph support here (and chart.js looks great) then I would like to implement a 'stock level history' graph - and I'm sure there are other areas it could be used too!

Excited to see what you come up with here :)

_Originally posted by @SchrodingersGat in https://github.com/inventree/InvenTree/issues/1482#issuecomment-821979111_

matmair:
@SchrodingersGat I would love to implement a stock level history graph.
One problem I see is that one can have multiple stock items of the same part in the same location-> should those be summed up or should we keep it per stock item?

SchrodingersGat:
I think there would be a case for a "two-level" stock history:

### Stock Item History

Currently we already have stock item history (although this is in *serious* need of rework). So we could have a graph of "stock level" for a single StockItem

### Part History

Individual stock items come and go, and may well be entirely deleted from the system. So I think that each time a stock item quantity is adjusted, the parent "Part" object also should create a new `PartStockHistory` database entry. That way, independent of what happens to the actual stock items, we have a historical stock level record for the `Part` itself.

matmair:
@SchrodingersGat I just sketched the *Stock Item History Graph* out and realised something: the way Stock Items work they will (al)most never get restocked. That is because every time you buy new stock a new stock item will be created.
I think it does not make sense to implement it. I already have a rough outline in my fork because I realised to late so if you think it still makes sense, I can finish it.

SchrodingersGat:
I think you are probably right. The effort required to track and graph stock level for each individual stock *item* is probably not worth it, given how ephemeral the individual stock items can be.

Instead, I think that "stock level history" should be calculated against each *Part*.

Each time a stock item is updated (here "updated" means "created", "edited", "deleted"), we should run `part.calculate_stock_level()` which enters a new row into the "stock_history" table for that part.

Some considerations:

- If the stock level remains unchanged then no need to add a new row
- Bulk stock actions (e.g. competing a build and updating 100x stock items) will make a *lot* of these calls! We can offload these to the background worker
- Will need to add "signals" to the StockItem model to handle the cases where stock items are created / deleted

SchrodingersGat:
Actually @matmair the individual stock item history already is present, look at the `StockItemTracking` model. This stores a "quantity" for each time the StockItem is updated.

[...]

PR #1642, opened 2021-06-07
Part - persistant stock history

matmair:
for #1484


SchrodingersGat:
@matmair can you provide some more context details here, what you are trying to achieve?

matmair:
@SchrodingersGat Sure
TLDR: This PR enables a graph of historical stock levels even when stock get deleted as you suggested in #1484.


This PR implements a new model `PartQuantityHistory` to which an entry is added each time a StockItemTracking entry is added (via signal). 

In this new model there are fields for the data tracked in StockItemTracking and total stock at time of creation (after the action triggering the tracking was processed). This information could be used for graphs but also be enhanced with more tracked properties to make more historical insights possible.

As the tracking models share a lot of similarities I made as many parts of them abstract/reusable as possible:
This included:
Python:  `ItemTracking` an abstract model as basis for tracking models, `TrackingMixin` a mixin for API views with tracking models, `TrackingSerializer` a serializer base class for tracking models.
JS: `loadTrackingTable` shared table-loading function for both StockIteTracking and PartQuantityHistory.

From a UI perspective there are new API and View-Url-Endpoints, a navbar entry and a view for viewing the tracking entries on the part-detail views - based on how it is done in StockItems.

matmair:
@SchrodingersGat do we even need this feature anymore? This was meant to fix #1484

SchrodingersGat:
@matmair sorry I haven't been giving this one much attention. I'm not sure that this feature fits in *right now* - I've been meaning to spend some time looking at refactoring the existing stock tracking history, which means I was overlooking this PR as I thought it would change it anyway.

I think at the moment you'd be better served focusing your attention on other components! If you're happy to close this issue for now, we can maybe come back to it at a later date. But currently I think just keeping it up to date with master is a waste of time :)

matmair:
@SchrodingersGat glad we feel the same way. I am also not really sure anybody needs this function. I certainly don't - would be nice but not necessary.

PR #1634, opened 2021-06-05
Internal price

matmair:
Adds internal prices (for usage in assembly pricing etc) and all needed views, tools etc. As in #1606.

At the moment missing:

- [x] tests
- [x] documentation
- [x] ~~better integration into pricing view (only one)~~ 
- [x] maybe a calculation tool

Issue #1606, opened 2021-05-24
Internal prices


matmair:
@SchrodingersGat what do you think of internal prices for parts?
The use case would be for calculations in assemblies as the sale-price normally includes a profit margin and only using the sub-part prices does not include the work-time.

SchrodingersGat:
I think this makes sense. What extra fields / information do you think would be added?

matmair:
@SchrodingersGat I would add a price range just like the sale-price and add tools for calculating / fitting those into the 'new' pricing view.

Issue #1592, opened 2021-05-16
Add weight to a part and sum in sales orders and assemblies to calculate shipping


G-Pereira:
Add a weight parameter in such a way that the price for shipping can be estimated for assemblies or multiple part sales orders.

SchrodingersGat:
@G-Pereira I'm going to close this one as "out of scope". You can now add extra line items to orders, for any extra costs like shipping, etc.

Issue #1577, opened 2021-05-12
Use stock item "purchase_price" in Part Pricing calculation


eeintech:
The `PartPricing` view does not seem to take into account the price of stock items (stored in the `purchase_price` field) but only supplier price breaks.

https://github.com/inventree/InvenTree/blob/07c7af1f2e1bb577aa92709d0055d70a9d02ffed/InvenTree/part/views.py#L1950

My proposal is to have a separate calculation for supplier pricing (from supplier price break) and purchased pricing (from stock items), to not overwrite the pricing of one or the other in the BOM calculation for example.

Any thoughts?

matmair:
@eeintech sounds great. And that is basically what I implemented in #1482. It shows all the prices of all stock items and quantities. Just in a chart.
I am considering also adding a break-even chart with its own permission.

eeintech:
@matmair Sorry I guess I missed the intention of #1482 which is more about price history tracking?

I'm looking into straight forward way to see purchase price directly into a BOM calculation, and maybe also indicated for every single item in a BOM. Will that #1482 cover it?

matmair:
@eeintech I made a few screenshots for reference
![grafik](https://user-images.githubusercontent.com/66015116/118177545-21e63980-b433-11eb-9185-3374eee0c7fc.png)
![grafik](https://user-images.githubusercontent.com/66015116/118177633-3aeeea80-b433-11eb-88ed-e8ff6b89e491.png)

The modal shows the total price range for the BOM calculated over all items and quantities, a column in the BOM table might indeed be more fitting.

eeintech:
Wow yeah that's way more fancy than what we need :smile: 

I just pushed those missing BOM columns in #1581

matmair:
I am not really happy with the Part Pricing view, it is quite inefficient (space wise) in my mind.

eeintech:
How do you imagine it? If it were to be improved?

matmair:
I am currently experimenting a bit internally what people think and suggest.
Merging the alert on top and the first table.
I am A/B testing just that and a matrix instead of a lot of headings.

matmair:
@eeintech have you seen the new pricing page? Anything I should add? Will work on this next, including #1592 and maybe #1539

eeintech:
@matmair I'm sorry I missed it. Is it in master already? I'll check it out soon!

eeintech:
I was trying to check out the new pricing page and ran into #1641

[...]

PR #1645, opened 2021-06-08
Pricing reformat

matmair:
@eeintech Is that more like you think it should be?

### DIFF
--- a/InvenTree/part/templates/part/order_prices.html
+++ b/InvenTree/part/templates/part/order_prices.html
@@ -119,8 +119,8 @@
 
 {% if price_history %}
     <hr>
-    <h4>{% trans 'Stock Pricing' %}<i class="fas fa-info-circle" title="Shows the prices of stock for this part
-the part single price shown is the current price for that supplier part"></i></h4>
+    <h4>{% trans 'Stock Pricing' %}<i class="fas fa-info-circle" title="Shows the purchase prices of stock for this part.
+The part single price is the current purchase price for that supplier part."></i></h4>
     {% if price_history|length > 1 %}
         <div style="max-width: 99%; min-height: 300px">
             <canvas id="StockPriceChart"></canvas>
--- a/InvenTree/part/templates/part/order_prices.html
+++ b/InvenTree/part/templates/part/order_prices.html
@@ -15,7 +15,16 @@
 {% block details %}
 {% default_currency as currency %}
 
-{% crispy form %}
+<form method="post" class="form-horizontal">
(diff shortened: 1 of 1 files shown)


eeintech:
Sorry but I am still confused by the mix of information in this part detail page... In https://github.com/inventree/InvenTree/issues/1577#issuecomment-856314057 you said:

> Stock prices alone are not very informative. the quantity serves to help the user to get a feel how much (in total) they paid for the stock while the difference helps to show how much prices differ from the calling price for that supplier price 

The working of the tab is "Order Pricing", so I will think this is pricing information from sales and purchase orders.

Then the graph is called "Stock Pricing", so I am assuming this is pricing for the current stock, one has to know that this price was either set by:
- a purchase order
- manually on the stock item
The graph also uses some pricing information from the supplier part data (price breaks).

And there is the "Part Pricing" modal view, which exclusively uses supplier part data.

You've done a great work with this graph, however pricing strategy is missing clarity. I understand well how the mechanics of InvenTree work but for "casual" users, like my colleagues, are getting lost.

I _think_ we need to revamp the pricing into a single view called "Pricing". On top of my mind, it should break down into 2 sections: "Supplier Pricing" and "Stock Pricing" (as opposed to "Order"). The first section will feature what's in the "Part Pricing" modal view and the second features the "Stock Pricing".

The cohesion will help explain the logic.

What do you think?

matmair:
@eeintech I am actually working towards one view put there are still a few pieces like internal prices(#1634 #1606 ) missing.
I already have a mock-up tested internally but there are functions missing in the system for making it all work right now. I envision graphs with the different kinds of prices to match them more easily. Also assistants for adding weighted (#1592 ) margins and alerts for adjusting sale and internal prices if the purchase prices or margins change. But all functions should be opt-in to make it slim by default. Plus revamped docs and permissions for prices.

I try to make the PRs smaller now as more than one change takes forever to discuss and get merged (see #1588).

matmair:
@SchrodingersGat do you prefer big PRs or small single-issue incremental changes?

SchrodingersGat:
@matmair smaller PRs are definitely easier to manage, where possible :)

SchrodingersGat:
@matmair is this PR still relevant or is it going to be rolled in to the "pricing page" strategy?

Bottom line should we merge this in now, or close and await #1656

matmair:
@SchrodingersGat  This was an intermidate fix for some valid ciritisms of the changed view - i would prefer this to be rolled in so that master is a bit more usable and cleaner.

PR #1653, opened 2021-06-11
display latest serial on all part-views #1648

### LINKED ISSUE #1648: Display "latest serial number" info on part page
- If the part is "trackable"
- Show the "highest" serial number for this part (part tree)

image

matmair:
grafik

Fixes #1648


SchrodingersGat:
Nice!

PR #1655, opened 2021-06-11
possible fix for #1628

### LINKED ISSUE #1628: [BUG] Bootstrap pagination: NaN to NaN of undefined rows
To reproduce:

1. Navigate to a part list under a category
2. Select "All" for the number of items displayed
3. Refresh the page, it will show as:

image

matmair:
Fixes #1628


SchrodingersGat:
Thanks @matmair - I'm on vacation for a few days so I'll have a look as soon as I get back home

matmair:
@SchrodingersGat great, enoy your time off! Will do some bigger stuff then till you are back.

eeintech:
I just tried that fix but does not work on Chrome... You guys are using Firefox right?

matmair:
@eeintech I can confirm that the bug seems to still persist in chrome but I failed so far to find what is causing it. Will investigate in the weekend.

SchrodingersGat:
@matmair are you going to continue with this one in a new PR?

matmair:
I have no idea what is causing it and a half working solution is kind of bad so I closed that.

eeintech:
@matmair Don't mind me using another browser, if it solves it for part of InvenTree users, probably worth it!

matmair:
@eeintech you represent 64% of the worlds browser users while we do under 4%
grafik
So this fix would mostly be of cosmetic nature.

eeintech:
4% better than 0%? It does not hurt is what I mean :) No worries any way

PR #1666, opened 2021-06-16
total price column as per #1660

matmair:

PR #1669, opened 2021-06-16
So total

matmair:
@SchrodingersGat just like you asked in #1666 -  calc is done just on a JS basis


SchrodingersGat:
Nice work, thanks for pulling this together so quickly.

matmair:
> 
> 
> Nice work, thanks for pulling this together so quickly.

This is core functionality of bootstrap-table so easy to do.

PR #1679, opened 2021-06-17
fix for part with no bom-price

### LINKED ISSUE #1678: BOM-price pie: Error when no Price on item
@matmair the error still exists, it errors out when you have one or more BOM items which do not have any pricing information:

```python
x = [True for a in bom_items if len(set(a['price'])) == 2]
```
But if `a['price'] is None`:

image

_Originally posted by @SchrodingersGat in https://github.com/inventree/InvenTree/issues/1634#issuecomment-863239204_

matmair:
Closes #1678


SchrodingersGat:
Thanks for the fix :)

PR #1682, opened 2021-06-17
Order modal show price

### LINKED ISSUE #391: When ordering parts, show the pricing for each supplier
Next to each available supplier, display the price calculated based on the provided quantity.

matmair:
Fixes #391

This PR adds the fuctionallity propsed by @SchrodingersGat in #391 in 2019.
In the part-order-modal the price of each supplier-part (based on the entered quantity) is displayed. A custom submit button enables reloading the supplier-part selection screen to update the prices.


InvenTree/order/templates/order/order_wizard/select_parts.html
```diff
@@ -49,7 +51,13 @@ <h4>
                         <select class='select' id='id_supplier_part_{{ part.id }}' name="part-supplier-{{ part.id }}">
                             <option value=''>---------</option>
                             {% for supplier in part.supplier_parts.all %}
-                            <option value="{{ supplier.id }}"{% if part.order_supplier == supplier.id %} selected="selected"{% endif %}>{{ supplier }}</option>
+                            <option value="{{ supplier.id }}"{% if part.order_supplier == supplier.id %} selected="selected"{% endif %}>
+                                {% if show_price %}
+                                    {% call_method supplier 'get_price' part.order_quantity as price %}
+                                    {% include "price.html" with price=price %} -
```
SchrodingersGat, line 57: image

This line throws an error immediately for me, I think you need to pass currency to price.html ?

matmair: @SchrodingersGat that is done in line 7, I think the merge fixed if, could you try again?

SchrodingersGat:
image

image

Still seeing the error, not sure what I'm doing wrong, this part has prices set for multiple suppliers as shown in the first screenshot.

matmair:
@SchrodingersGat right. I could reproduce and fix it with the fixtures loaded. Will start doing that instead of using my own testing database. Should work now.

SchrodingersGat:
@matmair still having some issues:

Firstly, I definitely have supplier pricing set:

image

However when I click the "order" button (on the same page):

image

Also, when I hit "Update Prices" I get the following screen:

image

Note that this part is marked as purchasable:

image


Also, the "Update Prices" button seems to persist on other modals once it is drawn:

image

Looks like these extra modal buttons need to be explicitly reset

SchrodingersGat:
@matmair that's fixed the zero-pricing issue, nice work.

However the extra button still persists on the next modal you launch after closing that one:

image

matmair:
@SchrodingersGat I found a few other problems while debugging this, will ping you when I ironed out everything.

matmair:
I fixed the button always appearing but still working on another the quantity-zero issue and on errors not finished multi-part modals reappear.

SchrodingersGat:
ok no worries, let me know when you are again ready for review.

matmair:
> Also, when I hit "Update Prices" I get the following screen:
> 
> image

@SchrodingersGat turns out that dropping all parts with a quantity of zero is no bug but hardcoded behaviour. I think that makes sense. The question is if the start-quantity should be 1 by default (currently 0) to prevent confusion.

SchrodingersGat:
Fair enough, this makes sense. Thanks for your work on this.

PR #1701, opened 2021-06-21
Fix supplier part view/edit without a manufacturer

atwalsh:
Fixes the following issues when a supplier part was created with an MPN but no manufacturer was assigned:
- Viewing a supplier part stock item
- Editing a supplier part stock item

---

<details>
<summary>Translation discussion...</summary>

@SchrodingersGat - I ran `invoke translate` in 402a7da, but this resulted in a massive diff. The only new translatable text was the `'No manufacturer set'` text on the supplier part stock details page. Here's the lines in the en translation: https://github.com/inventree/InvenTree/blob/402a7dabf1a0a44e668aee01722b380240e5fadd/InvenTree/locale/en/LC_MESSAGES/django.po#L5743-L5745

Screen Shot 2021-06-21 at 11 12 54 AM

Let me know how you would like to proceed with the translations. Maybe I ran the translation script incorrectly?

</details>

### DIFF
--- a/InvenTree/company/views.py
+++ b/InvenTree/company/views.py
@@ -623,7 +623,8 @@ class SupplierPartEdit(AjaxUpdateView):
         supplier_part = self.get_object()
 
         if supplier_part.manufacturer_part:
-            initials['manufacturer'] = supplier_part.manufacturer_part.manufacturer.id
+            if supplier_part.manufacturer_part.manufacturer:
+                initials['manufacturer'] = supplier_part.manufacturer_part.manufacturer.id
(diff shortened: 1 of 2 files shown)


eeintech:
@atwalsh You shouldn't run `invoke translate`, the translation is handled by [Crowdin](https://crowdin.com/) now, you can revert your last commit :wink:

atwalsh:
Sounds good. Do you have a preference on reset+force push or a revert?

eeintech:
I would do `git reset HEAD^` then commit?

SchrodingersGat:
Thanks for the fix @atwalsh

PR #1688, opened 2021-06-20
Adds "parameters" for manufacturer parts

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/1434

# TODO

- [x] Admin interface
- [x] DRF API
- [x] Web forms
- [ ] Documentation
- [ ] Unit tests
- [x] Update python interface - https://github.com/inventree/inventree-python/pull/61

PR #1703, opened 2021-06-21
Pseudo-fix for API validation error

eeintech:
@SchrodingersGat In #1688 this line was added in the `clean` method of the `Part` model:

https://github.com/inventree/InvenTree/blob/f90730d866ec93b016ddca782212534a60feb179/InvenTree/part/models.py#L674

And API could not update parts anymore:
``` bash
WARNING:root:Bad response (400) - PUT 'http://127.0.0.1:8000/api/part/3/' - {"name":["Part must be unique for name, IPN and revision"],"IPN":["Part must be unique for name, IPN and revision"],"revision":["Part must be unique for name, IPN and revision"]}
```

It seems like when the validation of the serialized data is kicking-in, the part associated to the data is not loaded properly and `self.pk` is `None`. I'm not sure why exactly, I have searched around to figure out if there was a better way to retrieve the instance  inside the `ModelSerializer` class:
https://github.com/inventree/InvenTree/blob/f90730d866ec93b016ddca782212534a60feb179/InvenTree/InvenTree/serializers.py#L52

But have not found anything...

Maybe the check for the `pk` is enough? It does the job for me.

Also, I think using the built-in `UniqueConstraint` instead of manual unique validation is cleaner.


SchrodingersGat:
Ah good catch @eeintech 

See my notes in #1705 - turns out this has been a long-running bug but has gone unnoticed until I made those changes.

SchrodingersGat:
Should be fixed now - please confirm @eeintech

eeintech:
@SchrodingersGat Works great! :+1:

PR #1704, opened 2021-06-21
I think a fix for #1663

matmair:
Closes #1663

Issue #1663, opened 2021-06-14
Add ability sort "allocated stock" table by "allocated" quantity


SchrodingersGat:
Currently cannot sort by allocated quantity which makes it difficult to immediately see which parts have not yet been allocated (or fully allocated).

SchrodingersGat:
@matmair fix was a bit more complex - see #1706

PR #1706, opened 2021-06-22
Bug fix for build order allocation sorting

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/pull/1704
Fixes https://github.com/inventree/InvenTree/issues/1663

### DIFF
--- a/InvenTree/templates/js/build.js
+++ b/InvenTree/templates/js/build.js
@@ -418,6 +418,7 @@ function loadBuildOutputAllocationTable(buildInfo, output, options={}) {
             sub_part_detail: true,
             sub_part_trackable: trackable,
         },
+        disablePagination: true,
         formatNoMatches: function() { 
             return '{% trans "No BOM items found" %}';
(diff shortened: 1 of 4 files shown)


InvenTree/templates/js/build.js
```diff
@@ -686,14 +688,12 @@ function loadBuildOutputAllocationTable(buildInfo, output, options={}) {
                     return makeProgressBar(allocated, required);
                 },
                 sorter: function(valA, valB, rowA, rowB) {
+                    console.log("sorting!");
```
matmair, line 691: @SchrodingersGat  is that there on purpose or from testing?

matmair (commented):
A bit 😅 more complex. I just fixed the sorting function.

InvenTree/templates/js/build.js
SchrodingersGat: Thanks @matmair was just a debug statement

Issue #1213, opened 2021-01-08
Investigate DRF Forms


SchrodingersGat:
DRF is already used extensively for the API.

And it appears that it also supports form management:

https://www.django-rest-framework.org/topics/html-and-forms/

If this is the case, then the existing form framework could be ported to using the API, which would *significantly* reduce existing code duplication.

eeintech:
Would you lose the crispy form layer?

SchrodingersGat:
Good question! I'm not sure.. I was just reading through the DRF documentation and saw they supported web forms. Not something I plan on tackling for quite some time anyway.

rcludwick:
Most people these days develop an api layer and then develop a front end based upon that.  So you could write an alternative front end for IOS, Android, or integrate it internally with your own applications.  Your auth layer would likely need to change though.

SchrodingersGat:
@rcludwick absolutely. A lot of the web interface currently makes use of the API. For example, all the dynamic tables are loaded using AJAX

Issue #707, opened 2020-04-10
Stock "adjustment" actions should be ported to the API


SchrodingersGat:
the adjument actions (move / add / remove / count) are duplication between the django web forms and ajax API interface.

They should be consolidated to use just the API and some jQuery on the web front end

Issue #1687, opened 2021-06-20
Observe "default" values for blank fields when creating objects via API


SchrodingersGat:
It appears that when creating objects via the API (e.g. with the python interface), then any fields not explicitly specified will have the value set to "blank" or "false", rather than the value specified as "default" for that field in the model..

Needs to be investigated!

SchrodingersGat:
After research, it appears that the behaviour I "expected" to work is not yet supported in DRF:

Ref: https://github.com/encode/django-rest-framework/pull/8002

Hopefully this is merged soon

eeintech:
Good catch @SchrodingersGat, hope the feature request gets in soon!

Issue #1683, opened 2021-06-17
Front End Roadmap


SchrodingersGat:
The InvenTree project, from rather humble beginnings, has accrued a very large feature set and frankly the front-end experience is suffering. I'm not by any means a  web developer, and the UX design has been leveraged and hacked together from various components, without much thought for consistency or "best practice"

I would like to gauge interest and suggestions from users, and especially those with experience in front-end web design. Following are some key points I would like to see changed / improved in future revisions of InvenTree

### UX Consistency

Refactor many pages for a common "feel" or UX approach across the entire site. Currently each page has its own quirks.

### Modernize

I don't mean "bloat" ;) Current front end stack is pretty much just jquery and bootstrap, with some small libraries thrown in. If we don't need to use a more modern js framework, that's great! But some conversations about the CSS and "look and feel" need to take place

### Too Many Clicks

Many operations currently require too many "clicks" from the user. 

- Click on a button
- Open a dialog box
- Click on a text entry
- Enter data
- Click "Accept"
- Refresh page

I would like many actions to move to "inline" (i.e. data can be directly edited where it is displayed on the page).

To achieve this, we would need to make further use of the REST API

### REST API

The REST API is currently used somewhat throughout the web interface. Primarily for rendering the various live data tables. I'd like to see this used more:

- Remove duplicate code paths (forms rendered in django and also serialized in DRF)
- Move all validation code to the REST API for simplification
- More use of seamless data validation (with ajax requests)
- Less reliance on pop-up forms
- Less need to refresh an entire page - just reload the particular page parts

### SPA Design

The current "tabbed" pages provide a lot of data for a given view (e.g. "Part" page) but doing a page refresh every time you change tabs is annoying.

I would like to see the entire page loaded once, and swapping between tabs simply swaps out the content in the main page panel.

==================

Obviously a lot of work here! This is not going to happen right away, there's a lot of dev work to finish off first. But I'm keen for comments and input. I'll try not to take any criticism of my naive design choices too personally ;)

@matmair @eeintech I know you guys will want to have some input on this!

==================

### References

- https://github.com/inventree/InvenTree/issues/1673
- https://github.com/inventree/InvenTree/issues/1234
- https://github.com/inventree/InvenTree/issues/1213

matmair:
Full disclosure: I am an engineer. I do not have any training in UX-design. And I really like Bootstrap with pure JS if possible - because it is predictable.

[...]

PR #1716, opened 2021-06-23
[WIP] API Forms

SchrodingersGat:
This PR implements "forms" across the API.

The eventual goal here is to eliminate the current duplication of code which renders web forms from the django server, and also (separately) defines the API.

The DRF API will (when this PR is complete) provide *all* (or at least most) of the modal forms.

### Form Features

- [x] Field selection
- [x] Ignore certain fields
- [x] Set custom field ordering
- [ ] Custom "grouping" of form fields
- [ ] Custom form elements
- [x] Use pagination for select2 API requests - speed up results!
- [ ] Launch "secondary" modals (and handle the results)
- [ ] Allow "infinite" stacking modals (created at run-time)
- [x] Callbacks when fields are edited
- [ ] Option for custom client-side field validation (before sending data to the server)
- [x] Support custom filtering for foreign-key fields
- [x] Render optional "prefix" elements to form fields
- [ ] Inter `min_value` from numerical django model fields
- [ ] Infer `max_value` from numerical django model fields
- [x] Allow (non required) select2 fields to be cleared
- [ ] Correctly render file fields which already have a file, in "edit" mode (e.g. StockItemAttachment)
- [x] Framework form for "deleting" objects
- [ ] Display "warning" screens for deletion - i.e. "these things will also be deleted"
- [ ] Allow non-required file / image fields to be cleared
- [ ] Insertion of fields defined completely at "run time"
- [ ] Run custom code before the form is displayed, but after data is fetched
- [ ] Run custom code before form is submitted
- [x] Focus on a particular field when opening the modal (if desired)
- [ ] Detect if user does not have required permissions and hide secondary buttons accordingly
- [x] Record user when creating a PurchaseOrder
- [x] Record user when creating a SalesOrder
- [ ] Migrate settings forms to the API
### Bugs

- [ ] Launching a secondary modal with a select2 attached strips the select2 features from the first modal... 
- [x] Python bindings unit tests are broken - need to submit a PR there once this one is merged.

### Other Items

- [x] Ensure that all "attachment" tables use the REST API

### Supported Input Types

- [x] Model reference ForeignKey fields (probably the most complex one!)
- [x] Boolean fields
- [x] String fields
- [x] Integer fields
- [x] Float fields
- [x] URL fields
- [x] email fields
- [x] decimal field
- [x] date field
- [ ] datetime field
- [ ] time field
- [x] choice field
- [ ] multiple choice field
- [x] file upload
- [x] image upload

Ref: [django DRF metadata](https://github.com/encode/django-rest-framework/blob/e92016ac2e926483e05e296558fc3d1ea3279625/rest_framework/metadata.py#L36)

### Python Integration

- [ ] Python support

TODO: Add support to the python bindings to allow dynamic lookup of "queryset" for a given ForeignKey field.

### Unit Testing

- [x] Increase unit testing for the new OPTIONs endpoints
- [x] Ensure that the OPTIONS endpoints returns correct information for "related" fields
- [x] Test that the OPTIONS endpoint correctly returns 'default' values for fields with specified defaults
- [x] Ensure the OPTIONS endpoint correctly returns 'filters' for model fields
- [ ] Ensure the OPTIONS endpoint correctly returns min_value and max_value for fields
- [x] Ensure that OPTIONS includes 'help_text' for related fields
- [ ] Correctly handle invalid inputs for *all* MPTT fields (e.g. setting invalid parent node)
- [ ] Automatically remove invalid choices from MPTT fields (e.g. don't let user select invalid "parent" items)
- [ ] Save user information when creating or updating an attachment
- [ ] Save user information when creating or editing a StockItem 
- [ ] Ensure that the 'required' property is getting set correctly - both for fields and relations
- [ ] Ensure that 'read_only' property is getting set correctly - both for fields and relations - e.g. looks like this might not be the case for StockItemSerializer

### Documentation

- [ ] Add API docs for OPTIONS endpoints

### References

- https://github.com/inventree/InvenTree/issues/1683
- Fixes https://github.com/inventree/InvenTree/issues/1213
- Fixes https://github.com/inventree/InvenTree/issues/707
- Requires https://github.com/inventree/InvenTree/issues/1687 to be fixed
- Waiting on https://github.com/encode/django-rest-framework/pull/8002
- Waiting on https://github.com/encode/django-rest-framework/discussions/8003


SchrodingersGat:
Very early proof of concept screenshot:

Modal forms are now (well, will be) fully rendered on the front end.

This means that the foreign-key fields can be rendered however we want!

See the "Manufacturer" description below - the representation has been pulled from the JSON data provided by the API.

We could extend this to render images, or whatever data we want.

image

SchrodingersGat:
Example of custom rendering within the form, allowing images and fancy things.

image

matmair:
@SchrodingersGat nice! Might copy that idea for other projects ;-)

SchrodingersGat:
@matmair still just poking around the edges of the possibilities here but I think it's looking pretty promising!

Pretty happy with the minimal code required to render the arbitrary forms, still have to work out posting and validation!

SchrodingersGat:
Performance benchmark:

- Opening the "New Supplier" form with the server-side form rendering takes ~3,000ms
- Rendering the same form on the client side opens the form in ~30ms

SchrodingersGat:
@matmair @eeintech below is an example of the absolutely huge speed improvement we can get by using the API for form rendering.

I picked one of the "simplest" forms - creating a new `Company` object, so this is already one of the fastest forms to render.

I hope you can see from the captures below just how much more responsive the new interface approach is.

### Server Side Rendering (old approach)

server_side


### Client Side Rendering (new approach)

client_side

matmair:
@SchrodingersGat nice. I can see a lot of performance improvements here. Would be interesting to measure the performance of InvenTree overtime to see what real-world effect these kinds of changes have.

eeintech:
@SchrodingersGat This is amazing!

Sorry I was out for few days but the speed improvement is awesome, I have to dig more when I get the chance to see what it is like to build form on the client side, is this a built-in feature of DRF?

SchrodingersGat:
> is this a built-in feature of DRF?

I've had to hack a few things into the DRF framework to support this, but the main trick is leveraging the [metadata class](https://www.django-rest-framework.org/api-guide/metadata/) to provide information on the fields available at each API endpoint (using a [HTTP OPTIONS request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS)).

In the end you can write some pretty simple javascript:

```javascript
  $('#edit-part').click(function() {
      constructForm(
          '{% url "api-part-detail" part.pk %}',
          {
              method: 'PATCH',
              title: '{% trans "Edit Part Details" %}',
              fields: {
                  name: {
                      onEdit: function() {
                          console.log('Edited name field');
                      }
                  },
                  description: {
                      prefix: `<span class='fas fa-user'></span>`,
                  },
                  category: {
                      filters: {
                      },
                      secondary: {
                          label: '{% trans "New Category" %}',
                      },
                  },
                  active: {
                      onEdit: function() {
                          console.log('edited active field');
                      }
                  },
                  purchaseable: {},
                  salable: {},
                  component: {},
              }
          }
      );
  });
```

to render the following form:

image

If you want to see the nuts and bolts:

**Server Side**

- [serializers.py](https://github.com/inventree/InvenTree/blob/ed2f21f583358c7d6af1d4360ce14ae9d4c48c10/InvenTree/part/serializers.py)
- [metadata.py](https://github.com/inventree/InvenTree/blob/ed2f21f583358c7d6af1d4360ce14ae9d4c48c10/InvenTree/InvenTree/metadata.py)

**Client Side**

- [forms.js](https://github.com/inventree/InvenTree/blob/ed2f21f583358c7d6af1d4360ce14ae9d4c48c10/InvenTree/templates/js/forms.js)
- [model_renderers.js](https://github.com/inventree/InvenTree/blob/ed2f21f583358c7d6af1d4360ce14ae9d4c48c10/InvenTree/templates/js/model_renderers.js)

SchrodingersGat:
image

eeintech:
@SchrodingersGat I've only scratched the surface of where the metadata framework and OPTIONS requests gets you, however I have a few questions:
* Can **all** forms be migrated to this framework? I'm thinking the multi-step form implementation that we recently merged in and that I was thinking the BOM import would leverage and which still needs unit testing. Lemme know if you think this won't be impacted.
* In the long-term, can the "standard" Django views be obsoleted in favor of this new framework? Do you have plan on deleting them as part of this PR or right after?
* Do you anticipate any kind of user-facing functional changes with this implementation or would it be completely transparent (apart from the performance/speed improvement)?

SchrodingersGat:
@eeintech  all great questions :)

>  Can all forms be migrated to this framework? I'm thinking the multi-step form implementation that we recently merged in and that I was thinking the BOM import would leverage and which still needs unit testing. Lemme know if you think this won't be impacted.

I'm not sure yet how many forms can be migrated. I'd like to do as many as possible mostly to prevent code duplication. With regard to the import wizards - the functionality is not mirrored on the API like many of the simple forms. Also, these wizards to a *lot* more back-end processing than the regular forms.

I would say that the wizard forms will remain in their current implementation for a good while, even if they do (eventually) get migrated.

> In the long-term, can the "standard" Django views be obsoleted in favor of this new framework? Do you have plan on deleting them as part of this PR or right after?

Yes, for as many forms as possible I would move to delete the existing forms and replace with front-end forms. This PR has replaced a small number of forms already (the very simple ones).

This will be an ongoing process (probably not going to to do it *all* in this PR) - starting with the simpler forms to ensure there are no bugs and gauge user feedback.

> Do you anticipate any kind of user-facing functional changes with this implementation or would it be completely transparent (apart from the performance/speed improvement)?

Apart from maybe a few small visual differences, it would be pretty transparent to the user. One huge advantage is the custom rendering we can do on the front end, so users might start to see some improvements there.

InvenTree/order/serializers.py
```diff
@@ -335,13 +338,20 @@ def __init__(self, *args, **kwargs):
...
 
     allocated = serializers.FloatField(source='allocated_quantity', read_only=True)
     fulfilled = serializers.FloatField(source='fulfilled_quantity', read_only=True)
+    
+    sale_price = InvenTreeMoneySerializer(max_digits=19, decimal_places=4)
+
     sale_price_string = serializers.CharField(source='sale_price', read_only=True)
 
+    sale_price_currency = serializers.ChoiceField(
+        choices=djmoney.settings.CURRENCY_CHOICES,
```
matmair, line 351: @SchrodingersGat maybe use the new helper function `common.settings.currency_code_mappings` for this? Would provide support for setting choices in DB in the future and limit to accepted currencies.

matmair:
@SchrodingersGat a few things that I develop rn will need rewriting when this gets merged, when do you think it will be ready? Should I already pull this into my PRs?

SchrodingersGat:
> 
> 
> @SchrodingersGat a few things that I develop rn will need rewriting when this gets merged, when do you think it will be ready? Should I already pull this into my PRs?

yes this PR has grown a lot, and touches a lot of code! I am thinking I will merge this in sooner, rather than waiting until every wishlist item is implemented, otherwise it will be a nightmare to merge.

InvenTree/order/serializers.py
SchrodingersGat: Thanks, I've taken that on board

matmair:
> > 
> > 
> > @SchrodingersGat a few things that I develop rn will need rewriting when this gets merged, when do you think it will be ready? Should I already pull this into my PRs?
> 
> yes this PR has grown a lot, and touches a lot of code! I am thinking I will merge this in sooner, rather than waiting until every wishlist item is implemented, otherwise it will be a nightmare to merge.

@SchrodingersGat loots of great stuff in there, will make adding new forms more seamless. I will keep doing small stuff while you work on this and then finish the wishlist for pricing. Everything left there needs form rendering.

SchrodingersGat:
@matmair enough for now - remaining outstanding issues / features are documented here: https://github.com/inventree/InvenTree/issues/1755

eeintech:
It's so quick btw :smiley:

SchrodingersGat:
> @SchrodingersGat For the moment the "API-ready" forms are only for companies and orders? I've tried Parts, Categories and Stock location/items and still getting GET requests.

I have only migrated a handful of the existing forms - basically the ones without any complex server-side logic. Those will require some closer attention.

eeintech:
Sounds good!

PR #1718, opened 2021-06-24
Allow BOM pricing to be valid when using internal pricing

rgilham:
When Inventree pricing for the BOM is set to use internal pricing for BOM calculation it returns that a warning message: 
**Note: BOM pricing is incomplete for this part** 
even though all BOM components have valid internal pricing.
This PR will check if internal pricing is used.

### DIFF
--- a/InvenTree/part/models.py
+++ b/InvenTree/part/models.py
@@ -1465,16 +1465,16 @@ class Part(MPTTModel):
         return self.supplier_parts.count()
 
     @property
-    def has_pricing_info(self):
+    def has_pricing_info(self,internal=False):
         """ Return true if there is pricing information for this part """
-        return self.get_price_range() is not None
+        return self.get_price_range(internal=internal) is not None
 
     @property
     def has_complete_bom_pricing(self):
         """ Return true if there is pricing information for each item in the BOM. """
-
+        use_internal = common.models.get_setting('PART_BOM_USE_INTERNAL_PRICE', False)
         for item in self.get_bom_items().all().select_related('sub_part'):
-            if not item.sub_part.has_pricing_info:
+            if not item.sub_part.has_pricing_info(use_internal):
                 return False
 
         return True


InvenTree/part/models.py
```diff
@@ -1465,16 +1465,16 @@ def supplier_count(self):
         return self.supplier_parts.count()
 
     @property
-    def has_pricing_info(self):
+    def has_pricing_info(self,internal=False):
         """ Return true if there is pricing information for this part """
-        return self.get_price_range() is not None
+        return self.get_price_range(internal=internal) is not None
```
SchrodingersGat, line 1470: Should we have a fallback to BOM pricing if there is no internal price?

i.e.

```python
if internal_price is not None:
    return internal_price
elif bom_price is not None:
    return bom_price
else:
    return None
```


@matmair - you have more direct knowledge about the internal pricing structure, do you want to weigh in here?

rgilham: This change will only be effective when PART_BOM_USE_INTERNAL_PRICE is True. 
I was thinking that it might be sensible to use internal pricing when a part has an internal price (and configured to be used) and if no internal pricing is set then use the supplier pricing. Should the supplier not have pricing then the Error should still be displayed.  Your thoughts on this?

SchrodingersGat: @rgilham - agreed. Is this what your current implementation does?

matmair: @SchrodingersGat that check is already in place
https://github.com/inventree/InvenTree/blob/be85de4a7b5dc6d328fe1d519dd592fdba76b76e/InvenTree/part/models.py#L1584-L1587

matmair (changes requested):
@rgilham looks good, please fix the [small styling issue](https://github.com/inventree/InvenTree/pull/1718/checks?check_run_id=2900385759#step:5:7) and I think we are good to go.
Thanks for the contribution!

SchrodingersGat:
@rgilham please fix style issues and I can merge

Issue #1727, opened 2021-06-27
Add "Shipping Cost" to orders


SchrodingersGat:

G-Pereira:
This would be nice!

Issue #1656, opened 2021-06-12
Overview issue:  Pricing


matmair:
This is a overview issue for reorganising all pricing bits in InvenTree. Its goal is to focus all discussion into this issues (like #1577, #1645)

# Vision
A single page for all things pricing.
The following questions :
How much does it cost to buy the part from my registerd suppliers? How much does the BOM cost? How much do the different BOM-items contribute to the cost? For how much is it (internally in a BOM of an other part or externally) sold? How does the spread of these prices develop over different quantities?

# Tasks / Issues
## 1. Functional / API-updates
- [x] 1.1 Internal prices - as mentioned in #1606 - PR already in #1634
- [ ] 1.2 global and part-level sale-margins (with settings to enable - off by default) -> *up for discussion*
- [ ] 1.3 alerts for adjusting sale and internal prices if the purchase prices or margins change

## 2. UI - Updates
- [x] 2.1 just one view for all price breaks of a part - PR #1712 
- [x] ~~2.2 graph for showing stock prices (the purchase prices at which current stock was bought)  ->already present - maybe refactor *up for discussion*~~
- [x] 2.3 graph for buying and selling price ranges in different quantities
- [ ] 2.4 margin calculating assistant as in #1592 - calculates how much margin is present / adjust internal and sale pricing to keep margin

## 3. Ancelary stuff
- [x] 3.1 #391 show prices in ordering forms - PR #1657 / #1682
- [ ] 3.2 #1631 minimum order quantity for supplier parts
- [ ] 3.3 sale order total and adding a global weight for shiping, handeling etc (maybe also helpful for the question in #1592) - PR #1730
- [x] 3.4 docs for pricing
~~- [ ] 3.5 permissions to limit shown information~~ not necessary

matmair:
@eeintech do you have any input on this? Especially 1.2, 2.2
I will first work on 3.1, 1.1, 2.1, 2.3, 1.2 in this order

matmair:
Suggested merging order: #1588, #1657, #1645, #1634

sintech:
Let me give a view from my perspective. 
I tend not to use PO for adding new stock, so for me the only way to keep pricing info is `purchase_price` in `stockitem`, so what @matmair did in #1577 is a very good improvement. What I still missing is a unit price in Stock table.
Speaking of PO, It wold be nice to have "Total" column with quantity * unit price and "Grand Total" row with sum of all rows quantity and prices.

matmair:
[...]

PR #1730, opened 2021-06-28
sales order price and validation

### LINKED ISSUE #807: Add "empty" line to Purchase Order and Sales Order
Might want to add a line item against something not in the database. This could be simply because the item has not been entered, or because it requires a line for "consulting fee" or something non tangible.

matmair:
Adds items from #1727, #807 (like described in #1656 )

This PR adds:
- a `sale_price` for sale orders
grafik
- a `checksum` for the price
- a validation dialog for the price
grafik
- additional line items (including editiong dialog)
grafik
grafik

Closes #807


matmair:
Should also fix #807, 1592

matmair:
Closes #807

matmair:
Missing: predefined additional line items

matmair:
@eeintech mind looking at this? No hurry though, I can not find a bug preventing a merge anyway.

eeintech:
@matmair Looks like you need to add your new model `SalesOrderAdditionalLineItem` to the `users/models.py -> RULESET_MODELS` list, I believe it will be `order_salesorderadditionallineitem` or something similar

eeintech:
@matmair I took a look and reporting for duty!

1. I found that the sell price validation is persistent even when it is changed, to reproduce:
- Set the "Sell price" field with an initial value then validate the order
- Change the "Sell price" to different value, tag shows the sell price hasn't been validated (red)
- Reset the "Sell price" field to the initial value, tag shows the sell price has been validated (green) although there was a change in the sell price

Is that wanted? I would expect not.

2. Looks like the calculated price is not automatically updated after either a "Sales Order Item" or "Additional Order Line" is added, the page needs to be refreshed. Any chance the page can be reloaded automatically every time a new item is added?

3. Ran into issues with currencies, for instance when adding two additional line items with different currencies (CAD and EURO in this case) then refreshing the page I got this error:

image

I had to go in the Admin section to change the currency back to the default (CAD), I don't think the conversion is working on this table and the total price is based on the second currency instead of the default currency:

image

Same result with the "Sales Order Items" table too...

4. That one is a bug :bug: 

User can still add "Sales Order Items" and "Additional Order Lines" after the order was shipped:

image

5. That one too :bug: 

Sell price validation is reset after order is marked as "shipped" (the screenshot follows the one above, after page is reloaded):

image

6. That one is aesthetic only :smile: 

Why is my difference of $1 showing in green color? Feels like it should be red (as assuming user messed up!)

image

Looks like it's red when the "sale_price" is lower than calculated price though:
image

Also should "validate" be capitalized? (eg "Validate")

matmair:
@eeintech huge thanks, gone hash those out.
Regarding 6: The color is intended to warn if the users tries to sell below the calculated price. That's intended. The rest not really.

matmair:
1: this is expected, as the value is the same as was validated
2: Fixed, js-function now updates on each change
3: Fixed, currencies were not implemented in the total - function
4: Fixed in view - maybe should also be done in api - any ideas how @SchrodingersGat 
5.: Fixed
6: No fix required

SchrodingersGat:
>  Fixed in view - maybe should also be done in api - any ideas how @SchrodingersGat

Can you expand on that? I'm not exactly sure what you're askign

matmair:
> 
> 
> > Fixed in view - maybe should also be done in api - any ideas how @SchrodingersGat
> 
> Can you expand on that? I'm not exactly sure what you're askign

Where in the whole process should we stop user updating lines in a order that has been shipped?

SchrodingersGat:
Unless you have another suggestion I think the current option of disabling the "edit line item" buttons when the order has been shipped should be sufficient?

matmair:
@SchrodingersGat for applications using the API that could present a problem as they will need to check that and we need to trust that. I would feel more easy if it was not possible to change orders after they are marked shipped in any way.

SchrodingersGat:
Sure, well there are a couple of ways we could do this.. I would suggest setting certain fields as "read_only" if the referenced SalesOrder has been shipped. Could be done in the serializer or DetailView endpoint classes?

eeintech:
I don't know much of the meaning of "price validation" but I am still concerned that the "validation" has memory instead of being immediately deleted if sell price is set to another price, it feels like re-validation should always be done.

matmair:
@eeintech the validation ensures that the sell price is the same as when it was entered and that a few parameters that might influence the price are the same.
That's why I think this pattern makes sense: if I enter 200€ and validate that I expect only 200€ with the same delivery date and the same items at the same quantities and prices to be valid. With each change - which do not happen automatically - I expect to be asked again to validate the price.

matmair:
@eeintech @SchrodingersGat any idea how to solve [this error](https://github.com/inventree/InvenTree/pull/1730/checks?check_run_id=3108489350#step:5:289)? I have been fiddeling with this for a week.

SchrodingersGat:
@matmair I'll have a look into this one next week

eeintech:
@matmair Looks like the error is the extra `files` argument of the `patch` api test method:
https://github.com/inventree/InvenTree/blob/bcfb3ac067f28b0ec416fdb839829441f45dcd33/InvenTree/InvenTree/api_tester.py#L138

I don't see it in the `APIClient.path` method (rest_framework -> test.py):
``` python
def patch(self, path, data=None, format=None, content_type=None, follow=False, **extra):
```

Maybe we want something like this instead?

``` python
response = self.client.patch(url, data=data, format='json', **{'files': files}) 
```

SchrodingersGat:
@matmair try this:

```
---
 InvenTree/InvenTree/serializers.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/InvenTree/InvenTree/serializers.py b/InvenTree/InvenTree/serializers.py
index 58d33697..8ec98cd7 100644
--- a/InvenTree/InvenTree/serializers.py
+++ b/InvenTree/InvenTree/serializers.py
@@ -46,7 +46,7 @@ class InvenTreeMoneySerializer(MoneyField):
             amount = None
 
         try:
-            if amount is not None:
+            if amount is not None and amount is not empty:
                 amount = Decimal(amount)
         except:
             raise ValidationError(_("Must be a valid number"))
-- 
2.30.2.windows.1

```

SchrodingersGat:
@matmair I've fallen behind on this one - is it ready for review?

matmair:
> 
> 
> @matmair I've fallen behind on this one - is it ready for review?

@SchrodingersGat let me merge in the last changes and check again but should be.

Maybe there should be a more general discussion between you, @eeintech and me if this feature is ready for master like this or if the both of you would like to see additional stuff. I want to reduce the number of half-ready features I submit into master 👼. Just irritates users.

matmair:
@SchrodingersGat should be good to go technically

eeintech:
@matmair Just took another look.

1. Sell price has 4-digit while calculate price has only 2-digit. Also space out the validation label from the closing bracket (better yet would be at the right-hand side of the cell like the "admin" label above):

image

2. Total in sale order line items adds up prices with different currency **ignoring** conversion rate (here 14.5 CAD + 120.1 USD shouldn't give a result of $134.6 USD but rather closer to $131.7 USD, or better yet should be in the default currency which is CAD):

image

3. PO column shows "undefined" for stock unrelated to POs (instead of blank). Underlying stock item information shows "undefined" instead of the name of the location (URL points to location):

image

Other than that it looks good!

matmair:
@eeintech thanks for the testing, will fix!

SchrodingersGat:
@matmair I'm getting some new form errors when trying to add a line item to a sales order:

image

eeintech:
@matmair Seems like this PR is stalling a bit. Was it a highly demanded /desired feature or should it be pushed to `0.6.x` releases?

matmair:
> 
> 
> @matmair Seems like this PR is stalling a bit. Was it a highly demanded /desired feature or should it be pushed to `0.6.x` releases?

@eeintech there will be some unittesting required and maybe it should be redone so I vote to target 0.6.x and to make that about pricing.

tse-tarn-france:
@SchrodingersGat @matmair, don't know if this takes place here but for parts that are "Purchaseable" AND "Salable" price may appear automatically when creating a sale order line. For build parts of course price is almost never the same, however a "reference sale price" could also appear and be modified.

This is a nice to have, absolutly not a priority.

Regards.

eeintech:
Didn't want it anymore @matmair ?

matmair:
@eeintech I will rewrite the whole pricing system and this PR just got ugly with how long 0.6.0 took. I will take some inspiration but do it a bit cleaner. Hashing should be a generator for example - so plugins can use the same function as prices and BOMs to validate. Is going to take a while but the current system is just not performant enough.

PR #1734, opened 2021-06-29
option to hide related parts

### LINKED ISSUE #1733: Add option to hide "related parts" tab

matmair:
Fixes #1733

### DIFF
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -219,6 +219,13 @@ class InvenTreeSetting(models.Model):
             'validator': bool,
         },
 
+        'PART_SHOW_RELATED': {
+            'name': _('Show related parts'),
+            'description': _('Display related parts for a part'),
+            'default': True,
+            'validator': bool,
+        },
+
         'PART_INTERNAL_PRICE': {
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
Thanks!

eeintech:
@matmair @SchrodingersGat Do we actually want to keep this feature? I would say yes, we're using it lightly and was requested in #652 and implemented in #1048 but if all those "niche" features are clogging InvenTree I don't think we should keep them..?

SchrodingersGat:
I'm happy to keep it for now. Minimal effort to keep it hidden - perhaps it should be the default?

eeintech:
I'm okay with hidden as default, made me realize the "documentation" is buried in release notes: https://inventree.readthedocs.io/en/latest/releases/0.1.4/#related-parts

lookme2:
I use it pretty heavily here at my place.  I vote to keep it.

matmair:
Maybe we should make it a bit more visible in the docs.

I personally do not use this feature very much and have no strong feelings on how the default should be, just implemented the PR because I had some spare time after lunch.

Issue #1245, opened 2021-01-19
Add "depleted stock" to homepage


SchrodingersGat:
Add display of stock items which are "depleted" (in stock but quantity = 0)

SchrodingersGat:
Additionally add settings or which items are displayed on the home page

matmair:
@SchrodingersGat how did you imagine the settings? One per item? And per user or global?

SchrodingersGat:
This was a while ago, I think I was imagining that each item on the home page could be configurable (simple boolean setting).

PR #1747, opened 2021-07-02
Depleted stock on start page

matmair:
This PR adds the first item of #1245, a filter for depleted stock on the home-page.

grafik


SchrodingersGat:
Simple and elegant! Thanks :)

PR #1749, opened 2021-07-02
Extend build order table

### LINKED ISSUE #1356: Add "issued_by" and "resposible_for" fields to Build Order table

matmair:
Closes #1356

Still missing user_detail for responsible user because it throws errors when the `responsible` field is null.


SchrodingersGat:
`responsible` field references the `Owner` model - https://github.com/inventree/InvenTree/pull/1716 adds an API endpoint and read-only serializer for Owner which might be useful here.

matmair:
@SchrodingersGat thanks for the serializer, didn't even see that the field was not linked to User model ;-)

SchrodingersGat:
Nice work, thanks :)

PR #1750, opened 2021-07-03
Stock table add supplier part column

### LINKED ISSUE #1323: Add "Supplier Part" column to stock table

matmair:
Fixes #1323


SchrodingersGat:
Nice, that's been hanging around for a while

PR #1757, opened 2021-07-04
Stock previous / next serial 

### LINKED ISSUE #1081: Navigate to next/previous serial number button
When on the stock item page for a given serial number, a button to navigate to the next serial number stock item would be handy, especially if it stayed on the same subpage (e.g. tests).

matmair:
Closes #1081

This PR implements a previous / next button next to the serial number of a stock that navigates to the stock item with the shown serial with the current view.

### DIFF
--- a/InvenTree/build/templates/build/build_output.html
+++ b/InvenTree/build/templates/build/build_output.html
@@ -72,6 +72,7 @@ loadStockTable($("#stock-table"), {
     params: {
         location_detail: true,
         part_detail: true,
+        supplier_part_detail: true,
         build: {{ build.id }},
     },
     groupByField: 'location',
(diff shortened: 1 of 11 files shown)


matmair:
grafik
@jnewlands is that what you had in mind?

SchrodingersGat:
Looks pretty good, but I would like to have the current serial number more prominent somehow

matmair:
@SchrodingersGat do you think making it bold would do the job? I was not sure if that would make it to prominent. Or make the prev/next smaller?

SchrodingersGat:
Maybe both?

matmair:
@SchrodingersGat did both now

SchrodingersGat:
Thanks @matmair this looks pretty nice now. Happy with merging this in.

eeintech:
@matmair Sorry to be late to the party but it looks like it does not work for 2-digit serial number:

image

Can it be supported?

matmair:
@eeintech see #1802 👼

eeintech:
Nice 👍

PR #1760, opened 2021-07-05
Converted BOM import to new multi-step form framework

eeintech:
The BOM import now uses the same multi-step framework than:
- purchase order imports (ref #1561)
- part imports (ref #1588)


SchrodingersGat:
Seems to work pretty well, thanks @eeintech

PR #1761, opened 2021-07-05
Fixes for stock modal and notes propagation to tracking items

eeintech:
I noticed that the modal was losing stock item information if not valid in the `POST` request (see missing stock item up top):

image

This PR fixes it:

image

Also the "Notes" field would not be propagated to the stock tracking items during a "move" action, this PR fixes this too


SchrodingersGat:
Thanks :)

PR #1763, opened 2021-07-06
Fixed supplier part detail handling in stock table

eeintech:
This line would throw JS error when "Supplier Part" column was enabled on the stock tables that did **not** specify to the API to return `supplier_part_detail` data:

https://github.com/inventree/InvenTree/blob/c2df1fcd95c461bb7f1929368e96f45b1ad779b5/InvenTree/templates/js/stock.js#L697

Also there were a few stock table calls referencing `supplier_detail: true` instead of `supplier_part_detail: true` (`supplier_detail` is not a valid API filter).

Now should be good!

### DIFF
--- a/InvenTree/company/templates/company/detail_stock.html
+++ b/InvenTree/company/templates/company/detail_stock.html
@@ -23,7 +23,7 @@
         params: {
             company: {{ company.id }},
             part_detail: true,
-            supplier_detail: true,
+            supplier_part_detail: true,
             location_detail: true,
         },
         buttons: [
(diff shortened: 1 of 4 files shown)


matmair:
@eeintech great catch 👍 I never made a PR for that fix 👼

eeintech:
@matmair You are doing so much more than my little fixes haha!

matmair:
> 
> 
> @matmair You are doing so much more than my little fixes haha!

@eeintech I try to but life gets in the way of my ToDo-lists 😆

Issue #1791, opened 2021-07-09
Combine "Manufacturers" and "Suppliers" tab for Part view


SchrodingersGat:
Annoying to switch between the two, especially as the tables are highly related.

PR #1793, opened 2021-07-09
Consolidate Manufacturers and Suppliers views

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/1791


eeintech:
Makes sense, thanks @SchrodingersGat !

PR #1812, opened 2021-07-13
Custom SupplierPart object manager and prefetch related models

### LINKED ISSUE #1809: Admin interface can be super slow
Some pages on the admin interface takes ages (like, 10 seconds) to render.. 

Need to profile them to work out where the inefficiencies are.

eeintech:
Should fix #1809 and maybe improve performance a bit when lookups to SupplierPart objects are made :smiley:


SchrodingersGat:
Thanks for the improvement. We should try to profile the other admin pages too.

PR #1829, opened 2021-07-18
Initial amount

### LINKED ISSUE #1796: Feature: The way to create initial stock amount from "Create New Part" dialogue.
Each time I create a new part, I have to go to the Stock Items section and create new stock. It would be faster, if directly on "Create New Part" would be an option to set an initial number of stock items in default location. Just a number, and if it is 0, the initial stock will be ignored and the system work as it is now

matmair:
This PR adds the ability to create initial stock when creating a part. If a default location is set it is used as the location for the part, there is a setting to enable this feature, defaluted to false.
Closes #1796

### DIFF
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -233,6 +233,13 @@ class InvenTreeSetting(models.Model):
             'validator': bool,
         },
 
+        'PART_CREATE_INITIAL': {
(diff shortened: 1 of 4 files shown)


SchrodingersGat:
Thanks @matmair this seems like a pretty simple way of implementing this.

PR #1838, opened 2021-07-18
Search add autocomplete

### LINKED ISSUE #280: [Long term] Display live search results in the search box at top right
- Use ajax
- Display results
- ???
- It's an older meme sir

matmair:
This PR adds autocomplete to the search box. Selecting a result takes the user to the part directly, if enter or the search button is used the search-page appears.

Closes #280


matmair:
@SchrodingersGat I do not know how to fix the styling issue, I get either a W503 or W504 error with the operators but have to break the line somewhere

SchrodingersGat:
@matmair instead of adding a new View for this, the live search results should be retrieved using the existing API search functionality.

e.g. `/api/part/?search="widget"` - this already searches across the name, description, keywords, IPN fields and I'd not like to introduce a duplicate method.

matmair:
@SchrodingersGat the idea was to create a very light endpoint that serves very light responses and could be extended to search more later. While the new view gets me a response in 20ms it takes the normal part-search 500ms - the idea behind a typeahead / autocomplete would be to be as fast as possible.

matmair:
New method 20ms: https://github.com/inventree/InvenTree/pull/1838/commits/4013abe4de1d3cff06d28fc3aa334016fa550d50
Old methid ~500ms: https://github.com/inventree/InvenTree/pull/1838/commits/43f1e2b8f9162475395e08785bbcfb28d5f234b6

SchrodingersGat:
@matmair if the API for part search is taking ~500ms then I think we need to look at where the inefficiencies are there.

SchrodingersGat:
It actually seems pretty performant on my end.

I would suggest you limit the number of responses in the API request, so that on a generic query it doesn't return heaps of results (and slow down the response).

`/api/part/?search=widget&limit=25&offset=0` (for example)

SchrodingersGat:
Looks really good, I'd like to see a couple of visual tweaks:

- Increase spacing slightly between search results
- Prepend small thumbnail image to each result

matmair:
> 
> 
> It actually seems pretty performant on my end.
> 
> I would suggest you limit the number of responses in the API request, so that on a generic query it doesn't return heaps of results (and slow down the response).
> 
> `/api/part/?search=widget&limit=25&offset=0` (for example)

At least in my dev enviroment I do not go below an average of 500ms even with limits set
grafik
My database contains a test-dataset of around 400 parts

SchrodingersGat:
Made some significant improvements to the part API speed in #1850  - check that out once merged...

matmair:
@SchrodingersGat the UI stuff should be finished

SchrodingersGat:
Nice work @matmair this looks pretty great now!

eeintech:
Amazing stuff! Bravo @matmair  :clap:

PR #1846, opened 2021-07-19
Allow BOM pricing to be valid when using internal pricing - style fix

matmair:
This PR is #1718 but without the style error as the author did not react for 6 days.

Closes #1718


eeintech:
Nice :+1:

PR #1852, opened 2021-07-20
Improved handling of PO items destination

### LINKED ISSUE #1832: Purchase order clarification
Since there is a lack of documentation about PO I would like to ask some (maybe silly) questions.

Prerequisites:
- Parts with filled manufacturer MPN and/or supplier SKU.

What is a suggested manual PO workflow?
- Create PO
- Use "Order Parts" dialog from Part to add a line to PO.
- (optional) Add order lines manually.
- Click "Place order"
- Make real order on supplier(s) website. Wait for shipping.
- Click "Receive Parts", fill actual received amount and status
- Click "Complete Order"

Is it a correct Upload File for Purchase Order UseCase?
- Create PO
- Make order on supplier website
- Receive parts and accounting + inventory docs
- Upload inventory sheet to PO. Automatically fill order item lines, quantities, prices.
- Click "Place order"
- Click "Receive Parts", fill actual received amount and status
- Click "Complete Order"

Questions:
- Currently it looks like I have to fill part price even before making real supplier order, is it correct?
- What if price has changed during making real order? But as I already clicked "Place order" I can't change item price.
- What is the purpose of "destination" field in order item line, if I have to choose "Destination" each time I click "Receive Parts"?
- Does it make sense to keep "Pricing ranges" in inventree because supplier prices could change pretty often.
- Does it make sense to fill order item price when parts are received (not before they are ordered)?

eeintech:
Fixes #1832

* Adds "Destination" as select field in the receive part form
* Made the global "Destination" an optional field (stock location is optional for stock items)
* If the "destination" value is set for purchase order item then use it
* If the global "destination" field is set, override values of each line in the form

### Order with items set to different destinations
image

### Receive form
image

### DIFF
--- a/InvenTree/order/forms.py
+++ b/InvenTree/order/forms.py
@@ -84,9 +84,9 @@ class ReceivePurchaseOrderForm(HelperForm):
 
     location = TreeNodeChoiceField(
         queryset=StockLocation.objects.all(),
-        required=True,
+        required=False,
         label=_("Destination"),
-        help_text=_("Receive parts to this location"),
+        help_text=_("Set all received parts listed above to this location (if left blank, use \"Destination\" column value in above table)"),
     )
 
     class Meta:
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
Thanks @eeintech

PR #1856, opened 2021-07-21
Add instance-specific filters to API OPTIONS data

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/pull/1851

Adds a separate field to the OPTIONS endpoint for ForeignKey fields, called `instance_filters`

If the API endpoint is associated with a particular model instance, then that model has the opportunity to provide custom filters *for that model instance*

- [x] Part
- [x] PartCategory
- [x] StockItem
- [x] StockLocation
- [x] BuildOrder

This does not require hard-coded filters to be added to the client-side code, it all happens on the server. It also allows a future path for other instance-specific filters to add other functionality.

### DIFF
--- a/InvenTree/InvenTree/metadata.py
+++ b/InvenTree/InvenTree/metadata.py
@@ -32,6 +32,9 @@ class InvenTreeMetadata(SimpleMetadata):
 
     def determine_metadata(self, request, view):
         
+        self.request = request
+        self.view = view
+
         metadata = super().determine_metadata(request, view)
 
         user = request.user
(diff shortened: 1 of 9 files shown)


SchrodingersGat:
@eeintech give this a go

eeintech:
@SchrodingersGat It looks good to me.

One question: what if one wants to override the `instance_filters`, how would you do it?

SchrodingersGat:
> One question: what if one wants to override the instance_filters, how would you do it?

In the client side, you can easily override the values being supplied to the "filters" parameter for each field.

For any filter key (e.g. `exclude_tree` is the new one in this PR) the filter values are determined as follows (in decreasing order of priority):

1. "filters" parameter supplied by the calling code on the client side
2. "instance_filters" supplied by the OPTIONS endpoint
3. "filters" supplied by the OPTIONS endpoint

The calling javascript can *always* override the filters if it thinks it knows better, e.g.:

```js

constructForm("/api/url/pk/", {
    fields: {
        my_field: {
            filters: {
                category: 7
            }
        }
    }
});
```

eeintech:
Ok nice, thanks @SchrodingersGat

PR #1858, opened 2021-07-22
Exchange rate as worker task

eeintech:
@SchrodingersGat I hope this is the way it was intended to be run as a task.

I added a generic function to run task either by the worker cluster or as a blocker.
Let me know what do you think.

Now anybody can add a method to the `InvenTree/tasks.py` file and run it using the worker cluster simply with:
``` python
from InvenTree.tasks import run_task

# 'blabla' method needs to be implemented in InvenTree/tasks.py
run_task('blabla')
```

I've tested it with `update_exchange_rates` method and it runs fine both by the cluster and as blocker.


SchrodingersGat:
@eeintech what's the intended use here? The currency rates are already updated periodically by the background worker, and I think it is better to leave the process "blocking" here so that the server responds with the latest values.

eeintech:
Maybe `update_exchange_rates` is not the best use for it but, as said in the PR description, this is a helper function to make it easy to run tasks from anywhere? That way when more and more tasks methods are implemented, there wouldn't be too much overhead in each call.
I guess you are suggesting a way to force use of blocking vs using the cluster?

SchrodingersGat:
Is this functionally different from the existing `offload_task` function? 

https://github.com/inventree/InvenTree/blob/529742b5203005d5d71921c1ee32cd1bc540af4d/InvenTree/InvenTree/tasks.py#L54

eeintech:
@SchrodingersGat I guess I kinda overlooked `offload_task` however I think this one will error out if:
* function defined by `taskname` does not exist
* if there are no workers running, it might just stay in the queue forever and the user won't know

In `run_task` I am handling those two checks:
* if the task function does not exist or is not found in `tasks.py`, user gets a warning
* if workers are not running, the task is run synchronously

Do you want me to re-use `offload_task` and add the checks there?

SchrodingersGat:
> if workers are not running, the task is run synchronously

This would be a good addition to the `offload_task` function, to ensure that the function does actually get run even if the background worker is not active. 

> if the task function does not exist or is not found in tasks.py, user gets a warning

Makes sense to add to `offload_task` too.

eeintech:
@SchrodingersGat Take a look at the updated `offload_task` code which includes the two checks. Now running `update_exchange_rates` synchronously using `force_sync` option :+1: 

Should be good to merge, I'm actually starting to look how can we use workers for distributor API, starting with Digi-Key.
Digi-Key unfortunately does not provide any permanent token so I'm hoping we can get a worker to keep on refreshing the token in the background every hour or so and don't have to go through the app authorization phase again.

InvenTree/InvenTree/tasks.py
```diff
@@ -51,22 +51,54 @@ def schedule_task(taskname, **kwargs):
...
+
+        If workers are not running or force_sync flag
+        is set then the task is ran synchronously.
     """
 
+    # Get task list
+    tasks = get_task_list()
+
+    # Check if task exists
+    if taskname not in tasks:
```
SchrodingersGat, line 71: I don't want to limit this to *only* tasks within this particular file. We may in the future have a lot of async tasks split between the various apps, e.g.

`stock.tasks.doSomeTask`

eeintech: Umm not sure how can we guarantee the task is implemented in this case, any idea?

SchrodingersGat: You should have a look at how django-q manages this

eeintech: @SchrodingersGat Please take a look at the new implementation, I'm sure it will be much more to your liking :smiley:

SchrodingersGat:
Yeah this looks pretty solid now! Thanks @eeintech

PR #1860, opened 2021-07-23
fix for phantom migration warning

### LINKED ISSUE #1775: Migration warns for phantom part changes
Here is the warning:

```
Your models in app(s): 'part' have changes that are not yet reflected in a migration, and so won't be applied.
Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

Running `manage.py makemigrations` does **not** generate new migration file...

Running `manage.py showmigrations part` shows all part migrations are complete.

I found this warning both with PostGreSQL and SQLite3, if that has anything to do with backend dependency.

matmair:
Closes #1775

@eeintech So turns out the phantom part changes shown are the result of me forgetting to also check if a migration is run, not only if migrations are generated. This PR fixes that.

### DIFF
--- a/InvenTree/InvenTree/fields.py
+++ b/InvenTree/InvenTree/fields.py
@@ -55,7 +55,7 @@ class InvenTreeModelMoneyField(ModelMoneyField):
     
     def __init__(self, **kwargs):
         # detect if creating migration
-        if 'makemigrations' in sys.argv:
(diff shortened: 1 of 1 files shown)


matmair:
@G-Pereira this also caused the error you had

eeintech:
@matmair Nice find, thanks!

PR #1887, opened 2021-07-30
settings fixes

matmair:
Fixes two major problems that I found with the new settings.js if the database was migrated a lot:
- empty key fields are possible in the setting; fix: only include setting in allValues() that have a key
- old settings stay in the database; fix: a clean-up function

### DIFF
--- a/InvenTree/InvenTree/management/commands/clean_settings.py
+++ b/InvenTree/InvenTree/management/commands/clean_settings.py
@@ -21,7 +21,6 @@ class Command(BaseCommand):
 
         # check if key exist and delete if not
         for setting in db_settings:
-            print(setting.key)
             if setting.key not in model_settings:
                 setting.delete()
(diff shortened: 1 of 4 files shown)


SchrodingersGat:
Good catch. Let me know when it is ready

sintech:
Please also add check for unescaped chars in string values. 
I had an apostrophe in site name which caused rendering issues.

matmair:
@sintech will look into that. Normally all strings should be escaped by Django but maybe that's turned off somewhere

matmair:
@sintech escaping was not considered in the code till now, I added it for settings via databases. Please check out if your case works now and report what makes errors if not.

sintech:
@matmair, yes it works fine now. Apostrophe got HTML entity encoded in settings.js. 
`INVENTREE_INSTANCE: 'Sintech&#x27;s InvenTree server'`

PR #1892, opened 2021-08-02
Re-enabled installing stock items into others

### LINKED ISSUE #1770: Feature: Add installed stock items after build creation
Include an "Add stock item" button to the install stock items tab for a stock item.

Currently installed stock items can only be added whilst in production through a build order. Would be beneficial to be added after production as well.

eeintech:
Intent is to fix #1770 

This implementation manages both:
* installing the current stock item into another stock item (action item "Install)

image

* installing other stock items into the current stock item ("Installed Items" tab)

image

### DIFF
--- a/InvenTree/stock/forms.py
+++ b/InvenTree/stock/forms.py
@@ -241,15 +241,20 @@ class InstallStockForm(HelperForm):
         help_text=_('Stock item to install')
     )
 
-    quantity_to_install = RoundingDecimalFormField(
-        max_digits=10, decimal_places=5,
-        initial=1,
-        label=_('Quantity'),
-        help_text=_('Stock quantity to assign'),
-        validators=[
-            MinValueValidator(0.001)
-        ]
+    to_install = forms.BooleanField(
+        widget=forms.HiddenInput(), 
+        required=False
     )
+
+    # quantity_to_install = RoundingDecimalFormField(
+    #     max_digits=10, decimal_places=5,
+    #     initial=1,
+    #     label=_('Quantity'),
+    #     help_text=_('Stock quantity to assign'),
+    #     validators=[
+    #         MinValueValidator(0.001)
+    #     ]
+    # )
 
     notes = forms.CharField(
         required=False,
@@ -261,7 +266,7 @@ class InstallStockForm(HelperForm):
         fields = [
             'part',
             'stock_item',
-            'quantity_to_install',
+            # 'quantity_to_install',
             'notes',
         ]
 
--- a/InvenTree/stock/forms.py
+++ b/InvenTree/stock/forms.py
@@ -245,16 +245,6 @@ class InstallStockForm(HelperForm):
         required=False,
     )
 
-    # quantity_to_install = RoundingDecimalFormField(
-    #     max_digits=10, decimal_places=5,
-    #     initial=1,
-    #     label=_('Quantity'),
-    #     help_text=_('Stock quantity to assign'),
-    #     validators=[
-    #         MinValueValidator(0.001)
-    #     ]
-    # )
-
     notes = forms.CharField(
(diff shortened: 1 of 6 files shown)


wolffam:
@eeintech , cool!! Thanks for taking a look at this!
Does this also allow for stock items to be uninstalled from a stock item?  My read of the code says that should be the case but I don't remember how it looks on the front end.

eeintech:
@wolffam Yes you can definitely do that too, this feature was actually already there.

* Uninstall an "installed" stock item:

image

* Uninstall stock items from list of installed stock items:

image

Note that installing/uninstalling currently only supports serialized stock items, which I believe is the main use-case for this.

wolffam:
great! From my read it looks like it brings back the functionality we had before.  Thanks!

InvenTree/stock/forms.py
```diff
@@ -241,16 +240,21 @@ class InstallStockForm(HelperForm):
...
-        help_text=_('Stock quantity to assign'),
-        validators=[
-            MinValueValidator(0.001)
-        ]
+    to_install = forms.BooleanField(
+        widget=forms.HiddenInput(),
+        required=False,
     )
 
+    # quantity_to_install = RoundingDecimalFormField(
```
SchrodingersGat, line 248: Please remove these commented-out lines

InvenTree/stock/forms.py
```diff
@@ -261,7 +265,7 @@ class Meta:
         fields = [
             'part',
             'stock_item',
-            'quantity_to_install',
+            # 'quantity_to_install',
```
SchrodingersGat, line 268: Remove

InvenTree/stock/views.py
```diff
@@ -558,11 +595,17 @@ def get_initial(self):
         if items.count() == 1:
             item = items.first()
             initials['stock_item'] = item.pk
-            initials['quantity_to_install'] = item.quantity
+            # initials['quantity_to_install'] = item.quantity
```
SchrodingersGat, line 598: Remove

SchrodingersGat:
Looks pretty comprehensive @eeintech - just a few small points to address.

eeintech:
@SchrodingersGat Sorry I commented out the old code which was setting the quantity through the form, mostly for comparison with the new one. It's gone now!

SchrodingersGat:
Great, thanks for your work on this!

wolffam:
@eeintech , once again thanks for the work on this!
Was there a specific reason why you limited the install to "serialized" stock items as opposed to "trackable" stock items (I think that's how it was before the initial regression).  We have a workflow that sometimes involves replacing stock items that are trackable (by batch number) but don't have specific, individual serial numbers

eeintech:
@wolffam I don't really understand how would that work..? Do you have a real life example?

To me installing an item into another means you need to know exactly which items you are referring to, not just an "idea" (eg. a batch) of which items are going into which group of items. Or that'd become more complex I guess and might come outside the scope of InvenTree :confused: 

Let us know how would see this be implemented in a neat way and we can try to get that done.

wolffam:
For my use case, if I have an assembly with 10 screws (all from batch number 7) and I need to replace those screws (after so many reinstalls), I would like to remove those 10 screws (batch 7) and install 10 new screws of a new batch.  It's not a common workflow but one we have hit in the past.  
There may be an analog to the more common use case of Inventree with electronic components but I can't think of one off the top of my head.

wolffam:
With that said, if there isn't a clear use case for the majority of users, I can implement some workarounds on my end that will allow us to handle those cases

eeintech:
I'd have to look into this more, might get pretty ugly to handle such niche case... If you find a neat workaround, do you mind posting your own PR?

PR #1895, opened 2021-08-03
'has_ipn' filter method did not return queryset

eeintech:
With this PR it does :smiley:


matmair:
@eeintech nice catch; intresting that nobody found that in the last month

SchrodingersGat:
Thanks @eeintech

Issue #1889, opened 2021-08-01
Use purchase_price for BOM cost calculation


sintech:
Current BOM pricing approach implies that all BOM items should be supplier parts with pre-filled price breaks. And it gives you the cost of item if you buy all parts from corresponding suppliers.
While in my case, I assemble my projects from parts I already have. So I would like to know how much it will cost to assemble this item based on parts I have in stock (purchase price).

Please consider replacing supplier price with purchase price for the BOM pricing view. This could be the global/user settings option.
@matmair what do you think?

matmair:
@sintech that is not quite as simple as it sounds. A lot of EE people collect parts from different vendors at different price points. So if you have a part of which you bought 5 at 4ct, 10 at 6ct and 40 at 5ct - how much should InvenTree charge for 3 pieces? Just 4 because it is the middle? A weighted middle? Should that change every time you make a PO? What if you use parts like this in assemblies and you have a complex part with 100s of subassemblies? Should we calculated the price of each part bottom up? Would anything ever have a fixed price?

Our solution to that currently is the internal price: If activated you can set a price that should be used for BOM calculations. The pricing section of a part provides you with a graph of the cost at which you bought inventory/stock so you can choose a good middle-ground you should charge.

matmair:
@sintech if you have a idea how the problems stated above can be handled I am happy to review a PR.

sintech:
@matmair, I got you point and it is fair one.
But can we do just min and max price as it is currently done in code. As I can see, `get_supplier_price_range` loops over all suppliers, get prices based on quantity and finally return one min and max price.

Internal price is a good tool for something like workforce or price override. I don't want to abuse it by copying data I already have in DB.

eeintech:
@matmair I have to admit that the current pricing data present in the current version of InvenTree is quite complex and is not documented... I personally just discovered that the internal pricing option was disabled on our instance and found it in the setting. But still unsure what it does!

If I look at a BOM, I see we have 3 columns for pricing: purchase price average, purchase price range and buy price (stock price?).
And no idea what the "dollar sign" button calculation form uses for part pricing...

The pricing graph is still unclear too:

![image](https://user-images.githubusercontent.com/4020546/127882617-241a39a0-363f-4587-bb61-24e2ae306f1c.png)

"Single Price" vs "Part Single Price?
Information icon does not overlay any information.

"Part Pricing" form is also not telling where does the price comes from, even though I _know_ it is supplier price break.

[...]

PR #1897, opened 2021-08-03
updating language to be clearer

matmair:
see https://github.com/inventree/InvenTree/issues/1889#issuecomment-891901070

cleaner and clearer language is important for UX ;-)


eeintech:
Nice work @matmair !

PR #1907, opened 2021-08-04
translation improvement

### LINKED ISSUE #1896: Tranlsation improvments
@SchrodingersGat we have a lot of translations that are barely started. I would like to try fixing that with two small UI changes that might encourage users to submit translations:
- [x] showing the percentage of translated strings int he language selection box
- [x] Adding the information how to translate stuff to the selection box itself (maybe with a setting to disable that)

A more proactive step would be to ask admin-users with a language that is not translated a lot to consider submitting translations - I do not think that would find support but would be a possibility.  
I just think we should do something because languages with 10% translations are not helping anyone.

matmair:
Like described in #1896 this PR adds more information about translations and adds a call to action.

Closes #1896


SchrodingersGat:
Seems to work pretty well, thanks @matmair

Issue #1755, opened 2021-07-03
Outstanding tasks for API Forms


SchrodingersGat:
Remaining tasks / features / bugs to be implemented from https://github.com/inventree/InvenTree/pull/1716

### Form Features

- [x] Custom "grouping" of form fields
- [x] Custom form elements
- [x] Launch "secondary" modals (and handle the results)
- [x] Allow "infinite" stacking modals (created at run-time)
- [x] Option for custom client-side field validation (before sending data to the server)
- [ ] Inter `min_value` from numerical django model fields
- [ ] Infer `max_value` from numerical django model fields
- [ ] Correctly render file fields which already have a file, in "edit" mode (e.g. StockItemAttachment)
- [ ] Display "warning" screens for deletion - i.e. "these things will also be deleted"
- [ ] Allow non-required file / image fields to be cleared
- [x] Insertion of fields defined completely at "run time"
- [ ] Run custom code before the form is displayed, but after data is fetched
- [x] Run custom code before form is submitted
- [ ] Detect if user does not have required permissions and hide secondary buttons accordingly
- [ ] Migrate settings forms to the API

### Bugs

- [x] Launching a secondary modal with a select2 attached strips the select2 features from the first modal... 

### Supported Input Types

- [ ] datetime field
- [ ] time field
- [ ] multiple choice field

### Python Integration

- [ ] Python support
- [ ] Add support to the python bindings to allow dynamic lookup of "queryset" for a given ForeignKey field.

### Unit Testing

- [ ] Ensure the OPTIONS endpoint correctly returns min_value and max_value for fields
- [x] Correctly handle invalid inputs for *all* MPTT fields (e.g. setting invalid parent node)
- [x] Automatically remove invalid choices from MPTT fields (e.g. don't let user select invalid "parent" items)
- [x] Save user information when creating or updating an attachment
- [ ] Save user information when creating or editing a StockItem 
- [x] Ensure that the 'required' property is getting set correctly - both for fields and relations
- [ ] Ensure that 'read_only' property is getting set correctly - both for fields and relations - e.g. looks like this might not be the case for StockItemSerializer

### Documentation

- [ ] Add API docs for OPTIONS endpoints

### References

- https://github.com/inventree/InvenTree/issues/1830
- https://github.com/inventree/InvenTree/issues/1683
- https://github.com/inventree/InvenTree/issues/1213
- https://github.com/inventree/InvenTree/issues/707
[...]

PR #1912, opened 2021-08-05
Re-added ability to link manufacturer part to supplier part via API

eeintech:
@SchrodingersGat In #1836, the ability to link a manufacturer part to a supplier part using:
- manufacturer ID
- manufacturer MPN

was removed.

The idea was to create a shortcut in the creation of the supplier part, not having to find out which manufacturer parts are associated to the same part on the client side. In this case the validation of link is done on the server side and, if both information above are correct, the supplier is automatically connected to the corresponding manufacturer part.

Did you intentionally intend to force the client to retrieve the list of manufacturer parts, then get the correct manufacturer part id, then link it directly to the supplier part? Or are you open for shortcuts in this process? (if you have any better idea than the one presented here, let me know!)

### DIFF
--- a/InvenTree/company/models.py
+++ b/InvenTree/company/models.py
@@ -9,7 +9,9 @@ import os
 
 from django.utils.translation import ugettext_lazy as _
 from django.core.validators import MinValueValidator
+from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.utils import IntegrityError
 from django.db.models import Sum, Q, UniqueConstraint
 
 from django.apps import apps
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
The removal was intentional, as part of the migration to the new "API forms" approach. 

While I was happy with the original implementation at the time, I don't like that it's "different" to how all other forms are managed - i.e. a single form could touch two different database tables in an opaque manner. In particular if the user makes a slight typo in the MPN, they will not know that it will result in a new manufacturer part being created.

The "ideal" shortcut here would be to re-implement the "secondary" modal forms in the API forms. i.e. if the required MPN data is not already in the available list, you can simply launch a secondary modal form on top of the current one to create the new manufacturer part, and then once that form is completed, the data is filled back into the original form.

Re-implementing the secondary forms has run into a road-block and is an outstanding task here - https://github.com/inventree/InvenTree/issues/1755

Once I've worked out how to implement that, it should be relatively simple to fix that here.

eeintech:
@SchrodingersGat I understand you are concerned about "API forms".

However my original question is related to external API calls using the `inventree-python` package.
I would have to review the code again but I don't think it was creating any new Manufacturer part on its own, it was just a way to identify a manufacturer part filling in both the name and part number.
When you don't have the UI form, it makes it easier to pass on this data (that you get from as spreadsheet or a supplier search) than having to fetch the manufacturer part ID, the fetching of the manufacturing part ID can be done on the server side.
Where do you think we can add this, if not in the serializer and `save` method?

SchrodingersGat:
Fair enough, I guess I did not fully understand the use case. If you want to submit a PR to re-introduce that feature into the API endpoint I'd be happy to merge back in.

PR #1925, opened 2021-08-08
Fix for missing part filter

### LINKED ISSUE #1915: Internal price is not filtered by part
Is this a normal behavior?
It looks like API call is missing `part` filter. So all parts have the same `Internal Cost` table.
image

image

matmair:
Fixes #1915


matmair:
@SchrodingersGat I just noticed that the part price-break API-Endpoints do not supply the price-break values as a number. This PR adds that - but is a API breaking change

SchrodingersGat:
Please bump the API version number here, then, and add a doc string:

https://github.com/inventree/InvenTree/blob/9565b3004d0d9fd51ab9a04c5fc746f37d24e0b7/InvenTree/InvenTree/version.py#L13

PR #1928, opened 2021-08-09
Localization of item.quantity removed

erkutalakus:
Localization of quantity for different cultures(turkish in my case) using comma(,) instead of dot(.) leads syntax error in javascript code and prevents stock item history to load.


SchrodingersGat:
Thanks for the contribution @erkutalakus 

Localisation is a difficult problem to get right!

Issue #1737, opened 2021-06-30
"permission denied" error when starting background worker


BansheePrime:
I'm followint the  docker development server setup guide and when I say:
`docker-compose -f docker-compose.dev.yml up -d inventree-dev-worker`

I got:
```
Starting inventree-dev-worker ... error

ERROR: for inventree-dev-worker  Cannot start service inventree-dev-worker: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/home/inventree/docker/start_dev_worker.sh": permission denied: unknown

ERROR: for inventree-dev-worker  Cannot start service inventree-dev-worker: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/home/inventree/docker/start_dev_worker.sh": permission denied: unknown
ERROR: Encountered errors while bringing up the project.
```

I'm only learning and your help will be much appreciated.

SchrodingersGat:
Have you copied the `docker-compose.yml` file exactly, or have you made alterations?

BansheePrime:
> Have you copied the `docker-compose.yml` file exactly, or have you made alterations?

`docker-compose.yml` is changed by me in {volumes}  section with path to local data. As I understand error with `permission denied` is result of attempt to execute script  `start_dev_worker.sh` in container and there we acting as root or I missing something very obvious? Local sudo'ing gives same error. I'm working with WSL2 on Win10 so let me try to setting up project on my unix-box.

Thank you for taking your time to answer my questions. I'm learning and definitely missing something here.

SchrodingersGat:
As with https://github.com/inventree/InvenTree/issues/1735 - are you manually building the dockerimage or downloading from dockerhub?

The inventree-worker and inventree-server contains should normally run "out of the box" without such issues...

SchrodingersGat:
can you please try again now that #1735 has been fixed?

BansheePrime:
Please excuse me for interrupting your work with weekend ahead. 
Right now in app Dockerfile line 1 is:
`FROM alpine:3.13 as base`

It worked for me only if I write:
`FROM python:alpine3.13 as base`

If you don't mind I'll take little more time to properly check things out.

SchrodingersGat:
Ok, let me know what you discover!

SchrodingersGat:
@BansheePrime did you ever get anywhere with this?

BansheePrime:
> @BansheePrime did you ever get anywhere with this?

I'll do my best to show you results in a few days.

[...]

PR #1935, opened 2021-08-10
Feature/simple startup

markdedeuge:
Proposal/Discussion:

- [ ] Migrate the _should change_ environment variables out of the docker-compose files. These can be supplied by a .env in one place
- [ ] Provide a simple 'one-liner' or set of Makefile commands to run the setup using the .env vars

Initial InvenTree setup:
- [ ] Pass through the database name (postgres only for now)
- [ ] Pass through the superuser credentials 

This could break users current workflows (as they may have hardcoded their config into the compose files)


markdedeuge (commented):
This fixes #1737

matmair:
@markdedeuge nice idea. Before releasing this we should def. update the [docs](https://github.com/inventree/inventree-docs) because the docker section conflicts with this PR.

@SchrodingersGat just tried it and it breaks setups made to the 'old' docks so on release there should be some kind of warning because a lot of docker-setups will break.

eeintech:
@markdedeuge Any plan on pushing forward in the coming days?

markdedeuge:
@SchrodingersGat could you let me know what you'd like to see changed/done here?

Issue #1673, opened 2021-06-17
Redesign of Part main page


sintech:
I propose to reduce the number of clicks needed to get all part info by changing the contents of Part page.
Current `Part Details` pane have info and elements rarely used during part lifetime, so it could be moved to one of the secondary pages. While the other info, such as: stock, parameters, notes and attachments should be available right on the main page.
Tables could be limited to display only N rows with link to corresponding pane to get everything else. Or vertical scroll-bar could be used instead.
Please find attached a mockups of main page for wide and narrow browser windows.
What do you think?

![part_details_wide](https://user-images.githubusercontent.com/10057327/122350997-61ce9e00-cf56-11eb-8046-6644ed2c218f.png)
![part_details_narrow](https://user-images.githubusercontent.com/10057327/122350995-61360780-cf56-11eb-98e4-4da45c4a80be.png)

SchrodingersGat:
In general I agree that the part detail page could be more "information dense"

I'd like to see a "simplified" version of the stock table here, with the existing stock table kept on a separate tab - otherwise i think it would look too cluttered.

eeintech:
It's a fair asking, the difficulty relies on which information to display first, each and every InvenTree user would have their own preference...

sintech:
> "simplified" version

Entirely agree with this, all tables could be just static html tables with minimum necessary information. 

![part_details_lite](https://user-images.githubusercontent.com/10057327/122400694-9fe2b680-cf84-11eb-8d93-a796b8de0cbb.png)

SchrodingersGat:
I certainly don't want to have to support "configurable" displays, unless someone puts up their hand to manage the template / javascript coding for that!

Any changes here would have to be more than simple aesthetic preferences, otherwise we'd have to manage multiple stylistic changes all the time!

@sintech I like the simplified version you've outlined above. Not very intrusive at all. And most of the "important" information is front and center.

matmair:
@SchrodingersGat sounds like a nightmare with the current frontend-stack 👼

@sintech proposal makes sense, maybe keep the current info-header for the more detailed views and just change `/part/<pk>`?

eeintech:
I like the proposal too. One warning is about the size of the front tables, it should be limited to 2-3 elements max, preferably filter in "last updated" fashion. If you have 10-15 parameters per parts (like we do), tons of attachments and stock items, it will make it worse!

sintech:
> size of the front tables

[...]

PR #1819, opened 2021-07-15
Cleaner part details

eeintech:
To build up on the part page refactoring started in #1811 

## Screenshots

image

image

image

PR #1943, opened 2021-08-10
Part summary panel

sintech:
Implementation of mockup from #1673
image

My first idea was to get data right from prerendered tables but then I found `load-success.bs.table` event and decided to take all info from tables JSON. Lazy loading almost broke all my work, so just for these three tables I preload data by triggering `fadeInStarted` event. Is it ok?
I'm slightly confused by how `{% settings_value ` work, what is the scope of this variable?

@SchrodingersGat what do you think about this?

### DIFF
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -946,6 +946,13 @@ class InvenTreeUserSetting(BaseInvenTreeSetting):
             'default': 10,
             'validator': [int, MinValueValidator(1)]
         },
+        'PART_DISPLAY_SUMMARY': {
+            'name': _('Part Summary'),
+            'description': _('Display Part Summary'),
+            'default': False,
+            'validator': bool,
+        },
+    
     }
 
     class Meta:
--- a/InvenTree/part/templates/part/detail.html
+++ b/InvenTree/part/templates/part/detail.html
@@ -10,6 +10,42 @@
 {% endblock %}
 
 {% block page_content %}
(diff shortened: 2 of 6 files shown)


SchrodingersGat:
@sintech you have stock of the STM32F446? Lucky you!

SchrodingersGat:
to be honest I am not sure this redesign is needed now that:

- The collapsible "part details" display has been added
- The previous "details" display has been "slimmed"
- Various tabbed views can be switched dynamically without reloading the entire page

@eeintech @matmair any thoughts on this one?

eeintech:
Any STM32 stock is worth more than bitcoin right now :smile: 

Sorry @sintech I have to agree with @SchrodingersGat I don't think this change brings much to the table...

sintech:
@SchrodingersGat @eeintech I'm humbly accept your opinion.
But at least please consider making Part description always visible b1bf02235b1151f3bd1ec290e2de2abcd7f538d5. I really liked pre #1819 part view, no offense @eeintech, currently it occupies almost 20mm from my vertical screen space only to show Part name.

The proposed changes are optional, you can switch them of if you don't like it.

If you don't accept this PR, please describe my options to keep it on my inventree instance. Is it possible to do via plugins or I have to apply this as patch after every update?

eeintech:
@sintech Closing this PR, you can re-open another PR to shrink the part name if that bothers you :)

PR #1949, opened 2021-08-12
Feature/dockerfile streamline

markdedeuge:
Prep for enabling a horizontally scalable inventree deployment

Change order matches commit messages.


matmair (commented):
Nice work already @markdedeuge docker is always a bit like blackmagic to me but this makes it very clear and mass-deployable

markdedeuge:
@SchrodingersGat what is the expected behaviour of the docker-compose.dev.yml startup when no ./dev virtual env exists?
Do we expect the worker to constantly restart while it waits for that venv to complete?

SchrodingersGat:
The worker should definitely not be started until the venv exists, if that is what you are asking? Perhaps a warning and sleep until it's present?

markdedeuge:
@SchrodingersGat I have put the startup work into the entrypoint scripts, but the procedure for waiting for the venv, db, etc means that you want to add that wait-for script before the entrypoint. 

So i'll make a change to keep those init scripts `start-worker` and `start-server` but call them `init-worker` `init-server`

This will mean that a the docker-compose file would have something like:

`services:
    server_init:
        ...
        command: ["/wait-for.sh", "--timeout=0", "db:{DBPORT}", "--", 
                             "/server-init.sh", 
                             "gunicorn", .... ]

    server_slave:
         ...
         command: ["/wait-for.sh", "--timeout=0", "server_init:8000", "--",
                             "gunicorn" ... ]

`
I'll push the final version of this PR before 12 today.

markdedeuge:
Please note:
This is a chicken&egg PR. In order to test the docker-compose* behaviour, the image must be built with the Dockerfile contained in this PR. 
See the

docker/Dockerfile
```diff
@@ -54,59 +54,74 @@ RUN addgroup -S inventreegroup && adduser -S inventree -G inventreegroup
...
+RUN echo "Downloading InvenTree from ${INVENTREE_GIT_REPO}" \
+  &&  git clone --branch ${INVENTREE_GIT_BRANCH} --depth 1 ${INVENTREE_GIT_REPO} ${INVENTREE_HOME}
 
 # Checkout against a particular git tag
 RUN if [ -n "${INVENTREE_GIT_TAG}" ] ; then cd ${INVENTREE_HOME} && git fetch --all --tags && git checkout tags/${INVENTREE_GIT_TAG} -b v${INVENTREE_GIT_TAG}-branch ; fi
+RUN chown -R inventree:inventreegroup ${INVENTREE_HOME}/*
+
+WORKDIR ${INVENTREE_HOME}
+#temporary entrypoint copy test. Can be removed after merge
+RUN if test -f docker/init-server.sh; then \
```
SchrodingersGat, line 90: What's the purpose of this?

docker/docker-compose.yml
```diff
@@ -112,4 +128,5 @@ volumes:
             o: bind
             # This directory specified where InvenTree data are stored "outside" the docker containers
             # Change this path to a local system path where you want InvenTree data stored
-            device: /path/to/data
\ No newline at end of file
+            device: /Users/markdedeuge/data
```
SchrodingersGat, line 131: probably don't want to commit this one...

docker/init-server.sh
```diff
@@ -0,0 +1,85 @@
...
+    python3 -m venv ./dev/env
+
+    # Activate the virtual environment
+    source ./dev/env/bin/activate
+
+    echo "Installing required packages..."
+    pip install --no-cache-dir -U -r ${INVENTREE_HOME}/requirements.txt
+fi
+
+if [[ "${INVENTREE_MIGRATE:-True}" = "True"  || "${INVENTREE_CREATE_SUPERUSER:-True}" = "True" ]]; then
```
SchrodingersGat, line 37: What is the advantage of running particular workflows based on the environment variables? This is different to the current docs guide - https://inventree.readthedocs.io/en/latest/start/docker_prod/ - should we support *both* approaches or will the install guide need to be re-written?

docker/docker-compose.dev.yml
```diff
@@ -16,44 +16,46 @@ services:
...
+        environment:
+            INVENTREE_DEV_VENV: "True"
         ports:
             - 8000:8000
+        user: root
         volumes:
-            # Ensure you specify the location of the 'src' directory at the end of this file
-            - src:/home/inventree
+            # Ensure you specify the location of the 'INVENTREE_SRC_DIR' in the dev-config.env file
+            - ${INVENTREE_SRC_DIR:?inventree_src_dir}:/home/inventree
```
SchrodingersGat, line 29: What is ?inventree_src_dir signifying here?

markdedeuge: configs like: `${SOMEVAR:?some_missing_var_hint}` will interpolate the SOMEVAR environment variable into the docker-compose.yml file. The ? indicates that the variable is required and has no default in the compose file. If that var is missing, then docker-compose will spit out the hint after the ?.

docker/docker-compose.yml
markdedeuge: Good catch.

docker/Dockerfile
markdedeuge: Its to allow someone to test the compose files. The current prod repo that is checked out will not contain these files. (Unless my understanding of the workflow scripts is wrong [probably is]. 

That test can be removed if the build scripts use the correct branch when running the `production` image build.

docker/init-server.sh
markdedeuge: Both work from my local tests.

The migrate and createsuperuser stage does make the startup slightly more streamlined. For scaling out to additional django servers there is a need to disable/not-run that migrate stage or when workers start/restart they would attempt to run it.

docker/docker-compose.yml
markdedeuge: @SchrodingersGat should I add the docker-compose parameterisation into this PR?

It would replace these variables, ie `/path/to/xxx` would be replaced with an `${INVENTREE_SOME_VAR:?inventree_some_var}` and the user would then only look in the default_env file to see what needs updating.

SchrodingersGat: I'd like to get this one merged first, it's quickly blowing out into a huge change!

SchrodingersGat: But moving these things to the .env file is a good idea.

docker/init-server.sh
SchrodingersGat: If these vars are not set then the operation of the dockerfile / docker-compose would not otherwise be affected, right?

markdedeuge: In the single node configuration mentioned in the docs (where only one inventree-server runs), a user shouldnt notice a change. 
When they're not set, the `:-True` provides the default value which should match what was currently happening in the start_* scripts, or mentioned in the docs. 

- The pre-PR configuration runs the migrate stage each time unless the var INVENTREE_MIGRATE is set to something besides True
- The createsuperuser command will try and continue if the user is present, or create one if the DJANGO_SUPERUSER_* set of vars is set.  The dev_config.env now contains the defaults mentioned in the docs for the superuser. 

I tried to retain the default behavior as much as possible

SchrodingersGat:
Tried pulling this down and rebuilding the development docker image:

```
docker-compose -f .\docker\docker-compose.dev.`yml up -d
ERROR: Missing mandatory value for "volumes" option interpolating ['${INVENTREE_SRC_DIR:?inventree_src_dir}:/home/inventree'] in service "inventree-dev-server": inventree_src_dir
```

And the contents of the `dev-config.env` file:

```
INVENTREE_HOME=/home/inventree
INVENTREE_DB_ENGINE=sqlite3
INVENTREE_DB_NAME=/home/inventree/dev/inventree_db.sqlite3
INVENTREE_MEDIA_ROOT=/home/inventree/dev/media
INVENTREE_STATIC_ROOT=/home/inventree/dev/static
INVENTREE_CONFIG_FILE=/home/inventree/dev/config.yaml
INVENTREE_SECRET_KEY_FILE=/home/inventree/dev/secret_key.txt
INVENTREE_DEBUG=true
INVENTREE_WEB_ADDR=0.0.0.0
INVENTREE_WEB_PORT=8000
#Do not use for production!
DJANGO_SUPERUSER_USERNAME=admin
DJANGO_SUPERUSER_EMAIL=admin@abc.com
DJANGO_SUPERUSER_PASSWORD=password
INVENTREE_SRC_DIR=../

#--
#Add these variables to your .env file
#--

# Change this path to a local system path where you want InvenTree postgres data stored
#INVENTREE_DEV_DB=/some/data/path/for/db

```

If I manually alter the docker-compose file "volumes" entries to:

```
            # Ensure you specify the location of the 'INVENTREE_SRC_DIR' in the dev-config.env file
            - ../:/home/inventree
```

then I get this error:

error from sender: open \\?\C:\inventree\env\lib64: Access is denied.
ERROR: Service 'inventree-dev-server' failed to build : Build failed
```

File: docker/init-server.sh (hunk 3 above)
Comment (markdedeuge) [reply docker/init-server.sh:37]: The switches mean adding extra servers to the current set of inventree-server set would look like:

    services:
        inventree-server-x:
            ...
            environment:
                INVENTREE_MIGRATE: "False"
                INVENTREE_CREATE_SUPERUSER: "False"
                ....
            command: [ "wait-for-it.sh", "--timeout=0",  inventree-server:8000", 
                                   "/init-server.sh", 
                                   "gunicorn" ... ]

This config would require setting up a loadbalancer to either: 
- route directly to the inventree-server containers instead of nginx
- route directly to inventree-server containers that integrate/contain nginx directly

**Comment by markdedeuge** [comment]
You cannot use relative links for directories AFAIK. It must be absolute. 
What command are you using to build? `docker-compose -f docker/docker-compose.dev.yml build` ?

**Comment by SchrodingersGat** [comment]
If you use the docker-compose.dev.yml file from master, that uses a relative dir:

https://github.com/inventree/InvenTree/blob/d8eefec0653959d1a416517031684ea74c6908a8/docker/docker-compose.dev.yml#L59

**Comment by markdedeuge** [comment]
Interesting. Im probably wrong on that count then. Does a fully qualified path work or do you get the same error?

File: docker/docker-compose.dev.yml [hunk 5]
```diff
@@ -16,44 +16,46 @@ services:
...
         env_file:
             # Environment variables required for the dev server are configured in dev-config.env
             - dev-config.env
 
         restart: unless-stopped
 
+        command: ["/init-server.sh", "python3", "manage.py", "runserver", "0.0.0.0:8000"]
```
SchrodingersGat, line 36: If we specify an "entrypoint" and "cmd" under the "dev" target in the dockerfile, do we need to specify the command here in the docker-compose file?

It would seem to me that coding the correct entrypoints / commands into the dockerfile targets means that we can write simpler docker-compose files?

SchrodingersGat:
> 
> 
> Interesting. Im probably wrong on that count then. Does a fully qualified path work or do you get the same error?

Same error(s) for a hardcoded path.

If I have a hardcoded path in the .env file then it fails with the first error.

If I put the hard-coded path in the docker-compose file it fails with the second

markdedeuge:
What path are you launching the docker-compose up command from?

SchrodingersGat:
> 
> 
> What path are you launching the docker-compose up command from?

I have tried both from top level and docker dirs.

i.e.

`docker-compose -f docker/docker-compose.dev.yml`

and

```
cd docker
docker-compose -f docker-compose.dev.yml
```

docker/docker-compose.dev.yml
markdedeuge: these are more placeholders for the following PR. Some of these variables like the port could/should? be interpolated by the compose file. 

I should have updated the CMD entries in the Dockerfile to match those currently present in the docker-compose.yml file .

markdedeuge:
What is the output of `docker-compose -f docker/docker-compose.dev.yml config`

SchrodingersGat:
```
PS C:\inventree> docker-compose -f .\docker\docker-compose.dev.yml config
ERROR: Missing mandatory value for "volumes" option interpolating ['${INVENTREE_SRC_DIR:?inventree_src_dir}:/home/inventree'] in service "inventree-dev-server": inventree_src_dir
```

SchrodingersGat:
Can you actually specify a volume in the .env file? Isn't that just for env vars *within* the docker context?

markdedeuge:
If the `docker-compose ... config` doesnt interpolate, add the missing ENVs to dev-config.env. So here it needs 
INVENTREE_SRC_DIR=`C:\inventree\` (or the correct method of adding windows paths). 

Once you get an output, can you also run the command for the working docker-compose.dev.yml from master please?

markdedeuge:
> Can you actually specify a volume in the .env file? Isn't that just for env vars _within_ the docker context?

The variable will be interpolated by compose into that position. So it should be equivalent to manually typing it in place in the file. The `config` command builds out the full compose file, so we can check between what is in this PR and what is in master and see whats happening to those paths.

SchrodingersGat:
@markdedeuge there's a lot of new stuff for me to look at here, if you can hold off on pushing new additions, I'll try to work my way through it. You've already helped me fix a couple of docker bugs :)

markdedeuge:
Sure, I'll submit a few small gunicorn optimisations. If you submit those outputs of docker-compose config that would be helpful :)

matmair:
@markdedeuge as #1970 is merged now can we close this?  Do you want to see some of the remaining stuff in here implemented in master?

SchrodingersGat:
There are still a number of improvements that @markdedeuge has implemented here that I want to see merged in, #1970 was more background work. I'm away this weekend but will discuss further with @markdedeuge next week

eeintech:
@SchrodingersGat and @markdedeuge Have you guys talked about testing and merging this soon?

SchrodingersGat:
I believe @markdedeuge had some further great ideas in addition to what has already been piecemeal integrated in.

I'll close this PR for now, @markdedeuge can open a new with the other features as needed.

PR #1955, opened 2021-08-12
Always return right base currency

matmair:
The dynamic base currency sometimes leads to messages like below and unneeded API calls to exchangerate.host
grafik

This PR fixes that


markdedeuge:
How does the BASE_CURRENCY function control the currency?
I couldnt find it being called anywhere.

SchrodingersGat:
Same question from me @matmair :)

matmair:
In InvenTree it doesn't, we use a function to represent that base currency. But external libraries like djmoney use the settings-value.
Because of that the exchangerate backends for example always thought falsely that the base currency changed (from USD to EUR) on startup and updated the rates.

This patch fixes this rare edge-case - mainly because I currently debug and profile InvenTrees external API-calls and this bugged me.

matmair:
@markdedeuge good question btw

SchrodingersGat:
Does django-money handle the case where `DEFAULT_CURRENCY` is a callable?

matmair:
@SchrodingersGat have not tried that, I only tested it with `BASE_CURRENCY` because that is called by the exchange-rates.

SchrodingersGat:
@matmair oops, I actually meant to type `BASE_CURRENCY`. So, to clarify, you have tested that this new `BASE_CURRENCY()` function is correctly called by django-money?

markdedeuge:
From what I could see in django-money it does not expect functions

SchrodingersGat:
@matmair any update on this?

matmair:
@SchrodingersGat I tested a lot around with getters / overriding the settings class but I think the small log message will have to stay for now 👎

PR #1957, opened 2021-08-13
BOM - show purchase price

### LINKED ISSUE #1889: Use purchase_price for BOM cost calculation

matmair:
Shows purchase prices for a BOM in the pricing modal and prices tab.
Suggested by @sintech

Fixes #1889


sintech:
Thank you @matmair looks exactly what I needed.
Unfortunately it gives me an error on my test BOM Part
image
image

sintech:
Thank you @matmair looks exactly what I needed.
Unfortunately it gives me an error on my test BOM Part:
image

It fails for sub-parts without purchase price, I think some check is needed.

matmair:
@sintech should work now.
The question is if there should be some kind of warning if not all parts are covered by purchase prices. The current method is to then use BOM and buy prices.

sintech:
> The question is if there should be some kind of warning if not all parts are covered by purchase prices. The current method is to then use BOM and buy prices.

I think it should be ok to have parts with no pricing info at all and purchase price in particular. They just have to be excluded from BOM price calculation.

eeintech:
@matmair So in this example:

image

Maybe we should have some kind of warning stating that the "Cost" lines are based on incomplete pricing information?
I guess same for the "Purchase Price" line?

matmair:
@eeintech @sintech so a small tooltip with information what information was used for calculation would be ok? I would include multiple levels of information depending on the settings.

sintech:
@matmair I think a tooltip about what prices was used for pie chart is also needed. I mean the precedence: supplier cost > internal > purchase.
And that `Unit cost` includes supplier cost + internal cost.

matmair:
@sintech moved this tooltip FR into a new issue

matmair:
@sintech does this PR work for you now?

sintech:
@matmair it is strange but it gives me an error
image

Maybe it's because I dont have any stock item for this part anymore :(
image

sintech:
@matmair seems to be working now. I had git issues...
It is pity that all purchase price info has lost with deletion of stock...

matmair:

SchrodingersGat:
Thanks @matmair this looks good on my end.

PR #1970, opened 2021-08-18
Entrypoint

SchrodingersGat:
Simplifying docker entrypoint functionality

Ref: https://github.com/inventree/InvenTree/pull/1949

This PR does the following:

- Replaces all entrypoint scripts with a simple init.sh script
- Makes greater use of the "invoke" commands to ensure other required commands are run

PR for new documentation: https://github.com/inventree/inventree-docs/pull/157


docker/Dockerfile
```diff
@@ -36,18 +36,15 @@ ENV INVENTREE_GUNICORN_WORKERS="4"
...
 
 LABEL org.label-schema.schema-version="1.0" \
       org.label-schema.build-date=${DATE} \
       org.label-schema.vendor="inventree" \
       org.label-schema.name="inventree/inventree" \
       org.label-schema.url="https://hub.docker.com/r/inventree/inventree" \
-      org.label-schema.version=${INVENTREE_VERSION} \
-      org.label-schema.vcs-url=${INVENTREE_REPO} \
-      org.label-schema.vcs-branch=${BRANCH} \
-      org.label-schema.vcs-ref=${COMMIT}
```
markdedeuge, line 50: I think this is good info to embed. I'd keep the version and commit labels

SchrodingersGat: The INVENTREE_VERSION and COMMIT vars are not actually specified anywhere. Branch and repo we could do.

markdedeuge: They should be easy enough to add to the workflow.yml builds though no?

markdedeuge:
Would you like me to remove #1949 as it conflicts with this ?

SchrodingersGat:
@markdedeuge don't remove yet - I'm using it as a reference point! Sorry to be cheeky and re-implement it :)

markdedeuge:
I'd have been happy to make any changes you needed :(

docker/Dockerfile
SchrodingersGat: Added these back in. Not sure how to add the inventree version code though?

SchrodingersGat:
> 
> 
> I'd have been happy to make any changes you needed :(

This is more so I can understand (piece by piece) what's going on :)

SchrodingersGat:
The new ENTRYPOINT / CMD setup is much more streamlined. The same entrypoint command runs in all cases (server and worker):

- Ensures correct directory structure is created
- Copies `config.yaml` file across to the required place
- Activates python venv (if required)

Previous functionality of waiting for the db to be ready (etc) is now handed off to the various `invoke` commands

SchrodingersGat:
The development docker setup has now been well tested on my setup. Much simpler workflow now :)

SchrodingersGat:
Verified that the "production" setup works in accordance with the new simplified build instructions

SchrodingersGat:
@markdedeuge most of this stuff is somewhat separate from your PR, but your PR caused me to look closely at the docker setup and find some existing issues!

markdedeuge:
Its about 90% the same as the Dockerfile + init approach isnt it?

PR #1978, opened 2021-08-19
Fix translations stats - task

matmair:
This PR fixes the errors brought up in https://github.com/inventree/inventree-docs/issues/156.

It adds a task for collecting the translations stats and runs that instead of translate - emitting the problem with all the new translation files in the git-tree.
@eeintech  thanks for pointing the error out.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -173,7 +173,18 @@ def static(c):
     manage(c, "collectstatic --no-input")
 
 
-@task(post=[static])
+@task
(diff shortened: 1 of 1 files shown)


eeintech:
Works great! Thanks @matmair :+1:

matmair:
> 
> 
> Works great! Thanks @matmair 👍

You are welcome - my feature my fix 👼

PR #1982, opened 2021-08-19
Enable Inventree to be deployed on a Kubernetes Cluster

nwns:
This project is moving too quickly for me to keep up with a Kubernetesified version of Inventree out of tree.  I think this merge request contains most of the required changes to enable running the existing docker image directly on a Kubernetes Cluster.

What would be required for the project to merge these changes?

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -12,19 +12,22 @@ database setup in this file.
 """
 
 import logging
-
 import os
 import random
-import string
 import shutil
+import socket
+import string
 import sys
 from datetime import datetime
 
 import moneyed
-
 import yaml
-from django.utils.translation import gettext_lazy as _
 from django.contrib.messages import constants as messages
+from django.utils.translation import gettext_lazy as _
+from psycopg2.extensions import (
+    ISOLATION_LEVEL_READ_COMMITTED,
+    ISOLATION_LEVEL_SERIALIZABLE,
+)
 
 
 def _is_true(x):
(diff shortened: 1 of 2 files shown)


matmair:
@nwns thanks for the PR!
A few form things that jumped out to me: please pin the dependencies if possible - that is important for reliable testing results. Also, I think the `requirements.txt` is missing `psycopg2`.
Regarding the technical core, I am not qualified in containers but maybe @markdedeuge has insights on the technical details.

It would be great of you to document the usage of the additions in our doc repo if/after the PR gets merged.

markdedeuge:
I think it would be best to split this up into two/three phases to ensure all commits work more generally. 
1. PR for sorting the requirements file (should be an easy one to get through)
2. PR adding the externalised caches (django-redis, django-storage), with an appropriate docker-compose test for running those + changes to the settings.py file. This is probably going to need some degree of iteration here and input from @SchrodingersGat 
3. Final kubectl type commit that contains all the required kubernetes config + an example deployment.

matmair:
@markdedeuge good idea. I would suggest adding the requirements that are needed in the same PR. Just merging in requirements feels wrong.

SchrodingersGat:
@nwns there is a lot of different things here going on at once - I'd be more comfortable adding in individual features via separate PRs.

As @markdedeuge says above, at the very least a single PR for supporting configurable cache options would be a good idea. In particular I would like to closely look at what goes into the `settings.py` file and what could be broken out into configuration options (either env vars or external config file).

I'm going to close this PR for now - please feel free to re-submit the individual components as I'd be happy consider each one on its own.

PR #1998, opened 2021-08-22
Fix for production docker environment

bobek:
Values are passed as written resulting in `"WARNING"` being passed to python logger, which will complain and panic. Fix is simply to remove `"` from the value.


SchrodingersGat:
Good catch, thanks!

PR #1999, opened 2021-08-22
Disable functions in navbars

### LINKED ISSUE #1945: Add possibility to disable unused functions
I started using Inventree for my home inventory, there are many alternatives out there but none are simple to use.
The combination of the app for easy scanning codes and the server works really well.
For my use case I don't need the selling part and maybe also the buying part as I don't need to track every order I make, also because you need to add first a manufacturer and a seller for every part, this can be time-consuming.
I suggest then to add in the settings the possibility to enable every section as a module, like is structured Odoo for example.
This will make the interface cleaner and expandable, maybe with plugins.

matmair:
This PR enables disabling some core in the navbars. It just stops showing stuff as a link - really shutting down stuff in the backend would be a major change and refactor if done right.

Closes #1945

### DIFF
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -787,37 +787,37 @@ class InvenTreeSetting(BaseInvenTreeSetting):
         # enable/diable ui elements
         'BUILD_FUNCTION_ENABLE': {
             'name': _('Enable build'),
-            'description': _('Enable build functionally in InvenTree interface'),
+            'description': _('Enable build functionality in InvenTree interface'),
             'default': True,
             'validator': bool,
         },
         'BUY_FUNCTION_ENABLE': {
             'name': _('Enable buy'),
-            'description': _('Enable buy functionally in InvenTree interface'),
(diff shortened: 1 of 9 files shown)


matmair:
@SchrodingersGat Would be ready for merge - the question is if you want a deeper integration or hiding is enough.

InvenTree/common/models.py
```diff
@@ -783,6 +783,44 @@ class InvenTreeSetting(BaseInvenTreeSetting):
             'description': _('Prefix value for purchase order reference'),
             'default': 'PO',
         },
+
+        # enable/diable ui elements
+        'BUILD_FUNCTION_ENABLE': {
+            'name': _('Enable build'),
+            'description': _('Enable build functionally in InvenTree interface'),
```
SchrodingersGat, line 790: "functionally" -> "functionality" (multiple instances)

SchrodingersGat:
I think this simple approach is the best one. Trying to change the behavior under the hood would be extremely difficult!

PR #2005, opened 2021-08-23
Updated color of `new` and `add` UI buttons

eeintech:
While working on the doc for orders (ref https://github.com/inventree/inventree-docs/pull/158) I noticed the color of "New Order" buttons was set to blue (primary CSS class) instead of green (success CSS class) like in the rest of the UI for all other elements (parts, categories, stock, locations, build, companies).
Trying to make the UI color coding more consistent :smiley: 

EDIT: also updated the color of all "add" buttons to green/success class


SchrodingersGat:
Nice find, thanks @eeintech

Issue #2007, opened 2021-08-23
Part view: "Sales Orders" tab always enabled


eeintech:
To reproduce:

1. Create a "Component" only part: disable all other part attributes
2. Check the navbar on the part view: "Sales Orders" tab is shown and "Sales Orders"/"Sales Orders Allocation" tables can be loaded when clicked on

![image](https://user-images.githubusercontent.com/4020546/130487950-9cc0e762-af5a-47dc-8b9d-57d28f9eb596.png)

This view should be restricted to "Salable" parts only.

PR #2010, opened 2021-08-24
Hide "Sales Orders" tab if part is not salable

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/2007


eeintech:
Thanks!

Issue #2006, opened 2021-08-23
Part default attributes are shown in "Edit Part" form


eeintech:
One way to reproduce:

1. Create a part with "Assembly" and "Salable" checked in the part attributes, leave others off
2. Once the part is created, click on "Part actions" -> "Edit Part", scroll down to part attributes

This is what is shown:

![image](https://user-images.githubusercontent.com/4020546/130487018-b37c72df-1320-468d-a0fd-b073f06593d4.png)

Only "Component" is checked and it is not the actual value (default maybe?).

Setting the part attributes differently and submitting the form:

![image](https://user-images.githubusercontent.com/4020546/130487224-b907d66a-3cf2-420b-8be4-a845b76b4d23.png)

Will **not** fix:

![image](https://user-images.githubusercontent.com/4020546/130487331-ffa4c71b-60fc-4aba-96bc-78dcb7ee3566.png)

PR #2011, opened 2021-08-24
Fix for "part" form fields

SchrodingersGat:
- Specify "default" rather than overriding "value"

Fixes https://github.com/inventree/InvenTree/issues/2006


eeintech:
Thanks @SchrodingersGat

Issue #2016, opened 2021-08-24
[BUG] MPN sorting is broken in `Purchase Order Items` table


eeintech:
1. Go to a purchase order details and click on "Order Items" tab.
2. Sort the `Purchase Order Items` table by MPN.

## Before sorting by MPN

![image](https://user-images.githubusercontent.com/4020546/130660558-0363ed63-af15-4259-8b36-6e9cdbf5c15c.png)

## After sorting by MPN

![image](https://user-images.githubusercontent.com/4020546/130660587-3027d82c-6e47-4a15-a034-efea1c83c0af.png)

** Javascript error **

![image](https://user-images.githubusercontent.com/4020546/130660777-2aa7921f-0a3d-46d5-86e8-e21077c59d0d.png)

**Version Information**
This was tested on master.

PR #2020, opened 2021-08-25
Mpn sorting fix

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/2016

There's a fair bit going on under the hood here:

### DIFF
--- a/InvenTree/order/templates/order/purchase_order_detail.html
+++ b/InvenTree/order/templates/order/purchase_order_detail.html
@@ -294,7 +294,7 @@ $("#po-table").inventreeTable({
         {
             field: 'part',
             sortable: true,
-            sortName: 'part__part__name',
+            sortName: 'part_name',
             title: '{% trans "Part" %}',
             switchable: false,
             formatter: function(value, row, index, field) {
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@eeintech as you reported this bug can you please have a review?

SchrodingersGat:
There may be other tables which suffer from this original bug too - we can fix this by replacing calls to `OrderingFilter` with `InvenTreeOrderingFilter`

InvenTree/order/templates/order/purchase_order_detail.html
```diff
@@ -358,6 +358,7 @@ <h4>{% trans "Order Notes" %}</h4>
         {
             sortable: true,
             field: 'purchase_price',
+            sortName: 'price',
```
eeintech, line 361: Unfortunately that one does not work...

eeintech (commented):
Everything looks good, except sorting by `price` (eg. "Unit Price"). "Total Price" sorting creates some weird API query too: `&ordering=X` or `&ordering=-X` with X being an integer.

@matmair What do you think?

Issue #1985, opened 2021-08-20
Regression: tables options do not stick anymore


eeintech:
User constantly has to re-enable columns in table views:

![image](https://user-images.githubusercontent.com/4020546/130279071-cd4107a4-b067-44ad-83a0-980dd8210f49.png)

SchrodingersGat:
Seems to be working on my end:

![image](https://user-images.githubusercontent.com/10080325/130357928-a3c37a9e-4fb1-4ced-a1e5-0cef6d91be8b.png)

Can you open up the developer view in your browser and see if there are any javascript errors being flagged?

eeintech:
@SchrodingersGat Sorry forgot to mention there is no visible JS errors and I have to be honest I am not sure how to reproduce... it happens too often though to me, and I still have not found a pattern. Will report with more info (like tracing back my steps) next time that happens.

eeintech:
Also I did notice that those settings are sticky at server level, not a user level (eg. enabled table columns are the same for all users). Is that expected?

eeintech:
@SchrodingersGat I was able to reproduce this issue (thank god, it was driving me crazy). Try those steps:

1. Go to list of purchase orders and enable all columns:

![image](https://user-images.githubusercontent.com/4020546/130659612-08d8d869-ac71-4747-8f30-c7d16ecc9619.png)

2. Now click on a purchase order. Click on "Order Items" tab. Enable all columns:

![image](https://user-images.githubusercontent.com/4020546/130659711-a054e9fa-aa01-40b5-b654-fd0920f52ad3.png)

3. Go back to the purchase order list (step 1), all table options have been disabled:

![image](https://user-images.githubusercontent.com/4020546/130659789-68da558a-0be5-48bf-ab0d-1fbdbbd0b635.png)

4. Re-enable all of them, then go back to the purchase order, the "Purchase Order Items" table has lost all of its options...

Have yet to 

Here is my install information:

# Version Information:
InvenTree-Version: 0.5.0 pre
Django Version: 3.2.4
Commit Hash: 5349addc0
Commit Date: 2021-08-24
Database: postgresql
Debug-Mode: False
Deployed using Docker: False

PR #2021, opened 2021-08-25
Change name of purchaseorder line item table

SchrodingersGat:
- Was conflicting with purchaseorder table
- Saved column selections were being overridden

Fixes https://github.com/inventree/InvenTree/issues/1985

### DIFF
--- a/InvenTree/order/templates/order/purchase_order_detail.html
+++ b/InvenTree/order/templates/order/purchase_order_detail.html
@@ -28,7 +28,7 @@
             {% endif %}
         </div>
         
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@eeintech the two tables in question had the same "name" - this name is used to store the column selections in your browser local storage.

Can you please test this and let me know if the issue is addressed?

eeintech (commented):
I kinda suspected those two tables were sharing the same local cache / same id or name. I'm glad you found this and yes it is working great now!

PR #2026, opened 2021-08-28
Js linting

SchrodingersGat:
Adds linting for rendered javascript files.


InvenTree/templates/js/dynamic/inventree.js
```diff
@@ -132,43 +132,24 @@ function inventreeDocReady() {
...
-function isOnlineTransfer(transfer) {
-    /* Determine if a drag-and-drop transfer is from another website.
-     * e.g. dragged from another browser window
-     */
-
-    return transfer.items.length > 0;
-}
-
-
-function getImageUrlFromTransfer(transfer) {
```
matmair, line 152: @SchrodingersGat doesn't this remove drag and drop?

matmair (commented):
@SchrodingersGat cool idea, will make testing also easier. How about the removed code - is all of that unused?

PR #2028, opened 2021-08-31
fix html templates

matmair:
Makes parsing for screenreads easier and has better w3c compliance.

### DIFF
--- a/InvenTree/part/templates/part/part_pricing.html
+++ b/InvenTree/part/templates/part/part_pricing.html
@@ -9,11 +9,11 @@
 
 <table class='table table-striped table-condensed table-price-two'>
     <tr>
-        <td><b>{% trans 'Part' %}</b></td>
+        <td><strong>{% trans 'Part' %}</b></td>
         <td>{{ part  }}</td>
     </tr>
     <tr>
-        <td><b>{% trans 'Quantity' %}</b></td>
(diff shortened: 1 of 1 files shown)


InvenTree/part/templates/part/part_pricing.html
```diff
@@ -9,11 +9,11 @@
 
 <table class='table table-striped table-condensed table-price-two'>
     <tr>
-        <td><b>{% trans 'Part' %}</b></td>
+        <td><strong>{% trans 'Part' %}</b></td>
```
SchrodingersGat, line 12: Should the `<strong>` be closed with `</strong>` and not `</b>`?

matmair: @SchrodingersGat you absolutley right, seems like I forgot to push that

SchrodingersGat:
Thanks. Are you planning on making these changes for more files, or just this one?

matmair:
@SchrodingersGat generally I would like to fix it everywhere but would wait till we have the new merging strategy and are stable. So in 0.6.0 or so.

I started using a source-analyzer (if you are interested you can find it [here](https://sonarcloud.io/dashboard?id=matmair_InvenTree)) I already use at work for my commits to InvenTree recently and it complained about those tags.

PR #2030, opened 2021-09-01
Html code smells

matmair:


SchrodingersGat:
Thanks @matmair 

I'd love to have HTML linting / style checking as part of a CI step.. Might be complex as we'd need to first pass the HTML files through the templating engine

matmair:
@SchrodingersGat I don't think that would be worth the effort. Sonarcloud and other providers are already equipped for these checks so if you want something like that I suggest one of these providers.

matmair:
@SchrodingersGat just realized I did not even PR this😇 just made it to see how it affected our code-style / debt rating

PR #2039, opened 2021-09-07
Re-add gunicorn to requirements file

SchrodingersGat:
Fixes https://github.com/inventree/inventree-docs/issues/165


eeintech:
Even though InvenTree can run without it?

SchrodingersGat:
It's a small price to pay for the convenience of simplifying the install experience for everyone I think

eeintech:
I guess it is difficult to customize requirements per installation method, fair enough!

Issue #1421, opened 2021-03-27
Integrating with 3rd party PO/SO generation


miniPCR-Jesse:
First off, really impressive piece of software that fills a large hole in the open source community for SMB needs!  I have a question regarding 3rd party integration.  I would assume a large amount of people using InvenTree are small companies that use QB Online to handle purchase orders and Shopify/WooCommerce to handle sales orders.  I know InvenTree has an API, but I wasn't able to find any discussions on how people have specifically integrated InvenTree with other common applications.

Has anyone come up with a script to push PO's from Inventree to QB online, and/or pull sales orders from Shopify/WooCommerce into Inventree?  Or the more general question would be: how are people handling POs/SOs in multiple systems?

eeintech:
Hello @miniPCR-Jesse 

Good questions and definitely relevant topic. My company does have a online shop but it's not yet connected to our inventory. It would be very interesting to develop tools to connect shopping systems to InvenTree.
I'm sure in the near future this is something we're going to want, and I'm going to volunteer to make it happen :smile: 

In the meantime, could you share your experience in the details of such implementation such as:
* connection between commerce/book-keeping platforms and InvenTree's API (do commerce platform have APIs too?)
* chain of events after a sales order is placed on commerce platform
* chain of events after a purchase order is recorded on InvenTree

SchrodingersGat:
@miniPCR-Jesse glad you are excited about the project :)

Currently I am working on a [branch](https://github.com/inventree/InvenTree/pull/1398) which adds a very crucial feature required for third party integration. It adds support for a background process runner, which can handle periodic tasks like checking third-party services for updates, and performing other actions like sending email, etc.

Once that branch is merged, then considerations of other third party integrations will be much easier!

In the mean time, as @eeintech  says, any information on how you might want to see such features integrated would be helpful.

miniPCR-Jesse:
I've been trying to get detailed info from our operations person, but he's been out for a bit.  Figure in the meantime I'll write out my best guess.

We use WooCommerce to handle all of our orders (orders are manually entered into WooCommerce if they're not made online) and Quickbooks Online for our accounting.  Both seem to have pretty thorough APIs implemented (documentation [here](https://developer.intuit.com/app/developer/qbo/docs/develop) and [here](https://woocommerce.github.io/woocommerce-rest-api-docs/)).  QB has pseudo inventory-management but it's laughably bad (just a "qty on hand" field).

[...]

PR #2049, opened 2021-09-12
Webhook receiver

### LINKED ISSUE #2036: [3rd party integration] Webhook receiver
Add a general webhook receiver for plugins.

- [x] general endpoint
- [x] HMAC verification
- [x] fast response -> use worker
- [x] unittests
- [x] messge-logging

Part of #1421

matmair:
Closes #2036


SchrodingersGat:
@matmair could you provide a little bit more background on how this webhook architecture would be used? Is this intended to be used directly for third-party integration?

I'm afraid I've fallen behind on my reading of the related issues, sorry if I'm asking you to reiterate this.

matmair:
@SchrodingersGat no problem. This is one of the building blocks for #1421 - 3rd party integration.

I will give you an example: I currently develop a IntegrationPlugin that connects InvenTree with Shopify.

On startup the plugin checks that it has deployed a webhook-receiver for getting notifications from Shopify. Then it registers the random URL for this endpoint with Shopify's webhook ressource for a couple of events. So each time a order gets created in Shopify a JSON message with all information will be received on the endpoint. The plugin then gets fed this information and uses is to create a SO in InvenTree.

The webhook receiver supports inheritance so you only need 4 loc to implement one in a plugin. Because of its usage of the database to define those you can have multiple webhooks per plugin and action as some 3rd parties might require that.

matmair:
The sample Plugin should be PR ready in a few days so you can try out all pieces working together.

InvenTree/common/api.py
```diff
@@ -5,5 +5,153 @@
...
+
+class VerificationMethod:
+    NONE = 0
+    TOKEN = 1
+    HMAC = 2
+
+
+class WebhookView(CsrfExemptMixin, APIView):
+    """
+    Endpoint for receiving webhoks.
```
eeintech, line 44: Typo

eeintech:
@matmair Wondering how to test the webhooks are functional? I've created an end-point but using the DRF form I'm not able to post a message into that webhook:

image

matmair:
@eeintech the unit tests in `common/test.py` show best how it works. You need to log into the admin site and create an endpoint.

Then there are three methods of Authentication:
- If no token is set the webhook just accepts everything
- If a token is set the token needs to be supplied as a header named 'Token'
- If a secret is set you need to supply a header named 'Token' with a valid HMAC calculated with the secret and the payload

I personally use postman to test the webhooks.

eeintech:
I'm not sure why the DRF form is not working but it seems straightforward otherwise.

> You need to log into the admin site and create an endpoint.
> 
> Then there are three methods of Authentication:
> 
> * If no token is set the webhook just accepts everything
> * If a token is set the token needs to be supplied as a header named 'Token'
> * If a secret is set you need to supply a header named 'Token' with a valid HMAC calculated with the secret and the payload

All good information that need to go into the doc :wink:

matmair:
@eeintech my idea would be to hide the whole function from admins because this only safely works if you know what you are doing.  With a badly made Plugin one could do arbitrary code execution via a webhook.

eeintech:
@matmair Oh okay so what you are saying is to have hardcoded webhooks instead? Wouldn't it be more flexible to be able to add them dynamically?

matmair:
> 
> 
> @matmair Oh okay so what you are saying is to have hardcoded webhooks instead? Wouldn't it be more flexible to be able to add them dynamically?

My idea would be to remove the adding function on this specific admin page.

matmair:
> I'm not sure why the DRF form is not working but it seems straightforward otherwise.
> 
> > You need to log into the admin site and create an endpoint.
> > 
> > Then there are three methods of Authentication:
> > 
> > * If no token is set the webhook just accepts everything
> > * If a token is set the token needs to be supplied as a header named 'Token'
> > * If a secret is set you need to supply a header named 'Token' with a valid HMAC calculated with the secret and the payload
> 
> All good information that need to go into the doc :wink: 

@eeintech did you get it to run btw?

eeintech:
@matmair Nope but haven't tried harder though

matmair:
@SchrodingersGat this is included and improved in #2074

SchrodingersGat:
@matmair ok great, that makes things simpler!

I should have some time starting next week to sit down and look at this closely :)

PR #2084, opened 2021-09-29
Show part delete button when part still active

### LINKED ISSUE #2072: [FR] Deleting a part is not discoverable
**Problem**
At first glance, there is no way to delete a part. Requires marking a part as inactive first before revealing the "delete part" button. Makes deleting a part very undiscoverable in the ui.

**Suggested solution**
Show the delete part button even when a part is still active. When selected, prompt the user to mark the part as inactive first (or suggest the user leave the part as inactive instead of deleting).

 * A comment in the part attributes "active" toggle switch to indicate making the part inactive will allow you to delete it.
 * Update the docs on how to delete a part.

eeintech:
Fixes #2072

### DIFF
--- a/InvenTree/part/templates/part/part_base.html
+++ b/InvenTree/part/templates/part/part_base.html
@@ -508,7 +508,8 @@
         launchModalForm(
             "{% url 'part-delete' part.id %}",
             {
-                redirect: {% if part.category %}"{% url 'category-detail' part.category.id %}"{% else %}"{% url 'part-index' %}"{% endif %}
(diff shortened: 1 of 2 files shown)


matmair (approved):

SchrodingersGat:
@eeintech looks great, I'd suggest you remove the "Submit" button though

image

SchrodingersGat:
@eeintech I think this is the last one we are waiting on for `0.5.0` :)

eeintech:
My bad, it's done now!

Let's do this!

SchrodingersGat:
Awesome, thanks :)

PR #2097, opened 2021-10-04
Fix plural name for Companies in Admin interface

eeintech:
Was:
image

With fix:
image


SchrodingersGat:
@eeintech looks like you forgot to add the migration file :)

eeintech:
@SchrodingersGat Yup, sorry about that!

Issue #2104, opened 2021-10-06
[FR] Add option to display stock quantity in "part" display (within forms)


SchrodingersGat:
![image](https://user-images.githubusercontent.com/10080325/136137354-81fad1f1-ec45-4b0e-8da9-b36c40ea50cf.png)

Here it would be useful to know what stock is available for each part...

matmair:
Nice idea @schrodingersGat the data gathering function should check if the user has the rights to access this kind of information before displaying it.

PR #2114, opened 2021-10-06
Display stock count in part forms

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/2104

image


matmair (approved):
Nice and easy change - how hard would it be to make a global setting to control default settings?

matmair:
> > Nice and still small change - how hard would it be to make a global setting to control default settings?
> 
> 
> 
> Can you expand on this? What would the global setting be for exactly?

Globally enable / disable viewing the stock so the admin can control the default behavior.

Issue #2096, opened 2021-10-04
[BUG] After updating to 0.5.0 only the Parts tab is shown


G-Pereira:
![image](https://user-images.githubusercontent.com/15521303/135897960-3674e989-1886-49fa-aa5f-c8be276df8b8.png)

eeintech:
@G-Pereira Are other tabs enabled?

![image](https://user-images.githubusercontent.com/4020546/135912278-d263c0ea-66d1-4641-8cb2-59f0c3b57821.png)

SchrodingersGat:
@G-Pereira did you stop the server before performing the update? Try restarting the server process

G-Pereira:
@SchrodingersGat I didn't, but afterward I always reboot the system and this time was no exception.

@eeintech That did the trick weird how the update changes a thing like that. @SchrodingersGat was that intentional?

G-Pereira:
Also I've noticed that enabling Sales orders does not enable Sales so enabling SO without enabling Sales is pointless.

SchrodingersGat:
This is not an intended behavior, no. It needs addressing.

eeintech:
@SchrodingersGat It happened to me too... see #2014 (this is a duplicate)

G-Pereira:
Closing this since it's a duplicate! Thanks @eeintech

eeintech:
@G-Pereira Do you remember from which exact version were you updating from? Or the commit hash?
Which backend do you use?

G-Pereira:
> @G-Pereira Do you remember from which exact version were you updating from? Or the commit hash?
> Which backend do you use?

Yes, it was a tagged release as well 0.4.5
I am using a venv on bare metal at the moment

G-Pereira:
Oh you've meant the database engine, MySQL

eeintech:
Ok I don't know why it would but there could be in relation to the type of backend... it's quite rare error though and not reproducible...

matmair:
@SchrodingersGat I would say revert #1999 and leave this behind us.

PR #2145, opened 2021-10-12
Revert 1fb76b9

SchrodingersGat:
Removes global functions to enable / disable entire sections of functionality

Ref: https://github.com/inventree/InvenTree/issues/2096


eeintech:
@matmair @SchrodingersGat Thanks for agreeing on reverting this :+1: it makes a lot of sense

Issue #1738, opened 2021-07-02
Mark multiple build outputs as complete with single step


SchrodingersGat:
- Add checkbox to "build output" display
- Select multiple (or all) pending outputs
- Press "Complete Selected Outputs"
- Launches confirmation form
- Select destination / status for the outputs

PR #2159, opened 2021-10-14
Build output improvements

SchrodingersGat:
This PR addresses a number of shortcomings with the current build output stock allocation and completion interface.

- Moves to "modern" API driven approach
- Refactor and cleanup of tables

Closes https://github.com/inventree/InvenTree/issues/1738


matmair:
[Open in Gitpod](https://gitpod.io/from-referrer/)

PR #2161, opened 2021-10-14
Onboarding group

matmair:
This PR:
- adds the option to add a new user to an initial group (otherwise the user is not seeing anything)
- changes the grouping in the SSO settings so they make more sense

### DIFF
--- a/InvenTree/InvenTree/forms.py
+++ b/InvenTree/InvenTree/forms.py
@@ -23,6 +23,7 @@ from common.models import InvenTreeSetting
 
 logger = logging.getLogger('inventree')
 
+
 class HelperForm(forms.ModelForm):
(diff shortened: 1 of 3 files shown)


matmair:
@SchrodingersGat I enabled the already written code for settings with choices to have a callable (functions) as 'choice' argument. Any reason it was still #TODO known?

SchrodingersGat:
@matmair no I cannot remember why that was a `# TODO` ... If it is all working for you as expected, I'm happy with that!

matmair:
> 
> 
> @matmair So this is a setting to select a default group for new user when then sign-up? Might be good to document that, unfortunately don't have much time to check it out right now...

already working on all the doc updates but the getting started section takes a lot of time.

PR #2163, opened 2021-10-14
Sort requirements.txt

nwns:
Lets sort requirements.txt since the order doesn't really matter as discussed in #1982 .


matmair:
@nwns nice seeing you here! Are you still thinking about fixing the stuff needed for Kubernetes? Would be a great addition for the project!

SchrodingersGat:
Thanks @matmair

PR #2164, opened 2021-10-14
Set some sensible postgres timeouts

nwns:
Set some sensible timeouts for connecting to a database server.

Only affects PostgreSQL at this time.  I am not familiar with how MySQL/MariaDB handles these same issues though I assume it has exactly the same problem.

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -457,32 +457,101 @@ Ref: https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-OPTIONS
 """
 
 # 'OPTIONS' or 'options' can be specified in config.yaml
-db_options = db_config.get('OPTIONS', db_config.get('options', {}))
+# Set useful sensible timeouts for a transactional webserver to communicate
+# with its database server, that is, if the webserver is having issues
+# connecting to the database server (such as a replica failover) don't sit and
+# wait for possibly an hour or more, just tell the client something went wrong
+# and let the client retry when they want to.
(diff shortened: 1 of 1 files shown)


InvenTree/InvenTree/settings.py
```diff
@@ -25,6 +25,10 @@
 import yaml
 from django.utils.translation import gettext_lazy as _
 from django.contrib.messages import constants as messages
+from psycopg2.extensions import (
+    ISOLATION_LEVEL_READ_COMMITTED,
+    ISOLATION_LEVEL_SERIALIZABLE,
+)
```
matmair, line 31: I suggest putting the imports into line 499ff to make sure it does not require the pgsql bindings if not using postgres

nwns: Yeah, that would work as well. I went a slightly different way though.

SchrodingersGat:
@nwns I like the idea here, however I'd like to see some changes with the way it's implemented within settings.py

Have a look at the implementation in https://github.com/inventree/InvenTree/pull/2168. It is not a full implementation of what you've done here, but it is a framework for how I'd like to see it done, to support options for other database backends too.

I tried putting these thoughts into a review of your PR but it was getting too complicated to convey my ideas

Once https://github.com/inventree/InvenTree/pull/2168 is merged, please merge master back into this branch, and add in the extra options!

Also, please prefix any relevant environmental variables for database configuration with `INVENTREE_DB_`

Thanks for adding these parameters in! I'll check out your redis PR soon too :)

SchrodingersGat:
@nwns thanks for updating this!

PR #2196, opened 2021-10-24
Price refactoring

### LINKED ISSUE #2042: [BUG] BOM Item API is unreasonably slow
**Describe the bug**
When viewing a BOM table, the API request for the various BOM items takes a significant amount of time.
For example in my test instance, a BOM with only 25 items takes 1050ms to load.

Need to investigate why the BOM serializer / API is so slow.

### Query Speed Tests

- There are 2100 BomItem objects in the database

| Condition | Query Time |
| --- | --- |
| List all items | 54 seconds |
| Limit to 100 results | 15 seconds |
| Limit to 10 results | 15 seconds |
| Limit to 1 result | 15 seconds |

(For comparison, there are ~2000 parts in the database, and querying a single *Part* result (`?limit=1&offset=0`) returns in ~67ms)

Clearly something is not right here...

### Queryset Annotations

The queryset annotations which calculate pricing info are the culprit here:

https://github.com/inventree/InvenTree/blob/54d770927acb8d0e024b9d1288d6d4e0ae00bdb2/InvenTree/part/api.py#L1144

Removing these queryset annotations reduces the query time down to **32ms**

**(This is 99% reduction in query time)**

@eeintech  @matmair I think we need to consider a better way of managing pricing information. Calculation of pricing "on the fly" is very expensive and really impacts query speed, especially for larger tables such as the BOM lists.

Perhaps we should be pre-calculating and storing pricing information, and re-calculating the prices any time a down-stream price changes. The background worker could be used for this if it is computationally expensive.

Thoughts on this?

matmair:
This PR will fix a few performance problems with prices.
- [x] Price calculations on the run -> refactor and prefetching prices for different scenarios; these prefetched prices will be updated each time a price-break is updated.
- [ ] Purchase price history statistics -> each time a stock item with prices is added the related price history statistics are updated.
- [ ] Migrations to add the new prefetched values into the databases

Closes #2042

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2196"></a>


matmair:
This is a 7 month old test. Nothing that should be used here.

PR #2208, opened 2021-10-30
Email notification for low stock

### LINKED ISSUE #2093: [FR] Email notifications for low stock
When a part stock level falls below the "low stock" threshold, email any users who have "starred" that part.

rocheparadox:
This Pull Request

-  facilitates email notifications when a part's stock quantity falls below the minimum quantity. 

closes #2093

### DIFF
--- a/InvenTree/InvenTree/tasks.py
+++ b/InvenTree/InvenTree/tasks.py
@@ -12,6 +12,7 @@ from django.utils import timezone
 from django.core.exceptions import AppRegistryNotReady
 from django.db.utils import OperationalError, ProgrammingError
 from django.template.loader import render_to_string
+from django.utils.translation import gettext_lazy as _
 
 
 logger = logging.getLogger("inventree")
@@ -321,21 +322,18 @@ def notify_low_stock(stock_item):
 
     if len(starred_users) > 0:
         logger.info(f"Notify users regarding low stock of {stock_item.part.name}")
-        body = f'Hi, {stock_item.part.name} is low on stock. Kindly do the needful.'
         context = {
             'part_name': stock_item.part.name,
             # Part url can be used to open the page of part in application from the email.
             # It can be facilitated when the application base url is accessible programmatically.
             # 'part_url': f'{application_base_url}/part/{stock_item.part.id}',
 
-            'message': body,
-
             # quantity is in decimal field datatype. Since the same datatype is used in models,
             # it is not converted to number/integer,
             'part_quantity': stock_item.quantity,
             'minimum_quantity': stock_item.part.minimum_stock
         }
-        subject = f'Attention! {stock_item.part.name} is low on stock'
+        subject = _(f'Attention! {stock_item.part.name} is low on stock')
         html_message = render_to_string('stock/low_stock_notification.html', context)
         recipients = starred_users.values_list('email', flat=True)
-        send_email(subject, body, recipients, html_message=html_message)
+        send_email(subject, '', recipients, html_message=html_message)
--- a/InvenTree/InvenTree/tasks.py
+++ b/InvenTree/InvenTree/tasks.py
@@ -11,6 +11,7 @@ from django.utils import timezone
 
 from django.core.exceptions import AppRegistryNotReady
 from django.db.utils import OperationalError, ProgrammingError
+from django.template.loader import render_to_string
 
 
 logger = logging.getLogger("inventree")
@@ -52,7 +53,7 @@ def schedule_task(taskname, **kwargs):
         pass
 
 
-def offload_task(taskname, force_sync=False, *args, **kwargs):
+def offload_task(taskname, *args, force_sync=False, **kwargs):
     """
         Create an AsyncTask if workers are running.
         This is different to a 'scheduled' task,
@@ -290,7 +291,7 @@ def update_exchange_rates():
     Rate.objects.filter(backend="InvenTreeExchange").exclude(currency__in=currency_codes()).delete()
 
 
-def send_email(subject, body, recipients, from_email=None):
+def send_email(subject, body, recipients, from_email=None, html_message=None):
     """
     Send an email with the specified subject and body,
     to the specified recipients list.
@@ -306,4 +307,35 @@ def send_email(subject, body, recipients, from_email=None):
         from_email,
         recipients,
         fail_silently=False,
+        html_message=html_message
     )
+
+
+def notify_low_stock(stock_item):
+    """
+    Notify users who have starred a part when its stock quantity falls below the minimum threshold
+    """
+
+    from allauth.account.models import EmailAddress
+    starred_users = EmailAddress.objects.filter(user__starred_parts__part=stock_item.part)
+
+    if len(starred_users) > 0:
+        logger.info(f"Notify users regarding low stock of {stock_item.part.name}")
+        body = f'Hi, {stock_item.part.name} is low on stock. Kindly do the needful.'
+        context = {
+            'part_name': stock_item.part.name,
+            # Part url can be used to open the page of part in application from the email.
+            # It can be facilitated when the application base url is accessible programmatically.
+            # 'part_url': f'{application_base_url}/part/{stock_item.part.id}',
+
+            'message': body,
+
+            # quantity is in decimal field datatype. Since the same datatype is used in models,
+            # it is not converted to number/integer,
+            'part_quantity': stock_item.quantity,
+            'minimum_quantity': stock_item.part.minimum_stock
+        }
+        subject = f'Attention! {stock_item.part.name} is low on stock'
+        html_message = render_to_string('stock/low_stock_notification.html', context)
+        recipients = starred_users.values_list('email', flat=True)
+        send_email(subject, body, recipients, html_message=html_message)
--- a/InvenTree/InvenTree/tasks.py
+++ b/InvenTree/InvenTree/tasks.py
@@ -310,30 +310,3 @@ def send_email(subject, body, recipients, from_email=None, html_message=None):
         fail_silently=False,
         html_message=html_message
     )
-
-
-def notify_low_stock(stock_item):
-    """
-    Notify users who have starred a part when its stock quantity falls below the minimum threshold
-    """
-
-    from allauth.account.models import EmailAddress
-    starred_users = EmailAddress.objects.filter(user__starred_parts__part=stock_item.part)
-
-    if len(starred_users) > 0:
-        logger.info(f"Notify users regarding low stock of {stock_item.part.name}")
-        context = {
-            'part_name': stock_item.part.name,
-            # Part url can be used to open the page of part in application from the email.
-            # It can be facilitated when the application base url is accessible programmatically.
-            # 'part_url': f'{application_base_url}/part/{stock_item.part.id}',
-
-            # quantity is in decimal field datatype. Since the same datatype is used in models,
-            # it is not converted to number/integer,
-            'part_quantity': stock_item.quantity,
-            'minimum_quantity': stock_item.part.minimum_stock
-        }
-        subject = _(f'Attention! {stock_item.part.name} is low on stock')
-        html_message = render_to_string('stock/low_stock_notification.html', context)
-        recipients = starred_users.values_list('email', flat=True)
-        send_email(subject, '', recipients, html_message=html_message)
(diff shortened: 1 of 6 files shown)


InvenTree/InvenTree/tasks.py
```diff
@@ -306,4 +307,35 @@ def send_email(subject, body, recipients, from_email=None):
...
+            # 'part_url': f'{application_base_url}/part/{stock_item.part.id}',
+
+            'message': body,
+
+            # quantity is in decimal field datatype. Since the same datatype is used in models,
+            # it is not converted to number/integer,
+            'part_quantity': stock_item.quantity,
+            'minimum_quantity': stock_item.part.minimum_stock
+        }
+        subject = f'Attention! {stock_item.part.name} is low on stock'
```
matmair, line 338: This needs to be translated

InvenTree/InvenTree/tasks.py
```diff
@@ -306,4 +307,35 @@ def send_email(subject, body, recipients, from_email=None):
...
+    """
+    Notify users who have starred a part when its stock quantity falls below the minimum threshold
+    """
+
+    from allauth.account.models import EmailAddress
+    starred_users = EmailAddress.objects.filter(user__starred_parts__part=stock_item.part)
+
+    if len(starred_users) > 0:
+        logger.info(f"Notify users regarding low stock of {stock_item.part.name}")
+        body = f'Hi, {stock_item.part.name} is low on stock. Kindly do the needful.'
```
matmair, line 324: This needs to be translated

InvenTree/stock/models.py
```diff
@@ -1651,6 +1652,17 @@ def before_delete_stock_item(sender, instance, using, **kwargs):
...
 
+@receiver(post_save, sender=StockItem, dispatch_uid='stock_item_post_save_log')
+def after_save_stock_item(sender, instance: StockItem, **kwargs):
+    """
+    Check if the stock quantity has fallen below the minimum threshold of part. If yes, notify the users who have
+    starred the part
+    """
+
+    if instance.quantity <= instance.part.minimum_stock:
+        inventree_tasks.notify_low_stock(instance)
```
matmair, line 1663: Maybe this should be called as Async_task? Thinking of allocating etc . WDYT @SchrodingersGat ?

matmair (changes requested):
Looks good overall

InvenTree/stock/models.py
rocheparadox: @matmair I thought of making it an AsyncTask but made it a sequential one since the `offload_task()` function which eventually gets called by `notify_low_stock()` runs the task asynchronously .

InvenTree/InvenTree/tasks.py
rocheparadox: should something be done from my side?

matmair: If you know how django handles it (its the _ function) yes, ping me if I should do it

InvenTree/InvenTree/tasks.py
matmair: @rocheparadox I would prefer if we used a template file for this as well - makes it easier for users to override the messages without messing with functional code.

InvenTree/InvenTree/tasks.py
rocheparadox: I shall do a little research and see if I can do it.

InvenTree/stock/models.py
matmair: It is still a db-access but we can come back to this if we see perf problems with stock changes,

matmair (approved):

matmair:
@rocheparadox @SchrodingersGat I think this is ready

InvenTree/InvenTree/tasks.py
```diff
@@ -306,4 +307,35 @@ def send_email(subject, body, recipients, from_email=None):
         from_email,
         recipients,
         fail_silently=False,
+        html_message=html_message
     )
+
+
+def notify_low_stock(stock_item):
```
SchrodingersGat, line 315: This function should be moved to `InvenTree/part/tasks.py` as it relates to the part model.

SchrodingersGat, line 315: You should pass a `Part` object to the function, not a `StockItem`.

The reason is that the *Part* is what we are concerned about running low on stock.

Say you have a capacitor with a low stock threshold of 2,000 units. You have the following stock items:

- A) One reel of 4,000
- B) One reel of 4,000
- C) One reel of 2,500

You remove 600 units from C) which brings it down to 1,900 units. This individual stock item is below the threshold, but there is still sufficient stock of the Part.

The "low stock threshold" should only be activated when the **total stock** for a given *Part* dips below the low stock threshold.

InvenTree/InvenTree/tasks.py
```diff
@@ -306,4 +307,35 @@ def send_email(subject, body, recipients, from_email=None):
...
+            'message': body,
+
+            # quantity is in decimal field datatype. Since the same datatype is used in models,
+            # it is not converted to number/integer,
+            'part_quantity': stock_item.quantity,
+            'minimum_quantity': stock_item.part.minimum_stock
+        }
+        subject = f'Attention! {stock_item.part.name} is low on stock'
+        html_message = render_to_string('stock/low_stock_notification.html', context)
+        recipients = starred_users.values_list('email', flat=True)
```
SchrodingersGat, line 338: Need to add a check to ensure that each of the starred users has a valid email

InvenTree/InvenTree/tasks.py
SchrodingersGat: Refer to the django translation docs - https://docs.djangoproject.com/en/3.2/topics/i18n/translation/

- In python code we use the ugettext_lazy function
- In template files we use `{% trans %}`

InvenTree/InvenTree/tasks.py
```diff
@@ -290,7 +291,7 @@ def update_exchange_rates():
     Rate.objects.filter(backend="InvenTreeExchange").exclude(currency__in=currency_codes()).delete()
 
 
-def send_email(subject, body, recipients, from_email=None):
+def send_email(subject, body, recipients, from_email=None, html_message=None):
```
SchrodingersGat, line 295: What's the difference between "body" and "html_message" here?

InvenTree/stock/models.py
```diff
@@ -1651,6 +1652,17 @@ def before_delete_stock_item(sender, instance, using, **kwargs):
...
 
 
+@receiver(post_save, sender=StockItem, dispatch_uid='stock_item_post_save_log')
+def after_save_stock_item(sender, instance: StockItem, **kwargs):
+    """
+    Check if the stock quantity has fallen below the minimum threshold of part. If yes, notify the users who have
+    starred the part
+    """
+
+    if instance.quantity <= instance.part.minimum_stock:
```
SchrodingersGat, line 1662: Refer to my note above regarding "StockItem quantity" vs "Part quantity".

Here you should be checking something along the lines of:

```python
if instance.part.total_stock <= instance.part.minimum_stock:
```

Although I would suggest that this is written as a method of the `Part` model and called here.

InvenTree/stock/models.py
SchrodingersGat: I think offload_task is OK to use here

InvenTree/stock/templates/stock/low_stock_notification.html
```diff
@@ -0,0 +1,27 @@
+<p>{{ message }}</p>
```
SchrodingersGat: Rather than passing `{{ message }}` through to the template, we should construct the message in the template file itself.

SchrodingersGat: Please move this template to the "part" app (i.e. `InvenTree/part/templates/part/low_stock_notification.html`)

InvenTree/stock/templates/stock/low_stock_notification.html
```diff
@@ -0,0 +1,27 @@
...
+    </tr>
+
+    <tr style="height: 3rem">
+        <td style="text-align: center">{{ part_name }}</td>
+        <td style="text-align: center">{{ part_quantity }}</td>
+        <td style="text-align: center">{{ minimum_quantity }}</td>
+    </tr>
+
+    <tr style="background-color: aliceblue;height: 4rem;">
+        <td colspan="3" style="padding-top:1rem; text-align: center">You are receiving this mail because you have starred the part {{ part_name }} in
```
SchrodingersGat: These strings will need to be placed within a `{% trans "string" %}` block

@matmair in the case of sending an email, how does the server know which locale to encode in? With a web request this is included in the HTML headers, or we use the user specified locale...

InvenTree/stock/models.py
SchrodingersGat: I would suggest that you add a method to the `Part` model called `check_low_stock()` which you can then call from the post_save hook

SchrodingersGat:
In addition to a `post_save` hook we'll also need a `delete` hook (on the `StockItem` model). If a stock item is deleted, the available stock for the parent `Part` is also reduced.

This is another reason to add the `check_low_stock` method to the `Part` model - it can be then reused in the multiple hooks.

InvenTree/InvenTree/tasks.py
rocheparadox: **starred_users** is an `allauth.account.models.EmailAddress` QuerySet, so the variable contains the list of available emails of the users who have starred the part in question. The variable name is the cause of confusion I guess. It should have been **starred_users_email**. I shall change it to the latter.

InvenTree/InvenTree/tasks.py
rocheparadox: HTML code will be displayed as a plain text if the template is sent as the second argument to `send_mail()` function which is the `message` argument (we call it `body` here). `html_message` keyword argument is used to render the html template inside the email body. [reference](https://docs.djangoproject.com/en/3.2/topics/email/)

InvenTree/stock/templates/stock/low_stock_notification.html
rocheparadox: I changed the same in my latest commit based on the suggestions my @matmair .
Kindly refer the [commit](https://github.com/inventree/InvenTree/pull/2208/commits/6ec2801fcea15d0784e0ad57f6000e8b568aa05b)

rocheparadox: I changed the same in my latest commit based on the suggestions my @matmair .
Kindly refer the [commit](https://github.com/inventree/InvenTree/pull/2208/commits/6ec2801fcea15d0784e0ad57f6000e8b568aa05b)

InvenTree/stock/templates/stock/low_stock_notification.html
rocheparadox: I changed the same in my latest commit based on the suggestions my @matmair .
Kindly refer the [commit](https://github.com/inventree/InvenTree/pull/2208/commits/6ec2801fcea15d0784e0ad57f6000e8b568aa05b)

matmair: @SchrodingersGat it uses default locale(English), we currently have not implemented reports or messages with a correct language parsing; is on my todo after I get the base for plugins finished

InvenTree/InvenTree/tasks.py
SchrodingersGat: Thanks!

SchrodingersGat:
@rocheparadox this is looking pretty good now! Can you also please add a post_delete hook for the StockItem model?

rocheparadox:
@SchrodingersGat I thought of doing adding `post_delete` hook but was quite unsure of what to do inside the hook function. Can you explain a lil bit?

SchrodingersGat:
I would just duplicate the code that you have here, but into the "delete" hook:

image

rocheparadox:
Now I get it. I was reading the parent `Part` as parent `StockItem` the whole time :face_with_head_bandage: and had no idea what that meant. My bad.

rocheparadox:
@SchrodingersGat , I have added the `post_delete` hook. Kindly merge the changes.

SchrodingersGat:
Looks good @rocheparadox :)

PR #2209, opened 2021-10-30
Update README.md

matmair:
Added a button in READMEto open up a env in gitpod directly; Also testing if PR button works

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2209"></a>


matmair:
@SchrodingersGat so I can confirm it works. Could you check if the rebuilds run under https://gitpod.io/projects? Should look something like this:
grafik

SchrodingersGat:
Seems to be working here too - https://github.com/inventree/InvenTree/pull/2214

PR #2271, opened 2021-11-08
Fix missing actions for stock item

SchrodingersGat:


SchrodingersGat:
Fixes bug in recent UI overhaul which removed stock item actions for users who did not have "superuser" status

PR #2288, opened 2021-11-10
Add setting for enabeling reports

### LINKED ISSUE #2278: [FR] Turn off reports by default
**Problem**
A lot of report buttons do nothing on a fresh install

**Suggested solution**
Make reports a global setting and disable them by default.

We could also do nothing

Yes

matmair:
Fixes #2278

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2288"></a>

### DIFF
--- a/InvenTree/build/templates/build/build_base.html
+++ b/InvenTree/build/templates/build/build_base.html
@@ -34,6 +34,7 @@ src="{% static 'img/blank_image.png' %}"
 {% include "admin_button.html" with url=url %}
 {% endif %}
 <!-- Printing options -->
+{% if report_enabled %}
 <div class='btn-group'>
     <button id='print-options' title='{% trans "Print actions" %}' class='btn btn-outline-secondary dropdown-toggle' type='button' data-bs-toggle='dropdown'>
         <span class='fas fa-print'></span> <span class='caret'></span>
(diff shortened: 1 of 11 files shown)


matmair:
@SchrodingersGat you wanted to add something about reports? The PR for disabling the buttons would be ready else.

matmair:
Any idea how to detect if the instance is being set up new or just updated? Would need that for a migration to keep printing enabled if the instance is just updated.

SchrodingersGat:
> Any idea how to detect if the instance is being set up new or just updated? 

Good question. We would also need this for the "getting started" guide, I would imagine.

This PR looks OK apart from this, your solution is a lot more elegant than the one I was thinking of :)

matmair:
> 
> 
> > Any idea how to detect if the instance is being set up new or just updated?
> 
> Good question. We would also need this for the "getting started" guide, I would imagine.
> 
For getting started I thought of adding that in in parallel with a unique identifier file that I will need for some advanced security features and just probe if that is present. I can add that now. The sec feature will take a while as it will depend on the plugin system. Keyword: Federation 😇.
>
> This PR looks OK apart from this, your solution is a lot more elegant than the one I was thinking of :)
It does not disable the functionality as that will mess with labels but that would be only a few lines if you want it to be there.

eeintech:
> > Any idea how to detect if the instance is being set up new or just updated?
> 
> Good question. We would also need this for the "getting started" guide, I would imagine.

Thought there could be an easy tip from Django doc on how to do this but don't see nothing... A custom flag would have to do I guess?

SchrodingersGat:
It might be more effort than it is worth. 

Perhaps we just add a note in the docs that this setting must be turned on to enable reporting? We might have to answer a few bug reports about this, but I feel like that might be a simpler solution than potentially introducing a new bug vector.

matmair:
Ready for merge then!

SchrodingersGat:
@matmair could you please add a quick PR to the docs about this new feature?

https://inventree.readthedocs.io/en/latest/report/report/#report-options

PR #2302, opened 2021-11-12
Add support for es-419 (spanish, latin america)

SchrodingersGat:


SchrodingersGat:
@matmair I believe you have had some experience with the translation selection. I'm having some issue here adding a specific variant of spanish (name "latin american", es-419)

I have added the new code to settings.py:

image

However in the language selection, it is simply duplicating the `es` language code...

image

Any ideas here?

SchrodingersGat:
It appears the the "country code" part (i.e. `ll-cc`) of the language code is being stripped out...

I added some debug info to the HTML template and got this:

image

SchrodingersGat:
image

Headers indicate "regular" spanish is being provided

SchrodingersGat:
If I specify `es-mx` (Mexican Spanish) this is provided as a separate item. Perhaps it is an issue with the numeric code `-419`...

image

matmair:
> 
> 
> If I specify `es-mx` (Mexican Spanish) this is provided as a separate item. Perhaps it is an issue with the numeric code `-419`...
> 
> image

As far as I am aware gettext prefers ll_cc formatting. I am not sure if the general latin america regional code is covered. Maybe try 'es_419'?

SchrodingersGat:
@matmair I think that is the case, we may not be able to support es-419

(Note this came from a user request on the crowdin site)

matmair:
> 
> 
> @matmair I think that is the case, we may not be able to support es-419
> 
> (Note this came from a user request on the crowdin site)

@SchrodingersGat so the language code with an underscore is not working?

SchrodingersGat:
> so the language code with an underscore is not working? 

You cannot use an underscore in LANGUAGES configuration in `settings.py` - it throws an error.

matmair:
> 
> 
> > so the language code with an underscore is not working?
> 
> You cannot use an underscore in LANGUAGES configuration in `settings.py` - it throws an error.

ok then it is not supported. Regional codes must be declared with an underscore - countries also can use '-'

matmair:
@SchrodingersGat well we could use just some other country code in the area to support users in that region. From other projects I do not think there is difference greater then a few percent. European Spanish takes over if nothing is present in the right locale.

rjusher:
@SchrodingersGat If my experience orients me in the right direction, is because the translation machinery of django, can't handle a numeric country code. In latin america is kind of weird that the region was defined with a number, I can definitely work with ES-MX, but it would be better if you could create the ES-GT code, so I can translate with the business lingo that we use locally, another translator can take that translation and use 80% to port it to their specific country.

SchrodingersGat:
@rjusher after trying to get `es-gt` to work, it appears that django supports the following spanish language variants:

https://github.com/django/django/blob/8d9827c06ce1592cca111e7eafb9ebe0153104ef/django/conf/global_settings.py#L78

Looks like `es-gt` is not supported by default. If you're happy with `es-mx`, and happy to provide translations, I'll add in support for that :)

rjusher:
Let´s try to do es-mx then.

Thanks!!

PR #2315, opened 2021-11-16
Small tweaks

SchrodingersGat:


SchrodingersGat:
Hmmm. even thought the function "attach_file" is no longer used for anything, it is referenced in an early migration file.

matmair:
> Hmmm. even thought the function "attach_file" is no longer used for anything, it is referenced in an early migration file. 

Maybe add the code to the migration?

PR #2074, opened 2021-09-23
New Plugin definition

matmair:
This is a heavy PR for #2037.
It adds a new plugin system with database controlled plugins, wide integration into InvenTree, a pluggable mixin system so that integrations are easily made, live-reloading of apps so no server admin is needed for plugin changes, a maintenance mode while plugins reload (errors out to 503), ...
Plugins can be loaded from a directory or from the 'inventree_plugins'  entry_point. Metadata about the latest git commit or metadata in the plugin class are used to inform admins about the different plugins.
Initial integrations are:
- navigation
- urls
- global settings
- full Django apps

Includes #2049 as that is needed for a few showcase mixins.
Closes #2037
Closes #2218
Closes #2295
Closes #2294
Closes #2293
Closes #2292
Closes #2305
Closes #2318
Closes #2308
<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2074"></a>

PR #2336, opened 2021-11-19
Webhooks

### LINKED ISSUE #2036: [3rd party integration] Webhook receiver

matmair:
Fixes #2036
This is in preparation for discussion and merge after #2074 is done.

See #2049
This PR adds a Webhook receiver under the `/api/webhook/<endpoint-id>/` URL. The view is designed to be inherited so each step is split into actionable functions. While there is the possibility for the (admin) user to create webhook-endpoints this is in preparation for integrating 3rd party software via plugins.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2336"></a>


SchrodingersGat:
@matmair I guess now that the plugins architecture is merged, now we should look at this?

Looking through the code here, it appears that this is somewhat "separate" from the plugin architecture? 

Unless I'm reading this wrong (and please correct me if so):

I would like to see the web-hook system be a sub-component of the plugin architecture - each "plugin" can accept custom web hooks to perform actions.

Can you run me through your thinking here? Does it line up with above?

matmair:
@SchrodingersGat it is not depended on the plugin system but receiving webhooks without the plugin system dors not really make sense as the hook should trigger some custom code which should be in a plugin or something similar.
Plugins can be full apps so they can contain a model that inherits from the webhook base model. This mechanism ensures that the webhook will not be dead when the plugin is deactivated so that more strict senders do not kill your webhook-ability. Once the plugin is back up it can process through the missed messages at startup or through a task.
That is the whole reason to have it in common and not plugin.

InvenTree/common/models.py
```diff
@@ -1395,6 +1403,184 @@ def is_valid_choice(cls, user_color_theme):
         return False
 
 
+class VerificationMethod:
+    NONE = 0
+    TOKEN = 1
+    HMAC = 2
+
+
+class WebhookEndpoint(models.Model):
```
SchrodingersGat, line 1412: Should the "WebhookEndpoint" have a foreignkey relation to "PluginConfig"? This way, when the plugin is uninstalled, so are the corresponding endpoints. Also, it would enable plugins to keep track of their own endpoints?

InvenTree/common/models.py
```diff
@@ -1395,6 +1403,184 @@ def is_valid_choice(cls, user_color_theme):
...
+        help_text=_('Name for this webhook')
+    )
+
+    active = models.BooleanField(
+        default=True,
+        verbose_name=_('Active'),
+        help_text=_('Is this webhook active')
+    )
+
+    user = models.ForeignKey(
```
SchrodingersGat, line 1452: Is the "user" the latest user to make use of the endpoint, or the only user who is allowed to use it

SchrodingersGat:
@matmair so if I understand correctly:

- Webhook objects register a new webhook URL
- Messages are received at that URL, and stored in the database
- The background worker process extracts a message from the database, and passes it to the plugin to run?

SchrodingersGat:
@matmair I'm going to merge this now so we can have some time to play around with it before the 0.6.0 release

PR #2339, opened 2021-11-19
Protected settings

matmair:
This PR adds protected settings. These settings retract their value on all endpoints. This way you can place API-Keys etc. in settings.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2339"></a>

### DIFF
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -108,7 +108,9 @@ class BaseInvenTreeSetting(models.Model):
         for key, value in settings.items():
             validator = cls.get_setting_validator(key)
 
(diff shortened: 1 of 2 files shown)


matmair:
This is pulled from #2074

SchrodingersGat:
Nice addition. Please note that there is also support for "hidden" settings (which are not exposed to the API, for example)

matmair:
> Nice addition. Please note that there is also support for "hidden" settings (which are not exposed to the API, for example)

@SchrodingersGat the difference is that you can check for the existence of the setting. I think i will document all settings options for the plugins.

PR #2349, opened 2021-11-22
BOM import / export improvements

SchrodingersGat:


SchrodingersGat:
Closing for now, this approach is not working...

matmair:
@SchrodingersGat what was the problem / painpoint?

SchrodingersGat:
The current BOM import is incredibly slow, especially when you have a db with lots (> 1000) of parts.

The existing import process iterates through every single part, for  each imported line item. It takes enough time to timeout our production server at work. 

There are a few different approaches I've been considering, I want to reduce code duplication as much as possible also. I've been trying to work out how we can use `django-import-export` to do the bulk of the heavy lifting here.

matmair:
@SchrodingersGat I meant with the PR 😆

SchrodingersGat:
Ah, I wanted to investigate some other options...

matmair:
I tried using import-export for something similar a while ago and gave up on it. Anything per-line is problematic. I found it is faster to do this in multiple subsequent steps. So create all parts as a bulk operation, then run all operations that need to run on the new objects. Also optimising the save() overrides and signals held a lot with my last project. MAybe this is a option for us too.

matmair:
@SchrodingersGat I am quite full this week but I would love to give it a try next week. Could you add a foo file that reflects the structure that makes the problem you want to be solved? So we can have a baseline to run tests against?

SchrodingersGat:
@eeintech the third step is the worst offender, true.

PR #2350, opened 2021-11-22
Fix sidebar translation

matmair:
This PR fixes the missing translations for sidebar items
Fixes inventree#2347

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2350"></a>


SchrodingersGat:
Ah yes, I see the problem now with the way the translations are performed here!

matmair:
> Ah yes, I see the problem now with the way the translations are performed here!

Yeah my approach broke something too - working on a fix asap. We should really try to get code coverage up

matmair:
@SchrodingersGat now it should work. Just waiting for the checks to go through

matmair:
@SchrodingersGat @eeintech you should really try replacing all translation strings with xxx
grafik

SchrodingersGat:
> you should really try replacing all translation strings with xxx

Can we make a "testing" language which does this automatically?

PR #2351, opened 2021-11-22
Respect rule W293

### LINKED ISSUE #2335: [FR] Confirm to W293
@SchrodingersGat I think we should confirm to W293 and probably review all exceptions in the setup.cfg:
https://github.com/inventree/InvenTree/blob/6a235c9f15d3731a736148eb25fedaaa2921e39d/setup.cfg#L2-L14

especially naming conventions. 0.6.0 is already a major refactor so we could fix those things as well. I would volunteer to do this.

matmair:
This PR removes rule W293 from the ignore list and fixes it in the code base
Fixes #2335 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2351"></a>

### DIFF
--- a/InvenTree/InvenTree/api_tester.py
+++ b/InvenTree/InvenTree/api_tester.py
@@ -46,7 +46,7 @@ class InvenTreeAPITestCase(APITestCase):
             self.user.is_staff = True
 
         self.user.save()
-        
+
         for role in self.roles:
             self.assignRole(role)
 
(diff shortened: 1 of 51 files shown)


matmair:
@SchrodingersGat on closer inspection I think we should keep all other exceptions till all PRs are closed and then refactor at least the function names N802. Everything else would make merges really difficult.

SchrodingersGat:
@matmair yes agree, we should leave the other rules until we are close to releasing `0.6.0`. Thanks for tackling this one :)

PR #2388, opened 2021-11-30
Part details refactor

SchrodingersGat:
Visual refactor for "part details" panel

Back by popular request!

Ref: https://github.com/inventree/InvenTree/discussions/2375#discussioncomment-1718697

image

image

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2388"></a>


matmair:
https://github.com/inventree/InvenTree/discussions/2375#discussioncomment-1718697

matmair (approved):
I think that restores that functionality. Should it maybe be an option to either have the tab or the panel? Maybe a per-user display option? I find the 'new' approach much cleaner on denser displays - on smaller Tablets on the workshop floor for example - than the 'old' way

SchrodingersGat:
> Should it maybe be an option to either have the tab or the panel? 

I think that is needless complexity for no real benefit. I'm happy to return to the "old way", but there's no point keeping both IMO :)

matmair:
Ok then let’s scrap this one😇

matmair:
9FA861D9-1268-440C-8EC6-0213B3B43EBF
The change makes this pretty unreadable on anything that’s not a cosy big office monitor like a 12“ tablet

SchrodingersGat:
@matmair yes that is not very good when the screen is narrowed. Any idea how we can fix that with some nicer CSS?

matmair:
> @matmair yes that is not very good when the screen is narrowed. Any idea how we can fix that with some nicer CSS?

Some stuff can be optimized for sure but it always will take a lot of space. The redundant part name could be removed and maybe a responsive overflowing table would reduce the space needs.

eeintech:
I like it! Thanks @SchrodingersGat 

The problem with small screens is that the "Available stock" table column should collapse below the part image, currently it doesn't, that should solve it.

PR #2401, opened 2021-12-01
Load minified fa

matmair:
currently, we load hull js but also ship minified js in static -> so why not load it too?
Increases the first load imes and reduces traffic a bit.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2401"></a>

### DIFF
--- a/InvenTree/templates/base.html
+++ b/InvenTree/templates/base.html
@@ -121,7 +121,6 @@
     </div>
     {% include 'modals.html' %}
     {% include 'about.html' %}
-    {% include "notifications.html" %}
 </div>
 
 <!-- Scripts -->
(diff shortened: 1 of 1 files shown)


InvenTree/templates/base.html
```diff
@@ -121,6 +121,7 @@
     </div>
     {% include 'modals.html' %}
     {% include 'about.html' %}
+    {% include "notifications.html" %}
```
SchrodingersGat, line 124: Looks like pollution from your other PR?

matmair: I hate it when that happens to me, thanks for the heads up!

SchrodingersGat:
OK to merge?

PR #2402, opened 2021-12-01
Fix to big / small references

### LINKED ISSUE #2400: [BUG] Out of range Reference_ID field error when creating Purchase Order
I was entering an purchase order, when a got a non-descript internal error flash up on screen.

Appears to be an out of range error on the Reference_ID field.

Running Main branch on bare metal, with MariaDB.

It would be good to create a more user friendly error message, in place of this:
Error Message

Here's the purchase order data I entered on the form:
Purchase Order Entry

I haven't found our why the field is out of range. The supplier record appear to be fine.
Attached is the traceback from console.
[purchase order error traceback.txt](https://github.com/inventree/InvenTree/files/7636736/purchase.order.error.traceback.txt)

matmair:
Fixes #2400

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2402"></a>


matmair:
@SchrodingersGat any idea why the bindings are failing now?
https://github.com/inventree/InvenTree/runs/4388317058?check_suite_focus=true#step:6:61

SchrodingersGat:
@matmair all your PRs seem to be failing on the HTML and javascript checks, which use npm:

image

Any ideas? Have you touched any npm stuff?

matmair:
No, even stuff like webhooks fail. Which is just pulled from upstream master directly

SchrodingersGat:
Can you try removing `package.json` from .gitignore (and git add'ing if it need be) to this PR, and see if the issue goes away?

matmair:
> Can you try removing `package.json` from .gitignore (and git add'ing if it need be) to this PR, and see if the issue goes away?

Will do!

matmair:
@SchrodingersGat the workflow now works again. no idea however why it broke. I fixed the node version and packages now so hopefully, it won't break so easily.
What do you think about grouping/chaining workflows? For example version numbering, style checks, java and html templates could be one group and only after these pass the other actions are run. That would keep our resource pool from overflowing with runs for PRs that will have to be rereun anyway.

SchrodingersGat:
Also, this PR is looking good - can you also provide a similar fix for the "serial number" field of the StockItem model? This uses a similar setup to PurchaseOrder / SalesOrder / BuildOrder models

matmair:
> I think there was a recent npm installer update, that might be the cause.
> 
> I like the idea of chaining workflows with the less resource intensive ones first! Do you think you would have time to submit a PR on this?

Already working on it. My latest pull on my branches consumed 500 minutes in a few clicks. That is not too sustainable.

matmair:
> Also, this PR is looking good - can you also amend this PR to provide a similar fix for the "serial number" field of the StockItem model? This uses a similar setup to PurchaseOrder / SalesOrder / BuildOrder models

@SchrodingersGat the fix is in

SchrodingersGat:
Great, thanks for the quick fix here :)

SchrodingersGat:
@matmair I've changed the milestone to 0.6.0 because the `reference_int` field was not added until the 0.6.x branch

matmair:
@SchrodingersGat 0.5.4 should be pretty complete, right?

SchrodingersGat:
Yeah, [0.5.4 has been released](https://inventree.readthedocs.io/en/latest/releases/0.5.4/)

I have not deleted the milestone here because I was hoping (for 0.6.0) to link the release notes page back to the milestone page on github. There are *so many* changes for 0.6.0 that I don't want to enumerate them all on the release notes page.

However I don't think that there is a way to "hide" milestone pages on GitHub once they are complete, so that may simply clutter up the github repo, which I don't want...

matmair:
> Yeah, [0.5.4 has been released](https://inventree.readthedocs.io/en/latest/releases/0.5.4/)
> 
> I have not deleted the milestone here because I was hoping (for 0.6.0) to link the release notes page back to the milestone page on github. There are _so many_ changes for 0.6.0 that I don't want to enumerate them all on the release notes page.
> 
> However I don't think that there is a way to "hide" milestone pages on GitHub once they are complete, so that may simply clutter up the github repo, which I don't want...

I mean if we know that why not -> it is marked as being in the past so it is somewhat clear

SchrodingersGat:
I'm mostly worried about it being quite cluttered in the future if a large number of (old) milestones stick around. Happy to leave "as is" for now and deal with it when it becomes a problem :)

PR #2416, opened 2021-12-03
Workflow remaster

matmair:
Fixes the skipping, maybe squash this @SchrodingersGat 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2416"></a>

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v2
       - name: Check version number
-        if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
Happy to squash and merge? :)

SchrodingersGat:
Hmm still has the same outcome:

image

Maybe we should just put the version checks into their own action, (revert to how that file was previously) and then keep all the other ones in QC checks?

matmair:
@SchrodingersGat I think it is time to cut our losses and do that. Also can we please switch to azure devops? I deploy with that system at work in I spend more time doing this pipeline than this whole year at work.

SchrodingersGat:
I'd rather not change the CI pipeline again (moving from travis to github was already enough work). The current pipelines work well (even better now that you've refactored them!)

PR #2417, opened 2021-12-04
Pipeline refactor ... again

matmair:
Should fix #2416, the problems seem to not be reproducible on forks so good luck to us

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2417"></a>


.github/workflows/qc_checks.yaml
```diff
@@ -44,9 +44,7 @@ jobs:
 
   pep_style:
     name: PEP style (python)
-    needs: check_version
```
SchrodingersGat, line 47: I think you also need to delete the "check version number" check on line 37?

SchrodingersGat:
Thanks @matmair  let's see how we go this time :)

Issue #2422, opened 2021-12-05
[BUG] maintenance_mode_state.txt permissions


SchrodingersGat:
@matmair I've had some issues updating an InvenTree instance to latest master after the plugin merge.

The error was the user which runs the server does did not have permissions to write to this file:

`./InvenTree/InvenTree/maintenance_mode_state_.txt`

I sorted the permissions out, but I am wondering if this file needs to be brought out of the source directory as it is a "user configurable" (or at least run-time configurable) file.

For example we bring "config.yaml" out into a separate directory.

Could this file path be specified as part of the InvenTree run-time settings? Either as an environment variable or a variable within config.yaml?

matmair:
@SchrodingersGat I think your PR is a great fix!

PR #2423, opened 2021-12-05
Adjustments for maintenance-mode options:

SchrodingersGat:
- Brings the state into the same directory as the runtime config file

@matmair this seemed the simplest option to extract the maintenance-mode file from the source directories.

It will be (by default) stored in the same directory as the runtime config file. So, this depends on how the sysadmin has configured the system.

Closes https://github.com/inventree/InvenTree/issues/2422

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2423"></a>


InvenTree/InvenTree/settings.py
```diff
@@ -206,6 +209,16 @@ def get_setting(environment_var, backup_val, default_value=None):
...
 
+# Options for django-maintenance-mode : https://pypi.org/project/django-maintenance-mode/
+MAINTENANCE_MODE_STATE_FILE_PATH = os.path.join(
+    config_dir,
+    'maintenance_mode_state.txt',
+)
+
+MAINTENANCE_MODE_IGNORE_ADMIN_SITE = True
+
+MAINTENANCE_MODE_IGNORE_SUPERUSER = True
```
matmair, line 220: @SchrodingersGat this can lead to superusers receiving errors while plugins are reloaded. In the reload process all apps are reinitialized -> might cerate inconsistencies if someone (even superusers) access parts of InvenTree that are reloaded as the models and apps do not exsist

matmair (approved):

PR #2426, opened 2021-12-05
Q_CLUSTER is using the database for its broker

nwns:
This change partially reverts the changes of redis as cache because Q_CLUSTER is not using redis as the broker and doesn't need to.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2426"></a>

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -558,7 +558,7 @@ _cache_port = _cache_config.get(
 if _cache_host:
     # We are going to rely upon a possibly non-localhost for our cache,
     # so don't wait too long for the cache as nothing in the cache should be
-    # irreplacable.  Django Q Cluster will just try again later.
+    # irreplacable.
     _cache_options = {
         "CLIENT_CLASS": "django_redis.client.DefaultClient",
         "SOCKET_CONNECT_TIMEOUT": int(os.getenv("CACHE_CONNECT_TIMEOUT", "2")),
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
> Q_CLUSTER is not using redis as the broker and doesn't need to.

This wording confuses me a bit..

Are you saying this PR changes the q-cluster broker from redis back to the database?

nwns:
No, the broker for q-cluster was never changed from the database (even though I thought it was when redis was added as a cache).

This removes the configuration that half setup redis as the broker because there are advantages to keeping q-cluster using the Django ORM (such as the tasks appearing in the Django Admin) as the broker.  This just clarifies what q-cluster is using as the broker.

matmair:
@nwns actually noticed that when deploying to Kubernetes. If your nodes are somewhat fast it is not a problem. You were talking about changes to InvenTree to enable easier deployment. I have a few instances running in a private RKE cluster rn and did not change anything but a few DB settings and the exchange backend (using something custom). Is there anything remaining that you wanted to change to make it more optimized?

nwns:
@matmair Not that I know of right now.

I am working through documenting the process of deploying on K8s for dev so that I can verify the current master can be deployed to my liking.  My current production installs are so customised (and old) that I can't just change the image to verify the behaviour of the changes.

matmair:
@nwns some k8s docs would be very nice. I am a novice in this area so i feel it would be a nad idea for me to do this.

Have you ever done ci - testing of kubernetes deployments? Would be interesting to add this to the release pipeline.

matmair:
Also don't hesitate to open FRs for customisations you would like to see / options you miss. More ideas are always great!

PR #2427, opened 2021-12-05
Swallow Serialization failure while updating exchange rates

### LINKED ISSUE #2241: [BUG] could not serialize access due to concurrent update
could not serialize access due to concurrent update

and just crash....

nwns:
We can swallow the serialization exception because there is a scheduled
task that will update these later anyway.

Fixes #2241

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2427"></a>


SchrodingersGat:
@nwns yes, this seems like a clean way of handling this issue. Thanks for the fix!

PR #2431, opened 2021-12-06
Subcategory tree

### LINKED ISSUE #2250: [FR] Idea for subcategory tree view
**Problem**
As a part of Bootstrap 5 introduction (#2205), category treeview was removed in favor of "Subcategory tab".
This leads to unnecessary clicks when user wants to change category outside of currently opened branch.

**Suggested solution**
Add drop-down category list to the header of every part/stock view with the same content as in "subcategory tab". Active category should be highlighted and selected.
This list could be combined with current breadcrumbs.
<details>
<summary>Mockups:</summary>

изображение

изображение

</details>

Restore old category tree view.

Unfortunately it is above my head.

matmair:
This PR will add a new implementation of the old category tree. Based on jsTree and with partial tree loading.

Work in progress.
Closes #2250 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2431"></a>


matmair:
@SchrodingersGat could you check out the new side tree?
Three things are missing:
- start with the current branch opened - work ion that rn
- styling -> Did you like the old style or do you want something new?
- show parts -> should we show parts or stop at categories?

grafik

SchrodingersGat:
> styling -> Did you like the old style or do you want something new?

I would like the "tree toggle" button to be placed in front of "parts" (like it was on v0.5):

image
image

Styling wise I'm not super happy with the icons - can we change those?

> show parts -> should we show parts or stop at categories?

Maybe add this as a user-specific option which is off by default?

SchrodingersGat:
@matmair regarding comments by @sintech above - a single API query for *all* categories should not be too expensive

matmair:
grafik

matmair:
After about 500 categories the performance gets increasingly worse so we might have to add something for users that just shadow a suppliers categories.
@SchrodingersGat I would be done with this, anything missing?

Issue #2429, opened 2021-12-06
[BUG] RuntimeError: Model class django.contrib.sessions.models.Session doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS


cdwilson:
**Describe the bug**
Running `inv worker` repeatedly shows the following errors:

```
08:47:59 [Q] ERROR Failed [seven-uncle-tennessee-north] - Model class django.contrib.sessions.models.Session doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS. : Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.10/3.10.0_2/envs/inventree/lib/python3.10/site-packages/django_q/cluster.py", line 436, in worker
    res = f(*task["args"], **task["kwargs"])
  File "/Users/chris/Projects/inventree/InvenTree/InvenTree/tasks.py", line 240, in delete_expired_sessions
    from django.contrib.sessions.models import Session
  File "/opt/homebrew/Cellar/python@3.10/3.10.0_2/envs/inventree/lib/python3.10/site-packages/django/contrib/sessions/models.py", line 10, in <module>
    class Session(AbstractBaseSession):
  File "/opt/homebrew/Cellar/python@3.10/3.10.0_2/envs/inventree/lib/python3.10/site-packages/django/db/models/base.py", line 113, in __new__
    raise RuntimeError(
RuntimeError: Model class django.contrib.sessions.models.Session doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

**Steps to Reproduce**
1. `inv worker`

**Expected behavior**
No errors in the worker logs.

**Deployment Method**
- [ ] Docker
- [x] Bare Metal

**Version Information**
# Version Information:
InvenTree-Version: 0.6.0 dev
Django Version: 3.2.5
Commit Hash: b7f3a044d
Commit Date: 2021-12-06
Database: sqlite3
Debug-Mode: True
Deployed using Docker: False

matmair:
@cdwilson have you tried running `inv update` after pulling the update? Was this a fresh install or an upgrade from 0.5.x?

cdwilson:
I upgraded using `inv update` from 0.5.0 dev:

# Version Information:
InvenTree-Version: 0.5.0 dev
Django Version: 3.2.5
Commit Hash: b3c8bd777
Commit Date: 2021-09-24
Database: sqlite3
Debug-Mode: True
Deployed using Docker: False

[...]

PR #2435, opened 2021-12-07
Remove old task to delete expired sessions

SchrodingersGat:
- Does not apply any more with new session management

Fixes https://github.com/inventree/InvenTree/issues/2429

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2435"></a>


SchrodingersGat:
@matmair any ideas why the checks are failing here? It seems to be suggesting that the `django_q` app is not there, but I can run the same command manually and it works ok

image

matmair (approved):

Issue #2369, opened 2021-11-26
[FR] Add ability to assign multiple stock items to a customer


SchrodingersGat:
- Use the existing "multi-line" API forms
- Select multiple stock items in a table
- Or, a single stock item
- Select customer via dropdown

Ref: https://github.com/inventree/InvenTree/issues/2253

- [x] Prevent items being assigned if they are allocated (to build or sales orders)
- [x] Prevent items being assigned if they are installed in other items

PR #2436, opened 2021-12-08
Assign to customer

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/2369

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2436"></a>

### DIFF
--- a/InvenTree/build/migrations/0018_build_reference.py
+++ b/InvenTree/build/migrations/0018_build_reference.py
@@ -19,7 +19,8 @@ def add_default_reference(apps, schema_editor):
         build.save()
         count += 1
 
-    print(f"\nUpdated build reference for {count} existing BuildOrder objects")
+    if count > 0:
+        print(f"\nUpdated build reference for {count} existing BuildOrder objects")
 
 
 def reverse_default_reference(apps, schema_editor):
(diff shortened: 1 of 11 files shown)


rikimasorati:
what exactly is the Assign to customer for? I thought it would assign so i can later create a sale and ship those but i found that they immedialty count as shipped when assigned. and without being able to attach them to a sale. I dont really understand what this would be good for. 
Is there and similar option I can assign multiple serialized stock items to a sale?

SchrodingersGat:
The "assign to customer" option allows you to bypass the salesorder process, if required. Following the sales order process is the recommended way to get a stock item to a customer, though

turist79:
Ahh, so this is what "assigning to the customer" does... Frankly speaking i have expected the same as the original poster - some kind of customer reservation on our own stock.

I tried to revert this action by performing stock transfer on such an item but it seems to be impossible despite the fact that Stock transfer form opens correctly and allows to provide new stock location etc - it returns "StockItem cannot be moved as it is not in stock". Is there a way to get the part somehow back to the stock from  this state?

The ultimate end result of "assigning" part to the customer via SO and via "assign to customer" look the same - part is listed on the customer`s page under "Assigned stock items" (which is already misleading because of the word "stock", "Shipped items" would make much more sense here). Additional problem is that it seems to be impossible to differentiate which parts have been manually assigned and which - peoperly shipped since there is no reference to the SO or shipment on this page.

matmair:
I am locking this;
If you have an issue please raise a new issue or discussion, we will probably miss comments on a closed PR from 3 years ago more often then not.

PR #2445, opened 2021-12-10
Issue2250

SchrodingersGat:
- Replaces https://github.com/inventree/InvenTree/pull/2431
- Closes #2250 

@matmair I've expanded on your PR somewhat:

I have chosen to use [bootstrap-treeview](https://github.com/jonmiles/bootstrap-treeview) rather than jstree as:

- The library was already in the InvenTree code base
- More consistent with the bootstrap UI approach
- (to me at least) visually more appealing

I have also added the tree display feature back into the "stock" views too.

Further, the tree display is now below the breadcrumbs which stops the UI elements moving when you click on them 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2445"></a>

### Screenshots

image
image
image

PR #2451, opened 2021-12-12
Subcategory treeview improvements

sintech:
A couple of improvements for #2445
* Do not show icons for childless categories
* Propagate expanded state all the way to the root

@SchrodingersGat , current implementation of collapse/expand tracking does not allow me to implement the following usecase:
1. Click "hamburger" button to open tree
2. Find needed category, click on it.
3. New category is opened with **tree collapsed**

I don't see any point for tree to be always open because it consumes almost all my screen space. While for sidebar it was OK to have it always open.

What do you think about removing treeview state saving and make it collapsed by default?

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2451"></a>

### DIFF
--- a/InvenTree/templates/js/dynamic/nav.js
+++ b/InvenTree/templates/js/dynamic/nav.js
@@ -218,7 +218,6 @@ function enableBreadcrumbTree(options) {
                     collapseIcon: 'fa fa-chevron-down',
                 });
 
-                setBreadcrumbTreeState(label, state);
             }
         }
     );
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
> What do you think about removing treeview state saving and make it collapsed by default?

I'm fine with this, I think that if it shows the correct expanded state when it opens, that would be ok

SchrodingersGat:
@sintech thanks for the tweaks here, looks great :)

SchrodingersGat:
Please address QC issues :)

sintech:
> Please address QC issues :)

@SchrodingersGat , done.

PR #2487, opened 2021-12-29
Add new SN match group

### LINKED ISSUE #2484: [FR] Extend the syntax of Serial Number input to include continuing from the last assigned Serial Number
Hi,

After discussing with @matmair I wanted to request a feature to extend the special syntax when entering Serial Numbers to include the possibility of selecting the next Serial Number. InvenTree already knows what the next Serial Number will be but I haven't found a way to actually input it without typing manually, which is cumbersome cause the next Serial Number disappears as soon as you enter something in the field (it's the placeholder message)

**Problem**
I want be able to assign the next serial number when assigning a serial number to a newly built part.

**Suggested solution**
There's already a system in place to assign serial numbers by using a special syntax. Extend this syntax with a wildcard token or accept an empty string to assign the next serial number in a row.

I tried using `1+` or just `+` in the input field but it didn't do what I wanted.

I wouldn't mind adding it but I think it's easier for someone who has more knowledge in this part of the code to implement it. I think extending the syntax would not be a daunting task, as I would propose a single new wildcard token or just accepting an empty string.

matmair:
This PR adds a new symbol to serial number input: ~ for the next possible serial number

Closes #2484

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2487"></a>


SchrodingersGat:
Seems like a reasonable implementation :) thanks @matmair

Issue #2306, opened 2021-11-13
[Plugin] Plugin Settings should be in a separate table


SchrodingersGat:
@matmair with the implementation of multiple plugins, there is a chance that settings could clash. We also don't want plugin code to accidentally overwrite global settings

I suggest we create a new "Settings" table, separate to "GlobalSettings" and "UserSettings", called "PluginSettings".

Where the UserSettings are separated by the user ID, PluginSettings are separated by the "plugin key" field (which maps to the existing PluginConfig table). 

This way, multiple plugins can have the same setting name (e.g. "API_KEY") without conflict.

Ref: https://github.com/inventree/InvenTree/pull/2074

matmair:
> @matmair with the implementation of multiple plugins, there is a chance that settings could clash. We also don't want plugin code to accidentally overwrite global settings

This is by design so that plugins can override behaviour.
> 
> This way, multiple plugins can have the same setting name (e.g. "API_KEY") without conflict.

I think it would be beneficial if plugins can share settings but I can change the default behaviour.

SchrodingersGat:
There is nothing stopping plugins from accessing global settings too, but by default I think the plugin settings should be separated.

PR #2478, opened 2021-12-21
Adds settings plugin class

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/2306

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2478"></a>

PR #2492, opened 2022-01-01
Plugin settings refactor

SchrodingersGat:
Replaces https://github.com/inventree/InvenTree/pull/2478
Closes https://github.com/inventree/InvenTree/issues/2306

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2492"></a>


InvenTree/common/models.py
```diff
@@ -977,13 +976,6 @@ def save(self, *args, **kwargs):
             'validator': bool,
             'requires_restart': True,
         },
-        'ENABLE_PLUGINS_GLOBALSETTING': {
```
matmair, line 980: @SchrodingersGat do you want plugins to have the right to define settings without enabling it first?

You will need to change this registry to, it checks if the setting is enabled:
 
 ​        ​if​ ​settings​.​PLUGIN_TESTING​ ​or​ ​InvenTreeSetting​.​get_setting​(​'ENABLE_PLUGINS_GLOBALSETTING'​):

matmair (changes requested):
Looks great in general, pls check comment before merging.

InvenTree/common/models.py
SchrodingersGat: Now that the settings are quarantined in their own table, I don't think this is required - it is perhaps an extra impediment to configuration?

SchrodingersGat:
@matmair if all tests pass here, I *think* it is ready to be tested on your end :)

SchrodingersGat:
@matmair let me know if you can see any outstanding issues here :)

InvenTree/common/models.py
matmair: How should plugins share configs like website-endpoints etc. then? My idea is that if a plugin already exists for an integration with another service than a new plugin can extend on the functions of the first plugin an reuse settings set for it.
Do you want another mixin for marking settings globally available?

SchrodingersGat: good point, I had not considered that... Although there's nothing stopping a plugin for looking at settings provided by *other* plugins (just be key lookup)...

matmair: I think we will see in practice if that usecase pops up often enough to warrant a change / mixin.
I will rewrite my already existing plugins to use the new APIs and see if problems / code duplications arise.

SchrodingersGat: Fair enough, let's see how it plays out and refactor as necessary.

Issue #2493, opened 2022-01-01
[FR] DevOps improvments


matmair:
Important dirs in the following are:
- InvenTree/common/
- InvenTree/InvenTree/

Suggested devops code-changes:
- [x] enforce docstrings and a docstyle for all important dirs
~~- [ ] enforce real unit tests on important dirs (atm we test that each URL responds as part of the test cases so a lot of coverage is just from that branch)~~
~~- [ ] add unit tests (selenium or npm tests) for all generally reused js-functions~~
- [ ] add developer docs that ship with the codebase (prob. sphinx) so that class and implementation docs help with onboarding. Via napoleon / autodoc the docstrings can be used for that
- [x] check for docstrings optionally in CI -> #2652 / #2653 / #3089
- [x] rethink static code / third party js structure -> maybe add webpack / npm to deliver fewer small files and compress all deliveries. The 3rd party js / CSS code is currently a bit scattered and a lot of unneeded stuff is placed into the webservers static (would be intrestng if ppl. use CDNs / S3 as storage provider -> #2789 

QoL improvments:
- [x] add an automatic welcome message to new issue authors to provide links to docs and some general info on what kind of info helps with debugging
- [x] restructure CONTRIBUTING to add more info about architecture and workflows
- [x] add an automatic message for each new PR author to the contributing file

matmair:
This is very longterm

matmair:
After https://github.com/inventree/InvenTree/pull/3089 closes we can add inline docs and then close this one.

simonkuehling:
As found in #4421, it would be good to include `django-extensions` in the dev environment for graphing DB schemas etc.

Since I just tried it in the current dev-Docker image, I have gathered what is missing in requirements for this:
```
$ apt-get install libgraphviz-dev
$ pip3 install pygraphviz
$ pip3 install django-extensions
```
and in https://github.com/inventree/InvenTree/blob/master/InvenTree/InvenTree/settings.py
```
INSTALLED_APPS = [
[...]
    'django_extensions',
]
```

matmair:
Regarding python autodoc: https://github.com/mkdocstrings/mkdocstrings

SchrodingersGat:
@matmair wrt mkdocstrings - the intent here is that we can extract code documentation and include at docs.inventree.org on each release?

matmair:
Yes, in the end. I still have to look into it - just very short on spare time right now so I am documenting my ideas before they are gone again.

[...]

PR #2495, opened 2022-01-03
Mark stale PRs / issues

matmair:
This PR adds a check for stale PRs / issues starting from everything created 2022. This helps to keep the backlog clean of things that have no interested stakeholders.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2495"></a>


matmair:
@inventree/maintainer @inventree/triage this was not listed in #2493 but also helps to declutter. Mass closures of the backlog could hide away interesting ideas that were created a long time ago.
This would mark issues / PRs with 60 days without activity as stale and close them 7 days afterwards. Those are the default values by Github, I find them reasonable but am not set on them.
Before merging a new tag 'no-activity' would need to be added

eeintech:
> Mass closures of the backlog could hide away interesting ideas that were created a long time ago.

How do you think that closing all old, stale issues help "revealing" the interesting ideas? By forcing people to keep them open? I don't think that's going to work out well...

SchrodingersGat:
A lot of the "stale" PRs are due more to lack of developer time than anything else.. I think we should just keep working our way through them as we are now

matmair:
> > Mass closures of the backlog could hide away interesting ideas that were created a long time ago.
> 
> How do you think that closing all old, stale issues help "revealing" the interesting ideas? By forcing people to keep them open? I don't think that's going to work out well...

I think I did not formulae that well. Old issues would not be touched, only new things (starting 2022-01-01). Mass closures are bad in my opinion.
Forcing ppl. to interact all ~2 Months to keep something on a roadmap helped a lot in work projects I am involved with. It keeps things that were not well thought out to a minimum and reduces uneccesarry dev work.
Once something gets assigned a milestone or assignee it would not be affected by the action, only things that are not on a roadmap and are not 'owned' by any dev - so no one want's to do it and there is no timeline.

matmair:
> A lot of the "stale" PRs are due more to lack of developer time than anything else.. I think we should just keep working our way through them as we are now

I think there are a lot of issues that the OP is not even interested in anymore (see the none responses in #1421). I think bumping ppl. would help keep them engaged somewhat and keep the backlog clean of stuff without stakeholders.

eeintech:
> I think I did not formulae that well. Old issues would not be touched, only new things (starting 2022-01-01).

Oh I see, I guess in that sense I would be voting in favor :+1:

SchrodingersGat:
@matmair understood, yes now it seems like a very good idea!

PR #2515, opened 2022-01-07
[Plugin] Triggered Events

SchrodingersGat:
This PR adds the concept of "triggered events" which plugins can register against.

Events are triggered from certain server-side (back-end) events, e.g.

image

Plugins can register callbacks against such events by inheriting from the `EventsMixin` class, e.g.

image

Callback functions are processed by the background worker, and so will not block user operations

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2515"></a>


InvenTree/plugin/builtin/integration/mixins.py
```diff
@@ -180,6 +181,69 @@ def unregister_tasks(self):
...
+    # Override this in subclass model
+    EVENTS = []
+
+    class MixinMeta:
+        MIXIN_NAME = 'Events'
+
+    def __init__(self):
+        super().__init__()
+        self.add_mixin('events', 'has_events', __class__)
+        self.events = getattr(self, 'EVENTS', [])
```
matmair, line 215: It could be beneficial to use an (overridable) function to get the EVENTS variable so that a plugin could generate events dynamically or pass function references to the currently initialized instance of the plugin to the events (so that the plugin can have a state-machine / states and respect them when processing events). Totally not a must.

matmair: check L260 - 264 for an example

matmair (approved):

InvenTree/plugin/builtin/integration/mixins.py
SchrodingersGat: Good idea, this would allow events to be generated at 'run-time' based on configured settings.

matmair: Yes, that is kind of the idea. I wrote the plugin mixin code to be basically just a bunch of functions you can override to change everything. This is why the activation sequences are in the registry.py so they can not be mangled with and permissions to execute are respected.

SchrodingersGat:
@matmair the implementation has been greatly improved now, I think. Let me know if you like it :)

matmair (approved):

SchrodingersGat:
image

Working nicely now

matmair:
@SchrodingersGat you could just define a few methods and use this to reduce code duplication:
https://stackoverflow.com/a/32230933/17860466

matmair (approved):

SchrodingersGat:
Good tip, didn't know you could do that!

PR #2564, opened 2022-01-21
Template reduce duplication

matmair:


matmair:
UI-change => manuall testing needed

PR #2565, opened 2022-01-21
Fablab barcode plugin

Julian-Guinane:


matmair:
Hi @Julian-Guinane, @ammelab [I strongly suggest](https://inventree.readthedocs.io/en/latest/extend/how_to_plugin/#development-guidelines) using the packages for plugins rather than forking - this decouples you from upstream file location changes and makes updating in the future faster.

PR #2580, opened 2022-01-28
fix spelling

matmair:
Thanks @Stephano120
Good catch!

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2580"></a>


SchrodingersGat:
You'll need to run `invoke migrate` :)

matmair:
@SchrodingersGat seems like I found the next thing to automate ;-)

Issue #2585, opened 2022-02-01
[BUG] Inventree fiddles with files directly rather than using Django Storage api


nwns:
**Describe the bug**
Django provides an abstract [storage api](https://docs.djangoproject.com/en/4.0/ref/files/storage/#django.core.files.storage.get_storage_class) that should be used for all file accesses so that the server can be run from places that do not have a local filesystem and/or shared filesystem to store the media files.

**Steps to Reproduce**

Run Inventree API server on a node with a readonly filesystem and you will crash out in [label/app.py](https://github.com/inventree/InvenTree/blob/dd3ac58f38e0f27f18b48dcded7bc84043539d5f/InvenTree/label/apps.py#L70) because it cannot create the folders, no matter what the DEFAULT_STORAGE_CLASS is set to.

**Expected behavior**
I expect that inventree should honour the DEFAULT_STORAGE_CLASS for writing and reading its media files.

**Deployment Method**
- [x] Docker
- [x] Kubernetes
- [ ] Bare Metal

**Version Information**
Application fails to start up.  This using the master branch of the main Inventree repository.

matmair:
@nwns this is a known side effect of the current implementation of file access. The storage API can help with this but introduces a number of possible bugs/security issues with different storage providers as some of them (boto for example) do not respect the originally intended access rights.
We have to rewrite at least 30 code blocks, possible more, to get full compliance with the API and introduce extra security measures so that users do not just expose all their part files, photos, test results and reports to the world. Especially with cloud deployments like Kubernetes in mind.

So this is a rather sensitive task - contributions are very welcome. I would love to get to this but I am slammed with work rn.
When implementing this we should also document and test a few possible storage classes and recommended settings for those. Especially block storage services like Amazon S3 can be dangerous as they enable fast misconfiguration and could lead to circumvention of our privileges system.

nwns:
The changes for running in k8s will come with at least one example of a S3 like API for testing.

More importantly, we need to stop adding new code that directly fiddles with the filesystem and uses the Storages api, even if the only production supported storage is local files for now.

matmair:
@nwns I will have to disagree here. Current standard in the codebase is to use pythons built-in file-APIs (fiddle as you call it). Using 2 different APIs for the same thing sounds like a great source for strange, hard to reproduce, special case bugs.
[...]

PR #2586, opened 2022-02-01
Switch Maint Mode to use default storage backend

nwns:
It seems that Maintance Mode also fiddles with the filesystem directly. This changes the backed that Maintance Mode uses so that it will put the flag file into wherever the media files are stored, currently this is the media directory on disk as there is no way to configure the default storages in inventree.

#2585

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2586"></a>

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -172,12 +172,6 @@ if MEDIA_ROOT is None:
     print("ERROR: INVENTREE_MEDIA_ROOT directory is not defined")
     sys.exit(1)
 
-# Options for django-maintenance-mode : https://pypi.org/project/django-maintenance-mode/
-MAINTENANCE_MODE_STATE_FILE_PATH = os.path.join(
-    config_dir,
-    'maintenance_mode_state.txt',
-)
-
 # List of allowed hosts (default = allow all)
 ALLOWED_HOSTS = CONFIG.get('allowed_hosts', ['*'])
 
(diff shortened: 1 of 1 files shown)


matmair:
@nwns this makes current deploy docs invalid, the file needs to be in the config dir as others might be write protected.

nwns:
@matmair The issue is that config may be write protected.  This should move the file into the default media storage, which should always be writable by the server.

nwns:
This seems to have the desired effect.  `maintenance_mode_state.txt` is now in the /static directory which should always be writable.

SchrodingersGat:
@matmair please let me know if you are happy with this change :)

matmair (approved):

matmair:
seems to work. We should add something to the docs about this as more complex proxy / CDN configs might struggle with this. I deploy a bigger install in a corporate environment where there are multiple proxies that also proxy the media files. So in this case the maintenance lock could be duplicated around. I will patch this on my deploy fork but it could happen to other users as static and media often get proxied / uploaded to external filehosts in bigger django apps.

nwns:
As far as I can tell, duplicating the file around is exactly what is supposed to happen, so that every copy of the API server has a copy.

The file is not served by the API server (and it being served by the media file host is fine as the file literally contains a 1 if in maintenance mode or a 0 if not in maintenance mode).  Personally this should actually be stored in the cache if we don't want it to be stored in the database itself, but using the default media file storage is a decent default.

matmair:
@as there can be multiple caches there can be a situation where some API servers get the 0 and some get the 1 so the locking is not working -> this will lead to a 501 / 503 when the server is loading a plugin or restarting after an error in a plugin.
This is really not desirable - we can deal with it when it crops up, I already patched it in my deploy branch. It will come up on bigger / global / cloud deployments but that is not really our audience rn.

PR #2599, opened 2022-02-03
Hide part pricing history

SchrodingersGat:
Due to the many outstanding issues with part "pricing", the "Part" detail page can take a very long time (> 10 seconds) just to load templates if it has a complex (multi-level) Bill of Materials.

Pricing is getting a major refactor in 0.7.x release.

For now, this PR adds an option to hide pricing history (and calculation of pricing data) from the part detail page.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2599"></a>

### DIFF
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -776,6 +776,18 @@ class InvenTreeSetting(BaseInvenTreeSetting):
             'validator': bool,
         },
 
+        # 2022-02-03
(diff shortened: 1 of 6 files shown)


SchrodingersGat:
@matmair you may want to be aware of this one. I'm (by default) disabling part pricing display as it is really killing the browser (and the server!) for any level of BOM complexity..

We will have to re-address this in 0.7.x when we refactor the pricing. Going to be a daunting task I fear.

PR #2609, opened 2022-02-07
Child model permissions

### LINKED ISSUE #2409: [BUG] Incorrect permissions for delete action on BOM items
**Describe the bug**

Cannot delete a selection of BOM items even with Parts>Change permission enabled.

**Steps to Reproduce**

1. Navigate to part of assembly type
2. Load the Bill of Materials tab
3. Click on the edit BOM button
4. Select a number of items
5. Click on the delete button for the selection (not individual)
6. Click "Accept", nothing happens

**Expected behavior**

All selected BOM items should be deleted from the BOM.

**Deployment Method**
- [ ] Docker
- [X] Bare Metal

**Version Information**
# Version Information:
InvenTree-Version: 0.5.4
Django Version: 3.2.5
Commit Hash: 66a08fe91
Commit Date: 2021-11-03
Database: postgresql
Debug-Mode: False
Deployed using Docker: False

eeintech:
Fixes #2409 

This PR is a proposed _minimal_ patch for adding permissions to add/change/delete part parameters and BOM items if user has `change` permission on the part model. It enables a "bypass" path for parent-children models in `RULESET_CHANGE_INHERIT` and does not depend on role structure (but still relies entirely on group permissions).

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2609"></a>


SchrodingersGat:
@eeintech let's see if I understand this.

You would like to extend the "logical" functionality of the existing (admittedly basic) permissions system, along the lines of:

"If a user has the ability to *change* a **Part**, then they also have the ability to *change / delete / create* a **BomItem**"

I would imagine that this concept could / would be extended to other tables, based on some logical criteria that we (the designers) develop, rather than permissions that the users could adjust themselves.

Is this accurate?

eeintech:
@SchrodingersGat You are correct. It intends to hardcode relationships between models in a very rudimentary way and expand the permissions of children models, given the parent model can be changed by user. The goal isn't to expose this at the user config level, at least not for now. This is just a quick patch, I can propose more parent-child relations if you wish, BOM items and parameters always most annoyed me!

SchrodingersGat:
I think it makes sense for "us" to decide this rather than the users (or admin users).

This seems OK as a patch to fix the outstanding issue, we can give this some more thought (and create some more complex behaviours) after next release.

SchrodingersGat:
Does this PR fix #2409 entirely, or other work required?

matmair:
@SchrodingersGat This patch works great and fixes the issue in an efficient way. Nice work @eeintech 

I would love to rewrite the permission system for the next major release as the current system doesn't provide the fine grained permissions some enterprise use cases need. We maybe should discuss all requirements then - especially what @eeintech brought up here

SchrodingersGat:
Agreed, it needs some serious attention for the next release.

Nice with @eeintech

matmair:
> Happy to discuss requirements for next release, I know @matmair is interested into object-level permissions too, however besides virtual twins to real objects (eg. stock items) I don't see where would it apply.

Mostly multitenant / department support.

PR #2620, opened 2022-02-13
Template cleanup

matmair:
This PR:
- makes the usage of quotes in templates uniform
- removes unneeded spaces

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2620"></a>

PR #2645, opened 2022-02-20
Template standardisation

matmair:
This PR makes the HTML templates more similar:
-  makes the usage of quotes in templates uniform
- removes unneeded spaces

Rework of parts of #2620

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2645"></a>


SchrodingersGat:
I appreciate the work here but please try to make such PR's smaller in the future (break into multiple if required).

This is hard to review!

image

SchrodingersGat:
Looks like there are still some breaking changes here:

image

I would suggest you re-submit as a number of smaller, targeted changes. there is simply too much changed here to be able to properly review.

PR #2680, opened 2022-02-27
Add DO quickdeploy to readme

### LINKED ISSUE #2670: Request: Add droplet as deployment method
Hi there @schrodingersgat @eeintech!
I would like to add the DigitalOcean droplet I recently created as a deployment method. There is the possibility to add a referral code to the link. If it is ok with you I would create a PR that adds a button to the Readme that enables a secure deployment to DigitalOcean within 3-5 min (if you already have a DigitalOcean account - else it is around 10 from start to ready deployment).

https://marketplace.digitalocean.com/apps/inventree

The update to 0.6.0 is already in review. The referral terms are: I get 25$ once a new customer (came from the app) spend 25$ as a credit on their plattform. I currently use Linode so I would not have use for it but I will switch if one or two referrals come in. I would be open to porridge resources for hosting the demo instance if you want.

matmair:
Closes #2670

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,9 @@ InvenTree is supported by a [companion mobile app](https://inventree.readthedocs
 
 - [**Download InvenTree from the Apple App Store**](https://apps.apple.com/au/app/inventree/id1581731101#?platform=iphone)
 
+# Deploy to DigitalOcean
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
Nice work @matmair

PR #2689, opened 2022-02-28
Add user with variable

### LINKED ISSUE #2501: [FR] Non-interactive, environment variable based set-up in Docker deployments
No. It is a quality of life improvement.

**Problem**
As discussed on the side in #2454, Docker-based deployment could be simplified a little bit. Some systems, in particular Synology NAS, do not have docker-compose nor do they allow for easy terminal access.

**Suggested solution**
I could also see an approach where bootstrapping is done through environment variables which could be removed by the user once it's set up (and then heavily complains about it if the user doesn't unset them afterwards), e.g., `INVENTREE_ADMIN_USER` and `INVENTREE_ADMIN_PASSWORD`. A start-up script could check for the presence of these, run migrations and create user accounts accordingly. That would probably be a more generic approach that might also work for other non-Synology deployment situations.

None on a generic level; for Synology NAS systems, a Synology package could work.

**Examples of other systems**
Show how other software handles your FR if you have examples.

Please no, my bash is terrible 😅

matmair:
This PR adds the possibility to add admin users with environment variables or via the config.

Closes #2501

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2689"></a>


InvenTree/InvenTree/apps.py
```diff
@@ -138,3 +146,54 @@ def update_exchange_rates(self):
...
+        if set_variables < 3:
+            logger.warn('Not all required settings for adding a user on startup are present:\nINVENTREE_SET_USER, INVENTREE_SET_EMAIL, INVENTREE_SET_PASSWORD')
+            settings.USER_ADDED = True
+            return
+
+        # good to go -> create user
+        user = get_user_model()
+        try:
+            with transaction.atomic():
+                new_user = user.objects.create_user(add_user, add_email, add_password)
```
SchrodingersGat, line 191: The new user should be assigned superuser status, right?

InvenTree/config_template.yaml
```diff
@@ -154,6 +154,11 @@ static_root: '/home/inventree/data/static'
 # Use environment variable INVENTREE_LOGIN_ATTEMPTS
 #login_attempts: 5
 
+# Add new user on first startup
```
SchrodingersGat, line 157: Should these be called `admin_user` / `admin_email` / `admin_password` for improved clarity?

matmair:
@schrodingersGat added the changes you asked for!

matmair:
Replaced with #2706

PR #2709, opened 2022-03-03
Stock export refactor

SchrodingersGat:
This PR refactors the existing "export stock" functionality

- Removes duplication of functionality, uses the existing API endpoints
- Export respects the filtering already performed via the tables
- Adds some export options for more "stock" tables
- Will make it easier to add data exporters for other tables

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2709"></a>


InvenTree/templates/js/translated/tables.js
```diff
@@ -21,6 +22,62 @@ function reloadtable(table) {
...
+                choices: exportFormatOptions(),
+            }
+        },
+        onSubmit: function(fields, form_options) {
+            var format = getFormFieldValue('format', fields['format'], form_options);
+            
+            // Hide the modal
+            $(form_options.modal).modal('hide');
+
+            for (const [key, value] of Object.entries(query_params)) {
```
matmair, line 67: @SchrodingersGat would this make sense as a generalized helper function? Can imagine we need this on multiple places

matmair (approved):
Great code reduction! Any plans to add testing? Would be a good point to add coverage.

PR #2714, opened 2022-03-04
[FR] Add delivery cost (excluding unit cost that already exists) in PO

### LINKED ISSUE #807: Add "empty" line to Purchase Order and Sales Order

### LINKED ISSUE #1727: Add "Shipping Cost" to orders

### LINKED ISSUE #2694: [FR] Add delivery cost (excluding unit cost that already exists) in PO

No.

**Problem**

When components are purchased, sometimes delivery of components price is excluded in purchasing, can calculate it in unit cost but that will be wrong as it's not the normal unit cost and usually deliver pricing doesn't change as much as components cost. 

If I include the delivery cost into the unit cost, it damages the plugin I'm developing to automatically grab recent components pricing from multiple suppliers as the component cost doesn't state anything about delivery.

Nice to have: where suppliers and manufacturers are, can add logistics providers such as DHL,UPS etc and add their cost there per weight per item.

**Suggested solution**

1. Add another field of delivery cost (total) for the PO
2. Optional but cool to have is delivery supplier, which company delivers it
3. option to add a tracking number? to see where the PO is.

Adding the delivery cost into the unit price.

**Examples of other systems**

Don't have any examples.

if can be done as plugin not as core, would love to give it support.

matmair:
Fixes #1727
Fixes #2694
Fixes #807

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2714"></a>


Zontex:
@matmair thanks for your hard work !

SchrodingersGat:
Nice clean implementation. I assume there's a lot of front-end work still to come?

Can you please address the two migrations? It looks like they may include some other features you were working on. In particular, there are fields which are added in the first but then removed in the second. Please squash this down to a single migration file :)

matmair:
@SchrodingersGat this is based on work I did half a year ago on the 0.5.4 branch - still fixing stuff.
I will clean up migrations, docs etc. before marking it as ready.
There is a lot of duplication in the frontend - maybe a refactor could help there. I will submit as is and then we can look.

matmair:
@SchrodingersGat also thinking about how plugins can fit in here.
I imagine adding the possibility to run calculations over an order with a button in the detail page and change the price of line item based on the results.

matmair:
@SchrodingersGat could you take a look at this while I clean up?
Any architecture changes whised? Does the naming fit?

matmair:
@SchrodingersGat how would you like this to interface with the reporting engine?

SchrodingersGat:
@matmair I'm happy with the architecture overall

> how would you like this to interface with the reporting engine?

Do you mean how are the additional lines provided to the reporting templates? Something as simple as:

image

matmair:

SchrodingersGat:
@matmair a few thoughts here:

### Terminology

I think it would be clearer to explicitly say that these new lines are "additional"

image

### Purchase Orders

Should we add the same additional features for purchase orders, while we are at it?

e.g. shipping / service fees / etc 

Specifically this would be required to close out https://github.com/inventree/InvenTree/issues/2694

matmair:
@SchrodingersGat I changed the wording. Not sure if the 'additional' will not raise questions in the future. No clue what qould be better though 😇.

I am running the unit test for POs right now. Should be up soon.

matmair:
@SchrodingersGat any options for reducing the code duplications in 'order.js'? I would like to reduce the code there before mergin.

matmair:
@SchrodingersGat on a more general note: come classes are called 'PO', others 'PurchaseOrder' - same with 'SO'.
Should we maybe use only one naming schema?

SchrodingersGat:
> @SchrodingersGat any options for reducing the code duplications in 'order.js'? I would like to reduce the code there before mergin.

Can you point me to what you mean by the duplications?

SchrodingersGat:
> @SchrodingersGat on a more general note: come classes are called 'PO', others 'PurchaseOrder' - same with 'SO'. Should we maybe use only one naming schema?

Fair point, I'd prefer the longer naming (e.g. `PurchaseOrder`). Although it is lengthier, it is much clearer.

matmair:
> > @SchrodingersGat on a more general note: come classes are called 'PO', others 'PurchaseOrder' - same with 'SO'. Should we maybe use only one naming schema?
> 
> Fair point, I'd prefer the longer naming (e.g. `PurchaseOrder`). Although it is lengthier, it is much clearer.

@SchrodingersGat should that be a separate PR or in this one?

matmair:
> > @SchrodingersGat any options for reducing the code duplications in 'order.js'? I would like to reduce the code there before mergin.
> 
> Can you point me to what you mean by the duplications?

@SchrodingersGat most of the changes in order.js are duplicate.

SchrodingersGat:
@matmair I'm happy with that for now - perhaps a future refactor (once we have unit testing for javascript, for example) would be in order.

SchrodingersGat:
@matmair getting a javascript error when I press "Add Additional Order Items" on the sales order page:

image

SchrodingersGat:
@matmair I see that @eeintech raises a good point in the discussion of https://github.com/inventree/InvenTree/issues/807

Namely, we can already add line items to purchase orders which do not reference a "Part". This can already be used for items such as shipping, one-off fees, etc.

image


Maybe extending this (existing) mechanism to the sales orders is a cleaner approach than introducing new database models?

i.e. update the "SalesOrderLineItem" model to allow it to not have to reference a "part".


Thoughts?

matmair:
@schrodingersgat
I find a separate model - that can not be allocated to a nicer solution - esp. if additional logic via plugins should be added (automatic tax calculation). But we can do this either way - just give me a direction and I will change it accordingly.

matmair:
@eeintech with unknown states a lot of code will get more complicated. For example, if you add lines whose values get automatically calculated by plugins (which is the next step after this) - for taxes eg.

eeintech:
@matmair I'm not talking about unknown states, just that blank or incomplete line items can be ignored in the next step. That's why I vote for not making the part field mandatory, not a new model.

SchrodingersGat:
> @SchrodingersGat I find a separate model - that can not be allocated to a nicer solution - esp. if additional logic via plugins should be added (automatic tax calculation). But we can do this either way - just give me a direction and I will change it accordingly.

I can see your point here. If we go this way, then we should make the `Part` field of the `PurchaseOrderLineItem` model mandatory, I thik.

matmair:
> > @SchrodingersGat I find a separate model - that can not be allocated to a nicer solution - esp. if additional logic via plugins should be added (automatic tax calculation). But we can do this either way - just give me a direction and I will change it accordingly.
> 
> I can see your point here. If we go this way, then we should make the `Part` field of the `PurchaseOrderLineItem` model mandatory, I think.

That is a breaking change. I can provide a migration but isn't that restricting current users?

SchrodingersGat:
Yes, a data migration would be required in that case...

What complexities / issues do you think we would encounter (with regard to plugins, for example) if we simply allowed the "part" field to be null for the SalesOrderLineItem model?

I am leaning towards this solution for its simplicity, and we would not need any data migrations.

matmair:
@SchrodingersGat I do not see problems in leaving the part field being nullable -  I see a problem with having multiple iterating lines (taxes, shipping, handling) with parts on the iterating lines themself as that can become expensive fast.

SchrodingersGat:
> as that can become expensive fast

How are you defining "expensive" here? In terms of implementation / coding effort? Financially? Something else?

matmair:
> > as that can become expensive fast
> 
> How are you defining "expensive" here? In terms of implementation / coding effort? Financially? Something else?

@SchrodingersGat computational. With a changing lazy queryset it would have to reload the set on each automated line calculation.

SchrodingersGat:
@matmair I have discovered another issue which impacts our decision here.

Currently if you have a purchase order with "blank" line items (that is, line items without a referenced part) then you cannot actually receive that line. Pressing the "receive items" button throws a javascript error, but even if that were fixed, the back end logic expects that you want to create a new stock item. You can't make a new stock item against a "blank" part, so this does not make sense.

I think that the correct way forward is to:

1. Update the `PurchaseOrderLineItem` model to require a non-null `SupplierPart` reference.
2. Create the new `ExtraLineItem` model (as you have already done here) for *both* `PurchaseOrder` and `SalesOrder` models
3. Data migration for existing users:
  - Find any instances of `PurchaseOrderLineItem` with a null (empty) `SupplierPart` reference
  - Create a new `ExtraLineItem` instance for each said instance
  - Delete the original `PurchaseOrderLineItem` instances with empty supplier part references

While this requires an additional data migration, I think it is the "best" way forward, as it clearly defines the separate expected behaviours of "physical" line items and "additional" (non physical) line items such as shipping, fees, etc, etc.

matmair:
@SchrodingersGat I think that makes more sense. We have to be weary of possible performance problems as the user base gets faster.
I am currently working on a schema where I will provide (storage limited) permanent subdomain hosting for users if they are willing to share their usage data so we can get real world performance data. I will migrate over some existing users from my private instances so we get more insights. Still ongoing but in the summer I should get to a point where we can start analyzing the data. I think there are a few points where the logic needs to be changed / optimized but without data it is hard to say.

matmair:
@SchrodingersGat so should I make the required migrations? I am still researching ways to make tests for them but not getting anywhere tbh.

eeintech:
> @matmair I have discovered another issue which impacts our decision here.
> 
> Currently if you have a purchase order with "blank" line items (that is, line items without a referenced part) then you cannot actually receive that line. Pressing the "receive items" button throws a javascript error, but even if that were fixed, the back end logic expects that you want to create a new stock item. You can't make a new stock item against a "blank" part, so this does not make sense.
> 
> I think that the correct way forward is to:
> 
> 1. Update the `PurchaseOrderLineItem` model to require a non-null `SupplierPart` reference.
> 2. Create the new `ExtraLineItem` model (as you have already done here) for _both_ `PurchaseOrder` and `SalesOrder` models
> 3. Data migration for existing users:
> 
> * Find any instances of `PurchaseOrderLineItem` with a null (empty) `SupplierPart` reference
> * Create a new `ExtraLineItem` instance for each said instance
> * Delete the original `PurchaseOrderLineItem` instances with empty supplier part references
> 
> While this requires an additional data migration, I think it is the "best" way forward, as it clearly defines the separate expected behaviours of "physical" line items and "additional" (non physical) line items such as shipping, fees, etc, etc.

Makes a lot of sense, I like it!

SchrodingersGat:
Great! You happy to tackle the data migration as outlined above?

matmair:
@SchrodingersGat so should I just make the required migrations? I am still researching ways to make tests for them but not getting anywhere eaningfull tbh.

And does the phrasing 'additional' make sense? In the german translation it does - not sure about general english thought. Seems kind of clunky to me. But I am certainly no native speaker. @SchrodingersGat @eeintech

matmair:
> Great! You happy to tackle the data migration as outlined above? It will need some thorough unit testing - there is already a framework for performing unit testing between different database migration levels.

Certainly! Can you point me to an example? I can not think of a snippet of the top of my head.

SchrodingersGat:
> Certainly! Can you point me to an example? I can not think of a snippet of the top of my head.

Here is an example of testing between two specific migrations:

https://github.com/inventree/InvenTree/blob/master/InvenTree/order/test_migrations.py

And an example of migrating from the *oldest* to the *newest* migrations (without knowing which migrations they are)

https://github.com/inventree/InvenTree/blob/master/InvenTree/build/test_migrations.py

SchrodingersGat:
> And does the phrasing 'additional' make sense? In the german translation it does - not sure about general english thought. Seems kind of clunky to me. But I am certainly no native speaker. @SchrodingersGat @eeintech

No, I don't think it makes much sense at all. There should be a better word to indicate the point of these "non physical" line items, but I am having trouble thinking of what that might be.

matmair:
@SchrodingersGat thank you for the sample. Regarding the wording we should probably all think a bit about it before merging.

eeintech:
Extra or optional maybe?

matmair:
> Extra or optional maybe?

@SchrodingersGat do you like extra? Maybe 'extra line' not 'extra line item'?

SchrodingersGat:
I'm reasonably happy with "Extra Line", unless we come up with something better :)

matmair:
@SchrodingersGat ok I will finish migrations and then change the name everywhere.

matmair:
@SchrodingersGat still need to write migration tests - will do that in a few hours.

InvenTree/order/migrations/0064_purchaseorderextraline_salesorderextraline.py
```diff
@@ -0,0 +1,103 @@
...
+    print(f'Starting to convert - currently at {OrderExtraLine.objects.all().count()} {extra_line_ref} / {OrderLineItem.objects.all().count()} {line_item_ref} instance(s)')
+    for lineItem in items_to_change:
+        newitem = OrderExtraLine(
+            order=lineItem.order,
+            notes=lineItem.notes,
+            price=getattr(lineItem, price_ref),
+            quantity=lineItem.quantity,
+            reference=lineItem.reference,
+        )
+        newitem.context = serializers.serialize('json', [lineItem, ])
```
SchrodingersGat, line 29: What is the purpose of this "context" field - what data / fields will be not available if we need to convert backwards?

matmair: @schrodingersgat there are multiple values like target_date, fullfilment etc. I am also planning to use the context for more plugin related things.

InvenTree/order/models.py
```diff
@@ -1163,6 +1167,38 @@ def complete_shipment(self, user, **kwargs):
         trigger_event('salesordershipment.completed', id=self.pk)
 
 
+class SalesOrderAdditionalLineItem(OrderLineItem):
+    """
+    Model for a single AdditionalLineItem in a SalesOrder
+    Attributes:
+        order: Link to the SalesOrder that this line item belongs to
+        title: titile of line item
```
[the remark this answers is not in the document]
matmair: Will fix

InvenTree/order/migrations/0064_purchaseorderextraline_salesorderextraline.py
SchrodingersGat: If these fields are still going to be used, I'd prefer to keep them in the new model. If not, I feel they should be deleted.

Can you give me an example of how the context would be used in the plugin system, differently to just keeping the fields against the model as "normal"?

SchrodingersGat:
In `order/models.py`:

image

- Remove this #TODO entry
- Change the `on_delete=models.SET_NULL` to `on_delete=models.CASCADE` (now that the `part` field is non-nullable)

InvenTree/order/migrations/0064_purchaseorderextraline_salesorderextraline.py
matmair: we do not need them in the new model - my thinking here is more that it does not add significant computational cost and there could be real information loss if a user updates without realising the model changed. There could be shipping target dates saved in them for example.

If a plugin wants to save additional data for a line the context could be great for that. Example: A line for shipping. Each time the plugin calculates the shipping rate, provider etc. it saves this information into the context of the line itself. I imagine this is more for documenting than for recalculating as it could give insights even after a plugin is removed.

SchrodingersGat: Fair, I can see that might be useful.

matmair:
> In `order/models.py`:
> 
> image
> 
>     * Remove this #TODO entry
> 
>     * Change to `null=False, blank=False`
> 
>     * Change the `on_delete=models.SET_NULL` to `on_delete=models.CASCADE` (now that the `part` field is non-nullable)

@SchrodingersGat  I changed it to block the creation of blank entries but left it as set_null so that deleting a part does not remove all old lines - seems counter-productive for when you need to check order values and lines are missing because you took parts out of the catalogue.

matmair:
Should be ready for merge now!

matmair:
@SchrodingersGat any idea what is happening here:
https://github.com/inventree/InvenTree/runs/5744391281?check_suite_focus=true#step:5:357
Company is not an instance of Company?

SchrodingersGat:
That's very strange. I can replicate it locally but not sure why it is happening...

SchrodingersGat:
Getting a client-side JS error here when trying to add an "Extra Line" to a sales order

image

SchrodingersGat:
Also the "extra lines" table seems to be very broken for the "Purchase Order" view:

image

SchrodingersGat:
@matmair it would be great to get this one closed out - any updates on this PR?

matmair:
@SchrodingersGat not really, did not work on this. Will take a look tonight.

matmair:
@SchrodingersGat this should be *kind of* good to go if the tests go through. I fixed the rendering issue - was a wrong capitalisation. 

There is still a broken test for the migration with the 'Company' is not 'Company' error. I will try to find a resolution for it but I tried already nearly for a month.

matmair:
@SchrodingersGat had time to look at this?

SchrodingersGat:
@matmair sorry, I had missed your ping. This looks great now, happy to merge in!

PR #2743, opened 2022-03-11
[BUG] Wrong icon on the place order button

### LINKED ISSUE #2742: [BUG] Wrong icon on the place order button
Small thing only, but the icon mentioned in the documentation would be better I think.
Képernyőfotó 2022-03-11 - 8 43 36

Doc says:
Képernyőfotó 2022-03-11 - 8 45 00

matmair:
Use paper plane as symbol
Fixes #2742


matmair:
@rkalman I made the change - let's see if @SchrodingersGat likes it that way. I am very indifferent about it.

PR #2747, opened 2022-03-14
Add field with last updated to supplier price break

matmair:
This PR;
- Adds an 'updated' field to the supplier price break, that auto-updates when the entry is saved
- Adds the field to the endpoints for supplier breaks
- Adds the field to the UI ( table with supplier price breaks)
- Ups the API version by one


InvenTree/company/templates/company/supplier_part.html
```diff
@@ -268,6 +268,11 @@ <h4>{% trans "Pricing Information" %}</h4>
                 return html;
             }
         },
+        {
+            field: 'updated',
+            title: '{% trans "Last updated" %}',
+            sortable: true,
```
SchrodingersGat, line 274: Please ensure that this date value is rendered through the `renderDate` function, according to user preference.

e.g.

```
  formatter: function(value) {
      return renderDate(value);
  }
```

SchrodingersGat (commented):
Looks great, just one simple fix to do :)

PR #2750, opened 2022-03-16
Rename for everything SO/PO

matmair:
Uses SO / POs more consequential in names (not - as requested, I am just too lazy to write it out).
Suggested in #2714 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2750"></a>


matmair:
@SchrodingersGat please only merge after #2714

PR #2372, opened 2021-11-27
Notification center and refactor

matmair:
This PR refactors the notification system to support multiple delivery methods.
It also adds a notification indicator in the navbar as well as a notification flyout and a notification page.

Closes #2279 
Closes #2282 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2372"></a>

PR #2764, opened 2022-03-20
Notification settings

SchrodingersGat:
Follows https://github.com/inventree/InvenTree/pull/2372

Adds a "settings" section for notifications.

First setting is an option for each user to choose to enable or disable email notifications

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2764"></a>

### DIFF
--- a/InvenTree/common/notifications.py
+++ b/InvenTree/common/notifications.py
@@ -129,7 +129,7 @@ class EmailNotification(BulkNotificationMethod):
 
     def send_bulk(self):
         html_message = render_to_string(self.context['template']['html'], self.context)
-        targets = self.get_targets()
+        targets = self.get_targets().values_list('email', flat=True)
 
         InvenTree.tasks.send_email(self.context['template']['subject'], '', targets, html_message=html_message)
 
--- a/InvenTree/common/notifications.py
+++ b/InvenTree/common/notifications.py
@@ -8,15 +8,19 @@ from allauth.account.models import EmailAddress
 from InvenTree.helpers import inheritors
 from InvenTree.ready import isImportingData
 from common.models import NotificationEntry, NotificationMessage
+from common.models import InvenTreeUserSetting
+
 import InvenTree.tasks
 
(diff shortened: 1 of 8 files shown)


InvenTree/common/notifications.py
```diff
@@ -108,13 +110,26 @@ class EmailNotification(BulkNotificationMethod):
...
 
     def get_targets(self):
+        """
+        Return a list of target email addresses,
+        only for users which allow email notifications
+        """
+
+        allowed_users = []
+
+        for user in self.targets:
```
matmair, line 120: Could we just use a filter? Seems like this could start taking a long time with a lot of targets.

InvenTree/common/notifications.py
```diff
@@ -8,15 +8,19 @@
...
 from common.models import NotificationEntry, NotificationMessage
+from common.models import InvenTreeUserSetting
+
 import InvenTree.tasks
 
 
 logger = logging.getLogger('inventree')
 
 
-# region notification classes
```
matmair, line 17: I would advocate to leave the regions in. They help to structure code in a bigger file if opened with a lot of bigger IDEs like pycharm, visual studio or visual studio code (with a plugin, the bultin function was removed a few releases ago)

matmair (approved):

InvenTree/common/notifications.py
SchrodingersGat: We have to individually check if each user has email notifications active.. Can we do that with a filter?

matmair: Not in reach of my computer atm. But it should be possible with a reference filter. So you filter the queryset on the target users, annotate the queryset with the reference to the setting value and then filter on the annotated value. That should result in only 1 database hit, not 1 per target user.

SchrodingersGat: I'm happy to look at optimizing this if you can propose something better :)

PR #2765, opened 2022-03-20
Add markdown editor

### LINKED ISSUE #2264: [FR] Inline Markdown editor
**Problem**
Currently, there is no preview for markdown in textareas that have that enabled.

**Suggested solution**
A fast search brought out this MIT-licensed library that could be loaded by class and seems to be enough but not too much:
https://github.com/Ionaru/easy-markdown-editor

Maybe this is not needed - could be beneficial for new users / not devs

yes

matmair:
This PR adds a visual markdown editor.

Closes #2264

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2765"></a>


SchrodingersGat:
@matmair it looks like you are making the markdown editor "inline" with the modal forms?

I think it would make more sense to not use a modal form for the "notes" field, but have a native editor displayed on the notes page.. Please see my *extremely rough* mockup below:

image

matmair:
@SchrodingersGat I will make an implementation over the weekend; I am a bit concerned with loading times on huge notes but we will see how the browser handles it.

SchrodingersGat:
@matmair the browser already has to load the notes to render them, I think it should be fine.

I'd also suggest that we make use of the API here, to asynchronously load (and auto-save) the notes field as the user edits it. I believe that the easymde library supports that?

SchrodingersGat:
@matmair I'm closing this one out as easymde has now been integrated :)

PR #2769, opened 2022-03-24
Add 'plugins_enabled' information to the root API endpoint

SchrodingersGat:


matmair (approved):

matmair:
@SchrodingersGat would it make sense to also provide an array with info about the currently loaded plugins?

SchrodingersGat:
> @SchrodingersGat would it make sense to also provide an array with info about the currently loaded plugins?

That information is already available at `/api/plugins/` if required :)

matmair:
> 
> That information is already available at `/api/plugins/` if required :)

@schrodingersgat I know - I was just thinking about reducing the number of needed API requests.

PR #2791, opened 2022-03-29
Preserve orders if companies are deleted

### LINKED ISSUE #2788: [BUG] Deleting a Customer Breaks Associated Sales Orders
**Describe the bug**

After deleting a Customer, any existing Sales Orders associated with the Customer receive the name "Invalid Customer", and cannot be opened using the left-side link in the Sales Order table.

**Steps to Reproduce**

1. Go to page for specific Customer.
2. Access admin panel using blue icon with tooltip "View in Administration Panel".
3. Scroll to bottom of admin panel page and delete customer.
4. Return to Sales Order list page.
5. Click on Sales Order which now says "Invalid Customer".

**Expected behavior**

Open Sales Order window where line items, shipments and allocations can be viewed.
If this expected behaviour is incorrect, please let me know an alternative for handling access to the broken SO.

**Screenshots**

image

I believe the following line in `InvenTree/order/models.py` is the cause of the issue as it references `None` since the customer no longer exists.

image

**Deployment Method**
- [x] Docker
- [ ] Bare Metal

**Version Information**

InvenTree-Version: 0.6.2
Django Version: 3.2.12
Commit Hash: 950d185
Commit Date: 2022-03-10
Database: mysql
Debug-Mode: True
Deployed using Docker: True

matmair:
This PR adds the function `get_deleted_company` to the `Company` models class. This will always return/create and return a special company that can not be deleted. This serves to keep the API returning predictable elements and keep the business logic sane (eg. aggregating queryies using id).

Fixes #2788

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2791"></a>


SchrodingersGat:
@matmair I'm not sure I agree with your approach here, at the very least I don't understand it.

What is the advantage of adding a new "is_deleted" field, when we could simply set the `order` reference to None. If we are going to allow deletion of company objects and retain the order information, isn't this just adding more complexity?

matmair:
@SchrodingersGat the idea is that there will always be a company instance. If you delete a company all instances that refer to that company are set to a fallback company for all orders etc. with deleted companies so that the API, frontend etc. do not break. Also you retain all information but the company something was sold to.

SchrodingersGat:
This "deleted company" does not provide any extra detail, compared to simply setting the related field in the `Order` model(s) to 'null' on delete, though?

On top of that another field is being added to the database model, which is only ever set to "True" for the deleted instance, correct?

From an information theory point of view, we'd retain exactly the same amount of information if we just nulled out the company field, and for less effort / overhead.

SchrodingersGat:
@matmair any further work on this one? Please use the `SET_NULL` option for these fields, rather than adding the extra complication of the faux deleted company object.

matmair:
@SchrodingersGat I implemented the requested change - I will however explore my first suggestion a bit further and find a cleaner implementation.

SchrodingersGat:
Unless you can find a convincing argument I'd prefer to simply set the company field to `null`  :)

matmair:
@SchrodingersGat I will ping you once I find a better argument - I would say we can merge like this with `SET_NULL` for now.

SchrodingersGat:
@matmair I'm happy with that. There are still some outstanding issues with this PR - the most notable of which is that deleting the "company" referenced from a SalesOrder or PurchaseOrder will break the templates:

image

image

matmair:
@SchrodingersGat I fixed the templates and __str__ instances

InvenTree/company/templates/company/manufacturer_part.html
```diff
@@ -90,7 +90,11 @@ <h4>
     <tr>
         <td><span class='fas fa-industry'></span></td>
         <td>{% trans "Manufacturer" %}</td>
-        <td><a href="{% url 'company-detail' part.manufacturer.id %}">{{ part.manufacturer.name }}</a>{% include "clip.html"%}</td>
+        <td>
+            {% if part.manufacturer %}
+            <a href="{% url 'company-detail' part.manufacturer.id %}">{{ part.manufacturer.name }}</a>{% include "clip.html"%}
+            {% endif %}
```
SchrodingersGat, line 96: If no manufacturer is set, we should display this 

```html
{% else %}
<em>{% trans "No manufacturer information available" %}</em>
{% endif %}
```

InvenTree/company/templates/company/manufacturer_part.html
```diff
@@ -324,7 +328,9 @@ <h4>{% trans "Parameters" %}</h4>
 
     deleteManufacturerPart({{ part.pk }}, {
         onSuccess: function() {
+            {% if part.manufacturer %}
             window.location.href = "{% url 'company-detail' part.manufacturer.id %}";
+            {% endif %}
```
SchrodingersGat, line 333: If we delete the manufacturer part, but no manufacturer is specified, where do we redirect the user? Index page?

InvenTree/company/templates/company/supplier_part.html
```diff
@@ -85,7 +85,11 @@ <h4>
         <tr>
             <td><span class='fas fa-building'></span></td>
             <td>{% trans "Supplier" %}</td>
-            <td><a href="{% url 'company-detail' part.supplier.id %}">{{ part.supplier.name }}</a>{% include "clip.html"%}</td></tr>
+            <td>{% if part.supplier %}
+                <a href="{% url 'company-detail' part.supplier.id %}">{{ part.supplier.name }}</a>{% include "clip.html"%}
+                {% endif %}
```
SchrodingersGat, line 90: Please add "No supplier information available" in the other case

InvenTree/order/templates/order/order_base.html
```diff
@@ -110,7 +110,11 @@
     <tr>
         <td><span class='fas fa-building'></span></td>
         <td>{% trans "Supplier" %}</td>
-        <td><a href="{% url 'company-detail' order.supplier.id %}">{{ order.supplier.name }}</a>{% include "clip.html"%}</td>
+        <td>
+            {% if order.supplier %}
+            <a href="{% url 'company-detail' order.supplier.id %}">{{ order.supplier.name }}</a>{% include "clip.html"%}
+            {% endif %}
```
SchrodingersGat, line 116: else, "No supplier information available"

InvenTree/report/templates/report/inventree_po_report.html
```diff
@@ -74,7 +74,7 @@
 
     <div class='header-right'>
         <h3>{% trans "Purchase Order" %} {{ prefix }}{{ reference }}</h3>
-        {{ supplier.name }}
+        {% if supplier %}{{ supplier.name }}{% endif %}
     </div>
```
SchrodingersGat, line 78: same again, add a message if the supplier has been deleted

SchrodingersGat:
@matmair a few small suggestions but pretty close!

InvenTree/company/templates/company/manufacturer_part.html
matmair: Index

SchrodingersGat:
@matmair ready for review again?

InvenTree/company/templates/company/supplier_part.html
```diff
@@ -85,7 +85,14 @@ <h4>
         <tr>
             <td><span class='fas fa-building'></span></td>
             <td>{% trans "Supplier" %}</td>
-            <td><a href="{% url 'company-detail' part.supplier.id %}">{{ part.supplier.name }}</a>{% include "clip.html"%}</td></tr>
+            <td>{% if part.supplier %}
+                <a href="{% url 'company-detail' part.supplier.id %}">{{ part.supplier.name }}</a>{% include "clip.html"%}
+                {% else %}
+                <em>{% trans "No supplier information available" %}</em>
+                {% endif %}
+                {% endif %}
```
SchrodingersGat, line 93: Is this a duplicate `{% endif %}` here?

matmair: good eye!

SchrodingersGat (approved):

SchrodingersGat:
Thanks for your work on this :0

Issue #2741, opened 2022-03-11
[BUG] Items remain allocated after shipment completed


rkalman:
**Describe the bug**

After a shipment is completed for a sales order, these items are also considered as "allocated to orders".

**Steps to Reproduce**

1. create new sales order
2. create new shipment, add line item, allocate it from stock. It looks like this which is correct:
<img width="1500" alt="Képernyőfotó 2022-03-11 - 8 09 56" src="https://user-images.githubusercontent.com/580623/157821111-03f122de-d175-4783-aa50-5cddbeae4001.png">

4. complete shipment
<img width="1126" alt="Képernyőfotó 2022-03-11 - 8 10 25" src="https://user-images.githubusercontent.com/580623/157821190-89d09daf-f2b4-4320-a496-1fd20c1d91d6.png">
5. on the part page, counters shows up like this: 
<img width="1701" alt="Képernyőfotó 2022-03-11 - 8 12 52" src="https://user-images.githubusercontent.com/580623/157821581-5a881069-0758-4e54-9231-9f870cdfe149.png">
I think "required for sales orders" and "allocated to orders" should be 0, because the required quantity already shipped.

But go forward:
6. Complete the sales order
7. Now the "allocated to orders" line disappears on the part page, it is correct 
8. create a new sales order, create a shipment, add the same line item with quantity=1, allocate it from stock
9. part page now shows like this: 
<img width="1705" alt="Képernyőfotó 2022-03-11 - 8 18 23" src="https://user-images.githubusercontent.com/580623/157822376-ddbf52fd-708f-49d4-a64c-6960ec0718ee.png">
SO0001 already shipped, but its items count as they would be in a pending order

**Expected behavior**

Shipped item should lost its allocation status, because it leaves the stock

**Version Information**

tested using demo.inventree.org

SchrodingersGat:
Bug confirmed, thanks for reporting @rkalman

SchrodingersGat:
@rkalman should be fixed in latest master now

rkalman:
Thank you again, Oliver!

rkalman:
Hi Oliver, just one more flaw regarding this.
After the shipment completed, on the part page the "available stock" counter seems wrong. In this case it should be 14 instead of 14-3=11. Tested on demo.inventree.org:
[...]

PR #2796, opened 2022-03-31
Show available stock in part BOM table instead of stock

### LINKED ISSUE #2795: [BUG] BOM table available stock calculation
**Describe the bug**
BOM table shows "in stock" in "Available" column instead of "Available stock". So this field excludes allocations.

**Steps to Reproduce**
Képernyőfotó 2022-03-31 - 17 28 27
Képernyőfotó 2022-03-31 - 17 29 35

**Expected behavior**
Show available stock in the available column.

**Version Information**
demo.inventree.org

If you agree I can submit a PR for this to review.

rkalman:
Fixes #2795

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2796"></a>

### DIFF
--- a/InvenTree/part/serializers.py
+++ b/InvenTree/part/serializers.py
@@ -246,6 +246,8 @@ class PartBriefSerializer(InvenTreeModelSerializer):
 
     stock = serializers.FloatField(source='total_stock')
 
+    available = serializers.FloatField(source='available_stock')
+
     class Meta:
         model = Part
         fields = [
@@ -263,6 +265,7 @@ class PartBriefSerializer(InvenTreeModelSerializer):
             'purchaseable',
             'salable',
             'stock',
+            'available',
             'trackable',
             'virtual',
             'units',
--- a/InvenTree/templates/js/translated/bom.js
+++ b/InvenTree/templates/js/translated/bom.js
@@ -798,7 +798,7 @@ function loadBomTable(table, options={}) {
     });
 
     cols.push({
-        field: 'sub_part_detail.stock',
+        field: 'sub_part_detail.available',
         title: '{% trans "Available" %}',
         searchable: false,
         sortable: true,
@@ -808,7 +808,7 @@ function loadBomTable(table, options={}) {
             var text = value;
 
             if (value == null || value <= 0) {
-                text = `<span class='badge rounded-pill bg-danger'>{% trans "No Stock" %}</span>`;
+                text = `<span class='badge rounded-pill bg-danger'>{% trans "Not available" %}</span>`;
             }
 
             return renderLink(text, url);
@@ -903,7 +903,7 @@ function loadBomTable(table, options={}) {
                 var can_build = 0;
 
                 if (row.quantity > 0) {
-                    can_build = row.sub_part_detail.stock / row.quantity;
+                    can_build = row.sub_part_detail.available / row.quantity;
                 }
 
                 return +can_build.toFixed(2);
@@ -914,11 +914,11 @@ function loadBomTable(table, options={}) {
                 var cb_b = 0;
 
                 if (rowA.quantity > 0) {
-                    cb_a = rowA.sub_part_detail.stock / rowA.quantity;
+                    cb_a = rowA.sub_part_detail.available / rowA.quantity;
                 }
 
                 if (rowB.quantity > 0) {
-                    cb_b = rowB.sub_part_detail.stock / rowB.quantity;
+                    cb_b = rowB.sub_part_detail.available / rowB.quantity;
                 }
 
                 return (cb_a > cb_b) ? 1 : -1;


SchrodingersGat:
@rkalman thanks for submitting a patch for this, however there are a few subtleties here with how the API is handled.

The existing `available_stock` function (for the `Part` model) is not very "efficient" when it comes to performing database queries. It's "fine" for running against a single Part instance, but once you are performing a large lookup of data (e.g. an entire BOM) then referencing this `available_stock` parameter quickly becomes expensive.

Your approach of adding an 'available_stock' field to the API serializer is good one, the data just needs to be calculated and annotated as part of the initial database query.

I will submit a new PR which addresses this and tag you in it, it might be easier to see what I mean by showing rather than telling! 

Addressing DB query efficiency has been a big learning curve for me, and it gets pretty complicated :)

SchrodingersGat:
For example here's how we calculate the "on_order" quantity in the API serializer context.

image

Note that it is different from how we calculate it for an individual `Part` instance:

image

SchrodingersGat:
Also, if we are making changes to the API the version number in InvenTree/version.py needs to be bumped, in addition to a description of what has been added :)

rkalman:
Yes, I really do not know the code in deep details as you see, a thank you for guiding me to the right direction!

However if the available stock calculation is expensive, then it can raise a bigger problem.
On many places the stock quantity is displayed only (like in the box table here, but also in the sales orders, in the quick search menu, ...) but basically it does not tell too much for the user when he uses the allocation system...
In my opinion all of this places should show the available quantity.

SchrodingersGat:
"Expensive" is relative :)

In simple terms the difference between the two approaches are:

## @rkalman method

- Retrieve multiple parts from the database in a single large query
- Iterate through each part (one at a time) and call the `available_stock` method
- Each call to `available_stock` requires multiple small database queries
- For "N" part objects, you are now performing "1 + N" database hits (this is the expensive part!)

## @SchrodingersGat method

- Pre-annotate the database query to *also* fetch allocation data for related stock items
- Retrieve multiple parts from the database in a single query (with annotations performed as part of the SQL query)
- Done!
- A single database hit is *much* cheaper, even if the query itself is a lot more complicated

Check out the implementation here: https://github.com/inventree/InvenTree/pull/2797

Once this change is merged in, please merge master into your branch, and use the "unallocated_stock" quantity (now available in the API) in the BOM table. Then push those commits back here!

If you have found other places where the "available" quantity should be used (rather than the "total") then that would be best done as a separate PR :)

SchrodingersGat:
Query annotation is performed as below:

image

rkalman:
Ok, thank you!
However the available quantity has one flaw, could you take a look on this: https://github.com/inventree/InvenTree/issues/2741#issuecomment-1084436693

SchrodingersGat:
@rkalman ok, that PR has been merged in now, which makes the field "unallocated_stock" available via the API.

Please merge the master branch into yours, and push the changes back to here :)

rkalman:
Ok, I'm stuck a little bit. I see your changes in PartSerializer, however this BOM table uses the PartBriefSerializer.
So I don't know how to use that new "unallocated_stock" field, because it seems to me it is only available in the PartSerializer, and not int the Part model itself, which is used by the PartBriefSerializer...

SchrodingersGat:
@rkalman fair point - I would suggest that as you need access to this extra information in the `BomItem` API, then change the `BomItemSerializer` class to use the `PartSerializer` rather than `PartBriefSerializer`:

image

rkalman:
yes, that was I have done, but where to call the static method partserializer.annotate_queryset?

SchrodingersGat:
Ah, ok on *closer* inspection, this is very interesting. I'm stumped - I am not sure how to annotate a queryset against related fields for the `part_detail` information - or if it is even possible.

Another interesting point is that you have helped me to understand at least part of why the [BOM API is slow](https://github.com/inventree/InvenTree/issues/2042).

Adding the `stock` field to the `PartBriefSerializer` is causing the "1 + N" query problem (described above) and post-calculating the available stock for each part separately.

I think that probably what we need here is a custom approach for serializing BOM items, with available quantities annotated to the BomItem serializer itself.

This would also let us do some other useful things, like adding calculated fields such as "substitute_stock" which takes into account *substitute* parts.

I'll have to mull over this one for a little bit, it's going to be reasonably complex to implement this *properly*.. I'll get back to you :)

SchrodingersGat:
@rkalman I have a [work in progress](https://github.com/inventree/InvenTree/pull/2806) to solve this, which adds both the "available_stock" and "available_substitute_stock" data to the BomItem API.

However I am stuck trying to add "available_variant_stock" due to the difficulty (maybe impossibility) of annotating stock items for all possible variants of parts, which exist as a tree structure and is hard to query as a flattened structure.

Please check out #2806 and let me know what you think so far.

rkalman:
I think the original idea to show the available stock instead of total stock in the BOM table is fulfilled nicely, it is much more relevant. As you know I am really new in python&django so unfortunately I cannot improve this any further.

SchrodingersGat:
@rkalman fair point, it is an improvement at least - I'll merge in for now and consider the other improvements later.

PR #2801, opened 2022-04-03
Use new unallocated_stock quantity in the part table and in the search results

rkalman:
Display available quantities in part table instead of in_stock quantities.
Also enhance the stock badge to show No Stock / Not available / Building xxx / On order xxx in the table.
Render the badge used in the search results using the same principle, but showing available/in stock quantities.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2801"></a>


SchrodingersGat:
Nice @rkalman thanks for the improvements here. Hopefully we can roll this into the BOM table soon too!

PR #2804, opened 2022-04-04
Quantity fixes

SchrodingersGat:
Tweaks for new part stock annotations

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2804"></a>

### DIFF
--- a/InvenTree/templates/js/translated/model_renderers.js
+++ b/InvenTree/templates/js/translated/model_renderers.js
@@ -31,6 +31,24 @@
  */
 
 
+// Should the ID be rendered for this string
+function renderId(title, pk, parameters={}) {
+
+    // Default = true
+    var render = true;
+
+    if ('render_pk' in parameters) {
+        render = parameters['render_pk'];
+    }
+    
+    if (render) {
+        return `<span class='float-right'><small>${title}: ${pk}</small></span>`;
+    } else {
+        return '';
+    }
+}
+
+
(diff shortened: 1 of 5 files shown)


InvenTree/templates/js/translated/model_renderers.js
```diff
@@ -31,6 +31,24 @@
  */
 
 
+// Should the ID be rendered for this string
+function renderId(title, pk, parameters={}) {
```
matmair, line 35: Nice abstraction.

InvenTree/templates/js/translated/part.js
```diff
@@ -491,13 +491,16 @@ function duplicateBom(part_id, options={}) {
...
+ * Construct a "badge" label showing stock information for this particular part
+ */
 function partStockLabel(part, options={}) {
 
     if (part.in_stock) {
         // There IS stock available for this part
 
         // Is stock "low" (below the 'minimum_stock' quantity)?
-        if (part.minimum_stock && part.minimum_stock > part.in_stock) {
+        if ((part.minimum_stock > 0) && (part.minimum_stock > part.in_stock)) {
```
matmair, line 503: very curious that we are only discovering this now

SchrodingersGat: I do not think it was causing a bug, actually. But, there's a *potential* to cause a bug :)

PR #2805, opened 2022-04-04
Plugins for notifications

### LINKED ISSUE #2385: [FR] Add that registers new notification types (and delivery options)
Add the option to register notification and delivery options via plugins.

matmair:
This PR adds optimisations for the notification API to make plugins that add notification methods easier by:
- Making use of settings for NotificationMethods easier
- Make user settings for notification (single pane!) easier
- optimise loading times for NotificationMethods - load only once

Closes #2385

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2805"></a>


matmair:
@SchrodingersGat this is finally ready for review!

SchrodingersGat:
@matmair I'm having some trouble wrapping my head around what's going on here. This looks very complex, can you run me through what's happening?

matmair:
@SchrodingersGat it is a pretty simple change but added another settings class that is pretty much the same as the others. Therefore I restored the settings a bit to reduce duplications.

The base concept is that notification methods will be fetched on startup and saved as a list in a custom object (`common.notifications.storage`) - just like the `registry` for plugins or the `apps` object in Django for app management.
Also, notification methods are now able to register global and per-user settings that work the same way as with plugins (it uses the same code mostly).  This also enables users to turn off specific notification methods just for them.

To ship a sample of how notifications should be built I refactored the integrated email notification into `InvenTree/plugin/builtin/integration/core_notifications.py`

matmair:
@SchrodingersGat had time to look at this?

SchrodingersGat:
@matmair understood, after a second read through it makes more sense now. It is crazy how complicate the "settings" code has become now!

SchrodingersGat:
image

Getting a duplicated item in the "notification settings" window

matmair:
@SchrodingersGat thank you for raising the issue - this is fixed now. Should be ready for final review and merge!

SchrodingersGat:
Ok, so some more issues :)

Firstly, when I try to change "enable email notifications" for the user, it throws an API error

image

It is trying to edit the *global* setting - I think that the duplicates vales before were because there is a global and a user setting with the same name. Instead of disallowing duplicates, we should not be displaying *global* settings in the user settings tab.

The other issue is the error that is being thrown:

image

image

image

It is throwing this error because setting **1** is actually the currency:

image

SchrodingersGat:
It would be useful to add "PluginSetting" and "NotificationUserSetting" integration to the admin display, too:

image

SchrodingersGat:
I think one issue is that you will need to add a new serializer class / API interaction for editing the "NotifcationUserSetting" class, as we have done for "PluginSetting"

image

matmair:
> It would be useful to add "PluginSetting" and "NotificationUserSetting" integration to the admin display, too:
> 
> image

We already have the settings (they are inline in the PluginConfigs)

matmair:
@SchrodingersGat this would be ready again

SchrodingersGat:
@matmair unfortunately still getting some errors:

### Plugin Settings

Server is throwing a 500 error when I try to change a plugin setting now:

image

image

**Note:** *This breaks configuration of all plugin settings*

### Notification Settings

Notification settings are also throwing a 500 error

image

### User Settings

Still seem to be working OK

### Global Settings

Still seem to be working OK

### Unit Testing

Please add some unit testing for the refactored "PluginSetting" and "NotificationSetting" classes, in addition to addressing the errors thrown above. Refer to the `GlobalSettingsApiTest` class (`./common/tests.py`)

SchrodingersGat:
Great, all seems to be working very well now!! Major effort on this one, thanks mate

PR #2812, opened 2022-04-06
Fix part and po files import

### LINKED ISSUE #2811: Exception while importing purchase order file
**Describe the bug**

When I try to submit a CSV to a purchase order right after I click "Upload File" I get an Internal Server Error [with this log](https://pastebin.com/sAsdZ8QU)

**Steps to Reproduce**

**Expected behavior**

A file choosing window to select which file on my computer should be uploaded

image

**Deployment Method**
- [x] Docker
- [ ] Bare Metal

**Version Information**

0.6.3

eeintech:
Fixes #2811

### DIFF
--- a/InvenTree/templates/patterns/wizard/upload.html
+++ b/InvenTree/templates/patterns/wizard/upload.html
@@ -1,3 +1,5 @@
+{% load i18n %}
+
 <div class='panel' id='{{ panel_id }}'>
     <div class='panel-heading'>
         <h4>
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
Thanks for the quick fix @eeintech

InvenTree/templates/patterns/wizard/upload.html
```diff
@@ -1,3 +1,5 @@
+{% load i18n %}
```
matmair, line 1: It is always the simple mistakes that are overseen ;-)

matmair (commented):
nice one @eeintech

PR #2833, opened 2022-04-15
Add 'available_variant_stock' to BomItem serializer

SchrodingersGat:
Adds information on variant stock to Part and BomItem serializers.

Allows for a more comprehensive picture of "available stock" when looking at BOM data or allocating items to a build order.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2833"></a>

### TODO

- [x] Unit tests
- [x] Update to use queryset annotation


matmair:
@schroedingergat this seems very expensive. Should this be a setting with a warning? I can imagine it taking quite a lot of ressources with larger databases.

SchrodingersGat:
Definitely it is expensive - and should be converted to a queryset annotation. I tried a number of approaches here but couldn't work it out.

The caveat here is that the extra database queries should only run *if* the individual BomItem has `allow_variants=True`. The calculation will short-circuit if this is set to `False` (which would be the majority of cases).

If it proves to be too intensive and is slowing down the Bom table load, we can reassess

matmair:
@schrodingergat I checked on a bigger install and if you have a warehouse full of stuff in your InvenTree it slows down considerable. Using redis as a database cacher helps a lot put I do not think a lot of ppl have that enabled.

SchrodingersGat:
Hmm, might need some further attention then. I think it's a very important addition (in particular I needed it for my work setup, as the "available-stock" parameter was calculating incorrectly for BOM lines where variants were used).

However setting this up as a database annotation looks like it may present some difficulties.

Current queryset annotations (at least the complex ones) make use of the [django-sql-utils](https://github.com/martsberger/django-sql-utils) addon.

e.g.

image

Functions such as `SubquerySum` make use of the related fields on the model instance.

However the "variants" are not related fields, they are a complex "tree" which exist under each model instance. It is possible (using [django-mptt](https://github.com/django-mptt/django-mptt) ) to construct a single query to return all variants of a given part as a "flat" query result, but I'm not sure how to connect the two things together.

Any ideas?

matmair:
@schroedingergat i have to look into this further to give a qualified answer. Will circle back when I have done that.

SchrodingersGat:
Thanks, appreciate it. Hopefully there's a workable solution.

matmair:
@SchrodingersGat this is a bit of a brain-nut. I can not get lower than 4 hits. Maybe someone else finds a way to do it in one hit.

SchrodingersGat:
@matmair in what might be my django magnum opus:

```python
"""
Queryset annotation to provide single-hit database lookup for all variant stock items for a part.

- We will get a queryset of all parts (with some filter)
- Annotate with the number of variants under that part
- Annotate with the total number of stock items *for all variants* under that part
- Annotate with the total stock quantity *for all variants* under that part

This took a *lot* of experimentation and searching, the solution was finally found here:
https://stackoverflow.com/questions/42543978/django-1-11-annotating-a-subquery-aggregate


Note here that as the 'variant tree' is defined using a MPTT structure, we need to make use of the additional fields,
to allow us to return a flattened sub-tree in a single database hit. 

It's not exactly pretty, but it works!

"""

from django.db.models import OuterRef, Subquery, IntegerField, FloatField
from django.db.models import F, Func

from part.models import Part
from stock.models import StockItem

# Construct an initial queryset to return parts
queryset = Part.objects.all()

# Construct a subquery to return all variant parts *below* (and not including) this one
variant_part_query = Part.objects.filter(
    tree_id=OuterRef('tree_id'),   # Parts in the same tree as the outer part
    lft__gt=OuterRef('lft'),       # Filter by 'lft' MPTT field
    rght__lt=OuterRef('rght'),     # Filter by 'rght' MPTT field
)

# Construct a subquery to return all stock items of any variant part *below* (and not including) this one
variant_stock_query = StockItem.objects.filter(
    part__tree_id=OuterRef('tree_id'),
    part__lft__gt=OuterRef('lft'),
    part__rght__lt=OuterRef('rght'),    
)

# Now we can annotate the original queryset with a bunch of Subquery calls:
queryset = queryset.annotate(
    n_variants=Subquery(
        variant_part_query.annotate(
            total=Func(F('pk'), function='COUNT', output_field=IntegerField())
        ).values('total'),
    ),
    n_stock_items=Subquery(
        variant_stock_query.annotate(
            total=Func(F('pk'), function='COUNT', output_field=IntegerField())
        ).values('total')
    ),
    q_stock_items=Subquery(
        variant_stock_query.annotate(
            total=Func(F('quantity'), function='SUM', output_field=FloatField())
        ).values('total')
    )
)

# Now, we can execute as a single query!
for p in queryset.values():
    print("Part:", p['name'])
    print("- Variants:", p['n_variants'])
    print("- Variant Stock Items:", p['n_stock_items'])
    print("- Variant Stock Quantity:", p['q_stock_items'])
    print("")
```

### References

- https://hansonkd.medium.com/the-dramatic-benefits-of-django-subqueries-and-annotations-4195e0dafb16
- https://pypi.org/project/django-sql-utils/
- https://docs.djangoproject.com/en/4.0/ref/models/expressions/
- https://stackoverflow.com/questions/42543978/django-1-11-annotating-a-subquery-aggregate

matmair:
@SchrodingersGat that is lean. Nice query!

PR #2841, opened 2022-04-19
X-forwarded-for

SchrodingersGat:
Forward client IP address through so that the "session" tracking makes sense

Currently the session IP tracks the proxy / load balancer

Ref: https://django-user-sessions.readthedocs.io/en/stable/installation.html#ip-when-behind-a-proxy

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2841"></a>


matmair (approved):
Should be good to go; The dependency is a bit old - I could not find a replacement with a fast search though

PR #2842, opened 2022-04-19
Fixed vulnerability in jquery-ui@1.12.1

awiswasi:
Affected versions of this package are vulnerable to Cross-site Scripting (XSS) when accepting the value of altField option of the Datepicker widget from untrusted sources, which may lead to execution of untrusted code. Vulnerability fixed in jquery-ui@1.13.0

### DIFF
--- a/InvenTree/InvenTree/static/script/jquery-ui/package.json
+++ b/InvenTree/InvenTree/static/script/jquery-ui/package.json
@@ -2,7 +2,7 @@
 	"name": "jquery-ui",
 	"title": "jQuery UI",
 	"description": "A curated set of user interface interactions, effects, widgets, and themes built on top of the jQuery JavaScript Library.",
-	"version": "1.12.1",
+	"version": "1.13.0",
 	"homepage": "http://jqueryui.com",
 	"author": {
 		"name": "jQuery Foundation and other contributors",


matmair:
@SchrodingersGat I would suggest a backport to 0.6.x

SchrodingersGat:
The edited file is not used as a manifest here, instead we have been (probably incorrectly) statically adding the js libs directly into the source code. So unless I'm mistaken, this change will not actually alter any of the JavaScript code?

@matmair you recently raised a valid point that we should be installing the JavaScript libs dynamically from a requirements file. Might be time to implement that.

matmair:
@SchrodingersGat I actually started with that but deemed it not important - I will search for the POC branch.

SchrodingersGat:
@awiswasi to fix this now, we will need to update all the js files for the entire package, not just the version number in `package.json`

We will (in a separate PR) work on making a proper dependency setup, but for now you'll have to overwrite the existing package files with the newer version (manually). Please do this and push the changes against this PR.

awiswasi:
Will do! I'm on it right now.

SchrodingersGat:
@awiswasi thanks for submitting this fix :)

PR #2844, opened 2022-04-19
WIP: Dev vendor files

matmair:
Example how files could be vendored in

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2844"></a>


matmair:
@SchrodingersGat the idea would be to use webpack to collect all the modules in the CI and package them together and remove the vendored files from the repo enitrely.

SchrodingersGat:
@matmair my understanding is that it would be similar to how the python packages are installed from `requirements.txt` - is this roughly accurate?

matmair:
@SchrodingersGat yes. We would just un npm install to gather all the libraries and then package them together into a bundle (single file) for easy redistribution. That gets placed into static and is shipped with the code.

PR #2849, opened 2022-04-20
Update requirements.txt

awiswasi:
Update Django to prevent a possible vulnerability to SQL injections, which was fixed in v4.0.4 that was released just 9 days ago. Also updating django-debug-toolbar to 3.2.1.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2849"></a>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # Please keep this list sorted
-Django==4.0.4                   # Django package
+Django==3.2.13                  # Django package
 bleach==4.1.0                   # HTML santization
 certifi                         # Certifi is (most likely) installed through one of the requirements above
 coreapi==2.3.0                  # API documentation
@@ -12,7 +12,7 @@ django-allauth-2fa==0.8         # MFA / 2FA
 django-cleanup==5.1.0           # Manage deletion of old / unused uploaded files
 django-cors-headers==3.2.0      # CORS headers extension for DRF
 django-crispy-forms==1.11.2     # Form helpers
-django-debug-toolbar==3.2.1     # Debug / profiling toolbar
+django-debug-toolbar==3.2.4     # Debug / profiling toolbar
 django-error-report==0.2.0      # Error report viewer for the admin interface
 django-filter==2.4.0            # Extended filtering options
 django-formtools==2.3           # Form wizard tools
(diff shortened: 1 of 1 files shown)


eeintech:
> Please fix conflicts with latest master :)

What are the conflicts? Seems like only the `requirements.txt` file has changed.
Is it possible to allow CI to run?

SchrodingersGat:
@awiswasi it appears that these critical bug fixes have been backported to `3.2.13`:

https://docs.djangoproject.com/en/4.0/releases/3.2.13/

So, rather than having to deal with the headache of updating to django-v4, please adjust this PR to only update `django==3.2.13`

eeintech:
Also the latest version of `django-debug-toolbar` is `3.2.4` and not `3.2.1` as pointed to in the updated `requirements.txt` file

matmair (changes requested):
This would be a major version change in the underlying framework - this needs to be checked by all core-devs before any work goes ahead!

matmair:
Ah I see everybody was here already.

@awiswasi thank you for looking out for security concerns!
Django is the framework everything builds on and follows semver - therefore there are a lot of breaking changes /depreciations associated with this PR. I can understand the suggested change but would very much prefer to switch to the patched version instead of a new major release.

awiswasi:
> @awiswasi it appears that these critical bug fixes have been backported to `3.2.13`:
> 
> https://docs.djangoproject.com/en/4.0/releases/3.2.13/
> 
> So, rather than having to deal with the headache of updating to django-v4, please adjust this PR to only update `django==3.2.13`

You're right. I missed that last night and just realized there was no need for v4.0.4 so I apologize.
I'm pretty new to version control and to be fair, I have only ever made a few contributions before as I am still learning, so don't mind me if I miss something haha. I'm a C++ developer but I've been learning Python for the past 4 years and I'm just getting into security and code optimization!

With that being said, my take on this is that we are going to go with `django@v3.2.13` and `django-debug-toolbar@v3.2.4` correct?

SchrodingersGat:
> With that being said, my take on this is that we are going to go with django@v3.2.13 and django-debug-toolbar@v3.2.4 correct?

Yep, this is correct thanks

awiswasi (commented):
django@v3.2.13 and django-debug-toolbar@v3.2.4

SchrodingersGat:
There are still conflicts as the requirements file has recently been changed. Please rebase this commit from the latest master

awiswasi:
I tried to rebase but I got "Updates were rejected because the remote contains work that you do not have locally."
Wasn't `requirements.txt` changed correctly? Sorry if I'm missing something

SchrodingersGat:
You will need to pull down the changes from the remote master/main branch before pushing back upstream

PR #2850, opened 2022-04-20
Docker refactor

SchrodingersGat:
Major refactor for the docker-compose scripts, which greatly simplifies docker setup, both for the production and development environments.

Docs: https://github.com/inventree/inventree-docs/pull/267

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2850"></a>

@matmair the biggest change here is that the user no longer has to change *anything* in the docker-compose files, it can now be done entirely from the `.env` file. It is a much simpler / cleaner setup process (and the docs are much easier to follow now too!)


matmair (approved):
Much simpler. We should takle static pwd for pgsql one day thought

SchrodingersGat:
> Much simpler. We should takle static pwd for pgsql one day thought

static pwd?

matmair:
We are setting a default database password in the .env - I recon a lot of ppl will not change that

SchrodingersGat:
Oh of course. We could make it blank, and throw an error if it is not set?
That way interaction is required by the user!

For the development server it may be easy to leave it unchanged but not for production.

PR #2852, opened 2022-04-20
Add sqlite-config file back in

SchrodingersGat:
- fixes CI bug introduced by #2850 
- required for CI steps
- might refactor this out later on


matmair:
@schrodingergat we could maybe change the location so the file is not used. Maybe to the workflow folder.

PR #2861, opened 2022-04-23
[FR] White labeling

### LINKED ISSUE #2301: [FR] White labeling
**Problem**
A lot of companies do not want it to be known which software they use. Either for not letting their competitor know what they use or for security reasons.

**Suggested solution**
Removing all tags and graphics that identify InvenTree would be a great option.
Also hiding normal debug and version info from users should be an option.

This is nice to have.

**Examples of other systems**
Most 'big vendor' solutions come with a very reduced custom login page (that they charge for like they invented HTML). That should be enough.

Yes

matmair:
This PR adds more customization options to Inventree:
- Add a message to the login page
- Add an info banner to the main page
- Exchange the logo
- Remove the InvenTree name from all user-facing places

Fixes #2301

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2861"></a>


matmair:
@SchrodingersGat this will require slight changes to the demo instance: the demo texts were removed from the code-base and moved to the config template. Please just remove the comments on the new lines there and everything should be back as before

InvenTree/InvenTree/settings.py
```diff
@@ -912,3 +913,21 @@ def _is_true(x):
...
 PLUGIN_FILE_CHECKED = False                    # Was the plugin file checked?
+
+# user interface customization values
+CUSTOMIZE = get_setting(
+    'INVENTREE_CUSTOMIZE',
+    CONFIG.get('customize', {}),
+    {}
+)
+
+CUSTOM_LOGO = get_setting(
```
SchrodingersGat, line 924: This option will (by default) throw a warning, right? as 'INVENTREE_LOGO' will not exist in the media dir.

Should the default value be an empty string (''), or the line commented out?

SchrodingersGat:
Looks pretty reasonable, I'll pull it down and have a closer look.

InvenTree/InvenTree/settings.py
matmair: Not exactly. As custom logo is false by default it is not checked. So everything keeps functioning as it used to be.

matmair: @SchrodingersGat can I mark this as resolved?

SchrodingersGat:
### Navbar Message

Navbar message alignment can be tweaked a bit, should match the current demo instance:

image

image

InvenTree/InvenTree/settings.py
```diff
@@ -912,3 +913,21 @@ def _is_true(x):
...
+    'INVENTREE_LOGO',
+    CONFIG.get('logo', False),
```
SchrodingersGat, line 926: I would suggest that the custom logo should be specified in the "customize" section of the config file, making this line

```python
CUSTOM_LOGO = get_setting(
    'INVENTREE_CUSTOM_LOGO',
    CUSTOMIZE.get('logo', False)
)
```

Then, the `config.yaml` file could look like:

```yaml
customize:
  navbar_message: My custom navbar message
  login_message: My custom login message
  logo: media/my_custom_logo.png
```

matmair:
> ### Navbar Message
> 
> Navbar message alignment can be tweaked a bit, should match the current demo instance, centred vertically and horizontally
> 
> image
> 
> image

I have changed the template example to use h6 tags so that the size fits the current method. The result between the demo and this PR are the same now.

SchrodingersGat:
@matmair this looks great now, thanks for the updates. Can you add some documentation when you have the time?

matmair:
> @matmair this looks great now, thanks for the updates. Can you add some documentation when you have the time?

@SchrodingersGat PR is already in

PR #2863, opened 2022-04-24
WIP: use storage API

### LINKED ISSUE #2585: [BUG] Inventree fiddles with files directly rather than using Django Storage api

matmair:
WIP!!!
This PR changes the codebase to use the storage API for writing and reading files. This is needed for portable installs, RO deployments etc.

Fixes #2585


matmair:
@nws would be happy for a quick check

matmair:
There is still something missing: secret key storage. I am not sure if it is safe to put that onto media as that might as well be S3

matmair:
This was just PoC to start a discussion; without an active user for reviews / code submissions this will go nowhere

PR #2865, opened 2022-04-24
Implement proxy based auth

### LINKED ISSUE #1693: Feature request: reverse proxy login support
I have traefik running as a reverse proxy in front of inventree.   The reverse proxy handles the AUTHENTICATION (and maybe AUTHORIZATION) before handing it off to any server running behind it.

In this case I have `https://inventree.{mydomain}.com`  pointing to the reverse proxy.  The reverse proxy authenticates with google, and then sets a cookie for auth.   The proxy then checks the presence of a valid auth cookie before handing the traffic off to inventree running in a docker container.  (This is all done with traefik, btw) 

In this scenario I have two login pages.  First for the google login page, and then the second for the inventree login page.  

So the reverse proxy will add a header to the request, in my case "X-Forwarded-User: me@gmail.com".  

It would be nice to have something along the lines in the config:

INVENTREE_REVERSE_PROXY_LOGIN=True
INVENTREE_REVERSE_PROXY_HEADER="X-Forwarded-User"

This would allow easier login for people/corporations that already have a reverse proxy system setup using LDAP/SSO/etc.  And Inventree wouldn't have to deal with complex auth mechanisms directly at all.  

Lastly Inventree would need something like this: 

INVENTREE_REVERSE_PROXY_AUTO_CREATE_USER=True/False

If True, an unknown user would be granted access (because both AUTHENTICATION of the user and AUTHORIZATION of the user accessing Inventree have been handled elsewhere) and Inventree should just add the new user into the User's table.

If False, AUTHENTICATION is handled by the reverse proxy, but AUTHORIZATION of the user accessing Inventree is controlled by the Django User table.  If a user exists in table and is_active=True then the user is granted access.  Otherwise the user gets a 403 Forbidden Page.

This way you can do fine grained control of user AUTHORIZATION in Inventree, yet still rely upon external AUTHENTICATION to validate who the person is.

matmair:
This PR adds proxy-based auth by HTTP header via the Django-native AuthBackend / Middleware.

Note: This is an old but still often used (mostly in enterprise) option. Similar was possible with a few lines before - this PR adds specific settings for adaptations users could do with overriding the auth and middleware stack. By default, it is turned off so no changes to default behaviour should be noticed.
Fixes #1693 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2865"></a>

### DIFF
--- a/InvenTree/InvenTree/middleware.py
+++ b/InvenTree/InvenTree/middleware.py
@@ -2,6 +2,9 @@ from django.shortcuts import HttpResponseRedirect
 from django.urls import reverse_lazy, Resolver404
 from django.shortcuts import redirect
 from django.conf.urls import include, url
+from django.conf import settings
+from django.contrib.auth.middleware import PersistentRemoteUserMiddleware
+
 import logging
 
 from rest_framework.authtoken.models import Token
@@ -112,3 +115,16 @@ class CustomAllauthTwoFactorMiddleware(AllauthTwoFactorMiddleware):
                 super().process_request(request)
         except Resolver404:
             pass
+
+
+class InvenTreeRemoteUserMiddleware(PersistentRemoteUserMiddleware):
+    """
+    Middleware to check if HTTP-header based auth is enabled and to set it up
+    """
+    header = settings.REMOTE_LOGIN_HEADER
+
+    def process_request(self, request):
+        if not settings.REMOTE_LOGIN:
+            return
+
+        return super().process_request(request)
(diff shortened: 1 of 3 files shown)


matmair:
The original FR #1693 also called for some sort of onboarding page if the user is unknown. This would be a new concept so I did not implement it.

SchrodingersGat:
@matmair ready for review?

matmair:
@SchrodingersGat yes, ready for review!

SchrodingersGat:
@matmair brilliant, looks very clean and allows configurability without changing default user experience :)

PR #2877, opened 2022-04-26
API Mixin cleanup

matmair:
This PR adds two small changes to make the API mixin safer to use:
- make header creation safer if not all settings are set
- add option `endpoint_is_url` if the whole URL is passed

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2877"></a>


matmair:
@SchrodingersGat some small changes to make it more difficult to shoot yourself into the foot with this mixin.

PR #2893, opened 2022-04-28
Build output table improvements

SchrodingersGat:
Significant overhaul of the "build output" screen:

- Displays "completed tests" for each build output
- Major speed improvement by optimizing API queries
- Table is more "responsive" (does not reload the entire table on every action)

**Example Page Load Speed Improvement**

- Old method: Approx. 6 seconds
- New method: Approx. 3 seconds

*(Using the same dataset)*

**Screenshot**

image

**Future Improvements**

The API queries should still be optimized here, there are probably some other improvements that could be made to improve page load speed.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2893"></a>

### DIFF
--- a/InvenTree/build/api.py
+++ b/InvenTree/build/api.py
@@ -442,6 +442,18 @@ class BuildItemList(generics.ListCreateAPIView):
         if part_pk:
             queryset = queryset.filter(stock_item__part=part_pk)
 
+        # Filter by "tracked" status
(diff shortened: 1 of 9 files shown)


eeintech:
Sorry @SchrodingersGat did not have the chance to look it up, looks good in the demo!

Issue #2314, opened 2021-11-15
[FR] Attachments for manufacturer parts


sur5r:
**Problem**
I would like to be able to attach datasheets, 3d models and similar files to *manufacturer parts*.

As a *part* can reference multiple *manufacturer parts*, I would like to keep my parts as generic as possible (like "100k 1% 1/10W 0603") and link multiple *manufacturer parts* and *supplier parts* to them. Datasheets, however, are manufacturer specific, so I would like attach them to the *manufacturer part*.

**Suggested solution**
Implement an attachment feature for *manufacturer parts* like the one for *parts*.

**Describe alternatives you've considered**
Keep my inventory as 1:1:1 mapping between part/supplier part/manufacturer part. But this makes the ability to link multiple *manufacturer parts* to a *part* unused. It also complicates managing exchangable/equivalent parts.

**Do you want to develop this?**
I would love to but I doubt my Python and or Django skills are good enough, yet. Given some pointers/hints I'm willing to give it a try, tought.

eeintech:
Makes a lot of sense :+1:

matmair:
@sur5r great idea. Do you want to dev it? Would be pretty straight forward as we already have the needed functions in the code so it more or less about adapting the example. We can provide you with pointers where the different components are.
@eeintech added the manufacturer parts if I remember correctly and I am happy to provide you a code-tour.

sur5r:
Given the code-tour, I would definitely give it a try. Thank you very much for the offer. How would you want to do the tour? BBB session maybe? Also, how much time would you recommend for the code tour?

SchrodingersGat:
Hi @sur5r thanks for the suggestion, it does make a lot of sense to add this feature

A brief rundown of what you'd need to do to implement this is below. Let us know if anything is missing or you need further pointers :)

The code base is getting pretty large now, so don't feel bad if it initially feels overwhelming!

## Branch Code

All features need to be on a separate code branch. Fork this repo, create a new feature branch.

## New Table / Model

You will need to create a new database table for storing attachments against a ManufacturerPart. This is actually pretty straightforward.

### Base Model

The base model which represents an attachment can be found here:

https://github.com/inventree/InvenTree/blob/4e09f0110b5d3ba280370a649406e9d2c3af74b4/InvenTree/InvenTree/models.py#L89

The key thing to note here is that the class is marked as "abstract" (line 195) which means that it does not represent a database table itself, but can be used as a starting point to define actual tables.

### Create new Class

[...]

PR #2912, opened 2022-05-01
Implement Attachments for manufacturer parts

sur5r:
Implements #2314

Open questions:
* Which kind of tests should be added for this?
* Squash beforce merge?

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2912"></a>


InvenTree/company/api.py
```diff
@@ -163,6 +164,32 @@ class ManufacturerPartDetail(generics.RetrieveUpdateDestroyAPIView):
     serializer_class = ManufacturerPartSerializer
 
 
+class ManufacturerPartAttachmentList(generics.ListCreateAPIView, AttachmentMixin):
```
matmair, line 167: @sur5r are you sure about this order? Mixins should be placed before the base view normally - could be wrong here too.

InvenTree/company/api.py
```diff
@@ -163,6 +164,32 @@ class ManufacturerPartDetail(generics.RetrieveUpdateDestroyAPIView):
...
+    filter_backends = [
+        DjangoFilterBackend,
+    ]
+
+    filter_fields = [
+        'manufacturer_part',
+    ]
+
+
+class ManufacturerPartAttachmentDetail(generics.RetrieveUpdateDestroyAPIView, AttachmentMixin):
```
matmair, line 184: Same as above

matmair (changes requested):
Seems to be good except for the comments

SchrodingersGat:
@sur5r this looks great, happy with the implementation :)

I would suggest adding some simple tests to https://github.com/inventree/InvenTree/blob/master/InvenTree/company/test_api.py - validating that the API functionality works checks everything top to bottom

- Retrieve a list of manfuacturer part attachments via the API
- Filter this list by manufacturer-part-id
- Update the attachment (comment field, etc)
- Upload a new attachment via the API

SchrodingersGat:
Also please merge in latest master :)

InvenTree/company/api.py
sur5r: No I'm not. I took `PartAttachmentList` and `PartAttachmentDetail` as an example.

I just swapped the order and noticed the uploading user was not saved with the old order but is with the new order.
So I guess you are right.

But that would hint at the order in `PartAttachment*` being wrong as well. And I just check and found the user field to be empty after uploading a (normal) part attachment.

matmair: @sur5r that is .... interesting. Do not know how something like this can happen - Mixins are one of the core concepts of the whole framework. Seems like we slept a bit on validation in the past.
Thank you for checking - I  suggest you change the order just in your new code and we fix the others in a separate PR that gets merged ASAP.

SchrodingersGat:
@sur5r I'd like to get this merged in, is it ready to go? Please mark as "ready for review" and @ me

SchrodingersGat:
Also, I forgot to mention, as we have added new functionality to the API, please increment the API version in [this file](https://github.com/inventree/InvenTree/blob/master/InvenTree/InvenTree/api_version.py) with a reference back to this PR and a brief description of what has changed (follow the convention in the file).

SchrodingersGat:
@sur5r any progress on this?

sur5r:
I haven't found time to implement the tests yet. Other than that, it seems to work fine so far. If you want to merge this now I can open a new PR for the tests.

SchrodingersGat:
@sur5r thanks, please make the api-bump fix and we can push the tests off to a new PR

SchrodingersGat (approved):

SchrodingersGat:
@sur5r thanks :) please do look into unit tests as soon as you can

PR #2920, opened 2022-05-02
Fix mixin loading order

matmair:
As brought up in #2912

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2920"></a>


matmair:
@SchrodingersGat no idea why the order of the mixin was that way

PR #2922, opened 2022-05-02
Clean up tags

matmair:
I would like to propose changing the labels into areas as documented in the PR and removing the following labels:
- email
- translation
- javascript
- python

The proposed areas are:
- feature -> features of InvenTree - mostly as the apps are structured
- type -> what kind of issues / PR is it
- state -> why was the issue / PR closed
- ecosystem -> not really regarding the code itself
- GH built-in -> GH tags (that are often used for search - so no change there)

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2922"></a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ The HEAD of the "main" or "master" branch of InvenTree represents the current "l
 
 **No pushing to master:** New featues must be submitted as a pull request from a separate branch (one branch per feature).
 
-#### Feature Branches
+### Feature Branches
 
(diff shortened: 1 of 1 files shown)


matmair:
@SchrodingersGat @eeintech

SchrodingersGat:
Seems reasonable, I'm happy with your suggested tags

SchrodingersGat:
Ah, just realized that there are a number of "tbd" strings in here.. might have merged this a bit too hastily :)

SchrodingersGat:
@matmair have updated the tags, and filled out the "tbd" strings

PR #2926, opened 2022-05-03
Use the tablename when generating automatic model events

SchrodingersGat:
Currently when a model instance is created / updated / deleted, an "instance.created" / "instance.saved" / "instance.deleted" event is triggered, which can be captured by a plugin and processed.

This PR changes this behaviour to use the name of the table as the event name, for better filtering by plugins.

e.g. creating a new stock item will generate a `stock_stockitem.created` event

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2926"></a>

### DIFF
--- a/InvenTree/plugin/events.py
+++ b/InvenTree/plugin/events.py
@@ -163,17 +163,15 @@ def after_save(sender, instance, created, **kwargs):
 
     if created:
         trigger_event(
-            'instance.created',
+            f'{table}.created',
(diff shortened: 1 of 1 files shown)


matmair:
@SchrodingersGat I would find it usefull if there was a kwarg for the action like action =  'instance.created' or 'created'.

matmair (approved):

PR #2937, opened 2022-05-06
[Feature] UI Plugins

SchrodingersGat:
This PR represents a first step towards integration of plugins directly into the web user interface.

Starting with the "Panel" mixin class, plugins can register custom 'panels' to be displayed on certain pages. However the framework is now in place for adding in other locations in the user interface where plugins could be *tastefully* integrated.

### TODO

- [x] Update documentation with information on new mixin class - https://github.com/inventree/inventree-docs/pull/276

### Panel Mixin

The "Panel Mixin" allows custom panels to be added to any page that has a "sidebar" menu. Each panel offers:

- Custom title
- Custom icon
- Rendered content
- Custom javascript code which runs when the panel is first displayed

### Screenshots

**Panel which is added to *every* page with a sidebar**

image

**Panel which only displays on the "Part" page**

image

**Panel which displays on a childless StockLocation page**

image

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2937"></a>


matmair (approved):

matmair:
@schrodingersgat I think we should provide a single import path (as with plugin.mixin) for all customizable views (maybe a plugin.view) to keep plugin devs sane in case we refactor. Else we are starting to cement the internal code structure because of the plugin interface. The same goes for models.

SchrodingersGat:
@matmair the example plugin now allows panel content to be rendered via a template or passed in directly. Please let me know if you think the template loading should be done differently? It feels pretty clean this way

InvenTree/build/views.py
```diff
@@ -29,7 +31,7 @@ def get_queryset(self):
 
     def get_context_data(self, **kwargs):
 
-        context = super(BuildIndex, self).get_context_data(**kwargs).copy()
```
matmair, line 32: Are you sure we do not need that? This was the object stays coupled

InvenTree/plugin/template.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+
+    def get_dirs(self):
+        dirname = 'templates'
+        template_dirs = []
+
+        for plugin in registry.plugins.values():
+            new_path = Path(plugin.path) / dirname
+            if Path(new_path).is_dir():
+                template_dirs.append(new_path)  # pragma: no cover
```
matmair, line 38: This can be removed now

matmair: I mean the no cover

InvenTree/build/views.py
SchrodingersGat: Do you mean the `super(BuildIndex, self)` call? It's now just `super()` (new style syntax) and the [MRO](https://realpython.com/python-super/#method-resolution-order) should sort it out.

InvenTree/plugin/template.py
```diff
@@ -35,7 +35,7 @@ def get_dirs(self):
         for plugin in registry.plugins.values():
             new_path = Path(plugin.path) / dirname
             if Path(new_path).is_dir():
-                template_dirs.append(new_path)  # pragma: no cover
+                template_dirs.append(new_path)
```
matmair, line 38: The more coverage the better! Nice that we now have an actual usage for this

InvenTree/build/views.py
matmair: @SchrodingersGat I mean the .copy()

SchrodingersGat: I see. If you search for `super().get_context_data` it does not seem to be consistent.

The [django docs](https://docs.djangoproject.com/en/4.0/topics/class-based-views/generic-display/#adding-extra-context-1) seem to indicate that the `copy()` is not required.

Should we remove all `.copy()` calls?

matmair: @SchrodingersGat I never use copy() with context as the object is created on each new request. But it should be consistent throughout the codebase.

matmair: This can be a separate issue too. The rest of the PR seems fine and a good improvment. The change just jumped out.

Issue #2925, opened 2022-05-03
[FR] Settings API - use KEY as URL identifier


SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/2805

Currently, we use the PK (primary key) of the setting instance to determine the API access URL

e.g. `/api/settings/1/`

However, there are a number of issues here:

- No way of accessing a setting object without knowing the PK value
- Settings get 'instantiated' on request, so we can't even read out the "default" value *until* it has a pk

### Suggestion

We should change the setting access API to use the unique KEY of the setting as the lookup URL, e.g.

- `/api/setting/ENABLE_REPORT/`
- `/api/setting/user/SEARCH_RESULTS_COUNT/`

This has some other advantages:

- Makes writing front-end access code (e.g. js / templates) much easier
- We can "create" a setting on request, if it does not exist in the database
- We can return an error code if an invalid KEY is requested

matmair:
@SchrodingersGat sounds like a good idea!

PR #2944, opened 2022-05-08
Settings by key

SchrodingersGat:
This PR changes the API interaction for settings, allowing lookup by unique key, rather than database ID

Closes https://github.com/inventree/InvenTree/issues/2925

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2944"></a>

### DIFF
--- a/InvenTree/common/api.py
+++ b/InvenTree/common/api.py
@@ -171,7 +171,7 @@ class GlobalSettingsDetail(generics.RetrieveUpdateAPIView):
 
         if key not in common.models.InvenTreeSetting.SETTINGS.keys():
             raise NotFound()
-        
+
         return common.models.InvenTreeSetting.get_setting_object(key)
 
     permission_classes = [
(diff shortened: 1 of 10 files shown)


InvenTree/common/api.py
```diff
@@ -158,9 +158,22 @@ class GlobalSettingsDetail(generics.RetrieveUpdateAPIView):
     - User must have 'staff' status to view / edit
     """
 
+    lookup_field = 'key'
     queryset = common.models.InvenTreeSetting.objects.all()
     serializer_class = common.serializers.GlobalSettingsSerializer
 
+    def get_object(self):
```
matmair, line 165: Can't we move this function into a mixin so the code only exists once?

InvenTree/common/api.py
```diff
@@ -213,9 +226,22 @@ class UserSettingsDetail(generics.RetrieveUpdateAPIView):
     - User can only view / edit settings their own settings objects
     """
 
+    lookup_field = 'key'
     queryset = common.models.InvenTreeUserSetting.objects.all()
     serializer_class = common.serializers.UserSettingsSerializer
 
+    def get_object(self):
```
matmair, line 233: It is basically the same code

matmair (approved):
Should we maybe move the API-endpoint for plugin settings into the common urls?

matmair (approved):

PR #2947, opened 2022-05-08
[Feature] Add setups / OOBE

### LINKED ISSUE #2284: [FR] settings setup guide
**Problem**
New installs can be configured a bunch - some stuff (like base url, name etc.) should really be set.

**Suggested solution**
- [ ] A setup modal if it is the first run
- [ ] Multipage wizard that guides through the experience
- [ ] Mixin that enables plugins to show setup page
- [ ] API endpoint to check if setups ran -> could be important for 3rd party systems
- [ ] Maybe a role for users/admins to specify who gets shown setup screens

Make a whole setup guide in the docs - I do not think that would be as effective.

**Examples of other systems**
OctoPrint has a great setup workflow

The structure yes, I would like to discuss the details of what to show in the core contributer group

matmair:
This PR implements setups as first class citizens. This required for the OOBE and setups in general.

Closes #2284

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2947"></a>


matmair:
@SchrodingersGat would you mind taking a look at the syntax for the setup files in this [sample file](https://github.com/inventree/InvenTree/blob/f461721259e89fb1321019c9026c8415fce713cd/InvenTree/oobe/setups/initial.yml)?

Not all functions are done rn; Namely missing is the data action at the end.
There are a few built-in limitations:
- no i18n included - or planned
- only a few specified inputs can be used
- no cross-session state saving - this is only implemented in-session as the whole framework also works without a database (initial setup for example)

SchrodingersGat:
@matmair I'm away for the weekend, will jump on this first thing next week :)

InvenTree/InvenTree/urls.py
```diff
@@ -152,6 +157,8 @@
     re_path(r'^stock/', include(stock_urls)),
     re_path(r'^supplier-part/', include(supplier_part_urls)),
 
+    re_path(r'^setup/', include(oobe_urls)),
```
SchrodingersGat, line 160: The api URL is `setups`, this one is `setup` - should we make this consistent?

matmair:
Missing:
- [ ] finishing action (yaml)
- [ ] next setup options (yaml)
- [ ] issue event after setup
- [ ] `SetupRun` model
- [ ] `first startup` flag
- [ ] settings field (forms)
- [ ] built-in setups for config file and use-case

Issue #2845, opened 2022-04-19
[FR] Identification Plugin System


SchrodingersGat:
Add a new plugin system for "identifiying" or "locating" components. Could be used from the web or the app.

This has been implemented (custom) with great effect already, it would be fantastic to add proper support for this via the plugin ecosystem.

### References

- https://github.com/inventree/InvenTree/discussions/2678
- https://www.reddit.com/r/raspberry_pi/comments/t0c99r/raspberry_pi_cm4_smart_inventory_management/

matmair:
@schrodingersgat how do you plan to do this? I would suggest adding a model that holds an identifier (token, address etc) and can be assigned to (multiple) stock location(s).
If you identify a stock item via ui or api a simple lookup is performed and all required identifiers are handed over to a plugin function. The plugin then needs to handle the rest.
I would love to write a small demo with an Adafruit MagTag and neopixels once you get the code done.

SchrodingersGat:
The plan is to rely on the plugin as much as possible - much like the way the "label printing" plugin works. Simply pass the StockLocation object to the plugin, and leave it at that. The plugin could call an external network API or something?

matmair:
@SchrodingersGat I would do the association / datahandeling in InvenTree to make that portable (change the illumination plugin for example / use multiple plugins at the same time (one system for the workflow and another for R&D. The association between location and identifier could be handled in the admin.

PR #2957, opened 2022-05-09
Adds plugin mixin to "locate" items

SchrodingersGat:
This PR adds yet another plugin mixin class, this time to allow "location" of stock items / locations.

Ref: https://github.com/inventree/InvenTree/discussions/2678

Closes https://github.com/inventree/InvenTree/issues/2845

App Integration: https://github.com/inventree/inventree-app/pull/124

This would allow, as a first class feature, development of external plugins to identify / locate / retrieve stock items or locations.

For example, flashing an LED at the location, or retrieving the stock item from a robotic storage system.

This PR also allows native integration of this feature with the app!

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2957"></a>

### Web Integration

image

### App Integration

image

### TODO

- [x] Add JSON field to StockLocation model for use by plugins
- [ ] Documentation


SchrodingersGat:
What do you mean by a "position reference" here?

matmair (approved):

SchrodingersGat:
@matmair yeah, that makes total sense. 

So, add a "reference" field to the `StockLocation` model? Could be a JSON field, perhaps?

matmair:
@SchrodingersGat yes that would be good. Maybe referenced to the plugin slug.

matmair (approved):

matmair:
@SchrodingersGat this will need some updates with the plugin refactor. Ping me if you want pointers / a PR

SchrodingersGat:
@matmair now that https://github.com/inventree/InvenTree/pull/2971 has merged I've got some conflicts to fix, working on that now.

Regarding the custom field for plugin access - I was thinking a "metadata" JSON field on the `StockLocation` model (which would not be used be internal application logic) could be provided solely for plugins to store custom data against a location. However, this would mean that each plugin would have to take care not to "clash" with whatever data it stores in that field.

Alternatively, we could have a new table which associates metadata between plugins and other tables, e.g. stock location. Then we'd want to expand this to other internal tables, such as the "part" table or "Company" table.

I can see this getting out of hand pretty quickly, and the simpler approach would be to have a single JSON field per location / per part / per customer / etc.

What are your thoughts on this one?

matmair:
@SchrodingersGat I would recommend moving the new API endpoints and their tests to /plugin/api.py

The JSON field seems to be much simpler. If we provide a helper function to access the data that uses the plugin slug it might help with collisions.

InvenTree/plugin/samples/integration/locate_sample.py
```diff
@@ -6,18 +6,22 @@
 
 import logging
```
matmair, line 7: I would recommend moving this to InvenTree/plugin/samples/locate/locate_sample.py

PR #2959, opened 2022-05-09
Fix usage of old html tags

matmair:
This PR fixes the usage of old HTML tags in a sample template.


SchrodingersGat:
You don't let me get away with anything ;)

matmair:
@SchrodingersGat I would not mind but my CI checks for W3C compliance and fails till upstream is fixed

SchrodingersGat:
Should we add that to our CI here?

matmair:
@SchrodingersGat If you like never passing CIs - sure. TBH I would not recommend but I need it for compliance reasons.

PR #2966, opened 2022-05-10
Add code128 template

matmair:


SchrodingersGat:
It'd be a good idea to add a unit test which renders this template, just to check it renders without error

SchrodingersGat:
Should we rename the barcode directory, or place it within a subdirectory of InvenTree? What other options do we have for resolving namespace issues?

matmair:
@SchrodingersGat I moved the barcode stuff all into the plugin app. A bit of refactoring needs to be done but it should be better this way.

matmair:
@SchrodingersGat could you review this and manually trigger postgres / mysql tests?

SchrodingersGat:
I can't manually trigger those until it's merged, but looks good, thanks for the quick fix

PR #2971, opened 2022-05-11
Plugin app refactor

matmair:
Heavy WIP!

This PR will refactor the internal python API for the plugin app - imports should stay the same if the general imports are used:
https://inventree.readthedocs.io/en/latest/extend/how_to_plugin/#development-guidelines

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2971"></a>


matmair:
IntegrationPluginBase is renamed to InvenTreePlugin -> there is a class to bridge the conversion - a deprecation warning is added

matmair:
@SchrodingersGat I would like to change the following tags:
`    PLUGIN_NAME = ''

    PLUGIN_SLUG = None

    PLUGIN_TITLE = None`

to remove `PLUGIN`

SchrodingersGat:
> @SchrodingersGat I would like to change the following tags: ` PLUGIN_NAME = ''
> 
> ```
> PLUGIN_SLUG = None
> 
> PLUGIN_TITLE = None`
> ```
> 
> to remove `PLUGIN` -> should be provide a depreciation warning for the old ones?

Is this adding something or just a personal preference? If it requires some extra work to deprecate, I'd vote to leave them as they are

matmair:
@SchrodingersGat these 3 attributes are the only ones carrying the prefix. Everything else is without.

matmair:
@SchrodingersGat this would be ready for merge. The internal code for plugins is now more concentrated in /plugin. The functions and tests for plugins are placed into their own subdirectories:
grafik
Builtin plugins (including actions and barcodes) are in the builtin dir:
grafik
And samples are separated the same way:
grafik
this way it is easier to find tests for the samples and imports are cleaner
grafik
The whole app is much cleaner that way and new big plugins can be added in a cleaner way that is not exploding the main /plugin dir

SchrodingersGat (approved):

SchrodingersGat:
@matmair a big change but I think it makes the plugin structure much better. Ready to merge, or more work to do here?

matmair:
@SchrodingersGat is ready to merge.
I will create some additional tests for plugins soon-ish. I somewhat feel like the code owner for users and plugins - so 100% coverage is the goal.

PR #2972, opened 2022-05-11
Do not try to create a new admin user if one already exists

SchrodingersGat:
@matmair a simple check to prevent an error being thrown if the admin user already exists

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2972"></a>

### DIFF
--- a/InvenTree/InvenTree/apps.py
+++ b/InvenTree/InvenTree/apps.py
@@ -190,7 +190,8 @@ class InvenTreeConfig(AppConfig):
         user = get_user_model()
         try:
             with transaction.atomic():
-                new_user = user.objects.create_superuser(add_user, add_email, add_password)
+                if not user.objects.filter(username=add_user).exists():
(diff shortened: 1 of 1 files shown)


InvenTree/InvenTree/apps.py
```diff
@@ -190,7 +190,8 @@ def add_user_on_startup(self):
         user = get_user_model()
         try:
             with transaction.atomic():
-                new_user = user.objects.create_superuser(add_user, add_email, add_password)
+                if not user.objects.filter(username=add_user).exists():
+                    new_user = user.objects.create_superuser(add_user, add_email, add_password)
             logger.info(f'User {str(new_user)} was created!')
         except IntegrityError as _e:
```
matmair, line 198: shouldn't this integrity error trigger if the user already exsists?

matmair (approved):

PR #2976, opened 2022-05-12
Do not redirect requests for media / static / api / js files

SchrodingersGat:
- For these paths, just return a 401
- This is necessary to stop unauthorized calls to the API or to request media files from redirecting to the login page

Recent testing for the inventree python bindings revealed some shortcomings, in particular trying to download a file (when no authenticated) returned a 200 status and redirected to the login page.

A redirect should only happen for web views which are being loaded by an actual user.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2976"></a>

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -157,7 +157,7 @@ jobs:
         run: |
           cd ${{ env.wrapper_name }}
           invoke check-server
-          coverage run -m unittest -s test/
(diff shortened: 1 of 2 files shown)


matmair (approved):

matmair:
Maybe the /auth endpoint should receive some attention too. Every Authenticator user gets access to all media rn if there are not extra steps taken by the admin.

PR #2982, opened 2022-05-12
Run all checks on PRs

SchrodingersGat:
Postgres and mysql checks are currently failing on push. 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2982"></a>

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -207,7 +207,7 @@ jobs:
     name: Postgres
     needs: ['javascript', 'html']
     runs-on: ubuntu-latest
-    if: github.event_name == 'push'
+    #if: github.event_name == 'push'
 
     env:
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@matmair as soon as we disable these checks, we actually needed them. I think this is related to your recent changes to the label templates, any idea what is going on here?

matmair:
@SchrodingersGat I think we have some problem with files that are generated and tests. I can’t get multiple PRs running because of that. Maybe we yank this test and create a collection of failing tests that we need to fix?

SchrodingersGat:
Yeah happy with that approach. I'm away from my computer for a few days so if you want to submit a PR which does that, it'd be great.

I think we also should run the `invoke test` step as part of the docker CI build, before pushing. We currently test bare metal but not within a docker context

PR #2984, opened 2022-05-12
Unit tests for auth stack

### LINKED ISSUE #2980: Add unittests for auth stack
Add full coverage for https://github.com/inventree/InvenTree/pull/2976
And the full auth stack in the middleware

matmair:
This PR:
- adds tests for the auth stack
- removes dead code in the auth stack

Closes #2980 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2984"></a>

### DIFF
--- a/InvenTree/InvenTree/middleware.py
+++ b/InvenTree/InvenTree/middleware.py
@@ -71,10 +71,6 @@ class AuthRequiredMiddleware(object):
 
             # No authorization was found for the request
             if not authorized:
-                # A logout request will redirect the user to the login screen
-                if request.path_info == reverse_lazy('account_logout'):
-                    return HttpResponseRedirect(reverse_lazy('account_login'))
-
                 path = request.path_info
 
                 # List of URL endpoints we *do not* want to redirect to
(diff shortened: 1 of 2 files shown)


InvenTree/InvenTree/middleware.py
```diff
@@ -71,10 +70,6 @@ def __call__(self, request):
 
             # No authorization was found for the request
             if not authorized:
-                # A logout request will redirect the user to the login screen
```
SchrodingersGat, line 74: Is this no longer required?

SchrodingersGat (approved):

SchrodingersGat:
These tests look good. Should we add some for `/media` files too?

InvenTree/InvenTree/middleware.py
matmair: @schrodingergat this will never trigger as all /account paths are excepted. And after logout you are always redirected to /index which sends you driectly to /account/login if you are not authenticated.

matmair:
@SchrodingersGat i had to remove tests for /media and /static as they block on postgres and mysql. I will try to add them in when I get the whole system running with files on the machine.

SchrodingersGat:
Fair enough, happy to add this as-is then.

PR #2942, opened 2022-05-07
Add Color-Theme tests

matmair:
This PR adds tests for the ColorTheme code.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2942"></a>

Issue #2989, opened 2022-05-14
[BUG] Tests with static data fail on postgres and sqlite


matmair:
@SchrodingersGat I think we have some problem with files that are generated and tests. I can’t get multiple PRs running because of that. Maybe we yank this test and create a collection of failing tests that we need to fix?

_Originally posted by @matmair in https://github.com/inventree/InvenTree/issues/2982#issuecomment-1126603940_

matmair:
See last PR in #2982

PR #2991, opened 2022-05-14
Static test fixes

matmair:
See #2989
Fixes #2942

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2991"></a>


matmair:
@SchrodingersGat any idea what is the difference between sqlite and postgres/mysql tests? I can't get my head around why static data is not accessible even though it is clearly collected on startup.

SchrodingersGat:
@matmair not sure, I have found in the past that PK values are not auto-assigned in a consistent manner when running postgres  mysql tests. They seem to be repeatable for sqlilte. Maybe it is something to do with that?

matmair:
@SchrodingersGat everything is working except on sqlite sometimes. I included a check for that. The coverage is much better this way though so I would suggest merging if you agree.

InvenTree/InvenTree/test_middleware.py
```diff
@@ -32,6 +32,10 @@ def test_AuthRequiredMiddleware(self):
         # logout
         self.client.logout()
 
+        # check that static files go through
+        # TODO @matmair reenable this check
+        # self.check_path('/static/css/inventree.css', 302)
```
SchrodingersGat, line 37: Does this still need to be re-enabled for this PR?

InvenTree/label/apps.py
```diff
@@ -42,6 +43,16 @@ def create_labels(self):
         """
         Create all default templates
         """
+        # Test if models are ready
+        try:
+            from .models import StockLocationLabel
+            obj = StockLocationLabel.objects.first()
+            print(obj)
```
SchrodingersGat, line 50: Remove this debug?

SchrodingersGat (approved):

InvenTree/InvenTree/test_middleware.py
matmair: @schrodingergat I would like to leave it in so we don't forget to fix this behaviour. Todos with my name attached stick out in my CI so I am reminded about them regularly.

InvenTree/label/apps.py
matmair: Absolutely!

matmair:
@SchrodingersGat requested changes are in.

SchrodingersGat (approved):

SchrodingersGat:
Nice work. I might have to employ some of these changes to get https://github.com/inventree/InvenTree/pull/3037 to work

PR #2994, opened 2022-05-15
Bump python version

matmair:
This PR bumps the python version for the translation workflow to fix the currently failing test.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2994"></a>

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -12,7 +12,7 @@ on:
       - l10*
 
 env:
-  python_version: 3.8
+  python_version: 3.9
   node_version: 16
 
   server_start_sleep: 60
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
If we are changing to 3.9 here should we upgrade this across the board?

matmair:
I bumped the check action as well

matmair:
@SchrodingersGat 3.8 would be fine too. 3.7 is from 2018 and needs fills for a bunch of stuff.

SchrodingersGat:
Would you agree that we should also:

- update python version in docker image
- update min required python version in the docs (for bare-metal install)

matmair:
@SchrodingersGat alpine 3.13 is containing python 3.8.8 currently - so it should be fine.  PR updating the docs is otw.

SchrodingersGat (approved):

matmair:
Hopefully, the CI now runs through.

Issue #2992, opened 2022-05-15
[BUG] Missing tests for plugins


matmair:
**Describe the bug**
There are tests missing for multiple mixins:
- events
- ~~custom panels~~ - https://github.com/inventree/InvenTree/pull/3034
- ~~locate plugin~~ - https://github.com/inventree/InvenTree/pull/3027
- printing - https://github.com/inventree/InvenTree/pull/3037

**Steps to Reproduce**
Current coverage graph

**Expected behaviour**
There are **no** functional tests - they can be broken by accident easily. Coverage at least for the functions is expected.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/66015116/168453439-03170ff8-5ab4-4080-9cc2-b3a3384b93df.png)

**Deployment Method**
- Any

**Version Information**
current master

SchrodingersGat:
@matmair are you happy with the coverage for the `events` plugin system, or would like to see more work on it before `0.7.0` release?

matmair:
@SchrodingersGat I would appreciate if we could cover all mixins and samples - but this is a soft target for me.

PR #3014, opened 2022-05-16
Add more tests for offload_task

### LINKED ISSUE #3005: Testing - Add unit test for background worker processes
https://github.com/inventree/InvenTree/pull/3004 has revealed a weakness in the unit testing processes.

We currently have no tests for functions that rely on the background worker. This has allowed a critical bug to creep into the code in the recent plugin refactor.

matmair:
This PR enables offload_task to use function references and adds more tests.
Closes #3005

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3014"></a>

### DIFF
--- a/InvenTree/plugin/base/event/events.py
+++ b/InvenTree/plugin/base/event/events.py
@@ -37,7 +37,7 @@ def trigger_event(event, *args, **kwargs):
     logger.debug(f"Event triggered: '{event}'")
 
     offload_task(
-        'plugin.events.register_event',
+        register_event,
         event,
         *args,
         **kwargs
@@ -72,7 +72,7 @@ def register_event(event, *args, **kwargs):
 
                         # Offload a separate task for each plugin
                         offload_task(
-                            'plugin.events.process_event',
+                            process_event,
                             slug,
                             event,
                             *args,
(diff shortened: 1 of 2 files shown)


matmair:
@SchrodingersGat this is ready after the current tests run through. All tests in /InvenTree/InvenTree are covered now - including the offloading task.
I strongly suggest to use direct function references for offloaded tasks - this will prevent them from breaking in refactors.

SchrodingersGat (approved):

SchrodingersGat:
Nice work @matmair - moving away from calling functions based on the string name is a great step.

There are still some naked calls to `offload_task` being handled here: https://github.com/inventree/InvenTree/blob/master/InvenTree/plugin/base/event/events.py

In addition to converting these, I think it would be useful to add some unit tests to ensure that the *events* are correctly offloaded.

e.g. create a new Part instance, check that the event was registered to a plugin?

matmair:
@SchrodingersGat there are multiple plugins that were introduced with no coverage (see #2992). My suggestion is to tackle those separately as there are more unit and e2e things missing there.
This PR is more focuses on unit and e2e tests for general tasks.

SchrodingersGat:
Fair enough, happy to tackle plugin coverage as a separate PR. Will merge this one in once checks have passed

PR #3024, opened 2022-05-18
[Feature] Shipment Creation in Stock Allocation Modal

### LINKED ISSUE #2970: [FR] Shipment Creation in Stock Allocation Menu
No

**Problem**
I find that through SO fulfillment I often have to close the Line Item allocation menu to create a new Shipment in the Pending Shipments menu. This is somewhat inconvenient when entering SOs manually.

**Suggested solution**
Add the "New Shipment" button in the "Allocate Stock Items to Sales Order" modal to allow for the creation of a shipment within the allocation menu.

Provide the option to create a shipment automatically every time an SO is created, so the Pending Shipments menu only has to be visited. This option could be added to the Sales Orders menu in Settings. (This could be done together with the solution above to streamline the SO fulfillment process).

**Examples of other systems**
I only have experience with InvenTree with regard to this aspect of fulfillment.

No, as I don't have experience working with Django.

maksimstojkovic:
closes #2970 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3024"></a>


maksimstojkovic:
Current implementation works, but some work needs to be done to hide the order field in the secondary modal.

The issue seems to be linked to the `fields.order.hidden` value being set to true, causing the field to be rendered as an input tag with id "#id_order", rather than the expected secondary modal div tag with id "#div_id_order_1".

maksimstojkovic:
Additional improvements are also necessary to update the shipment number on consecutive presses of the "Add Shipment" button. The same shipment number appears despite the successful addition of a previous shipment via the same modal.

InvenTree/templates/js/translated/order.js
```diff
@@ -2491,6 +2491,57 @@ function allocateStockToSalesOrder(order_id, line_items, options={}) {
...
                 auto_fill: true,
+                secondary: {
+                    method: 'POST',
+                    title: '{% trans "Add Shipment" %}',
+                    fields: function() {
+                        var ref = null;
+
+                        // TODO: Refactor code for getting next shipment number
+                        inventreeGet(
+                            '{% url "api-so-shipment-list" %}',
```
matmair, line 2502: @SchrodingersGat I think an API endpoint for this would be the safer option/ do we already have something that could be used for that?

matmair:
@maksimstojkovic good structure!
Could be way simpler if we provided an API endpoint for the reference, let's wait for @SchrodingersGat opinion on that. He is more knowledgeable in our frontend APIs than me.

InvenTree/templates/js/translated/order.js
```diff
@@ -2491,6 +2491,57 @@ function allocateStockToSalesOrder(order_id, line_items, options={}) {
...
+                                }
+                            }
+                        );
+
+                        var fields = salesOrderShipmentFields(options);
+
+                        fields.reference.value = ref;
+                        fields.reference.prefix = global_settings.SALESORDER_REFERENCE_PREFIX + options.reference;
+                        
+                        fields.order.hidden = false; // TODO: Hide order id field
```
maksimstojkovic, line 2540: If `fields.order.hidden` is not set to `false`, the `POST` request fails as it cannot locate a div with id `#div_id_order_1`. When `fields.order.hidden` is `true`, a `<input>` tag is created with id `id_order`, rather than the expected secondary modal `<div>` tag with id `#div_id_order_1`.

InvenTree/templates/js/translated/forms.js
```diff
@@ -1538,11 +1538,18 @@ function addSecondaryModal(field, fields, options) {
         var url = secondary.api_url || field.api_url;
 
         // If the "fields" attribute is a function, call it with data
-        if (secondary.fields instanceof Function) {
+        if (secondary.fields instanceof Function || secondary.fieldsFunction instanceof Function) {
 
             // Extract form values at time of button press
             var data = extractFormData(fields, options);
 
+            // Backup and execute fields function in sequential executions of modal
```
maksimstojkovic, line 1546: This allows the value of the next valid shipment reference to be recalculated each time the secondary button is pressed while the primary modal is open.

Further tests may be necessary to ensure this doesn't conflict with other secondary modals.

InvenTree/templates/js/translated/forms.js
```diff
@@ -1982,7 +1989,7 @@ function constructField(name, parameters, options={}) {
 
     // Hidden inputs are rendered without label / help text / etc
     if (parameters.hidden) {
-        return constructHiddenInput(name, parameters, options);
+        return constructHiddenInput(field_name, parameters, options);
```
maksimstojkovic, line 1992: This caused hidden inputs to have an incorrect ID for depth > 0

maksimstojkovic:
@SchrodingersGat @matmair This feature is now functioning as intended.

Main points for review are the process of obtaining the next available shipment number, and whether the solution for executing the `fields` function every time the secondary modal button is implemented in the most robust way possible.

I'm not sure how the review process works, so please let me know if anything further is required, or if changes are necessary.

matmair:
@maksimstojkovic looks good to me. Oliver needs to take a look and approve a run and merge but that shouldn't take too long.

SchrodingersGat:
@maksimstojkovic thanks again for this feature addition :)

PR #3027, opened 2022-05-18
Fix broken calls to offload_task

SchrodingersGat:
The 'locate' plugin was broken, calls to `offload_task` were passed the incorrect function handle.

@matmair another candidate for a testing framework for the plugins - the "sample" locate plugin here should be used to validate functionality.


matmair:
@SchrodingersGat there are some plugings that need coverage. I suggest those are the once that will break if anything changes #2992 we just can't merge in such important places without having coverage - it always comes back to bite us.

matmair:
I will try to tackle that one today or tomorrow.

SchrodingersGat:
@matmair I'm doing coverage for the `locate` plugin now :)

SchrodingersGat:
@matmair FYI using the new 'metadata' field is a nice way to mock plugin functionality for unit testing!

PR #3028, opened 2022-05-18
[Feature] Create a default shipment when creating SO

maksimstojkovic:
Related to #2970 

Creates a default shipment with reference 1 when creating a new sales order.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3028"></a>


SchrodingersGat:
@maksimstojkovic this should not be handled by the front-end. It should be performed by the database, when a new shipment is created

SchrodingersGat:
Amazing! 

> Let me know if you think True is a more intuitive default, or whether the feature should be opt-in.

I think `false` is a good default - it preserves the current behaviour

To finalize this, we'll need to add a unit test for the new functionality - in particular test that the setting changes what happens when a new SalesOrder is created.

https://github.com/inventree/InvenTree/blob/master/InvenTree/order/test_sales_order.py

SchrodingersGat (approved):

SchrodingersGat:
Nice work :) please address PEP style checks and I'll merge in.

Note that you can run style checks locally with `invoke style` to save some time

SchrodingersGat:
Thanks for the contribution @maksimstojkovic

matmair:
Congratulations on the successful front- and backend merge @maksimstojkovic !

PR #3032, opened 2022-05-18
Fix manufacturer part detail template

rkalman:
This PR fixes the manufacturer part detail view which failed to render because of a typo in the template.
Also adds missing tests for company views.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3032"></a>


SchrodingersGat:
Nice catch, and great addition of tests, thanks!

PR #3000, opened 2022-05-15
DevOps improvements

matmair:
WIP
This PR adds some Developer experience improvements:
- Adds pre-commit to the toolchain. This tool hooks into git and runs a set of tests before committing to git (only on the changed code)
- Adds pre-commit config for fixing/highlighting:
   - trailing whitespaces
   - end of file newline missing
   - yaml missformed
   - too large files added
   - line-endings are not *ix
   - flake8 has errors (using the setup.cfg config)
 - Adds an inv task to set up the dev-environment (setup-dev) -> Which could be extended to check for a venv and load test-data
- Flake-8 docstring tests (including pre-commit)

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3000"></a>

PR #3041, opened 2022-05-20
Style fixes

matmair:
This is a huge diff as it was auto-generated to fix a few styling problems:
- fix trailing whitespaces
- fix (missing) file endings
- Small PEP fixes in support files
- resort all imports to isort standard

All this will be done automatically if the devs have installed the dev-enviroment (see #3000)

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3041"></a>


InvenTree/InvenTree/management/commands/rebuild_models.py
```diff
@@ -27,6 +28,7 @@ def handle(self, *args, **kwargs):
         try:
             print("Rebuilding PartCategory objects")
 
+            # InvenTree imports
```
SchrodingersGat, line 31: Can we disable these messages appearing throughout the code?

SchrodingersGat:
I like the cleanup, but could do without the addition of the import comments everywhere in the code. Can we disable that option with isort?

InvenTree/InvenTree/management/commands/rebuild_models.py
matmair: Yes, they can be configured

matmair:
@SchrodingersGat I removed the comments

SchrodingersGat (approved):

Issue #1273, opened 2021-02-04
Remove cairo dependency


SchrodingersGat:
Next WeasyPrint release will [not require cairo](https://www.courtbouillon.org/blog/00004-weasyprint-without-cairo-what-s-different)

wengtad:
Seems like WeasyPrint was removed the cairo dependency. Based on their comment [here](https://github.com/Kozea/WeasyPrint/issues/841#issuecomment-773941914).

SchrodingersGat:
@wengtad yes promising progress! Hopefully the remaining non-python deps will be dealt with soon too :)

SchrodingersGat:
Beta version is out - https://www.courtbouillon.org/blog/00008-weasyprint-53-beta

liZe:
> Hopefully the remaining non-python deps will be dealt with soon too :)

I hope so, but you’ll probably have to wait a little bit more! :wink:

rcludwick:
Don't forget to pull this out of the DockerFile.

SchrodingersGat:
Currently waiting for v2.0 of django-weasyprint to be released : https://github.com/fdemmer/django-weasyprint/issues/53#issuecomment-896122557

42CrMo4:
@SchrodingersGat fyi [django-weasyprint 2.1.0](https://github.com/fdemmer/django-weasyprint/releases) is released.

SchrodingersGat:
@42CrMo4 thanks for the update, was not aware that v2.x was out yet!

PR #2456, opened 2021-12-14
Update weasyprint version

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/1273

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/2456"></a>

PR #3042, opened 2022-05-20
Docker improvements

SchrodingersGat:
Experimenting with changing docker image base from alpine to python:slim

Also trying to remove requirement for cairo, and update to latest version of weasyprint.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3042"></a>

- Closes https://github.com/inventree/InvenTree/issues/1273
- Closes https://github.com/inventree/InvenTree/issues/3065

### Weasyprint updates

Changing to a `python:slim` base docker image has seemingly fixed the issues which were preventing an update to django-weasyprint - ref https://github.com/inventree/InvenTree/pull/2456 . 

### Image Comparison

| | Alpine | python:slim |
| --- | --- | --- |
| Size | 1.82GB | 932MB |
| Build Time | 390s | ~~232s~~ 151s |

### TODO

- [x] Update docker installation instructions to match new format - https://github.com/inventree/InvenTree/pull/3042
- [x] Get the git commit hash back into the production image (we no longer include the `.git` directory)
- [x] Simplify docker build-and-push process (consolidate to a single .github workflow)
- [x] Fix root user issues for production image


matmair:
@SchrodingersGat these numbers look much better. Do you think we could get under 500 MBs? That is the magical number for a lot of included private docker registries.

SchrodingersGat:
I think that 500MB would be an ambitious target, not sure what else we can trim at this point!

matmair:
@SchrodingersGat have you tried the python-alpine image? It seems to be only 90MB

SchrodingersGat:
@matmair any idea why the sqlite coverage checks are failing here? Nominally I have not changed any internal code or tests, only the docker image, but the test is not running under docker...

SchrodingersGat:
So, weasyprint removed support for exporting to non-PDF formats in [release v53.0](https://github.com/Kozea/WeasyPrint/releases/tag/v53.0) - this is the release which moves away from cairo 

We currently require the non-pdf export to provide images of labels to external label printer plugins. This is the *only* part of the code that requires this feature.

So we can either:

A. Revert the weasyprint version back, add "cairo" deps back into the docker image and go back to what we were doing
B. Provide a `.pdf` file to the label plugins, and let them deal with it
C. Add some other dependencies to convert a pdf to an image

I would prefer *B* as it is the slimmest and simplest approach. It would just mean that any label printing plugins have to handle pdf files, not png files...

@matmair any thoughts or preferences here?

matmair:
@SchrodingersGat it is probably a very common requirement to have a png for printing so I would prefer for us to implement it.

SchrodingersGat:
Are you aware of any good low-dependency libraries for converting pdf to image?

matmair:
@SchrodingersGat seems they all use Cairo - back to the roots. Are the updates in weasyprint significant?

SchrodingersGat:
I guess the question is do we want to be locked into an older version of weaysprint (especially when we start expanding on the report functionality) or duplicate the previous functionality with a new library...

matmair:
@SchrodingersGat I would implement the image functionality with the library so we can move forward

SchrodingersGat:
@matmair done, pdf2image was very easy to integrate in the end.

SchrodingersGat:
Should we move all our github CI tests to docker images? It might be useful to move away from bare-metal setups for our test environment:

- Reduce needless duplication of effort
- Increased testing of our docker images
- Potentially speed up testing though image snapshots

matmair:
@SchrodingersGat the configuration time for the environment is minimal, that won't speed up more than 10-15 sec - I tested that back when I restructured the CI.
I like the current setup because it ensures that the suggested bare metal setup from the docs really works. I deploy everywhere bare-metal for compliance and speed reasons and never have a problem because the pipelines mirror that. If we switch to fully docker based it exposes us to the different DBMS setups failing.
I would rather add docker as another pull test like mysql and postgres.

SchrodingersGat:
@matmair reasonable. I'll leave the CI stuff how it is.

SchrodingersGat:
@matmair I think this is getting pretty close now - LMK what you think. I would like to get this merged in before thinking about signing the docker image in a different PR

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,69 @@
...
+# - Check that the version number matches the current branch or tag
+# - Build the InvenTree docker image
+# - Run suite of unit tests against the build image
+# - Push the compiled, tested image to dockerhub
+
+name: Docker
+
+on:
+  release:
+    types: [published]
```
matmair, line 19: Great for automation

.github/workflows/qc_checks.yaml
```diff
@@ -114,7 +117,7 @@ jobs:
       - name: Enviroment Setup
         uses: ./.github/actions/setup
         with:
-          apt-dependency: gettext
+          apt-dependency: gettext poppler-utils
```
matmair, line 120: should we mabybe add this requirement to the docs?

.github/workflows/version.yml
```diff
@@ -1,21 +0,0 @@
-# Checks version number
-name: version number
```
matmair, line 2: The less workflows the better!

Dockerfile
```diff
@@ -49,82 +50,82 @@ LABEL org.label-schema.schema-version="1.0" \
       org.label-schema.vendor="inventree" \
       org.label-schema.name="inventree/inventree" \
       org.label-schema.url="https://hub.docker.com/r/inventree/inventree" \
-      org.label-schema.vcs-url=${INVENTREE_GIT_REPO} \
-      org.label-schema.vcs-branch=${INVENTREE_GIT_BRANCH} \
-      org.label-schema.vcs-ref=${INVENTREE_GIT_TAG}
```
matmair, line 54: @SchrodingersGat don't we need this tag?

InvenTree/InvenTree/tests.py
```diff
@@ -380,6 +380,30 @@ def test_comparison(self):
         self.assertTrue(v_d > v_c)
         self.assertTrue(v_d > v_a)
 
+    def test_commit_info(self):
+        """Test that the git commit information is extracted successfully"""
+
+        envs = {
```
matmair, line 386: Suggestion: To conform with google docstyle there eshould be  no newline here

docker/production/.env
```diff
@@ -16,6 +16,12 @@ INVENTREE_WEB_PORT=1337
 INVENTREE_DEBUG=False
 INVENTREE_LOG_LEVEL=WARNING
 
+# InvenTree admin account details
+# Un-comment (and complete) these lines to auto-create an admin acount
```
matmair, line 20: Leaving unused passwords in the env can be dangerous - should we maybe also add a note about this?

Suggestion:
```
Remove the entered credentials after the first run
```

requirements.txt
```diff
@@ -29,16 +29,16 @@ django-sslserver==0.22                  # Secure HTTP development server
...
 django-user-sessions==1.7.1             # user sessions in DB
-django-weasyprint==1.0.1                # django weasyprint integration
+django-weasyprint==2.1.0                # django weasyprint integration
 djangorestframework==3.12.4             # DRF framework
 django-xforwardedfor-middleware==2.0    # IP forwarding metadata
 flake8==3.8.3                           # PEP checking
 flake8-docstrings==1.6.0                # docstring format testing
 gunicorn>=20.1.0                        # Gunicorn web server
 importlib_metadata                      # Backport for importlib.metadata
-inventree                               # Install the latest version of the InvenTree API python library
```
matmair, line 39: Any idea why that was in here in the first place? Seems to be a non-dependency for deployment enviroments

tasks.py
```diff
@@ -94,7 +94,7 @@ def install(c):
     print("Installing required python packages from 'requirements.txt'")
 
     # Install required Python packages with PIP
-    c.run('pip3 install -U -r requirements.txt')
+    c.run('pip3 install --no-cache-dir --disable-pip-version-check -U -r requirements.txt')
```
matmair, line 97: Question: Won't `--no-cache-dir` increase the download traffic a lot?

docker-compose.yml
```diff
@@ -101,4 +101,4 @@ volumes:
             o: bind
             # This directory specified where InvenTree source code is stored "outside" the docker containers
             # By default, this directory is one level above the "docker" directory
-            device: ${INVENTREE_EXT_VOLUME:-../}
+            device: ${INVENTREE_EXT_VOLUME:-./}
```
matmair, line 104: Question: Any specific reason for the move?

deploy/supervisord.conf
```diff
@@ -1,46 +0,0 @@
-; # Supervisor Config File
```
matmair, line 1: Question: Why is this removed? I am pretty sure we are linking to this in the docs for bare metal installs

matmair (changes requested):
@SchrodingersGat there are a few open questions form my side - especially moving the dockerfile and removing deploy info.

.github/workflows/qc_checks.yaml
SchrodingersGat: Great point - I'll add this to https://github.com/inventree/inventree-docs/pull/295

Dockerfile
SchrodingersGat: The git commit information is now passed through to the docker image at build time - the "production" image does not contain the `.git` directory any longer

docker/production/.env
SchrodingersGat: Will do.

Dockerfile
matmair: The tags are exposed if the image is up - right? A user once told me that he uses the image labels for kubernetes automation.

requirements.txt
SchrodingersGat: I think it crept in during my testing and never got removed. No real reason to have it here at all.

tasks.py
SchrodingersGat: Hmm not sure about that. Might be worth checking. My understanding was that it reduced the final docker image size

requirements.txt
matmair: I am currently looking into pip-tools and using full dependency pinning - that might be more stable for testing and deployment. The secondary dependencies could vary the way we do it now.

tasks.py
matmair: Did not know that - always learning new stuff! Then I think this is the right step 👍

deploy/supervisord.conf
SchrodingersGat: Ah you're right. I did a cursory search and thought it was orphaned.. I'll add it back in.

SchrodingersGat:
@matmair thanks for the review. Have made a few changes as suggested

docker-compose.yml
SchrodingersGat: Yep. Previously the docker build was performed from within the `./docker` directory, and involved re-downloading the source code from git based on the specified tag. However as we are now building from the github workflow, we have already defined the context of the built image (latest / stable / tag) and so we can use the simpler process of copying source files into the image.

However this needs to be performed from the top-level directory as you cannot `COPY` files from a directory above

matmair (approved):

docker-compose.yml
matmair: Seems more efficient, I like it!

matmair:
> @matmair thanks for the review. Have made a few changes as suggested

@SchrodingersGat great changes! Should be fine now in my view.

SchrodingersGat:
@matmair regarding the `--no-cache-dir` option:

with `--no-cache-dir` option

185s / 989MB

without `--no-cache-dir` option

115s / 1.04GB

Issue #3013, opened 2022-05-16
[FR] Notify staff users of error logs


SchrodingersGat:
When an error occurs, it will be logged in the database.

We should add a plugin which notifies any staff users when this happens, and provide them with a link to the error log.

Also, a link to generate an issue report on github.

PR #3075, opened 2022-05-26
Improve error management for order price calculation

SchrodingersGat:
- If there are missing exchange rates, it throws an error
- Very much an edge case

@matmair in the case where exchange rates are not available or have bugged out, there was an unhandled exception in the `get_total_price` method. I happened across it randomly after reinstalling from a fresh db.

This PR adds some error handling, but the question is - what do we return as the "total price" when exchange rates are not available?

Currently the code simply returns zero. Is this sufficient, do you think, given that this is somewhat of an edge case anyway?

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3075"></a>

### DIFF
--- a/InvenTree/order/models.py
+++ b/InvenTree/order/models.py
@@ -4,7 +4,10 @@ Order model definitions
 
 # -*- coding: utf-8 -*-
 
+import logging
 import os
+import sys
+import traceback
 from datetime import datetime
 from decimal import Decimal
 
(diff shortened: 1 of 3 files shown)


matmair (commented):
Returning zero will break stuff in a few places: total price in orders and the pricing tab mainly. Not sure if we want that.

matmair:
@SchrodingersGat if this is merged implementing #3013 should follow before the next release to ensure this kind of error is surfaced ASAP.

SchrodingersGat:
> Returning zero will break stuff in a few places: total price in orders and the pricing tab mainly. Not sure if we want that.

What is a reasonable alternative? We could return "None" and display an error message on the page?

SchrodingersGat:
image

Now an error message is displayed if the price cannot be calculated

Issue #3078, opened 2022-05-26
[FR] Automated releases


matmair:
**Problem**
Releases are currently a very manual process. That covers multiple repos.

**Suggested solution**

- [x] Use GitHubs automatic release note creation (with attribution, new contributors, section etc) [the docs are great](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes)
- [ ] Create a release action that runs after a release is created on invenTree/InvenTree and then runs on inventree-docs and inventree-python #3100 
- [ ] Create a PR template for releases

**Do you want to develop this?**
This would be a team effort but would be happy to do the dev-ops side

matmair:
Inspiration for workflow: https://gist.github.com/riggaroo/d828c5ffecf261d3ccc7bde89817dba5

matmair:
https://github.com/gr2m/twitter-together for releasing tweets

matmair:
https://github.com/marketplace/actions/release-for-reddit for releasing reddit post

SchrodingersGat:
This is definitely on my radar, every release cycle takes so much work, and invariably I'll overlook one of the many steps

matmair:
@SchrodingersGat should we maybe start with a list of steps?

SchrodingersGat:
### Release Notes / Changes / Milestones

Currently we assign PRs and issues to milestones, so we can look back and see exactly what has been done for a certain release.

We then also copy / reproduce this information over on the documentation repo, e.g. https://inventree.readthedocs.io/en/latest/releases/0.7.0/

This is a major duplication of data, ideally there would be a way of automating this / removing duplication?

### Post Release Steps

There a number of post-release steps which are currently performed manually

- Twitter post
- Reddit post
- GitHub discussion post
- Tag and release matching [documentation version](https://github.com/inventree/inventree-docs)
- Push new release to 'stable' branch also
- Release new version of python bindings (if applicable)
- Release new version of mobile app (if applicable)

matmair:
@SchrodingersGat the GitHub automated release note creation could take care of the first part - the rest could be done by a workflow that runs on release publishing.

PR #3099, opened 2022-05-30
Adds release.yml file for auto-generating release notes

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/3078

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3099"></a>

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,24 @@
+# .github/release.yml
+
+changelog:
+  categories:
+    - title: Breaking Changes
+      labels:
+        - Semver-Major
+        - breaking
+    - title: New Features
+      labels:
+        - Semver-Minor
+        - enhancement
+    - title: Bug Fixes
+      labels:
+        - Semver-Patch
+        - bug
+    - title: Setup Changes
+      labels:
+        - docker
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@matmair it looks like this will require some further work... I tried to auto-generate a release, and it added notes for every commit since `0.4.0`! Hopefully there are some options I can tweak to get it to work better than that...

matmair:
@SchrodingersGat I will look into that

SchrodingersGat:
Looks similar to the issues here: https://github.com/github/feedback/discussions/5975

SchrodingersGat:
@matmair I have worked out what is happening. `0.4.0` was the last release which was tagged on the *master* branch. All subsequent releases have been branched off, and then tagged. So the release notes algorithm can't "see" those newer releases.

Going forward, a more consistent approach of tagging major releases on the master branch, and then creating release branches from there, should allow the system to work properly. So, this is how I'll do releases starting at `0.8.0`

matmair:
@SchrodingersGat why not start with the next point release? Then we can test before releasing the next minor.

PR #3133, opened 2022-06-05
Add setups / OOBE

### LINKED ISSUE #2284: [FR] settings setup guide

matmair:
This PR implements setups as first class citizens. This required for the OOBE and setups in general.

Ref #2947
Closes #2284

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3133"></a>


SchrodingersGat:
@matmair do we realistically think this will make it into `0.8.0` or should we push to `0.9.0` and focus on closing out the remaining small items?

matmair:
@SchrodingersGat I really want this as a building block for some further plugin stuff so I will sit down and do this next thing after the current huntr.dev thingy

matmair:
Waiting for #3901 for this

PR #3136, opened 2022-06-06
Adds callback for creation of an error log

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/3013

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3136"></a>

### DIFF
--- a/InvenTree/common/notifications.py
+++ b/InvenTree/common/notifications.py
@@ -299,11 +299,12 @@ def trigger_notification(obj, category=None, obj_ref='pk', **kwargs):
             delivery_methods = (delivery_methods - IGNORED_NOTIFICATION_CLS)
 
         for method in delivery_methods:
-            logger.info(f"Triggering method '{method.METHOD_NAME}'")
+            logger.info(f"Triggering notification method '{method.METHOD_NAME}'")
             try:
                 deliver_notification(method, obj, category, targets, context)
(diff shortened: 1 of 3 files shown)


InvenTree/common/notifications.py
```diff
@@ -299,11 +299,12 @@ def trigger_notification(obj, category=None, obj_ref='pk', **kwargs):
             delivery_methods = (delivery_methods - IGNORED_NOTIFICATION_CLS)
 
         for method in delivery_methods:
-            logger.info(f"Triggering method '{method.METHOD_NAME}'")
+            logger.info(f"Triggering notification method '{method.METHOD_NAME}'")
             try:
                 deliver_notification(method, obj, category, targets, context)
             except NotImplementedError as error:
-                raise error
```
matmair, line 306: @SchrodingersGat raising an error here was intentional - this way a wrongly implemented notification method sticks out directly and does not fail silently for a long time.
If you do not want this to raise errors I suggest at least showing a warning in the UI.

matmair (commented):
Nice implementation

InvenTree/common/notifications.py
SchrodingersGat: Fair enough, I can see the merit in that. I'll make a new PR to revert this bit.

PR #3140, opened 2022-06-06
Fix has_complete_bom_pricing logic errors

### LINKED ISSUE #3139: [BUG] part.has_complete_bom_pricing is broken by c6fd2281
**Describe the bug**

There are two code errors in https://github.com/inventree/InvenTree/commit/c6fd2281d689309007b06c46560e6f446ee931bd which completely break the part.has_complete_bom_pricing property (at least in head, and the current stable release... and afaict, the code was never correct, but I haven't gone back to check).

The issues are:
1) before the loop - calling common.models.get_setting instead of common_models.InvenTreeSetting.get_setting when trying to retrieve PART_BOM_USE_INTERNAL_PRICE - afaict there is no get_setting method at the module level and never has been, only in the class... 
2) within the loop passing an argument (use_internal) to the has_pricing_info property of the sub_part, e.g. attempts to call a property as a method, this is not supported, and Python raises an exception. 

These problems have been hidden, because the way this property is used (within template if statements, as a property checking for falseness) falls victim to Django's bug (https://code.djangoproject.com/ticket/17664) where the exception is wrapped and compared to the == False part of the statement.... so the fact an exception is being raised in this method is not highlighted. 

**Steps to Reproduce**

1. Create a part with no pricing info
2. Create a part with pricing info
3. Create a composite part with a BOM that includes each of the parts created at step 1 and 2
4. Navigate to the Pricing tab of part created in 3

**Expected behavior**

A warning should be shown below the BOM pricing table stating that BOM pricing is incomplete for this part. Instead no warning is shown because the method is failing and the exception is somehow being eaten within the template processing... 

**Deployment Method**
N/A

**Version Information**

Running from HEAD

mattbnz:
Closes #3139

There are two logic errors in this property method that have been present since c6fd228.

1) The get_setting method needs to be called on the InventTreeSetting class in common.models, not on the module itself.

2) You cannot call a property method directly passing an argument in Python, so the call to has_pricing_info is invalid and always fails. Given that has_complete_bom_pricing is/was the only user of that property anyway, fix this by just internalising the logic from that property in the method directly.

Also a couple of stylistic fixes to the invocation of the property in the template. 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3140"></a>

### DIFF
--- a/InvenTree/part/models.py
+++ b/InvenTree/part/models.py
@@ -1530,16 +1530,11 @@ class Part(MetadataMixin, MPTTModel):
         return self.supplier_parts.count()
 
     @property
-    def has_pricing_info(self, internal=False):
-        """Return true if there is pricing information for this part."""
-        return self.get_price_range(internal=internal) is not None
-
-    @property
     def has_complete_bom_pricing(self):
         """Return true if there is pricing information for each item in the BOM."""
-        use_internal = common.models.get_setting('PART_BOM_USE_INTERNAL_PRICE', False)
+        use_internal = common.models.InvenTreeSetting.get_setting('PART_BOM_USE_INTERNAL_PRICE', False)
         for item in self.get_bom_items().all().select_related('sub_part'):
-            if not item.sub_part.has_pricing_info(use_internal):
+            if item.sub_part.get_price_range(internal=use_internal) is None:
                 return False
 
         return True
--- a/InvenTree/part/templates/part/part_pricing.html
+++ b/InvenTree/part/templates/part/part_pricing.html
@@ -75,7 +75,7 @@
     {% endif %}
     {% endif %}
 
-    {% if part.has_complete_bom_pricing == False %}
+    {% if not part.has_complete_bom_pricing %}
     <tr>
         <td colspan='3'>
             <span class='warning-msg'><em>{% trans 'Note: BOM pricing is incomplete for this part' %}</em></span>
--- a/InvenTree/part/templates/part/prices.html
+++ b/InvenTree/part/templates/part/prices.html
@@ -83,7 +83,7 @@
                     {% endif %}
                     {% endif %}
 
-                    {% if part.has_complete_bom_pricing == False %}
+                    {% if not part.has_complete_bom_pricing %}
                         <tr>
                             <td colspan='4'>
                                 <span class='warning-msg'><em>{% trans 'Note: BOM pricing is incomplete for this part' %}</em></span>


mattbnz:
I hadn't run setup-dev when I made this commit... but I have now and afaict it's passing all the checks, so I think the above failure is something unrelated to me?

SchrodingersGat:
@mattbnz this is a good find, and a great first contribution! Thanks for the fix. Not sure why the "welcome" workflow isn't running but everything else looks great.

SchrodingersGat (approved):

matmair:
This change broke 4 of my plugins. @mattbnz Is there any unstated reason 'has_pricing_info' was removed? I would read it if not.

SchrodingersGat:
@matmair the only existing use of `has_pricing_info` was in the function that @mattbnz updated and it didn't make much sense to use it there IMO so I thought it was OK to remove it.

I can see that breaking plugins would be annoying. Can you work around it with other existing functionality?

mattbnz:
The reason was just what was in the commit message " Given that has_complete_bom_pricing is/was the only user of that property anyway, fix this by just internalising the logic from that property in the method directly." - I grepped for any other users and didn't find any, so given the usage, seemed contained it seemed safe to remove... 

I have no objection to bringing it back if there are external things depending on it - but for future reference, is the external API marked/denoted somehow? It seems suboptimal if literally every property/method is unable to be changed because it could be depended on?

mattbnz:
https://github.com/mattbnz/InvenTree/commit/8df090691f2859ff7f8ad8884ab22845a4ac1637 is what it would need to look like to bring it back... 

Let me know if you want me to open a pull request for it. 

It does seem like it would be cleaner to keep it out if it's not a formal part of the API and plugins can be reworked though.

matmair:
@SchrodingersGat I will readd it in my fork - I am not planning to make my pricing tools public anyway.
Just removing model properties will break a lot of plugins though. I am CI testing against master but not everybody will.

@mattbnz we have a plugin API that makes it possible to extend functions. Part of that are events - those get triggered on each change on an instance. So everything exposed with a instance should be considered to be used and removal is therefore breaking. There are currently only 2-3 regular devs so some things regarding inner workings are not fully laid out. We are working on dev docs.

SchrodingersGat:
@matmair I think anything we want to be guaranteed as operational in the code should be covered in unit tests. That way, at least the tests will break and start a discussion.

matmair:
@SchrodingersGat there is a reason users and plugins have ~99% coverage.

matmair:
> https://github.com/mattbnz/InvenTree/commit/8df090691f2859ff7f8ad8884ab22845a4ac1637 is what it would need to look like to bring it back... 
> 
> Let me know if you want me to open a pull request for it. 
> 
> It does seem like it would be cleaner to keep it out if it's not a formal part of the API and plugins can be reworked though. 

@mattbnz tha ks, I will readd it myself. There is currently not distinction into formal / unformal API. I tend to mark internal functions with an underscore at the beginning of the name but as models are exposed I would consider everything exposed by them API.

If we start breaking the python API regularly I do not see any value in providing or using it.

Issue #2507, opened 2022-01-06
[FR] Total value of everything in storage


Guusggg:
**Problem**
It's the beginning of a new year again and we are cleaning up our taxes. For this we need to know an estimate of how much value we currently have in storage.

**Suggested solution**
It would be cool if selected roles can see the total value of the supply in the System Information popup.

**Describe alternatives you've considered**
Currently I have written a little script that pulls every stock item from the inventree instance and sum up their prices.

**Examples of other systems**
I don't think it's an uncommon feature in these kind of systems however I wouldn't know of a example program at the moment...

**Do you want to develop this?**
I think I can do it, but I need help with making it efficient as to not overload the server too much.

matmair:
@Guusggg this seems like a great use case for a report as it is not needed too often.

Guusggg:
@matmair Oh right, a report! Totally forgot about them. It might even be possible to do it today, with all the variables being exposed to that API.

matmair:
@guusggg no problem, it is not documented a lot so I tend to also forget about it. I hope to build an easy mechanism to share and customize report templates this year. Something like plugins (maybe using them) and a WYSIWYG - type editor.

SchrodingersGat:
@matmair I look forward to seeing what you can come up with :) we use reports very extensively at work, but it is definitely poorly documented and under appreciated!

matmair:
@Guusggg can we close this for now as it most likely a custom report or would you be willing to submit it upstream?
@SchrodingersGat this is on the looooong-term TODO-list 👼

SchrodingersGat:
I'll leave it open for now, we can think about it for a bit

Guusggg:
@matmair I don't mind closing it, do it as you guys please ^^. I can submit the script I wrote upstream but it's nothing special...

```python
from inventree.api import InvenTreeAPI
from inventree.stock import StockLocation, StockItem

api = InvenTreeAPI(SERVER_ADDRESS, username=MY_USERNAME, password=MY_PASSWORD)
items = StockItem.list(api)

items_without_price = 0
total = 0
for item in items:
    item_price = 0

    if not item.in_stock:
        #print(f"{part.name} is not in stock! Quantity={item.quantity}, Pk={item.pk}")
        continue 

    if item.purchase_price:
        item_price = float(item.quantity) * float(item.purchase_price)
    else:
        part = item.getPart()

        print(f"{part.name} has no pricing! Quantity={item.quantity}, Pk={item.pk}")
[...]

PR #3141, opened 2022-06-06
Separate unit cost and purchase price in BOM

mattbnz:
(not opening a bug for this, as the current behaviour may be intentional, but even if it is, I think this change is reasonable and does not break anything). 

It's possible for a part to have purchase price information, but not supplier
pricing (aka unit prices), but the current BOM price display logic will only
show the purchase prices when supplier prices are also available.

It seems reasonable to separate these two pieces of information - even when no
supplier pricing is available, seeing the purchase prices for BOM components is
useful on its own.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3141"></a>

### DIFF
--- a/InvenTree/part/templates/part/part_pricing.html
+++ b/InvenTree/part/templates/part/part_pricing.html
@@ -60,6 +60,7 @@
         <td>Max: {% include "price.html" with price=max_total_bom_price %}</td>
     </tr>
     {% endif %}
+    {% endif %}
     {% if min_total_bom_purchase_price %}
     <tr>
         <td><strong>{% trans 'Unit Purchase Price' %}</strong></td>
@@ -82,6 +83,8 @@
         </td>
     </tr>
     {% endif %}
+
+    {% if min_total_bom_price or min_total_bom_purchase_price %}
     {% else %}
     <tr>
         <td colspan='3'>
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@mattbnz this does seem reasonable. FYI the entire "pricing" structure is going to have a major overhaul for the next major release:

- https://github.com/inventree/InvenTree/issues/1656
- https://github.com/inventree/InvenTree/issues/2507
- https://github.com/inventree/InvenTree/pull/2196

The current way we iterate over pricing is *extremely* expensive, and needs addressing! If you are interested in helping out with that, given your interest in the pricing structure, then let us know :) Help is always appreciated

mattbnz:
Thanks, I'd seen #1656, but the other two references are new to me, I'll have a look.

I don't know how much time I can commit, but if there's clear requests/tasks that you can lay out, I'm happy to try.

SchrodingersGat:
Thanks for the contribution @mattbnz

matmair:
Pricing needs a lot of love.
I tried to fix it but had to switch to an external timeseries database to get a reliable and fast backend for the daily operations I needed. If the codebase had been on today's level of coverage, documentation and cleanliness I might have tried to do it here but now I am in production with another system that I cannot just switch. Would love to get my code in here but that would make setup at least twice as difficult as it is now.

Issue #2253, opened 2021-11-04
[FR] Convert more server-side forms to the API


SchrodingersGat:
Convert the forms missed in https://github.com/inventree/InvenTree/pull/2198

Refactoring existing "StockItem" forms to use the API forms architecture

Closes https://github.com/inventree/InvenTree/issues/1905

- [x] ValidateBOM form - https://github.com/inventree/InvenTree/pull/2477
- [x] Merge stock items - https://github.com/inventree/InvenTree/issues/346
- [x] Assign to Customer form - https://github.com/inventree/InvenTree/issues/2369
- [x] Install stock item - https://github.com/inventree/InvenTree/pull/2640
- [x] Uninstall stock item - https://github.com/inventree/InvenTree/pull/2931
- [x] Add / improve unit testing for new forms
- [x] Add placeholder text for serial numbers when creating a new stock item - https://github.com/inventree/InvenTree/pull/2370
- [x] Add placeholder text for serial numbers when serializing a stock item
- [x] Reimplement unit testing for stock item serialization
- [x] Build cancel - https://github.com/inventree/InvenTree/pull/2931
- [x] Build complete - https://github.com/inventree/InvenTree/pull/2510
- [x] New build output - https://github.com/inventree/InvenTree/pull/2631
- [x] Delete build output(s) - https://github.com/inventree/InvenTree/pull/2615
- [x] Issue purchase order - https://github.com/inventree/InvenTree/pull/2931
- [x] complete purchase order - https://github.com/inventree/InvenTree/pull/2931
- [x] cancel purchase order - https://github.com/inventree/InvenTree/pull/2931
- [x] cancel sales order - https://github.com/inventree/InvenTree/pull/2931
- [x] ship sales order - https://github.com/inventree/InvenTree/pull/2199
- [x] duplicate BOM - https://github.com/inventree/InvenTree/pull/2477
- [x] export BOM - https://github.com/inventree/InvenTree/pull/2452
- [x] Add related part - https://github.com/inventree/InvenTree/pull/2361
- [x] Delete related part - https://github.com/inventree/InvenTree/pull/2361
- [x] Return from Customer form - https://github.com/inventree/InvenTree/pull/3120
- [x] Delete part - https://github.com/inventree/InvenTree/pull/3135
- [x] Delete build order - https://github.com/inventree/InvenTree/pull/3155
- [x] Delete part category - https://github.com/inventree/InvenTree/pull/3181
[...]

PR #3143, opened 2022-06-06
Refactor stock item test result form for API

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/2253

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3143"></a>

### DIFF
--- a/InvenTree/InvenTree/forms.py
+++ b/InvenTree/InvenTree/forms.py
@@ -16,13 +16,12 @@ from allauth.exceptions import ImmediateHttpResponse
 from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
 from allauth_2fa.adapter import OTPAdapter
 from allauth_2fa.utils import user_has_valid_totp_device
-from crispy_forms.bootstrap import (AppendedText, Div, PrependedAppendedText,
-                                    PrependedText, StrictButton)
+from crispy_forms.bootstrap import (AppendedText, PrependedAppendedText,
+                                    PrependedText)
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import Field, Layout
 
 from common.models import InvenTreeSetting
-from part.models import PartCategory
 
 logger = logging.getLogger('inventree')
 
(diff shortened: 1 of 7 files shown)


matmair:
@SchrodingersGat have you tested the performance implications if a lot of items are deleted at once?

SchrodingersGat:
It's not "too bad", if the user has a very large number of test results to delete, they'll be presented with the spinny wheel until they're all done.

It is the same question when we are using the API to delete a lot of stuff at once. The current solution (which is being rolled out to all the "multi delete" actions) is to sequentially send API requests to delete the items one at a time.

### Advantages

- No "new" API endpoints are required
- Simplistic and uses code we already have
- Sequential operations prevent the server from being "overloaded" with requests and failing some of them

### Disadvantages

- Non atomic (some of the operations might fail, but the others are not rolled back)
- Speed issues (a single API request and DB hit to delete multiple things at once is obviously faster)

Currently leaning heavily towards simplicity. This refactor has removed a lot of old crusty "delete" views which were 90% boilerplate. If there is another way of tackling this problem which gets around the disadvantages listed above, I'd be keen to hear it :)

matmair:
@SchrodingersGat I understand the approach but would like to point out that the non-atomic ness can lead to interesting side effects as results can change for multiple seconds - I would prefer something more atomic. Also this moves business logic to the frontend. So if we want to have the same in the app we need to replicate the logic there too.

An alternative solution would be to enable a delete action on the list endpoints - we already have endpoints that take parts[] or items[] as options (for printing for example).

SchrodingersGat:
> An alternative solution would be to enable a delete action on the list endpoints - we already have endpoints that take parts[] or items[] as options (for printing for example).

I like that idea a lot. We don't need to add multiple new endpoints, just make a mixin class.

I'll merge this PR in now, but will soon launch a new PR to refactor the existing "multi delete" functions to use the list endpoint

matmair:
@SchrodingersGat when the mixed n is created we could maybe standardize on using items as the argument for one or more items, not items, parts, etc. Makes it easier to code and keeps all calls similar.

SchrodingersGat:
100% was thinking the same thing :)

matmair:
> 100% was thinking the same thing :)

Will be breaking thought. Good to test the new release process.

PR #3144, opened 2022-06-06
Ensure an error gets logged when a delivery method fails

SchrodingersGat:
- Refactor existing code to log a custom error to the database
- Limit error notifications to UI

Ref: https://github.com/inventree/InvenTree/pull/3136#pullrequestreview-996354950

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3144"></a>

### DIFF
--- a/InvenTree/part/test_part.py
+++ b/InvenTree/part/test_part.py
@@ -631,7 +631,7 @@ class BaseNotificationIntegrationTest(InvenTreeTestCase):
         self.part.save()
 
         # There should be 1 notification
-        self.assertEqual(NotificationEntry.objects.all().count(), 1)
+        self.assertEqual(NotificationEntry.objects.all().count(), 2)
 
 
 class PartNotificationTest(BaseNotificationIntegrationTest):
--- a/InvenTree/part/test_part.py
+++ b/InvenTree/part/test_part.py
@@ -633,6 +633,10 @@ class BaseNotificationIntegrationTest(InvenTreeTestCase):
         self.part.save()
 
         # There should be 1 notification
+
+        nots = NotificationEntry.objects.all()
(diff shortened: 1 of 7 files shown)


InvenTree/part/test_part.py
```diff
@@ -631,7 +631,7 @@ def _notification_run(self, run_class=None):
         self.part.save()
 
         # There should be 1 notification
-        self.assertEqual(NotificationEntry.objects.all().count(), 1)
+        self.assertEqual(NotificationEntry.objects.all().count(), 2)
```
matmair, line 634: Question: Why should be this be higher?

matmair (approved):

InvenTree/part/test_part.py
SchrodingersGat: I believe it is because as part of the notification unit testing there's a broken notification, which logs an error, which in turn sends another notification.

matmair: The distinction in NotImplemented and Exception was so that Test ng etc. is kept somewhat predictable.

InvenTree/part/test_part.py
```diff
@@ -616,6 +616,8 @@ def _notification_run(self, run_class=None):
         # reload notification methods
         storage.collect(run_class)
 
+        NotificationEntry.objects.all().delete()
```
matmair, line 619: Simple but effective 👍

matmair (approved):

InvenTree/part/test_part.py
SchrodingersGat: Not actually all that effective though... These notification unit tests are really frustrating!

Issue #2950, opened 2022-05-09
[FR] Extend existing notifications 


SchrodingersGat:
Currently the available notifications are limited to "low_stock" - i.e. the available stock quantity for a given part has fallen below a configured level.

Now that the notification system has been expanded, and includes the ability to write custom notification plugins, we should expand the different notifications that can actually be sent.

This issue is open for general discussion, to determine what "events" should be available as notifications. 

### Ideas

- [x] Purchase order is overdue - https://github.com/inventree/InvenTree/pull/3114
- [x] Sales order is overdue - https://github.com/inventree/InvenTree/pull/3114
- [x] Build order is overdue - https://github.com/inventree/InvenTree/pull/3114
- [x] ~~Stock items(s) have not been updated (e.g. stocktake) in a certain amount of time?~~
- [x] Items have been received against a purchase order (notify assigned user) - https://github.com/inventree/InvenTree/pull/3259
- [x] Build order has been completed (notify user who issued the order, if not the same person as the responsible user) - https://github.com/inventree/InvenTree/pull/3255
- [x] Build order has been completed (notify user responsible for "upstream" build order) - https://github.com/inventree/InvenTree/pull/3255
- [x] Build order has been completed (notify user responsible for "upstream" sales order) - https://github.com/inventree/InvenTree/pull/3255
- [x] New sales order has been created and assigned to you (if you were not the user who created it) - https://github.com/inventree/InvenTree/pull/3145
- [x] New purchase order has been created and assigned to you (if you were not the user who created it) - https://github.com/inventree/InvenTree/pull/3145
- [x] New build order has been created and assigned to you (if you were not the user who created it) - https://github.com/inventree/InvenTree/pull/3145

SchrodingersGat:
@eeintech @matmair any thoughts on this one?

matmair:
@SchrodingersGat I do not really need additional notifications (I think but who knows till they are there).

eeintech:
When a BO or PO passed the target date maybe? Opening it to plugins is a great idea

SchrodingersGat:
Yeah I had some similar ideas @eeintech 

Plugins can already trigger notifications, adding some more "core" notifications would be a good idea

rkalman:
It would be great to get a notification if ordered items are arrived - sent to the responsible of the PO. (Sometimes somebody else receives the items, so you need to poll the open POs)

PR #3145, opened 2022-06-06
Notification on new orders

SchrodingersGat:
Emits notifications when new orders are generated. Notifications go to responsible users, and exclude the user who created them.

- Sales Orders
- Purchase Orders
- Build Orders

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3145"></a>

Ref: https://github.com/inventree/InvenTree/issues/2950


matmair:
@SchrodingersGat did you know that a function can receive multiple signals? I think the order notifications could be one per notification type, not three if we use that.

InvenTree/build/models.py
```diff
@@ -1049,6 +1050,38 @@ def after_save_build(sender, instance: Build, created: bool, **kwargs):
...
+        # Notify the responsible users that the build order has been created
+        if instance.responsible is not None:
+
+            targets = []
+
+            for owner in instance.responsible.get_related_owners(include_group=False):
+                user = owner.owner
+
+                if user != instance.issued_by:
+                    targets.append(user)
```
matmair, line 1062: Suggestion: this feels like something we need more often. I think it would warrant a change to get_related_user to exclude the supplied user(s).

InvenTree/build/models.py
```diff
@@ -1049,6 +1050,38 @@ def after_save_build(sender, instance: Build, created: bool, **kwargs):
...
+
+            if len(targets) > 0:
+                # Notify the responsible user(s) that the new build order has been created
+                name = _("New Build Order")
```
matmair, line 1066: Suggestion: the models verbose name could he used here

InvenTree/build/models.py
```diff
@@ -1049,6 +1050,38 @@ def after_save_build(sender, instance: Build, created: bool, **kwargs):
...
+                context = {
+                    'build': instance,
+                    'name': name,
+                    'message': _("A new Build Order has been created and assigned to you"),
```
matmair, line 1070: Suggestion: models verbose name

InvenTree/build/models.py
```diff
@@ -1049,6 +1050,38 @@ def after_save_build(sender, instance: Build, created: bool, **kwargs):
...
+                    'link': InvenTree.helpers.construct_absolute_url(instance.get_absolute_url()),
+                    'template': {
+                        'html': 'email/new_order_assigned.html',
+                        'subject': name,
+                    }
+                }
+
+                common.notifications.trigger_notification(
+                    instance,
+                    'build.new_build_order',
```
matmair, line 1080: Suggestion: models meta name

InvenTree/build/models.py
SchrodingersGat: I have refactored this out, the `trigger_notification` method now accepts a set of users to exclude from the notification, which makes the calling code here a lot simpler.

PR #3146, opened 2022-06-06
Adds API mixin for "bulk delete"

SchrodingersGat:
- Introduces a BulkDelete API mixin class
- Allows deletion of multiple items against a single API request
- Advantages are atomicity and speed
- Disadvantages are there are fewer things to debug

Ref: https://github.com/inventree/InvenTree/pull/3143#issuecomment-1147338897

### TODO

Refactor (and add unit testing) for the following models which currently implement the `inventreeMultiDelete` function

- [x] Stock Item Test Result
- [x] Attachments
- [x] BOM items
- [x] Supplier Parts
- [x] Manufacturer parts
- [x] Manufacturer part parameter
- [x] Stock items
- [x] Delete outdated `inventreeMultiDelete` function (JS)

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3146"></a>

### DIFF
--- a/InvenTree/InvenTree/api.py
+++ b/InvenTree/InvenTree/api.py
@@ -76,7 +76,7 @@ class BulkDeleteMixin:
         model = self.serializer_class.Meta.model
 
         # Extract the items from the request body
-        items = request.data.get('items', None)
+        items = request.data.getlist('items', None)
 
         if items is None or type(items) is not list or not items:
             raise ValidationError({
--- a/InvenTree/InvenTree/api.py
+++ b/InvenTree/InvenTree/api.py
@@ -1,11 +1,14 @@
 """Main JSON interface views."""
 
 from django.conf import settings
+from django.db import transaction
 from django.http import JsonResponse
 from django.utils.translation import gettext_lazy as _
 
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import filters, permissions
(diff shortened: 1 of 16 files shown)


InvenTree/InvenTree/api.py
```diff
@@ -50,6 +53,52 @@ def get(self, request, *args, **kwargs):
         return JsonResponse(data, status=404)
 
 
+class BulkDeleteMixin:
+    """Mixin class for enabling 'bulk delete' operations for various models.
+
+    Bulk delete allows for multiple items to be deleted in a single API query,
+    rather than using multiple APi calls to the various detail endpoints.
```
matmair, line 60: Suggestion: API not APi

InvenTree/stock/api.py
```diff
@@ -1074,7 +1074,7 @@ class StockItemTestResultDetail(generics.RetrieveUpdateDestroyAPIView):
     serializer_class = StockSerializers.StockItemTestResultSerializer
 
 
-class StockItemTestResultList(generics.ListCreateAPIView):
+class StockItemTestResultList(BulkDeleteMixin, generics.ListCreateAPIView):
```
matmair, line 1077: Suggestion: create a ListCreateDeleteAPIView with the same inheritance in InvenTree/InvenTree/api.py and use that as a single import everywhere we need it.

PR #3156, opened 2022-06-07
Export records update - allow overwriting existing files without user input

miggland:
As it is now, the export-records function can not easily be run unattended. I added a switch (-o) to the command line to signal that overwriting the existing file is OK.

Furthermore, changed the following minor things:
- I added a command to remove the temporary file after the process is finished - otherwise, the file remains in place and making any changes to it is useless.
- `invoke style` now also includes tasks.py in the checks.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -282,8 +282,8 @@ def content_excludes():
     return output
 
 
-@task(help={'filename': "Output filename (default = 'data.json')"})
-def export_records(c, filename='data.json'):
+@task(help={'filename': "Output filename (default = 'data.json')", 'overwrite': "Overwrite existing files without asking first (default False)"})
+def export_records(c, filename='data.json', overwrite = False):
     """Export all database records to a file"""
     # Get an absolute path to the file
     if not os.path.isabs(filename):
@@ -292,7 +292,7 @@ def export_records(c, filename='data.json'):
 
     print(f"Exporting database records to file '{filename}'")
 
-    if os.path.exists(filename):
+    if os.path.exists(filename) and overwrite is False:
         response = input("Warning: file already exists. Do you want to overwrite? [y/N]: ")
         response = str(response).strip().lower()
 
--- a/tasks.py
+++ b/tasks.py
@@ -283,7 +283,7 @@ def content_excludes():
 
 
 @task(help={'filename': "Output filename (default = 'data.json')", 'overwrite': "Overwrite existing files without asking first (default False)"})
-def export_records(c, filename='data.json', overwrite = False):
+def export_records(c, filename='data.json', overwrite=False):
     """Export all database records to a file"""
     # Get an absolute path to the file
     if not os.path.isabs(filename):
@@ -329,7 +329,7 @@ def export_records(c, filename='data.json', overwrite = False):
         f_out.write(json.dumps(data, indent=2))
 
     print("Data export completed. Removing temporary files")
-    
+
     os.remove(tmpfile)
 
 
--- a/tasks.py
+++ b/tasks.py
@@ -328,7 +328,9 @@ def export_records(c, filename='data.json', overwrite = False):
     with open(filename, "w") as f_out:
         f_out.write(json.dumps(data, indent=2))
 
-    print("Data export completed")
+    print("Data export completed. Removing temporary files")
+    
+    os.remove(tmpfile)
 
 
 @task(help={'filename': 'Input filename', 'clear': 'Clear existing data before import'}, post=[rebuild_models, rebuild_thumbnails])
--- a/tasks.py
+++ b/tasks.py
@@ -224,7 +224,7 @@ def update(c):
 def style(c):
     """Run PEP style checks against InvenTree sourcecode"""
     print("Running PEP style checks...")
-    c.run('flake8 InvenTree')
+    c.run('flake8 InvenTree tasks.py')
 
 
 @task


tasks.py
```diff
@@ -282,8 +282,8 @@ def content_excludes():
     return output
 
 
-@task(help={'filename': "Output filename (default = 'data.json')"})
-def export_records(c, filename='data.json'):
+@task(help={'filename': "Output filename (default = 'data.json')", 'overwrite': "Overwrite existing files without asking first (default False)"})
```
matmair, line 285: Nitpik: it should be 'default = False'

tasks.py
```diff
@@ -282,8 +282,8 @@ def content_excludes():
...
+def export_records(c, filename='data.json', overwrite=False):
```
matmair, line 286: Request: Please add type 'bool'

tasks.py
```diff
@@ -328,7 +328,9 @@ def export_records(c, filename='data.json'):
     with open(filename, "w") as f_out:
         f_out.write(json.dumps(data, indent=2))
 
-    print("Data export completed")
+    print("Data export completed. Removing temporary files")
+
+    os.remove(tmpfile)
```
matmair, line 333: Suggestion: I would appreciate adding another switch for deleting the tempfile: it contains the permissions - that might be useful for some users.

matmair (changes requested):
Looks good overall! My suggestion would be to use the dev setup instead of inv style as it contains more checks. Just run dev-setup once and it will run every time you commit automatically.

SchrodingersGat:
@miggland happy to merge once you've made the changes as suggested by @matmair

tasks.py
miggland: It's something I considered when looking at the code. In my case, looking through the tmp-file did not make any substantial changes concerning permissions - it did however change 'special characters' (ä,ö,...) and indentation of the file.

I don't understand enough about Drupal to judge myself - but shouldn't the permissions part be covered by the excludes defined in `content_excludes()`, instead of being removed afterwards?

Anyway, having two files, one called .tmp, is not really a transparent solution. Having a switch may make sense though.

tasks.py
miggland: Nitpik back: why is it default = ... with spaces around the = sign in the help, but without spaces otherwise according to the style check?
None of this is consistent within the file at the moment, by the way.. see import_records() and delete_data() as well.

matmair: To keep it better readable and it is defined that way on the docstyle we use.
Pep8 is without spaces so the code is that way.

tasks.py
matmair: We are currently typing out the whole code base. There is no enforcement but that will be introduced in a bit.

tasks.py
matmair: We are using django. Excluding permissions in the export leads to custom user-model implementations breaking outright. That is why it needs to be done afterwards as that only removes the specified classes.
We have a sizeable user base. I prefer not to break the cli for them.

tasks.py
miggland: If you mean that I should change the function definition to the following:
```
def export_records(c, filename: str = 'data.json', overwrite: bool = False):
```
then that won't work with the current tests `invoke`: it fails with the error `ValueError: Function has keyword-only parameters or annotations, use inspect.signature() API which can support them`:

```
  File "/home/ubuntu/.local/bin/invoke", line 8, in <module>
    sys.exit(program.run())
  File "/home/ubuntu/.local/lib/python3.10/site-packages/invoke/program.py", line 373, in run
    self.parse_collection()
  File "/home/ubuntu/.local/lib/python3.10/site-packages/invoke/program.py", line 465, in parse_collection
    self.load_collection()
  File "/home/ubuntu/.local/lib/python3.10/site-packages/invoke/program.py", line 696, in load_collection
    module, parent = loader.load(coll_name)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/invoke/loader.py", line 76, in load
    module = imp.load_module(name, fd, path, desc)
  File "/usr/lib/python3.10/imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "/usr/lib/python3.10/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 719, in _load
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/ubuntu/InvenTree/tasks.py", line 291, in <module>
    def export_records(c, filename: str = 'data.json', overwrite: bool = False):
  File "/home/ubuntu/.local/lib/python3.10/site-packages/invoke/tasks.py", line 357, in inner
    obj = klass(
  File "/home/ubuntu/.local/lib/python3.10/site-packages/invoke/tasks.py", line 76, in __init__
    self.positional = self.fill_implicit_positionals(positional)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/invoke/tasks.py", line 167, in fill_implicit_positionals
    args, spec_dict = self.argspec(self.body)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/invoke/tasks.py", line 153, in argspec
    spec = inspect.getargspec(func)
  File "/usr/lib/python3.10/inspect.py", line 1245, in getargspec
    raise ValueError("Function has keyword-only parameters or annotations"
ValueError: Function has keyword-only parameters or annotations, use inspect.signature() API which can support them
``

Putting the hint in the help text doesn't really make much sense, since it does not apply to the CLI (where the flag is either there, or not..)

tasks.py
miggland: Yes, Django is what I meant *facepalm*

What you say makes sense. I have added two flags to allow the behaviour I think makes sense, with flags, and the historical behaviour without the flags.

SchrodingersGat (approved):

tasks.py
matmair: Looks good. Thx!

matmair (approved):

matmair:
@SchrodingersGat would be a go for merge from me

PR #3170, opened 2022-06-10
Prevent auto-creation of SalesOrderShipment if we are importing data

SchrodingersGat:
- Fixes a bug which prevents importing of datasets

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3170"></a>

(Cherry picked from https://github.com/inventree/InvenTree/pull/3118)


SchrodingersGat:
Note that the python checks are failing because master branch has some new API features which are not supported on `0.7.x` branch

Issue #3167, opened 2022-06-09
[FR] User configurable pattern for order reference field


SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/discussions/3163

Allow reference fields for orders (Build / Sales / Purchase) to follow a configurable pattern.

Currently we allow a regex pattern to be specified for the `IPN` (internal part number) field, which could be extended here.

The added complication is that the order reference field is shadowed with an integer copy, to allow for natural sorting of the references. So any regex (or similar) implementation would need to take that into account.

SchrodingersGat:
Implementation idea:

- Add a global setting (one for each order type) such as `PURCHASE_ORDER_REFERENCE_REGEX`
- This is a regex pattern that allows enforcement of custom reference value 
- The setting must either be *empty* (default) or have exactly 1 capturing group
- e.g. `(\d{4})-(?:.{4})` would match the custom pattern requested by @miggland 
- As there is a single capturing group (enforced) then we know how to extract the "incrementing" part of the reference

SchrodingersGat:
Refer to https://github.com/inventree/InvenTree/pull/3267

PR #3172, opened 2022-06-10
[WIP] Implement regex matching for order reference fields

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/3167

### Support Multiple Reference Fields

- BuildOrder
- SalesOrder
- PurchaseOrder

### Pattern Matching

Support two distinct types of pattern matching:

- regex: More powerful, but more complex and confusion
- fnmatch: Simpler, but does not provide complex functionality

### Reference Extraction

InvenTree needs to be able to "know" which part of the reference is the part to be incremented. This is pretty trivially achievable with regex (using a matching group), but unsure how to implement with fnmatch

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3172"></a>

### Documentation

New features will need to be documented!


matmair:
@SchrodingersGat sorry for the late jumping in: I think regex might be a bit strict / hard to use for this.
Suggestion: We already have a function that uses jinja for something similar (part ipn) - why not use that?

SchrodingersGat:
Agreed that regex is weighty - and difficult for the average user.

There are a couple of different goals here, not necessarily the same but certainly related:

1. Allow the user to "enforce" a certain pattern for order reference fields
2.  Improve the auto-incrementing reference (see #3167)

Point 1. can be covered using regex (which is already implemented for build order references) or by a simpler pattern matching (e.g. [fnmatch](https://docs.python.org/3/library/fnmatch.html)). Using jinja templates does not really enforce anything, but it could be used instead to create a "default" reference field (without pattern enforcement)

Point 2. is where the difficulty lies. The particular use case by @miggland provides an example of this. Currently the "auto incrementing" mixin will have trouble auto-incrementing a complex field with multiple number values. How can we tell the system which field to increment?

Regex provides a way to do this - you could have a single capturing group in your regex pattern, and do a simple regex-replace.

e.g. you could have a regex pattern like `ABC-(\d{4})-XYZ` 

- Reference field must conform the to pattern as above
- The numerical number in the middle can be captured, and used as the 'incrementing' portion
- The "next" reference can be intelligently constructed

For example if the latest reference is `ABC-123-XYZ` we can determine that the next one will be `ABC-124-XYZ`

However, you are right that this is all pretty complicated... I like the conciseness of the regex solution, but perhaps it is too cumbersome for the average user..

Can you suggest an alternative solution which allows some way to specify which part of the reference field is the "auto-incrementing" portion?

miggland:
Where is this incrementing done now, and with which input data?

SchrodingersGat:
> Where is this incrementing done now, and with which input data?

e.g. https://github.com/inventree/InvenTree/blob/d9efe27f8af4309e9f986e8a5c6e9dacc0e7f42c/InvenTree/build/models.py#L42

SchrodingersGat:
@miggland @matmair ERPNext does something like this

image

https://discuss.erpnext.com/t/erpnext-starting-number-for-next-purchase-invoice-purchase-order-sales-invoice-sales-order-etc/916

A replacement scheme like this might be suitable? It would require a lot of custom pattern matching / validation / edge case consideration

miggland:
The current proposal here seems to be missing two things in my view:
- the [increment function](https://github.com/inventree/InvenTree/blob/d9efe27f8af4309e9f986e8a5c6e9dacc0e7f42c/InvenTree/InvenTree/helpers.py#L166) could be fed a pattern to use instead of the very general one it uses now. That way, the function could know which part of the number to increment.
- the get_next_build_number function should replace things like 'YYYY', 'MM', 'DD' with the corresponding numbers
- after that, the checks put in place here make sure that the generated pattern matches the expectations. If this pattern is autogenerated, the user will not have to care much about the regexp defition.

In the ErpNext example above:
- Define PO numbers format as: `SI-.YY.MM.-.#####`
- InvenTree generates RegExp: `^SI-\d{2}\d{2}-\d{5}$`
- get_next_purchase_order_number (or similar) replaces YY and MM with 22, 06 when called
- increment function realises it should increment the last part of the number by the pattern it is fed.

For the last part, see my suggestion in your branch, @SchrodingersGat 

Would there be many edge cases? The number of possibilities is limited in my view:
| | User input | RegExp |
|--|--|--|
|Free text (verbatim)| "Order", "SI", "PO" | The same text |
|Year (4 digit) | YYYY | \d{4} |
|Year (2 digit) | YY | \d{2} |
|Month (2 digit) | MM | \d{2} |
|Day (2 digit) | DD | \d{2} |
| Number (2 digit) | NN | (\d{2}) | 
| Number (3 digit) | NN | (\d{3}) |
| Number (4 digit) | NN | (\d{4}) |
| Number (5 digit) | NN | (\d{5}) |

SchrodingersGat:
@miggland interesting ideas!

I would like the replacement scheme to be a bit more "pythonic" - rather than implementing something ourselves to replace `YY` and `MM` we could support the built-in datetime format codes: 

https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes

e.g.

`PO-%Y-%m-%d-#####` becomes `PO-2022-06-14-#####` and then we only have to handle the `#` replacement

miggland:
@SchrodingersGat that sounds reasonable; I was more on a conceptual level and stuck with the ERP example - using datetime formats makes sense.

However, I don't think allowing all possibilities makes sense (for example `%p %z %Z %c %X %x`)

matmair (approved):
Should be a good solution. I am still thinking about how this can be done with jinja - might not be worth the effort though.

SchrodingersGat:
Please continue discussion in https://github.com/inventree/InvenTree/pull/3267

PR #3175, opened 2022-06-10
Clean up tasks

### LINKED ISSUE #3157: FR: Reduce tasks
At least the task 'style' is redundant with pre-commit. I think we should clean up the tasks and remove everything we do not need to make it easier to keep an overview.

matmair:
This PR cleans up the available tasks.

Runs translation stats on updates now.
Removes multiple tasks:
- check
- shell
- style
- ~~import_fixtures~~

Small doc and structure updates.

Closes #3157

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3175"></a>

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -358,6 +358,53 @@ def delete_data(c, force=False):
         manage(c, 'flush')
 
 
+@task(post=[rebuild_models, rebuild_thumbnails])
+def import_fixtures(c):
+    """Import fixture data into the database.
+
+    This command imports all existing test fixture data into the database.
+
+    Warning:
+        - Intended for testing / development only!
+        - Running this command may overwrite existing database data!!
+        - Don't say you were not warned...
+    """
+    fixtures = [
+        # Build model
+        'build',
+
+        # Common models
+        'settings',
+
+        # Company model
+        'company',
+        'price_breaks',
+        'supplier_part',
+
+        # Order model
+        'order',
+
+        # Part model
+        'bom',
+        'category',
+        'params',
+        'part',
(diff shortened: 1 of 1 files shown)


matmair:
@SchrodingersGat this should be ready now - do you wish to keep any of those tasks?

SchrodingersGat:
The only one I am happy to remove is `check` - the others are pretty useful.

**style** - helpful to be able to run this one manually rather that only relying on the pre-commit hooks (which some users might not want to set up)
**shell** - I don't use this often, but it can be useful to shell in
**import-fixtures** - This is used by multiple CI steps, including the `inventree-python` and `inventree-app` repos

matmair:
> **style** - helpful to be able to run this one manually rather that only relying on the pre-commit hooks (which some users might not want to set up)

@SchrodingersGat Then we need to heavily improve that one - only covers about a third of the rules applied to the code base in the state it is right now. pre-commit can also be called with 'pre-commit run' - I do not see why somebody should set up all the dependencies and then stop at flake8 and its caller script.
Also, we need to find other options to check file endings, extra whitespaces etc.

> **shell** - I don't use this often, but it can be useful to shell in

This is just a passthrough - I do not really see the added benefit compared to just calling manage.py

> **import-fixtures** - This is used by multiple CI steps, including the `inventree-python` and `inventree-app` repos

Was already back in when you reviewed it - the test was failing

SchrodingersGat:
I'm happy to drop `style` if there is an alternative. Also happy to drop `shell` as you are right, manage.py provides the same functionality.

PR #3177, opened 2022-06-10
Task restructure

matmair:
Follow up to #3175

Heavily breaking - renames nearly all tasks and puts them into groups.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3177"></a>

### DIFF
--- a/.github/actions/migration/action.yaml
+++ b/.github/actions/migration/action.yaml
@@ -8,10 +8,10 @@ runs:
       - name: Data Import Export
         shell: bash
         run: |
-          invoke migrate
-          invoke import-fixtures
-          invoke export-records -f data.json
+          invoke main.migrate
+          invoke data.import
+          invoke data.export -f data.json
           python3 ./InvenTree/manage.py flush --noinput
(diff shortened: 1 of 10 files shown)


SchrodingersGat:
@matmair not sure I see the benefit here. This breaks a lot of existing workflows (in this and other repos, e.g. python bindings / app / etc) and also will require serious overhaul of the documentation. Is the current task structure missing any functionality that this change brings?

matmair:
@SchrodingersGat the idea is to restructure the commands in logical groups so it is better maintainable if/when it grows. The grouping into multiple namespaces also makes it easier for users to extend the commandset for custom deployments.
If this is is to heavy handed I would like to provide a leaner PR that removes the style command.

SchrodingersGat:
> If this is is to heavy handed I would like to provide a leaner PR that removes the style command.

I think this would be a better idea for now. I'd rather not make the other changes

matmair:
@SchrodingersGat I will create something new soon.

Issue #3168, opened 2022-06-10
[BUG] BOM table takes ages to load


SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/discussions/3166

Example URL: https://demo.inventree.org/part/112/

Bill of Materials takes ~5 seconds to load for this part.

Need to profile the DB / API queries

PR #3178, opened 2022-06-11
Setting caching

SchrodingersGat:
This PR adds a "caching" framework around loading settings from the database.

There are some annotated methods (such as `Part.full_name()` which rely on accessing settings from the DB (e.g. `PART_NAME_FORMAT`). This means that for a LIST API request, a DB hit is required for accessing each item in the list..

*Ironically this re-introduces the **1 + N** bug that the `full_name()` function was originally introduced to address!*

So, now we first query the cache to see if the setting has been recently used, which should reduce DB hits and radically improve query times.

Ref: https://github.com/inventree/InvenTree/issues/3168

### TODO

- [x] Add unit testing for the "caching" functionality
- [x] Ensure that saving settings updates the cache properly

### Future Work

This PR will form part of a larger effort to improve API efficiency, including:

- Add unit testing to the API to ensure the correct number of queries are being called
- Use profiling tools to reduce the number of queries made at various endpoints
- Improve query efficiency of the various endpoints

### Profiling

| URL | Without Caching | With Caching | Improvement |
| --- | --- | --- |---|
| /api/part/ |  2.36s | 1.3s | 45% |
| /api/stock/ | 3.2s | 2.6s | 20% |
| /api/bom/ | 2.1s |  1.4s | 33% |
| /index/ | 650ms | 483ms | 25% |
| /settings/ | 700ms | 550ms | 21% |

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3178"></a>

### DIFF
--- a/InvenTree/part/settings.py
+++ b/InvenTree/part/settings.py
@@ -5,34 +5,34 @@ from common.models import InvenTreeSetting
 
 def part_assembly_default():
     """Returns the default value for the 'assembly' field of a Part object."""
-    return InvenTreeSetting.get_setting('PART_ASSEMBLY', cache=False)
+    return InvenTreeSetting.get_setting('PART_ASSEMBLY')
 
 
 def part_template_default():
     """Returns the default value for the 'is_template' field of a Part object."""
-    return InvenTreeSetting.get_setting('PART_TEMPLATE', cache=False)
+    return InvenTreeSetting.get_setting('PART_TEMPLATE')
 
 
 def part_virtual_default():
     """Returns the default value for the 'is_virtual' field of Part object."""
-    return InvenTreeSetting.get_setting('PART_VIRTUAL', cache=False)
+    return InvenTreeSetting.get_setting('PART_VIRTUAL')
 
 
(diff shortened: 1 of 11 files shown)


InvenTree/part/settings.py
```diff
@@ -5,34 +5,34 @@
 
 def part_assembly_default():
     """Returns the default value for the 'assembly' field of a Part object."""
-    return InvenTreeSetting.get_setting('PART_ASSEMBLY')
+    return InvenTreeSetting.get_setting('PART_ASSEMBLY', cache=False)
```
matmair, line 8: Question: this seems like a prime candidate to cache: why is it disabled?

matmair (commented):
Should speed things up a noch

InvenTree/part/settings.py
SchrodingersGat: Ah, good catch. I did this to work out why the tests were failing, forgot to fix it

matmair: There is a bunch of those.

PR #3204, opened 2022-06-15
more user input sanitation

matmair:
This PR:

    Changes all APIs that inherit from the generic classes to use built in shims as a base - this will enabe us to make changes to them easier
    Adds a new mixin on all api views that change data to clean the user input using mozillas bleach
    Updates all custom create() and update() functions in the API views to only use cleaned user data

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3204"></a>

### DIFF
--- a/InvenTree/InvenTree/mixins.py
+++ b/InvenTree/InvenTree/mixins.py
@@ -1,0 +1,90 @@
+"""Mixins for (API) views in the whole project."""
+
+from bleach import clean
+from rest_framework import generics, status
+from rest_framework.response import Response
+
+
+class CleanMixin():
+    """Model mixin class which cleans inputs."""
+
+    # Define a map of fields avaialble for import
+    SAFE_FIELDS = {}
+
+    def create(self, request, *args, **kwargs):
+        """Override to clean data before processing it."""
+        serializer = self.get_serializer(data=self.clean_data(request.data))
+        serializer.is_valid(raise_exception=True)
+        self.perform_create(serializer)
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+
+    def update(self, request, *args, **kwargs):
+        """Override to clean data before processing it."""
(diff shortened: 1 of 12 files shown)


InvenTree/InvenTree/mixins.py
```diff
@@ -0,0 +1,90 @@
...
+
+        if getattr(instance, '_prefetched_objects_cache', None):
+            # If 'prefetch_related' has been applied to a queryset, we need to
+            # forcibly invalidate the prefetch cache on the instance.
+            instance._prefetched_objects_cache = {}
+
+        return Response(serializer.data)
+
+    def clean_data(self, data: dict) -> dict:
+        """Clean / snatize data.
```
SchrodingersGat, line 38: "sanitize"

InvenTree/order/api.py
```diff
@@ -114,7 +116,7 @@ class PurchaseOrderList(APIDownloadMixin, generics.ListCreateAPIView):
 
     def create(self, request, *args, **kwargs):
         """Save user information on create."""
-        serializer = self.get_serializer(data=request.data)
+        serializer = self.get_serializer(data=self.clean_data(request.data))
```
SchrodingersGat, line 119: Nice work catching this, I was concerned that any custom calls to `create` may have been missed.

SchrodingersGat (approved):
Very nice work @matmair - a clean implementation that will serve us well.

matmair:
@SchrodingersGat ready to merge now!

InvenTree/order/api.py
matmair: @SchrodingersGat I went through every single method called create or update and each line in those

InvenTree/InvenTree/mixins.py
matmair: @SchrodingersGat can you fix this in #3205 ?

SchrodingersGat: Done

PR #3246, opened 2022-06-23
Fix translation issue with javascript

SchrodingersGat:
As reported on crowdin, under some translations the user interface is broken, due to the inclusion of quote characters (e.g. `'`, `"`) in the translated strings.

This PR adds a custom translation parser which strips out quotes (and other *problematic* characters) before rendering to translated javascript files.

This solution ensures that the current translated strings (both "inputs" and "outputs") are untouched.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3246"></a>

*Once merged, this will also be cherry-picked onto the 0.7.x stable branch*

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -57,8 +57,8 @@ jobs:
           python3 check_js_templates.py
       - name: Lint Javascript Files
         run: |
-          invoke render-js-files
-          npx eslint js_tmp/*.js
+          python InvenTree/manage.py prerender
+          npx eslint InvenTree/InvenTree/static_i18n/i18n/*.js
 
   html:
     name: Style [HTML]
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -58,6 +58,8 @@ jobs:
       - name: Lint Javascript Files
         run: |
           python InvenTree/manage.py prerender
+          echo "Checking translated javascript files:"
+          ls InvenTree/InvenTree/static_i18n/i18n/
           npx eslint InvenTree/InvenTree/static_i18n/i18n/*.js
 
   html:
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -58,8 +58,6 @@ jobs:
       - name: Lint Javascript Files
         run: |
           python InvenTree/manage.py prerender
-          echo "Checking translated javascript files:"
-          ls InvenTree/InvenTree/static_i18n/i18n/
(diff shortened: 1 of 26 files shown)


InvenTree/templates/js/translated/tables.js
```diff
@@ -1,4 +1,4 @@
-{% load i18n %}
+{% load js_i18n %}
```
matmair, line 1: @SchrodingersGat have you looked into overriding the tag? That should be possible and reduce custom code.

matmair (approved):

matmair:
Reference: https://stackoverflow.com/a/51706505/17860466

SchrodingersGat:
I have now simplified the approach here, simply overriding the existing `i18n.py` file. Now, all templates which use the i18n tags will make use of the "safe" translations

.github/workflows/qc_checks.yaml
```diff
@@ -57,8 +57,10 @@ jobs:
           python3 check_js_templates.py
       - name: Lint Javascript Files
         run: |
-          invoke render-js-files
-          npx eslint js_tmp/*.js
+          python InvenTree/manage.py prerender
+          echo "Checking translated javascript files:"
+          ls InvenTree/InvenTree/static_i18n/i18n/
```
matmair, line 62: Question: is this still needed?

InvenTree/part/templatetags/i18n.py
```diff
@@ -0,0 +1,104 @@
...
+                    option,
+                )
+            )
+        seen.add(option)
+
+    return CustomTranslateNode(message_string, noop, asvar, message_context)
+
+
+# Re-register tags which we have not explicitly overridden
+register.tag("blocktrans", django.templatetags.i18n.do_block_translate)
```
matmair, line 90: Suggestion: django.tenplate.defaulttags already registers the default tags - why not just reuse that? Would save on LoC

matmair (approved):
I had a few comments but it should be good as-is

.github/workflows/qc_checks.yaml
SchrodingersGat: Otherwise where are the javascript files linted?

matmair: @schroedingersgat but what does line 61 do in that respect? Maybe I am overlooking something but that just fills the logs with a directory  content list.

InvenTree/part/templatetags/i18n.py
matmair: @schrodingersgat I am mobile right now - will.provide a PR in a few hours.

matmair (approved):

PR #3250, opened 2022-06-24
Remove custom javascript from auth pages

SchrodingersGat:
- Unauthorized user cannot load these scripts
- Simply throws console errors

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3250"></a>

e.g.

image

### DIFF
--- a/InvenTree/templates/account/base.html
+++ b/InvenTree/templates/account/base.html
@@ -89,5 +89,21 @@
 <!-- general JS -->
 {% include "third_party_js.html" %}
 
+<script type='text/javascript' src='{% static "script/inventree/message.js" %}'></script>
+
+<script type='text/javascript'>
+
+$(document).ready(function () {
+
+    {% if messages %}
+    {% for message in messages %}
+        showMessage(messsage);
+    {% endfor %}
+    {% endif %}
+
+});
+
+</script>
+
 </body>
 </html>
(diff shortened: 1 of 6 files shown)


InvenTree/templates/account/base.html
```diff
@@ -89,30 +89,5 @@
...
-<script type='text/javascript' src="{% i18n_static 'notification.js' %}"></script>
-
-
-<script type='text/javascript'>
-
-$(document).ready(function () {
-    // notifications
-    {% if messages %}
-    {% for message in messages %}
-    showAlertOrCache(
```
matmair, line 102: We will be missing sign-out / going into maintenance alters without this. That is a heavy regression in my eye.

matmair (changes requested):

InvenTree/templates/account/base.html
SchrodingersGat: Might have to look into splitting the javascript files then, into once which are "safe" for non-logged-in-users to view.

SchrodingersGat: @matmair LMK what you think of the new updates. I have put the messages back in, but removed any ability for the user to see sensitive JS files which require login.

matmair (approved):

matmair (approved):

PR #3259, opened 2022-06-26
Adds notification when items are received against a purcahse order

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/2950

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3259"></a>


matmair (approved):
Seems like we could reduce the number of HTML templates further but looks good for merge in this state.

PR #3263, opened 2022-06-28
[BUG] Duplication of "base URL" setting

### LINKED ISSUE #3196: [BUG] Duplication of "base URL" setting
Currently there are two distinct places to set the base URL for the server, and the two values are functionally distinct:

image

image

The second option was added when added support for django-allauth, but did not consider the interplay with the existing setting!

What we should do:

- Remove the custom global parameter `INVENTREE_BASE_URL` in favour of the option provided by django-allauth
- Expose the django-allauth settings to the "settings" page (rather than requiring the user to navigate to the admin section)
- Refactor any existing code that uses `INVENTREE_BASE_URL` to use the domain name setting configured in django-allauth

Ref: https://github.com/inventree/InvenTree/discussions/3195

matmair:
Fixes #3196

Adds after_safe key to settings: This provided callable is executed after the setting is saved.
The current (=>new) setting object is supplied to the function as the first argument.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3263"></a>

### DIFF
--- a/InvenTree/InvenTree/tests.py
+++ b/InvenTree/InvenTree/tests.py
@@ -23,6 +23,7 @@ from stock.models import StockLocation
 
 from . import config, helpers, ready, status, version
 from .validators import validate_overage, validate_part_name
+from django.contrib.sites.models import Site
 
 
 class ValidatorTest(TestCase):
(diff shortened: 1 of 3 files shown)


matmair:
@SchrodingersGat I implemented your suggested solution. Better to have a 95% solution on this than nothing at all.

SchrodingersGat:
> @SchrodingersGat I implemented your suggested solution. Better to have a 95% solution on this than nothing at all.

Pragmatic :)

SchrodingersGat (approved):

PR #3228, opened 2022-06-18
Update requirements

matmair:
This PR upgrades the way dependencies are handled.
## What?
Dependencies are described with their version requirements. If there is a need to pin a package to a specific or under/over a specific version that is defined, else the newest is used.
## How?
 The requirements are defined in a *.in file and then compiled by pip-tools to a normal *.txt file with each version pinned. So each time you want to update all requirements you just run the command 'pip-compile' and the most current versions matching is written to the *txt file.

## Changes
- Split requirements
- Update to pip-tools
- Remove pinning to the newest version
- Rewrite the version pinning where it is known

## Open questions
Do we need this package?
- django-admin-shell (also update from 0.1.2 -> [1.0.0](https://pypi.org/project/django-admin-shell/))
- django-cleanup (also update from 5.1.0 -> [6.0.0](https://pypi.org/project/django-cleanup/))
- pygments (also update from 2.7.4 -> [2.12.0](https://pypi.org/project/Pygments/))

Bigger updates:
- coverage 5.3 -> [6.4.1](https://pypi.org/project/coverage/)
- coveralls 2.1.2 -> [3.3.1](https://pypi.org/project/coveralls/)
- flake8 3.8.3 -> [4.0.1](https://pypi.org/project/flake8/)
- django-crispy-forms 1.11.2 -> [1.14.0](https://pypi.org/project/django-crispy-forms/)
- django-filter 2.4.0 -> [22.1](https://pypi.org/project/django-filter/)
- django-import-export 2.5.0 -> [2.8.0](https://pypi.org/project/django-import-export/)
- ~~django-markdownify 0.8.0 -> [0.9.2](https://pypi.org/project/django-markdownify/)~~
- django-mptt 0.11.0 -> [0.13.4](https://pypi.org/project/django-mptt/)
- django-sql-utils 0.5.0 -> [0.6.1](https://pypi.org/project/django-sql-utils/)
- django-stdimage 5.1.1 -> [6.0.1](https://pypi.org/project/django-stdimage/)
- qrcode[pil] 6.1 -> [7.3.1](https://pypi.org/project/qrcode/)
- rapidfuzz 0.7.6 -> [2.0.11](https://pypi.org/project/rapidfuzz/)

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3228"></a>

PR #3273, opened 2022-06-29
Pin django-stdimage to sub 6.0.0

SchrodingersGat:
- 6.0.0 breaks model serialization for background worker task
- NFI why it does this

Ref: https://github.com/inventree/InvenTree/pull/3228

@matmair the recent update to requirements.txt boosted django-stdimage to [> 6.0.0](https://github.com/codingjoe/django-stdimage/releases) - which is a breaking change.

For some reason, this breaks model serialization (for django-q) for instances which do not have an image field (or have an empty image). This in turn causes task offloading to fail.

Very "lucky" to have a set of circumstances that conspired to even show this error up!

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3273"></a>


SchrodingersGat:
@matmair what do I need to do to get the pre-commit checks to work? I have updated the requirements.in file and the pre-commit checks work OK on my local PC..

matmair:
@SchrodingersGat and the requirements.txt was not updated? Maybe run the dev setup command again - will provide a PR to fix this branch.
Any idea why our test did not catch this one?

matmair:
@SchrodingersGat should be good to go

SchrodingersGat:
> Any idea why our test did not catch this one?

Our existing unit tests did not ever test task offloading. I have added a very simple unit test, which I have confirmed breaks with `django-stdimage==6.0.0`

matmair:
The more tests the better @SchrodingersGat 👍

PR #3282, opened 2022-07-02
Use metadata in projects for plugin info

### LINKED ISSUE #3090: [FR] Switch to pyproject.toml
**Problem**
The new style for python projects is to use a *.toml file instead of setup.py + setup.cfg

**Suggested solution**
Switch to the newer format. Also for all built-in plugins and add support for them in plugins with a mixin that get's all its metadata from there.

**Examples of other systems**
https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/

Yes

matmair:
This enables the usage of package metadata for plugin author, website, and description.

Closes #3090

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3282"></a>


matmair:
@SchrodingersGat would be cool to get this into the 0.8.0 release. Reduces code complexity for plugins a bit (no more version.py file - just package metadata).

SchrodingersGat:
@matmair please see my previous comment - what happens when the user has hard-coded these values e.g. `DESCRIPTION` - could this be used as a fallback?

matmair:
@SchrodingersGat sorry if I am looking in the wrong place but I can not see a previous comment from you either in this PR nor in the FR #3000 .

Currently behavior would be to overwrite the hardcoded values. I can see how that could be a problem. I will change it to be metadata fallback rather then take precedence.

matmair:
@SchrodingersGat should be ready now

SchrodingersGat (approved):

PR #3284, opened 2022-07-02
Add bleach

matmair:
* use shims for API view inheritation

* Add mixin for input sanitation

* fix clean operation to fix all string values

* Also clean up dicts
this is to future-proof this function

* Update docstirng

* proof custom methods against XSS through authenticated users

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3284"></a>


SchrodingersGat:
A whirlwind of activity but zero lines changed. Any input required from me?

matmair:
@SchrodingersGat we already merged those. For some reason I had a branch laying around that did not catch up.

PR #3296, opened 2022-07-05
Allow searching of BOM List API endpoint

SchrodingersGat:


matmair (approved):
Nice addition. Curios we did not have that before.

PR #3303, opened 2022-07-06
Use python executable we were invoked with

sur5r:
Relying on PATH will disrespect the virtualenv when called from outside.

I stumbled across this while writing an ansible role for InvenTree (yet to be published).

These changes allow for invocation of `invoke` without setting PATH first.

I was unsure whether to change the `pre-commit` calls as well, that's why this is marked as draft.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3303"></a>

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -79,8 +79,9 @@ def manage(c, cmd, pty: bool = False):
         cmd: Django command to run.
         pty (bool, optional): Run an interactive session. Defaults to False.
     """
-    c.run('cd "{path}" && python3 manage.py {cmd}'.format(
+    c.run('cd "{path}" && {python} manage.py {cmd}'.format(
         path=managePyDir(),
+        python=sys.executable,
         cmd=cmd
     ), pty=pty)
 
@@ -96,7 +97,7 @@ def plugins(c):
     print(f"Installing plugin packages from '{plugin_file}'")
 
     # Install the plugins
-    c.run(f"pip3 install --disable-pip-version-check -U -r '{plugin_file}'")
+    c.run(f"{sys.executable} -m pip install --disable-pip-version-check -U -r '{plugin_file}'")
 
 
 @task(post=[plugins])
@@ -105,7 +106,7 @@ def install(c):
     print("Installing required python packages from 'requirements.txt'")
 
     # Install required Python packages with PIP
-    c.run('pip3 install --no-cache-dir --disable-pip-version-check -U -r requirements.txt')
+    c.run(f'{sys.executable} -m pip install --no-cache-dir --disable-pip-version-check -U -r requirements.txt')
 
 
 @task
@@ -114,7 +115,7 @@ def setup_dev(c):
     print("Installing required python packages from 'requirements-dev.txt'")
 
     # Install required Python packages with PIP
-    c.run('pip3 install -U -r requirements-dev.txt')
+    c.run(f'{sys.executable} -m pip install -U -r requirements-dev.txt')
 
     # Install pre-commit hook
     c.run('pre-commit install')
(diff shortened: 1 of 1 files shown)


matmair:
@sur5r as far as I understand pre-commit it sets up its own venv for its calls. Am I wrong here?

Ansible roll sound like something awesome. I just execute the tasks with custom scheduled tasks and webhooks right now. Scales reasonable and you don't need to keep ssh open.

matmair (approved):

SchrodingersGat:
Thanks for this improvement @sur5r

SchrodingersGat:
Instrumenting the call to `invoke server` shows that the wrong python executable is being called when we use `sys.executable` in the docker context..

image

image

This *should be* `/home/inventree/dev/env/bin/python`

SchrodingersGat:
Ah, so in the docker context, `invoke` is installed into the system python path:

```
pip show invoke
Name: invoke  
Version: 1.7.1
Summary: Pythonic task execution
Home-page: https://pyinvoke.org
Author: Jeff Forcier
Author-email: jeff@bitprophet.org
License: BSD
Location: /usr/local/lib/python3.9/site-packages
```

This means that any calls to `invoke` are passing the system python through as `sys.executable`

As django is not installed in a path accessible from system python, this is failing! Not sure how we work around this.. any ideas?

For what it is worth, calling `python -m invoke` instead of `invoke` works, but that is clunky and I don't want to have to resort to that.

SchrodingersGat:
@sur5r unless you can provide a better suggestion I would like to roll this commit back. The current solution might work fine in your application but it causes some bigger issues with docker. Perhaps there is another way we can solve what you are trying to achieve? Please let me know ASAP

matmair:
@SchrodingersGat we could install invoke into the venv in docker. That should solve that problem.

SchrodingersGat:
That's true, I was considering this approach. As `invoke` is already installed as part of the base docker image, is there a way (using pip) to *force* another copy of invoke to be installed into the venv?

sur5r:
Uh, that's tricky, indeed. I always had invoke installed in the venv and used that one.

If I'm reading the Dockerfile correctly, it does system-wide `pip install` runs. Is there a reason why no virtualenv is used here?

sur5r:
> Yes, in the "development" setup the virtualenv is setup at run-time into a mounted volume. This is so you can make changes to the venv while developing.. It is intentional, and difficult to code around with this new change!

I only partially understand this but I obviously don't want to break your development workflow.

> Can you explain a bit more about your particular use case? Essentially this PR has pushed the problem from the python executable to the invoke entrypoint :p

I am creating an ansible role to set up and update InvenTree. This currently contains this as a handler:

```
- name: inventree_update
  become: true
  become_user: inventree
  community.general.django_manage:
    virtualenv: /srv/inventree/venv
    project_path: /srv/inventree/src/InvenTree
    command: "{{ item }}"
  with_items:
    - makemigrations
    - migrate --noinput
    - migrate --run-syncdb
    - check
    - prerender
    - collectstatic --no-input
    - clean_settings
```

As you can see, I basically rebuilt `inv update` in there. While this works now, it might break in subtile ways if  `inv update` changes. I consider this a workaround as I duplicated code here.

While ansible provides an elegant way to call django manage commands, it doesn't really do so for arbitrary venv commands.

I originally just called `/srv/inventree/venv/bin/invoke update` as a plain shell command there (with cwd being `/srv/inventree/src/`). While the shebang of that invoke script points to the venv, the shell commands will fall back to the system-wide python installation, which then fail as there's e.g. no system-wide django.

So in the end it seems the docker setup relies on two python environments to interact while this change enforces a single environment.

So unless there's a good way to adapt the docker setup to this change I think this is best reverted and I adapt my ansible role accordingly.

sur5r:
Makes sense. Sorry for the trouble. I will try and find a different way to solve this from the ansible side of things.

matmair:
@sur5r I think you found an interesting problem here...
A reliable way to run maintenance scripts etc. on the environment would be nice. Please let us know if you come up with a Ansible role, I would love to add a link to the docs.

PR #3322, opened 2022-07-14
Migration bug fix

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/3307

Fixes a bug in data migrations where integer coercion of string values could overflow the "maximum value" of the integer fields.
These values are now clipped in the data migrations

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3322"></a>


SchrodingersGat:
Closing, branched from the wrong commit..

PR #3328, opened 2022-07-14
Update unit tests for API plugin mixin class

SchrodingersGat:
- API at previous target URL has changed
- Simplier to use the github API as a test case

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3328"></a>

### DIFF
--- a/InvenTree/plugin/base/integration/test_mixins.py
+++ b/InvenTree/plugin/base/integration/test_mixins.py
@@ -192,6 +192,11 @@ class APICallMixinTest(BaseMixinDefinition, TestCase):
             API_URL_SETTING = 'API_URL'
             API_TOKEN_SETTING = 'API_TOKEN'
 
+            @property
+            def api_url(self):
+                """Override API URL for this test"""
+                return "https://api.github.com"
+
             def get_external_url(self, simple: bool = True):
                 """Returns data from the sample endpoint."""
                 return self.api_call('orgs/inventree', simple_response=simple)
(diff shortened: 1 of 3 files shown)


InvenTree/plugin/base/integration/test_mixins.py
```diff
@@ -232,30 +242,32 @@ def test_api_call(self):
...
+        result = self.mixin.api_call('https://api.github.com/orgs/inventree', endpoint_is_url=True)
         self.assertTrue(result)
 
         # api_call with post and data
         result = self.mixin.api_call(
-            'api/users/',
-            data={"name": "morpheus", "job": "leader"},
-            method='POST'
+            'repos/inventree/InvenTree',
+            method='GET'
```
matmair, line 261: @SchrodingersGat this test is not really doing anything valuable anymore as the previous implementation did a POST request to ensure that works too. Could we change that back to ensure all methods work?

matmair (commented):
@SchrodingersGat was afk for a while - I think these changes weakened coverage meaning for the tests.

InvenTree/plugin/base/integration/test_mixins.py
SchrodingersGat: Can do, I was needing the tests to pass and waiting for your return.

matmair: Should be resolved in https://github.com/inventree/InvenTree/pull/3328

PR #3338, opened 2022-07-15
Do not show low/no stock warning for template parts

sur5r:
This hides the "No Stock"/"Low Stock" label for parts marked as template.

AIUI, template parts will never exist in stock, so showing the warning makes no sense.

If this is not how template parts are meant to be used then this could depend on `part.active` instead.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3338"></a>

### DIFF
--- a/InvenTree/part/templates/part/stock_count.html
+++ b/InvenTree/part/templates/part/stock_count.html
@@ -3,8 +3,10 @@
 
 {% decimal total_stock %}
 
+{% if not part.is_template %}
 {% if total_stock == 0 %}
 <span class='badge badge-right rounded-pill bg-danger'>{% trans "No Stock" %}</span>
 {% elif total_stock < part.minimum_stock %}
 <span class='badge badge-right rounded-pill bg-warning'>{% trans "Low Stock" %}</span>
+{% endif %}
 {% endif %}
--- a/InvenTree/templates/js/translated/part.js
+++ b/InvenTree/templates/js/translated/part.js
@@ -578,7 +578,7 @@ function partStockLabel(part, options={}) {
         // There IS stock available for this part
 
         // Is stock "low" (below the 'minimum_stock' quantity)?
-        if ((part.minimum_stock > 0) && (part.minimum_stock > part.in_stock)) {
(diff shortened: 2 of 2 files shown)


matmair:
@sur5r I am not sure how other users use this but I have stock for template parts. A lot of my screws are template parts so that similar screws can be made faster for example.

So this should be a setting in my opinion.

SchrodingersGat:
I agree that this should be a setting. Something like `HIDE_LOW_STOCK_FOR_TEMPLATE_PARTS` (Default=False)

sur5r:
Makes sense. Would you like this to be a setting in the Web UI or in the YAML file? What are your thoughts about inactive parts? I think they could use a similar setting.

Speaking of the YAML file: I noticed the template is missing a section for the redis cache. Do you want me to add one in a separate PR?

matmair:
@sur5r this for sure would be something for the ui. We try to place everything possible in the database, not the yaml. The file can be difficult to change in bigger orgs.

If you have relevant settings we should place in the yaml please make a PR, we are always happy to discuss changes.

SchrodingersGat:
> Would you like this to be a setting in the Web UI or in the YAML file?

Web UI for sure. LMK if you need some help sorting out the code

> What are your thoughts about inactive parts? I think they could use a similar setting.

Agreed, this would make sense as a separate setting.

SchrodingersGat:
@sur5r do you need any further input on this from us?

sur5r:
I think I found everything, but it will be roughly a week from now before I can work on inventree again.

SchrodingersGat:
@sur5r please ping us if you need any further assistance on this.

SchrodingersGat:
Hey @sur5r any updates on this one?

sur5r:
I unfortunately don't have time to work on InvenTree ATM. I hope to find some time in December.

matmair:
not stale

matmair:
@sur5r would you be ok with me bringing this over the finishing line?

sur5r:
Yes. Sorry it didn't work out time-wise.

Issue #3327, opened 2022-07-14
[FR] :checkered_flag: Add last updated column for supplier parts


Zontex:
**Is your feature request the result of a bug?**

No

**Problem**

Some supplier links are constantly updated, need to replace the old link with the new link and/or add a new price for the component. if can go to the supplier parts page and see when each part was last updated it will give an indication over which parts need to be checked and when especially in a process of mass updating components if I lose track I have no idea to know which I already updated and which not.

<img width="1542" alt="Screen Shot 2022-07-14 at 14 22 35" src="https://user-images.githubusercontent.com/19342533/178972093-655cf3d0-a5be-488e-b5bd-e615750c8937.png">

**Suggested solution**

Add a last-updated column, I believe this kind of last-updated already exists in some inventory datasets inside the data for things like pricing, etc ...

**Describe alternatives you've considered**

No automated alternative for it using scripting as I see.

matmair:
@Zontex metadata could be added with a script / plugin to achieve this.
Would you be willing to submit this to core?

SchrodingersGat:
Adding an "updated_on" field to the `SupplierPart` table would be trivial, however this would only cover the fields in the model itself. Pricing data is stored in the `SupplierPriceBreak` table and would not be covered by this simple addition. Might need some more thought than this

matmair:
Not stale

Zontex:
Not stale

matmair:
@Zontex this should be fixed by #4214 - do you mind checking if this covers your request?

Zontex:
@matmair yes looks perfect, shall we close this issue?

matmair:
@Zontex great to hear! I war not sure if there was context discussed I forgot - therefore I tagged you. The issue should close automatically once the PR is merged.

PR #3345, opened 2022-07-17
Last updated tracking for supplier parts

nwns:
Add last updated tracking for supplier parts and the supplier price breaks as discussed in #3327.

Initial implementation was as a property on the model, which worked but was not sortable.  In the attempt to make the sortable part work, I am getting Internal Server Errors but I cannot figure out why the dev server is not printing the stack trace.

Even forcing Debug mode and Logging to DEBUG using `INVENTREE_LOG_LEVEL=DEBUG INVENTREE_DEBUG=True invoke server`, all it outputs is `Internal Server Error` and then the path.  The response to the broswer is slightly more useful in that it has the repr of the exception, but without a stack trace I can't find what is failing.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3345"></a>


InvenTree/company/models.py
```diff
@@ -647,6 +657,19 @@ def __str__(self):
 
         return s
 
+    # @property
+    # def last_updated(self) -> datetime:
```
matmair, line 661: Question: Why is this commented out?

InvenTree/company/serializers.py
```diff
@@ -250,29 +253,31 @@ class Meta:
 
         model = SupplierPart
         fields = [
-            'available',
             'availability_updated',
+            'available',
```
matmair, line 257: Question: Why were these fields reordered? There is no technical reason I know for ordering an array in python.

InvenTree/part/filters.py
```diff
@@ -71,6 +71,22 @@ def annotate_total_stock(reference: str = ''):
     )
 
 
+def annotate_last_updated(reference: str = ''):
```
matmair, line 74: Question: What is `reference` used for?

InvenTree/part/filters.py
```diff
@@ -71,6 +71,22 @@ def annotate_total_stock(reference: str = ''):
...
+
+    - Include supplier part's updated field
+    - Include supplier part's availability_updated field
+    - Include supplier part's price breaks
+    """
+
+    return models.Max(
+        F('updated'),
+        F('availability_updated'),
+        # TODO Add price breaks
```
matmair, line 85: Question: What do you want to order by on price breaks?

InvenTree/common/models.py
```diff
@@ -1664,6 +1664,10 @@ class Meta:
         help_text=_('Unit price at specified quantity'),
     )
 
+    updated = models.DateTimeField(
+        auto_now=True, verbose_name=_("last updated")
```
matmair, line 1668: Suggestion(nitpick - non-blocking): If you put each property on a separate line and add a komma at the end future diffs will be smaller / cleaner to read.

InvenTree/company/models.py
```diff
@@ -392,20 +392,26 @@ def get_queryset(self):
...
+    A Part may be available from multiple suppliers.
 
     Attributes:
-        part: Link to the master Part (Obsolete)
-        source_item: The sourcing item linked to this SupplierPart instance
-        supplier: Company that supplies this SupplierPart object
-        SKU: Stock keeping unit (supplier part number)
-        link: Link to external website for this supplier part
+        base_cost: Base charge added to order independent of quantity
+            e.g. "Reeling Fee"
```
matmair, line 403: Question(non-blocking): Why sit the description reordered?

matmair (approved):
Hi there @nwns 
Thanks for the contribution. I added several questions and I will look into the errors.
Suggestion (non-blocking): If you are able/willing to add some unstetestes for the newly added filter working that would be appreciated.

InvenTree/common/models.py
```diff
@@ -1664,6 +1664,10 @@ class Meta:
         help_text=_('Unit price at specified quantity'),
     )
 
+    updated = models.DateTimeField(
```
matmair, line 1667: @nwns I think I understand the issues: There should be the option for empty entries so that already existing entries are still valid. So something like the following should be added (please Errol the migrations after that):

```
blank=true,
null=true,
```

matmair (changes requested):

InvenTree/company/api.py
```diff
@@ -313,6 +313,12 @@ def filter_queryset(self, queryset):
             active = str2bool(active)
             queryset = queryset.filter(part__active=active)
 
+        # Order by last_updated?
```
SchrodingersGat, line 316: Ordering should not be handled in the `filter_queryset` method. You should add "last_updated" to the `ordering_fields` list (search for "ordering_fields" in `api.py`)

InvenTree/company/migrations/0048_alter_supplierpricebreak_updated.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.2.14 on 2022-07-17 15:14
```
SchrodingersGat, line 1: There is no need for two separate migrations here, if it can be helped:

The following steps should merge the two migrations into one.

```
cd InvenTree
python ./manage.py migrate company 0046
rm ./company/migrations/0047_supplierpart_updated.py
rm ./company/migrations/0048_alter_supplierpricebreak_updated.py
python ./manage.py makemigrations company
python ./manage.py migrate company
```

InvenTree/part/filters.py
nwns: No idea.  `annotate_total_stock` had it, so I did the same.  As noted in the original post, I don't know if I need it because the code crashes but doesn't give me a stack trace.

InvenTree/part/filters.py
nwns: Because a price break may be the most recent update to the supplier part.

InvenTree/common/models.py
nwns: Yeah, this is what my default black settings produced.  I will need to see if there is an option that will fix this (or I need to change my formatter to whatever Inventree uses).

InvenTree/common/models.py
nwns: I don't think this should be nullable.  There is always a last update.  Also, the migrations put the current time in if the field was NULL.

InvenTree/company/models.py
matmair: Please remove that code then as we tend to not merge not used code.

matmair:
@nwns not really sure but as the migrations are fundamentally breaking the whole thing I would start by fixing them.

matmair:
@nwns after fixing the migrations as suggested I get normal trackback written to the database and a notification informing me of the new error. To me the filter implementation seems to be the problem as @SchrodingersGat suggested.
image

matmair:
@nwns I made a PR to your branch that works. Do you intend to submit this or should I make a new rewritten PR?

SchrodingersGat:
@nwns ping! Waiting for feedback on a number of points here. Let us know when you're able to get back to this.

PR #3348, opened 2022-07-17
Allow installation of local plugins via docker

### LINKED ISSUE #3343: [FR] Allow installation of local plugins via docker
Local install of plugins (i.e. local source files rather than installation via PIP) is currently not possible for docker.

This is because the local plugins directory is fixed, relative to the location of the source code.

Add a PR to:

- Make the 'plugin' source directory configurable
- Place it in the external volume directory under docker
- Create the directory on startup if does not exist

matmair:
This PR adds the ability to mount a custom plugin dir to a docker deployment.
Adds the suggestions of https://github.com/inventree/InvenTree/discussions/3339

Closes #3343

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -55,6 +55,7 @@ services:
         volumes:
             # Ensure you specify the location of the 'src' directory at the end of this file
             - inventree_src:/home/inventree
+            - inventree_plugins:/home/inventree/InvenTree/plugins
         env_file:
             - .env
         restart: unless-stopped
@@ -69,6 +70,7 @@ services:
         volumes:
             # Ensure you specify the location of the 'src' directory at the end of this file
             - inventree_src:/home/inventree
+            - inventree_plugins:/home/inventree/InvenTree/plugins
         env_file:
             - .env
         restart: unless-stopped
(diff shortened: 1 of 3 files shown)


matmair:
@SchrodingersGat @nwns this should be tested a bit more... I am not really deploying with docker that much.

SchrodingersGat:
Looks good @matmair - have you tested this implementation? I did not have time to actually check if my suggestion worked as intented.

matmair:
@SchrodingersGat this works on my local testing server (debian, custom auth-stack)

luwol03:
Its working for me.

SchrodingersGat:
@matmair this has broken CI docker build: https://github.com/inventree/InvenTree/runs/7399026732?check_suite_focus=true

I am also wondering if there is a "cleaner" way to achieve this, I'd love to keep everything in a single external volume. I'll see if there's any way of simplifying this

matmair:
@SchrodingersGat we could also add the possibility to mount a plugin dir with an environment variable. The plugin system support multi-dir.

SchrodingersGat:
@matmair I would like that approach. The dockerfile already contains a skeleton for it (`INVENTREE_PLUGIN_DIR`).

I tried a naïve implementation yesterday which *almost* worked but it complained that the loaded directory was outside the core django path. I think it is a template issue?

matmair:
@SchrodingersGat that might he a namespace issue. Do you have some code? Probably easy to fix.

PR #3357, opened 2022-07-19
Docker compose plugins fix

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/3348

- This PR reverts #3348 to fix the docker builds in CI
- A new solution will be provided soon
- Also provides some simplification for the InvenTree dockerfile

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3357"></a>


Dockerfile
```diff
@@ -127,20 +127,9 @@ FROM base as dev
...
 # Location for python virtual environment
 # If the INVENTREE_PY_ENV variable is set, the entrypoint script will use it!
-ENV INVENTREE_PY_ENV="${INVENTREE_DEV_DIR}/env"
-
-# Override default path settings
-ENV INVENTREE_STATIC_ROOT="${INVENTREE_DEV_DIR}/static"
-ENV INVENTREE_MEDIA_ROOT="${INVENTREE_DEV_DIR}/media"
-ENV INVENTREE_PLUGIN_DIR="${INVENTREE_DEV_DIR}/plugins"
-
-ENV INVENTREE_CONFIG_FILE="${INVENTREE_DEV_DIR}/config.yaml"
```
matmair, line 141: @SchrodingersGat are you sure we do not need any of those variables?

matmair (approved):

Dockerfile
SchrodingersGat: These variables are already defined at the top of the dockerfile. We were (previously) re-defining them for the development build but there is no longer any need to do this.

Issue #3287, opened 2022-07-03
[BUG] Internal server error after activating barcodes plugin


sur5r:
I just enabled plugin support on an instance. After restarting both uwsgi and the worker, I enabled the builtin barcodes plugin.

Now this gets logged when trying to access settings:

```
Internal Server Error: /settings/
Traceback (most recent call last):
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/base.py", line 829, in _resolve_lookup
    current = current[bit]
TypeError: 'InvenTreeBarcodePlugin' object is not subscriptable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/core/handlers/base.py", line 204, in _get_response
    response = response.render()
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/response.py", line 105, in render
    self.content = self.rendered_content
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/response.py", line 83, in rendered_content
    return template.render(context, self._request)
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/base.py", line 170, in render
    return self._render(context)
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/loader_tags.py", line 150, in render
    return compiled_parent._render(context)
  File "/home/inventree/.virtualenvs/inventree/lib/python3.9/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
[...]

PR #3364, opened 2022-07-20
Allow plugin loading from external directory

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/3348

Fixes https://github.com/inventree/InvenTree/issues/3287

This PR allows plugins to be loaded from an external directory, which provides an easier workflow for users, especially when running under docker.

- Looks for the `INVENTREE_PLUGIN_DIR` environment variable
- Loads plugins from that directory (at runtime)

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3364"></a>

### TODO

- [x] Documentation - update plugin docs with information on external directory - https://github.com/inventree/inventree-docs/pull/322
- [ ] Unit testing for new features

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -15,7 +15,6 @@ import random
 import socket
 import string
 import sys
-from pathlib import Path
 
 import django.conf.locale
 from django.core.files.storage import default_storage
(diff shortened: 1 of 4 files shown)


SchrodingersGat:
@matmair I worked out the error, it was simply that the plugin metadata code expected the path to be relative. Added some extra code to handle exceptions to that..

matmair:
@SchrodingersGat looks great!
We should def add unit tests for this case.

InvenTree/InvenTree/settings.py
```diff
@@ -938,10 +938,44 @@ def _is_true(x):
...
+
+            if not os.path.exists(init_filename):
+                try:
+                    with open(init_filename, 'w') as init_file:
+                        init_file.write("# InvenTree plugin directory\n")
+                except Exception:
+                    logger.error(f"Could not create file '{init_filename}'")
+                    continue
+
+            if os.path.exists(pd) and os.path.isdir(pd):
```
matmair, line 972: Suggestion: `pathlib.Path` is a bit easier to read and handles linux / mac / win paths better- should we switch to it for this? I would be glad to provide a patch for code.

matmair (approved):
Clean change!

InvenTree/InvenTree/settings.py
SchrodingersGat: @matmair yes please add a patch to this branch

SchrodingersGat:
> @SchrodingersGat looks great! We should def add unit tests for this case.

I am not immediately sure how we could achieve this.. Do you have an appropriate set of tests in mind?

InvenTree/plugin/registry.py
```diff
@@ -187,6 +187,53 @@ def reload_plugins(self, full_reload: bool = False):
 
         logger.info('Finished reloading plugins')
 
+    def plugin_dirs(self):
```
matmair, line 190: Suggestion: @SchrodingersGat should we maybe move this out of the class? Would keep the functions cleaner.

matmair (approved):

InvenTree/plugin/registry.py
SchrodingersGat: Happy to, I moved it out of settings.py because that requires a full reload to update (and thus can't be unit tested).

Still having trouble with testing plugins from external dirs.

luwol03:
I suggest also documenting how to load local plugins now in docker in the documentation.

InvenTree/plugin/registry.py
matmair: @SchrodingersGat a lot of plugin stuff needs reloads - you can easily provoke that with `registry.reload_plugins()`.

matmair:
@luwol03 outside of development I really recommend to not use local plugin dirs. You will only have problems with updates, environments changing and general file system problems.

InvenTree/plugin/registry.py
SchrodingersGat: @matmair except that the only place where the environment variables are inspected is in `settings.py` - so you can't "mock" environment vars and then call `reload_plugins()` - because the environment vars are not re-processed. Does this make sense?

SchrodingersGat:
@matmair I'm going to merge in as is but we should revisit for adding unit tests.

PR #3370, opened 2022-07-20
Remove special characters restriction in category names

### LINKED ISSUE #3358: [FR]Special characters restriction in category names
"Hi folks,

I tried to copy over the category tree from a webshop called TME via API but I realized that some of the categories cannot be added because they have illegal characters in the names.
By looking the code I found the limitation:
https://github.com/inventree/InvenTree/blob/master/InvenTree/InvenTree/validators.py#L84
What is the motivation behind of them? Having category names like 'D/A converters' or comma separated category names could be useful I think in many cases.

matmair:
Fixes #3358

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3370"></a>


martonmiklos:
> Tested for XSS-safety: bleach works! @martonmiklos there are still _some_ restrictions for security reasons but it should be a bit more free. If you press the `open in Gitpod` button an enviroment with a set up InvenTree will be made available.

Sweet!

I just pulled all the 1661 categories from the TME with Kintree to my Inventree instance without any issues.
I am still novice to gitpod, so I did the testing with  the oldschool cherry-pick way.

matmair:
@martonmiklos you just need to press the button, it starts a web-instance of vs code and sets up inventree with the test dataset. It even starts the server and opens the site.

martonmiklos:
Hey @matmair

Do you know why got migrations generated for this change?
I am working on fixing #3852  which is quite similar, but I do not know how to generate migrations (and is it necessary).
Thanks in advance!

SchrodingersGat:
Looks like migrations were generated here because the validators were removed from the field. The validator function was left in, so that migrating from older instances does not fail.

martonmiklos:
Thank you guys! I managed to commit the migrate script as well.

PR #3372, opened 2022-07-20
Add test command

matmair:
This PR adds a new command to invoke to enable simpler loading of testing data. It also makes gitpod.io loading faster.


SchrodingersGat:
Seems like a reasonable addition, thanks @matmair

PR #3375, opened 2022-07-20
Backport #3278

nwns:


nwns:
A cursory glance does not show any other trivially backportable commits since the last release 0.7 release.  There are a number of changes could qualify but they also change the API version.

SchrodingersGat:
@nwns thanks :)

Issue #3371, opened 2022-07-20
[BUG] Completing shipment via API without date ships, but does mark as shipped


miggland:
**Describe the bug**
By trying to mark a shipment as completed without a date, it is possible to get too many items marked/counted as shipped.

**Steps to Reproduce**

Create shipment, allocate items. Then, use the API endpoint `/api/order/so/shipment/{id}/ship/` to mark the shipment as completed, with the following payload:
```
{'shipment_date': None, 'tracking_number': '', 'invoice_number': '', 'link': ''}
```
The SO line items are marked as shipped, but the shipment is not completed.

Now, mark the shipment as completed with the following payload:
```
{'shipment_date': '2002-05-24', 'tracking_number': '', 'invoice_number': '', 'link': ''}
```
The items are shipped again, ie the number of shipped items is double what it should be.

**Expected behavior**
It should either not be possible to use the API endpoint without defining a valid date, or it should choose a reasonable default (ie today's date)

**Deployment Method**
- [x] Docker
- [ ] Bare Metal

**Version Information**
InvenTree-Version: 0.8.0 dev
Django Version: 3.2.14
Commit Hash: 6e2a2f2b2
Commit Date: 2022-07-15
Database: sqlite3
Debug-Mode: True
Deployed using Docker: True

matmair:
Hi there @miggland ! Thank you for this report. I can reproduce this. Interesting behaviour that is *not* meant to be this way.
Would you be interested in trying to fix this (we are always here for questions)?

PS: I edited you initial text to remove a few uninteded wordmarks form our version info thingy.

miggland:
See https://github.com/inventree/InvenTree/pull/3381 .. looks like an easier fix than I first expected.

PR #3381, opened 2022-07-21
Fix issue 3371

miggland:
This should fix https://github.com/inventree/InvenTree/issues/3371

Added another check for shipment_date being None, and overwriting in this case.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3381"></a>


SchrodingersGat:
Thanks @matmair for this fix!

matmair:
@SchrodingersGat I think @miggland did it all by himself ;-)

SchrodingersGat:
Good point, I tagged the wrong person! Thanks @miggland

Issue #3382, opened 2022-07-21
[FR] In API, add SalesOrder information to Shipment.allocations


miggland:
**Is your feature request the result of a bug?**
No

**Problem**
Currently, the API returns something like this for the allocations of a shipment (`/api/order/so/shipment/{id}/`):
```
"allocations": [
        {
            "pk": 8,
            "line": 12,
            "quantity": 10.0,
            "location": null,
            "location_detail": null,
            "item": 993,
            "item_detail": {
                "batch": "",
                "belongs_to": null,
                "build": null,
                "customer": 33,
                "delete_on_deplete": true,
                "expiry_date": null,
                "is_building": false,
                "link": "",
                "location": null,
                "notes": null,
                "owner": null,
                "packaging": "",
                "part": 105,
                "purchase_order": null,
                "pk": 993,
                "quantity": 10.0,
                "sales_order": 6,
                "sales_order_reference": "SO0006",
                "serial": null,
                "status": 50,
                "status_text": "Attention needed",
                "stocktake_date": null,
                "supplier_part": null,
                "uid": "",
                "updated": "2022-07-21",
                "purchase_price": null,
                "purchase_price_currency": "USD",
                "purchase_price_string": "-"
            },
            "order": 6,
            "part": 105,
            "part_detail": {
                "pk": 105,
                "IPN": "",
                "default_location": null,
                "name": "Green Square Table",
                "revision": "",
                "full_name": "Green Square Table",
                "description": "A square table - with green paint",
                "thumbnail": "/media/part_images/square_table_green.thumbnail.png",
                "active": true,
                "assembly": true,
                "is_template": false,
                "purchaseable": false,
                "salable": true,
                "trackable": false,
                "virtual": false,
                "units": ""
            },
            "shipment": 6,
            "shipment_date": "2022-07-21"
        }
    ],
```

While part and item details are given, information assigned to the sales order line item, such as the sales price, the total quantity, the reference etc, are not included and have to be retrieved separately.

**Suggested solution**
Add `order_detail` section to the output

**Do you want to develop this?**
Pull request will follow.

miggland:
Nevermind this - see #3383

PR #3383, opened 2022-07-21
Include order_detail information in Shipment.allocations

miggland:
Fix for issue #3382 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3383"></a>

### DIFF
--- a/InvenTree/order/serializers.py
+++ b/InvenTree/order/serializers.py
@@ -731,6 +731,44 @@ class SalesOrderSerializer(AbstractOrderSerializer, InvenTreeModelSerializer):
         ]
 
 
+class SalesOrderLineItemMinimalSerializer(InvenTreeModelSerializer):
+    """Serializer for a SalesOrderLineItem object, minimal version."""
+
+    def __init__(self, *args, **kwargs):
+        """Initializion routine for the serializer"""
+
+        super().__init__(*args, **kwargs)
+
+    # Annotated fields
+    quantity = InvenTreeDecimalField()
+
+    sale_price = InvenTreeMoneySerializer(
+        allow_null=True
+    )
+
+    sale_price_string = serializers.CharField(source='sale_price', read_only=True)
+
+    sale_price_currency = serializers.ChoiceField(
+        choices=currency_code_mappings(),
+        help_text=_('Sale price currency'),
+    )
+
+    class Meta:
+        """Metaclass options."""
+
+        model = order.models.SalesOrderLineItem
+
+        fields = [
+            'pk',
+            'quantity',
+            'reference',
+            'notes',
+            'sale_price',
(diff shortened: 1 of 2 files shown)


InvenTree/order/serializers.py
```diff
@@ -912,7 +912,7 @@ class Meta:
 class SalesOrderShipmentSerializer(InvenTreeModelSerializer):
     """Serializer for the SalesOrderShipment class."""
 
-    allocations = SalesOrderAllocationSerializer(many=True, read_only=True, location_detail=True)
+    allocations = SalesOrderAllocationSerializer(many=True, read_only=True, location_detail=True, order_detail=True)
```
matmair, line 915: Suggestion: maybe make both location and order detail a flag to reduce unneeded traffic.

matmair (approved):

matmair:
@miggland sample for the suggested switches:
https://github.com/inventree/InvenTree/blob/96faa0d5f053bccfd005059c6e7ca89d0785d42a/InvenTree/order/serializers.py#L91-L98

SchrodingersGat:
@miggland I'm a bit confused here, isn't the detail you want already available at that API endpoint?

image

miggland:
You're right @SchrodingersGat , I obviously didn't check the result properly..

That's not the information I wanted. What I want, as described in #3382, is actually the corresponding order line information.

miggland:
This is what I wanted:
image

Adding the `SalesOrderLineItemMinimalSerializer` feels a bit hacky, but I don't directly see a better way of doing this (since `SalesOrderLineItemSerializer` depends on `SalesOrderAllocationSerializer`, it has to be defined afterwards, and using this method would lead to a loop..

SchrodingersGat:
You're correct about avoiding a loop. TBH this use case seems (to me) to be bordering on a situation where simply requesting the lines for a given order (only one extra API call) is sufficient. Can you describe your desired use case here?

miggland:
The background is that I was starting on creating packing lists - ie reports of what is in a single shipment. I want the report to include the same information as the order confirmation, but be specific as to what is in a certain shipment.

While working on this I realised the problem - but thought I'd finish it up with a possible solution anyway. I'm not convinced it's a good idea either.

The alternative is to ask for each line item specifically while looping through a shipment - not hard, but it's either more than one API call or a search in the results returned. I think it's possible to put this in the python interface instead, as a method of the Shipment class 'getOrderLines' or whatever.

SchrodingersGat:
@miggland I think given the (almost) recursion here, and the duplication of existing data, I'd prefer not to merge this one in. Having multiple ways to achieve the same thing via the API gets confusing and hard to maintain :)

miggland:
Fair enough, I agree.

Issue #3031, opened 2022-05-18
[FR] Export/subscribe to calendar views


miggland:
**Is your feature request the result of a bug?**
[Scheduling view is implemented](https://github.com/inventree/InvenTree/issues/1206)

**Problem**
It would be great if the calendar view (eg the one for orders) could be obtained in a "subscription calendar", so that it is can be seen in eg Thunderbird

**Suggested solution**
I would like to have a link, such as example.com/inventory/orders.ics, which I can subscribe to (read-only) via my calendar viewer.

**Describe alternatives you've considered**
Manually placing the items in the calendar - takes much longer, not automatically updated, prone to errors.

**Examples of other systems**
Share examples eg on iCalshare, such as NASA schedule: <http://icalshare.com/calendars/5143>

**Do you want to develop this?**
I have no clue where to start, but I am not totally incompetent at Python. I would need several pointers.

matmair:
@miggland this would be rather simple with this well known library:
https://github.com/jazzband/django-ical

I would suggest creating the url somewhere here: https://github.com/inventree/InvenTree/blob/8ceff063f86394cd1be9f5c57e6302b79782a526/InvenTree/order/urls.py#L42

SchrodingersGat:
A very nice idea! The django-ical library looks like it would integrate very easily! Happy to give you some pointers digging into the code base if you want to tackle this.

Would it be better to have separate ical endpoints for each type of calendar, or a single one which shares *all* events:

- BuildOrder
- SalesOrder
- PurchaseOrder

miggland:
Having had a quick look this indeed looks like something which is not too complicated to implement.

I would say that separate links for each type of event keeps things easy - then each user can decide for themselves which ones they want to subscribe to, without having a lot of settings/categories/... in the calendar. Also easy to show/hide in the calendar client by showing/hiding the whole calendar.

SchrodingersGat:
Sounds reasonable @miggland - do you want me to assign this to you?

miggland:
I would give it a try - but it may take a while...

matmair:
@miggland no worries - this is a nice-to-have not a must. Feel free to ping us if you want to have pointers or feedback

SchrodingersGat:
Take your time @miggland - we welcome any and all assistance!

miggland:
I've started working on this. I need some help with authentication!

[...]

PR #3391, opened 2022-07-22
Export Orders as iCal calendars

miggland:
This is related to issue #3031.

I'm adding this draft PR so that you can follow along in my code, I'll also need some input.

I've implemented an export for purchase orders only so far, available at: `/order/purchase-order/calendar.ics`.

I've attached an example output (as .txt, because of Github).
[calendar.txt](https://github.com/inventree/InvenTree/files/9169190/calendar.txt)

Problems/todo:
- [ ] This seems not to support authentication in a calendar viewer, such as Thunderbird, at the moment. 
- [x] The URL, which is automatically added, is correct except for the `http://https://`at the beginning. Any ideas? 
- [ ] Extend the exports to Sales Orders
- [ ] Extend the exports to Build Orders
- [ ] Apply InvenTree permissions, to make sure orders are not displayed to users without permissions to view.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3391"></a>


SchrodingersGat:
> The URL, which is automatically added, is correct except for the http://https://at the beginning. Any ideas?

Not sure, may it be to do with how `django_ical` is rendering here?

>  This seems not to support authentication in a calendar viewer, such as Thunderbird, at the moment.

The mail client will need to authenticate against the server somehow. Does `django_ical` support that?

miggland:
> Not sure, may it be to do with how `django_ical` is rendering here?

I'm guessing it's from django_ical - I haven't actually added a URL, which is an option. It could also be caused by the syndication feed framework. Looks like it defaults to the [django SITE_ID](https://docs.djangoproject.com/en/4.0/ref/settings/#std-setting-SITE_ID).

> The mail client will need to authenticate against the server somehow. Does `django_ical` support that?

Normally ICS files are just web resources, meaning they are protected by basic HTTP auth. The way it is now, the user get's forwarded to a login page instead - which is fine in a browser, but not for an automatic download.

At the moment, I can't even get around this, as every page (?) needs authentication.. my knowledge of the Django & InvenTree authentication methods is very limited.

Looks like an [approach similar to this](https://www.djangosnippets.org/snippets/2468/) could work, according to [this thread](https://stackoverflow.com/questions/152248/can-i-use-http-basic-authentication-with-django).

Further permissions are another question - the orders shouldn't be displayed to users who do not have access. I've added this to the list above..

matmair:
@miggland auth seems to work with outlook

matmair:
@miggland I submitted a few small fixes - please check if it works this way

miggland:
The URL issue is fixed, marked as such above.

matmair:
@miggland feeds are generally not authenticated. You can use the function you mentioned to filter for the authenticated user.
I am not really sure that you can add auth to a feed.

SchrodingersGat:
Is this what we want from a data security POV then? Am I understanding correctly that the `.ics` endpoint does not require login?

matmair:
@SchrodingersGat in this state no but once filtering is introduced you just get an empty response. I am also looking into wrapping the view rn.

miggland:
> @miggland feeds are generally not authenticated. You can use the function you mentioned to filter for the authenticated user. I am not really sure that you can add auth to a feed.

That makes sense - but how do I make the authenticated user actually be recognised? At the moment, it's unauthenticated even if I send authentication headers.

matmair:
@miggland I am having a hard time finding good answers. I might need some time to understand the whole feed framework better before I can help with this one.
@SchrodingersGat Suggestion: Move this to 0.9.0

miggland:
> @SchrodingersGat Suggestion: Move this to 0.9.0

It's OK to leave it at 1.0.0 as in the FR, even ;)

It's not extremely important, we should get it right. Doesn't need to be in the next release for sure.

SchrodingersGat:
Still important, thanks bot

miggland:
I've been trying around a bit more. It seems that this problem is not necessarily limited to the feed framework.

I modified InfoView to return the username and whether or not the user is authenticated:
`'user': f'Username: {str(request.user)}, logged in ({request.user.is_authenticated})',`

When I open this page (`http://ubuntuvm:8000/api/`) in the browser, where I am logged in to InvenTree, I get this:
```
{
  "server": "InvenTree",
  "version": "0.9.0 dev",
  "instance": "My very first InvenTree Instance",
  "apiVersion": 76,
  "worker_running": true,
  "plugins_enabled": true,
  "active_plugins": [],
  "user": "Username: TheUser, logged in (True)"
}
```

However, when I download the page from the command line, using either Basic or Token auth like this:
`curl -H 'Authorization: Token 87b----d885' http://ubuntuvm:8000/api/`
`curl -H 'Authorization: Basic bW.........pbg==' http://ubuntuvm:8000/api/`
this is returned:
```
{
   "server" : "InvenTree",
   "version" : "0.9.0 dev",
   "instance" : "My very first InvenTree Instance",
   "apiVersion" : 76,
   "worker_running" : true,
   "plugins_enabled" : true,
   "active_plugins" : [],   
   "user" : "Username: AnonymousUser, logged in (False)"
}
```

To change the behaviour for this API, it has to be based on a different class, RetrieveAPI - then I get the InfoView to show me if I am logged in or not, for basic and token authentication:
```
user@computer:~$ curl -H 'Authorization: Basic bW________bg==' http://ubuntuvm:8000/api/
{"server": "InvenTree", "version": "0.9.0 dev", "instance": "My very first InvenTree Instance", "apiVersion": 76, "worker_running": true, "plugins_enabled": true, "active_plugins": [], "user": "Username: TheUser, logged in (True)"}

user@computer:~$ curl -H 'Authorization: Token 87b______________85' http://ubuntuvm:8000/api/
{"server": "InvenTree", "version": "0.9.0 dev", "instance": "My very first InvenTree Instance", "apiVersion": 76, "worker_running": true, "plugins_enabled": true, "active_plugins": [], "user": "Username: TheUser, logged in (True)"}

user@computer:~$ curl http://ubuntuvm:8000/api/
{"server": "InvenTree", "version": "0.9.0 dev", "instance": "My very first InvenTree Instance", "apiVersion": 76, "worker_running": true, "plugins_enabled": true, "active_plugins": [], "user": "Username: AnonymousUser, logged in (False)"}
```





All the changes for the example can be seen here:
```
diff --git a/InvenTree/InvenTree/api.py b/InvenTree/InvenTree/api.py
index 8aa15be33..9e611aec4 100644
--- a/InvenTree/InvenTree/api.py
+++ b/InvenTree/InvenTree/api.py
@@ -10,7 +10,7 @@ from rest_framework import filters, permissions
 from rest_framework.response import Response
 from rest_framework.serializers import ValidationError
 
-from InvenTree.mixins import ListCreateAPI
+from InvenTree.mixins import ListCreateAPI, RetrieveAPI
 from InvenTree.permissions import RolePermission
 from part.templatetags.inventree_extras import plugins_info
 
@@ -20,7 +20,7 @@ from .version import (inventreeApiVersion, inventreeInstanceName,
 from .views import AjaxView
 
 
-class InfoView(AjaxView):
+class InfoView(RetrieveAPI):
     """Simple JSON endpoint for InvenTree information.
 
     Use to confirm that the server is running, etc.
@@ -38,6 +38,7 @@ class InfoView(AjaxView):
             'worker_running': is_worker_running(),
             'plugins_enabled': settings.PLUGINS_ENABLED,
             'active_plugins': plugins_info(),
+            'user': f'Username: {str(request.user)}, logged in ({request.user.is_authenticated})',
         }
 
         return JsonResponse(data)
```

I don't quite understand the difference between AjaxView and RetrieveAPI, but it seems that AjaxView relies on "SessionAuthentication" and does not include "BasicAuthentication" and "TokenAuthentication".

I have not found a way to get the ical feed class to use these two yet.

miggland:
BTW, ignore the new commits. I made a mess when trying to update to the new version.

And also, you don't have to think too much about what I wrote before.

Going back from the API to general views, I have now found the solution to the problem!

matmair:
@miggland so what was the solution? I am very intrigued, I tried to get this working myself for half an hour and gave up.

miggland:
Well, not a perfect solution (yet), but working partially. I was wrong though, it's in the API view. In the other view, it works with Token auth, and not Basic.

This version works with Basic auth in curl:
`curl -H 'Authorization: Basic bW___________bg==' "http://ubuntuvm:8000/api/order/calendar/purchase-order/"`

The solution is to overload the call method, and perform authentication there - login in again, if necessary.

This should probably be able to support Token login as well, if the header is handled differently.

```
def __call__(self,request,*args,**kwargs):
        """Overload call in order to check for authentication.
        This is required to force Django to look for the authentication,
        otherwise login request with Basic auth via curl or similar are ignored,
        and login via a calendar client will not work.
        
        See:
        https://stackoverflow.com/questions/3817694/django-rss-feed-authentication
        https://stackoverflow.com/questions/152248/can-i-use-http-basic-authentication-with-django
        https://www.djangosnippets.org/snippets/243/
        """
        from django.http.response import JsonResponse
        from django.contrib.auth import authenticate, login
        import base64
        # ~ print(f"{request.META = }")
        
        if request.user.is_authenticated:
            return super().__call__(request,*args,**kwargs)
        
        # No login yet - check in headers
        if 'HTTP_AUTHORIZATION' in request.META:
            auth = request.META['HTTP_AUTHORIZATION'].split()
            if len(auth) == 2:
                # NOTE: We are only support basic authentication for now.
                #
                if auth[0].lower() == "basic":
                    uname, passwd = base64.b64decode(auth[1]).decode("ascii").split(':')
                    user = authenticate(username=uname, password=passwd)
                    if user is not None:
                        if user.is_active:
                            login(request, user)
                            request.user = user
        
        # Check again
        if request.user.is_authenticated:
            return super().__call__(request,*args,**kwargs)
        
        # Still  nothing
        return JsonResponse({"message": "Login required"})
```

This is for testing, the imports shouldn't be there later of course..

matmair:
@miggland no, just create a new PR and link this one there.

PR #3392, opened 2022-07-23
Switch to pathlib

matmair:


SchrodingersGat:
Nice idea :)

matmair:
@SchrodingersGat need to fix a few things but this should also help with future deployment to S3 storage

SchrodingersGat (approved):

SchrodingersGat:
Thanks @matmair

PR #3406, opened 2022-07-26
Remove references to "plugins" volume in production docker-compose file

SchrodingersGat:
This is going to be handled by passing an environment variable through: https://github.com/inventree/InvenTree/pull/3364

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3406"></a>


matmair (approved):
This method will also still work but environment variable seems to be a cleaner solution.

PR #3413, opened 2022-07-26
Development improvements

matmair:
This PR:
- Adds flags to set up things with oneliners
- Adds more console info about current state in tasks
- Ammends CONTRIBUTING with new flags
- Renames .env to .env.example to stop it from auto-loading

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3413"></a>


matmair:
@SchrodingersGat some minor improvements to get ppl. set up faster.

With these changes in place picking up dev on a Debian (current 11 with any of the desktop ends) only requires 4 lines:
```
git clone https://github.com/inventree/InvenTree.git && cd InvenTree
python3 -m venv env && source env/bin/activate
pip install invoke
invoke setup-dev --tests
```

SchrodingersGat:
@matmair the `.env` file is required for the development *docker-compose.yml* file - [see here](https://docs.docker.com/compose/environment-variables/#the-env-file)

By renaming this file the docker setup (for dev) is complicated. We would need to update the docs.

Instead of 

```
docker-compose up -d
```

it would be come

```
docker-compose up -d --env-file .env.example
```

matmair:
> 
> The point of having a "default" env file is to simplify setup and reduce required user input.

@SchrodingersGat the problem is that in non-docker testing environments the env overrides the config.yaml by default - which is kind of hard to diagnose.

SchrodingersGat:
What environments use the `.env` file? I thought that was specific to docker?

matmair:
@SchrodingersGat the .env file is a general *ix thing and get often loaded if a new shell is started in a directory. Docker uses it because linux already used it. Also dotfiles normally get ignored by git.

SchrodingersGat:
Understood. I think I have a way around this :)

SchrodingersGat:
> Understood. I think I have a way around this :)

Actually my idea does not look like it will pan out. The `.env` file provides a very convenient way to control the docker-compose file.. Maybe an alternative would be to move it into the `docker` directory, so at least there's some context (based on the directory) that it is related to the docker setup?

matmair:
@SchrodingersGat we could simply name it anything else then .env - .docker.env for example. The setting what the file should be named is configurable in docker

SchrodingersGat:
This means that the user needs to specify the env file on each command: e.g. `--env-file docker.env` which takes away from the "simplicity" of the current solution (at least from the perspective of the user!)

matmair:
@SchrodingersGat would it be an option to move the env into a seperate folder? The first thing I do on each review is to rename the .env before i start a debugging session.
I will revert the name change and submit it seperatly so we can review/merge the rest

SchrodingersGat:
> would it be an option to move the env into a seperate folder?

This is how it works in the stable / 0.7.x branch. So yes, we could simply revert it to that. I'll look into what that would entail.

SchrodingersGat (approved):

SchrodingersGat:
@matmair found a much simpler workaround - see https://github.com/inventree/InvenTree/pull/3429

SchrodingersGat:
Also found a couple of bugs which will be fixed in https://github.com/inventree/InvenTree/pull/3429 too

Issue #3373, opened 2022-07-20
[BUG] Version info contains uninteded wordmarks


matmair:
**Describe the bug**
The `copy version info` button contains some unnecessary text

**Steps to Reproduce**

1. Copy version info with button
2. Paste somewhere and find copy instruction added at the end and a header at the beginning

**Expected behavior**
Clean text

**Deployment Method**
- [x] Docker
- [x] Bare Metal

**Version Information**
```
# Version Information:
InvenTree-Version: 0.8.0 dev
Django Version: 3.2.14
Commit Hash: afcd60b
Commit Date: 2022-07-20
Database: postgresql
Debug-Mode: False
Deployed using Docker: True

            
                 Versionsinformationen kopieren
```

You see what I did there? 😇 

**Ref:**
#3371

miggland:
Just checked, and this does not happen on my InvenTree running 0.7.6. It does happen on the demo, development branch 0.8.0. See results:

```
# Version Information:
InvenTree-Version: 0.7.6
Django Version: 3.2.13


Database: postgresql
Debug-Mode: False
Deployed using Docker: True

```

```
# Version Information:
InvenTree-Version: 0.8.0 dev
Django Version: 3.2.14
Commit Hash: 164846fa7
Commit Date: 2022-07-24
Database: postgresql
Debug-Mode: True
Deployed using Docker: True
#

            
                 copy version information
```

SchrodingersGat:
Fixed in #3425

PR #3425, opened 2022-07-27
Fix version-copy button according to #3373

miggland:
See issue #3373 

This behaviour was introduced recently, the change seems to fix it again.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3425"></a>


matmair:
@miggland works on all delployment methods, thank you!

PR #3426, opened 2022-07-27
Fit for unit test

SchrodingersGat:
- Write known value to setting before perfoming read test

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3426"></a>


matmair (commented):
Maybe that makes the tests less flaky 👍

PR #3429, opened 2022-07-29
Simplification of development docker-compose file

SchrodingersGat:
- Rename .env file
- Remove requirement for variable interpolation within the docker-compose file itself

Ref: https://github.com/inventree/InvenTree/pull/3413

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3429"></a>


SchrodingersGat:
@matmair lmk what you think here. This change means the user experience is unchanged for the most part.

.github/workflows/qc_checks.yaml
```diff
@@ -132,7 +132,7 @@ jobs:
       - name: Run Tests For `${{ env.wrapper_name }}`
         run: |
           cd ${{ env.wrapper_name }}
-          invoke check-server
+          invoke check-server --debug
```
matmair, line 135: Question: Is this intended?

CONTRIBUTING.md
```diff
@@ -1,15 +1,29 @@
...
+
+### Bare Metal
+
 ```bash
 git clone https://github.com/inventree/InvenTree.git && cd InvenTree
 python3 -m venv env && source env/bin/activate
 pip install invoke && invoke setup-dev --tests
 ```
 
+### Docker
```
matmair, line 16: Comment: Nice that it is 3 lines per method

CONTRIBUTING.md
```diff
@@ -1,15 +1,29 @@
...
+
+```bash
+git clone https://github.com/inventree/InvenTree.git && cd InvenTree
+docker compose run inventree-dev-server invoke setup-test
```
matmair, line 20: Suggestion: If we want ppl. to develop with the image we should also set up pre-commit. That can be done with setup-dev --tests or setup-test --dev

docker-compose.yml
```diff
@@ -23,12 +25,12 @@ services:
...
         image: postgres:13
         expose:
-            - ${INVENTREE_DB_PORT:-5432}/tcp
+            - 5432/tcp
         environment:
             - PGDATA=/var/lib/postgresql/data/pgdb
-            - POSTGRES_USER=${INVENTREE_DB_USER:?You must provide the 'INVENTREE_DB_USER' variable in the .env file}
-            - POSTGRES_PASSWORD=${INVENTREE_DB_PASSWORD:?You must provide the 'INVENTREE_DB_PASSWORD' variable in the .env file}
-            - POSTGRES_DB=${INVENTREE_DB_NAME:?You must provide the 'INVENTREE_DB_NAME' variable in the .env file}
+            - POSTGRES_USER=pguser
```
matmair, line 31: Comment: Much simpler, I like that!

tasks.py
```diff
@@ -518,25 +518,30 @@ def test(c, database=None):
     manage(c, 'test', pty=True)
 
 
-@task(pre=[update], help={'dev': 'Set up development enviroment at the end'})
-def setup_test(c, dev=False):
+@task(help={'dev': 'Set up development enviroment at the end'})
+def setup_test(c, ignore_update=False, dev=False, path="inventree-demo-dataset"):
```
matmair, line 522: Question: Why the rename? Was this path already in use?
Suggestion: This name should also be in the .gitignore

matmair (approved):
Looks great and removes the conflict!

luwol03:
@SchrodingersGat What editor/ide are you using?

Have you ever tried visual studio code? Because they offer a feature called dev containers via a official extension. This means you use vscode like you did before, but you connect to a docker container where everything happens. (The terminal is attached to the container, the folder, bridged ports, ...) this has the advantage, that you don't bloat your system with venv's, pip dependencies, ... Everything you have happens in the container)

luwol03:
@SchrodingersGat I hope this is not sunk. I would like to hear your opinion and recommendations.

matmair:
@luwol03 have you noticed the gitpod.io buttons on top of all PRs? Press any of them and enjoy the show.
We already have something like you suggest implemented. All the dev-tools run on each push in the CI so you do not need to set them up in short-lived development containers.

Also we do not use black at the moment - it actually clashes with our linting rules sometimes.

luwol03:
> @luwol03 have you noticed the gitpod.io buttons on top of all PRs? Press any of them and enjoy the show. We already have something like you suggest implemented. All the dev-tools run on each push in the CI so you do not need to set them up in short-lived development containers.
> 
> Also we do not use black at the moment - it actually clashes with our linting rules sometimes.

@matmair Yeah, was just an experiment. Yeah, indeed I noticed the gitpod button in the prs. But it is not free after an certain amount of used hours. And I also prefer develop locally. With these two files added its just as simple as open vscode and say open this folder in dev container and it will handle everything for you. Because I don't want to bloat my file system with all different Django dependencies and python stuff and still enjoy the features like its running bare metal.

Issue #757, opened 2020-04-27
Add option to "delete stock items" when deleting a stock location


SchrodingersGat:
Currently when deleting a StockLocation the contained StockItem objects are automatically moved up a level.

Two new things need to be added:

1. Provide the user an option to either move the StockItems up a level (default) or delete them. This option should also provide a warning about what stock items are going to be deleted. Also consider whether a separate option should be provided as to what to do with sub-locations

2. When deleting a stock location, and moving contained parts "upward" ensure that each StockItem is given a new tracking note

eeintech:
@SchrodingersGat What did you decide on in #2736 ?

SchrodingersGat:
@eeintech this one is a separate issue.

Consider the case where you have a hierarchy of stock locations A -> B -> C -> D , and location "C" has 10 items in it.

If you were to delete C, the following happens:

1. The 10 stock items are now located in "B" (the "parent" location for C)
2. The parent of D is now B
3. The hierarchy is now A -> B -> D

However, the user may have actually wished to delete C *and everything in it*. Currently this is not easy to do - you would have to delete each child location and stock item individually.

luwol03:
I was also searching for something like this and ended up with a little scripted version. Just wanted to post that here if somebody else needs it until this feature is implemented.

```py
from inventree.api import InvenTreeAPI
from inventree.stock import StockLocation

SERVER_ADDRESS = ''
MY_USERNAME = ''
MY_PASSWORD = ''

api = InvenTreeAPI(SERVER_ADDRESS, username=MY_USERNAME, password=MY_PASSWORD)

def delete_recursively(location: StockLocation):
    print(location.pathstring)
    for sub_loc in location.getChildLocations():
        delete_recursively(sub_loc)
    location.delete()

delete_recursively(StockLocation(api, pk=306))
```

But this method has a big downside, its very slow. I guess this is caused by the logic which checks if it has Childs and tries to move them up.

matmair:
Thank you @luwol03 ! This should really be part of the general API.

SchrodingersGat:
@luwol03 looks like this feature is coming into the UI as well : https://github.com/inventree/InvenTree/issues/3384

> But this method has a big downside, its very slow

Yep, multiple API certainly add up! That's why we added a `BulkDeleteMixin` so that *some* items can be deleted all at once. 

[...]

PR #3435, opened 2022-07-30
Add option to recursively delete part categories

### LINKED ISSUE #3384: [FR]Recursive part category deletion
No

**Problem**
I have imported a category tree from an external source (TME API), and I have categories with child categories what I would like to delete.
Deleting the topmost categories will result the child categories to be moved one level upward. 

**Suggested solution**
kép
I would recommend adding a two comboboxes in this area:
- What to do with the Child categories: move to parent/delete (together with associated parts)
- What to do with the items associated to this category: move to parent/delete

Using raw SQL commands ;)

**Examples of other systems**
I am not aware of any.

Sure.

martonmiklos:
Fixes #3384

Ref: https://github.com/inventree/InvenTree/issues/757

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3435"></a>


InvenTree/templates/js/translated/part.js
```diff
@@ -324,20 +324,39 @@ function editCategory(pk) {
...
+        {
+            value: 'delete',
+            display_name: '{% trans "Delete them" %}',
+        }
+    ];
     constructForm(url, {
         title: '{% trans "Delete Part Category" %}',
         method: 'DELETE',
+        fields: {
+            'child-parts-action': {
```
SchrodingersGat, line 347: Please do not use `-` characters here, as it makes field lookup difficult. `-` is generally not accepted in a variable name. Use `_` instead e.g. `child_parts_action`

InvenTree/templates/js/translated/part.js
```diff
@@ -324,20 +324,39 @@ function editCategory(pk) {
  * Delete a PartCategory via the API
  */
 function deletePartCategory(pk, options={}) {
-    var url = `/api/part/category/${pk}/`;
+    var url = `/api/part/category/${pk}/?child-categories-action=${pk}&child-parts-action=${pk}`;
```
SchrodingersGat, line 327: Why the extra URL? The fields as passed as json data in the DELETE operation, not as query parameters in the URL

InvenTree/templates/js/translated/part.js
martonmiklos: Yeah, I faced with it :)

InvenTree/InvenTree/mixins.py
```diff
@@ -81,10 +81,56 @@ class RetrieveUpdateAPI(CleanMixin, generics.RetrieveUpdateAPIView):
...
+    """
+    Destroy a model instance.
+    """
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        self.perform_destroy(instance, **kwargs)
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def perform_destroy(self, instance, **kwargs):
+        instance.delete(**kwargs)
```
matmair, line 108: Question: Why is this class defined twice?

SchrodingersGat: Ctrl-C Ctrl-V Ctrl-V

InvenTree/part/models.py
```diff
@@ -68,28 +67,47 @@ def delete(self, *args, **kwargs):
 
         This must be handled within a transaction.atomic(), otherwise the tree structure is damaged
         """
-        with transaction.atomic():
+        print(kwargs)
+        try:
+            if kwargs.get('no_rebuild') is None:
```
SchrodingersGat, line 73: You make this comparison multiple times, I would suggest doing:

```rebuild = kwargs.get('rebuild', True)```

And then use the `rebuild` argument, and pass `rebuild=False` to disable rebuild

File: InvenTree/part/models.py [hunk 5]
```diff
@@ -68,28 +67,47 @@ def delete(self, *args, **kwargs):
 
         This must be handled within a transaction.atomic(), otherwise the tree structure is damaged
         """
-        with transaction.atomic():
+        print(kwargs)
```
SchrodingersGat, line 70: Please remove this debug statement before finishing :)

InvenTree/templates/js/translated/part.js
```diff
@@ -329,15 +329,32 @@ function deletePartCategory(pk, options={}) {
...
     {% trans "Are you sure you want to delete this part category?" %}
-    <ul>
-        <li>{% trans "Any child categories will be moved to the parent of this category" %}</li>
-        <li>{% trans "Any parts in this category will be moved to the parent of this category" %}</li>
-    </ul>
     </div>`;
-
+    var subChoices = [
+        {
+            value: 'move_to_parent',
```
SchrodingersGat, line 335: I'm not a huge fan of passing string variables through the API and then comparing them, especially when these are essentially "boolean" checks.

I think a *cleaner* approach would be have boolean inputs for:

- delete parts
- delete child categories

And then pass true or false for these named parameters. Makes the backend cleaner too.

InvenTree/InvenTree/mixins.py
martonmiklos: > Ctrl-C Ctrl-V Ctrl-V

Yes indeed :)

InvenTree/templates/js/translated/part.js
martonmiklos: > I'm not a huge fan of passing string variables through the API and then comparing them, especially when these are essentially "boolean" checks.

I second the usage of booleans, however my impression is that the choice is the most intuitive way to present these decisions to the user. 

What shall we write to the label of the checkbox? 

If we write just "Delete parts" then effect of the turned off value will not be defined for the user. 
Writing "Delete parts (if turned off the parts will be moved to parent)" to the labels looks a bit weird for me.

tldr: I would keep the choices, but with True and False values as options.

luwol03: I don't think that these are booleans. Because if delete child's is true, move to parent can't be also true. So they are dependent. What if we later came up with a third deletion strategy? What about changing it from strings to numbers (move to parent -> 1, recursive delete ->2, ...)? Or is that an even dirtier solution than the strings?

SchrodingersGat: @maksimstojkovic I think your approach is a reasonable compromise. 

Regarding the "incompatible choices" raised by @luwol03 - the back-end should check that the choices are compatible and if not, raise an error rather than actually deleting anything. This will render an error on the form so the user gets feedback

SchrodingersGat:
@martonmiklos also we should ensure that we add some unit tests to cover these new features. Once the feature part is done, I can run you through how this should be implemented.

martonmiklos:
> @martonmiklos please stop constantly force pushing - all the notification emails and git automation stop working as the hashes change. The PR will be squashed at the end anyway and with the way you currently work none of the code comments make sense.

Roger that, I will keep committing without force.

matmair:
Thanks @martonmiklos I will append CONTRIBUTING

SchrodingersGat:
@martonmiklos currently the CI test are failing early due to style issues. Can you fix those and we can see if the other issues are still present?

martonmiklos:
Something messed up badly on my machine, I keep getting :

```
----------------------------------------------------------------------
  File "/home/mm/Projektek/inventree/src/InvenTree/InvenTree/api_tester.py", line 44, in setUp
    self.group = Group.objects.create(name='my_test_group')
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/models/query.py", line 453, in create
    obj.save(force_insert=True, using=self.db)
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/models/base.py", line 739, in save
    self.save_base(using=using, force_insert=force_insert,
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/models/base.py", line 776, in save_base
    updated = self._save_table(
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/models/base.py", line 881, in _save_table
    results = self._do_insert(cls._base_manager, using, fields, returning_fields, raw)
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/models/base.py", line 919, in _do_insert
    return manager._insert(
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/models/query.py", line 1270, in _insert
    return query.get_compiler(using=using).execute_sql(returning_fields)
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/models/sql/compiler.py", line 1416, in execute_sql
    cursor.execute(sql, params)
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/backends/utils.py", line 66, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/backends/utils.py", line 78, in _execute
    self.db.validate_no_broken_transaction()
  File "/home/mm/Projektek/inventree/env/lib/python3.10/site-packages/django/db/backends/base/base.py", line 447, in validate_no_broken_transaction
    raise TransactionManagementError(
django.db.transaction.TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.
```
with each commit I had different issues with previously.

SchrodingersGat:
@martonmiklos did you make any progress on this or still stuck on unit tests?

martonmiklos:
> @martonmiklos did you make any progress on this or still stuck on unit tests?

I managed to step one step upper in the rabbit hole: I have overcome my issues mentioned above, but the unit test is still failing, and I have no idea why. (Some parent changes stop working over time.)

matmair:
@martonmiklos put the style issues are passing  now?

martonmiklos:
Ok guys, after a long time and a lot of messy git commits:
- it passing on stylistic tests
- it has unit test for all the 4 delete parameter cases

So I think we can start review discussion on this.

InvenTree/InvenTree/metadata.py
```diff
@@ -93,11 +93,11 @@ def determine_metadata(self, request, view):
 
             # Add a 'DELETE' action if we are allowed to delete
             if 'DELETE' in view.allowed_methods and check(user, table, 'delete'):
-                actions['DELETE'] = True
+                actions['DELETE'] = {}
 
             # Add a 'VIEW' action if we are allowed to view
             if 'GET' in view.allowed_methods and check(user, table, 'view'):
-                actions['GET'] = True
+                actions['GET'] = {}
```
matmair, line 100: comment: this is pretty deep in our meta logic.
@SchrodingersGat could you please take a look?

InvenTree/part/test_api.py
```diff
@@ -293,6 +294,113 @@ def test_invisible_chars(self):
...
+    def test_category_delete(self):
+        """Test category deletion with different parameters"""
+
+        class Target(IntEnum):
+            move_subcategories_to_parent_move_parts_to_parent = 0,
+            move_subcategories_to_parent_delete_parts = 1,
+            delete_subcategories_move_parts_to_parent = 2,
+            delete_subcategories_delete_parts = 3,
+
+        for i in range(4):
```
matmair, line 306: Question(non-blocking): Would a smaller range also be sufficient to get full coverage? Our tests are already running quite long

InvenTree/part/test_api.py
```diff
@@ -293,6 +294,113 @@ def test_invisible_chars(self):
...
+                if delete_child_categories:
+                    for part in child_categories_parts:
+                        part.refresh_from_db()
+                        self.assertEqual(part.category, parent_category)
+
+            if delete_child_categories:
+                self.assertEqual(PartCategory.objects.count(), category_count_before)
+            else:
+                #  Check if all subcategories to parent moved to parent and all parts deleted
+                for child in child_categories:
```
matmair, line 400: Suggestion: A short comment on why the assertations are made would be good for future debugging

InvenTree/templates/js/translated/part.js
```diff
@@ -333,15 +333,33 @@ function deletePartCategory(pk, options={}) {
...
+            display_name: '{% trans "Delete them" %}',
+        }
+    ];
 
     constructForm(url, {
         title: '{% trans "Delete Part Category" %}',
         method: 'DELETE',
+        fields: {
+            'delete_parts': {
+                label: '{% trans "What to do with the parts associated to this category:" %}',
```
matmair, line 353: Suggestion(non-blocking): I feel like there is a shorter way to phase this but my English is a bit German (eq complicated). @SchrodingersGat any suggestions?

InvenTree/InvenTree/metadata.py
SchrodingersGat: @martonmiklos what is the rationale for this particular change?

matmair (approved):
@martonmiklos good and complicated PR with good test! Thank you very much;
Please take a look at my comments, most are non-blocking.
The tests add some runtime - is there an easy way to reduce the number of loops? Those add a lot of I/O to the database; If that would be a lot of work do not bother - this would be nice-to-have and can always be done later (though experience from my own PRs shows later might be never)

InvenTree/templates/js/translated/part.js
```diff
@@ -333,15 +333,33 @@ function deletePartCategory(pk, options={}) {
...
     {% trans "Are you sure you want to delete this part category?" %}
-    <ul>
-        <li>{% trans "Any child categories will be moved to the parent of this category" %}</li>
-        <li>{% trans "Any parts in this category will be moved to the parent of this category" %}</li>
-    </ul>
     </div>`;
+    var subChoices = [
+        {
+            value: 0,
+            display_name: '{% trans "Move them under the parent category" %}',
```
SchrodingersGat, line 340: I think "Move to parent category" is cleaner here.

InvenTree/templates/js/translated/part.js
```diff
@@ -333,15 +333,33 @@ function deletePartCategory(pk, options={}) {
...
+        },
+        {
+            value: 1,
+            display_name: '{% trans "Delete them" %}',
```
SchrodingersGat, line 344: Simply "Delete" is cleaner here

InvenTree/templates/js/translated/part.js
SchrodingersGat: "Action for parts in this category"

InvenTree/templates/js/translated/part.js
```diff
@@ -333,15 +333,33 @@ function deletePartCategory(pk, options={}) {
...
         title: '{% trans "Delete Part Category" %}',
         method: 'DELETE',
+        fields: {
+            'delete_parts': {
+                label: '{% trans "What to do with the parts associated to this category:" %}',
+                choices: subChoices,
+                type: 'choice'
+            },
+            'delete_child_categories': {
+                label: '{% trans "What to do with the child categories (and parts associated to them):" %}',
```
SchrodingersGat, line 358: "Action for child categories"

InvenTree/InvenTree/metadata.py
martonmiklos: > @martonmiklos what is the rationale for this particular change?

Well someone recommended me to do that ;)
https://github.com/inventree/InvenTree/issues/3384#issuecomment-1198868669

SchrodingersGat: @martonmiklos I thought it sounded familiar. Thanks for the reminder :)

martonmiklos: > @martonmiklos I thought it sounded familiar. Thanks for the reminder :)

Getting back to the business: I am not familiar with the django to say it is safe or not, please decide if it can be handled differently or it is acceptable.

martonmiklos:
> @martonmiklos good and complicated PR with good test! Thank you very much; Please take a look at my comments, most are non-blocking. The tests add some runtime - is there an easy way to reduce the number of loops? Those add a lot of I/O to the database; If that would be a lot of work do not bother - this would be nice-to-have and can always be done later (though experience from my own PRs shows later might be never)

Which loops do you refer to? We can create less sub parts/categories, but I doubt that it adds much runtime.
Having coverage for all the four parameter option combinations is necessary I think to prevent regressions.

InvenTree/part/models.py
```diff
@@ -63,31 +63,44 @@ class PartCategory(MetadataMixin, InvenTreeTree):
...
+        for child in self.children.all():
+            if kwargs.get('delete_child_categories', False):
+                child.delete_recursive(**dict(delete_child_categories=True,
+                                              delete_parts=delete_parts,
+                                              parent_category=parent_category))
+            else:
+                child.parent = parent_category
+                child.save()
+
+        super().delete(*args, **dict())
```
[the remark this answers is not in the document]
martonmiklos: Not yet, but please educate me a bit how can I trace the runtime of the queries in django. What I see now is when I run this test with --keepdb it takes ~6 seconds, but I do not know what percentage does it take of the setup/cleanup. I could trace it with some console printup, but I assume there are more elegant ways to do so in django.

I implemented the recursive method separately to be able to run all recursive queries in a single transaction (initiated in the delete method). TBH the fact that the transaction is works for recursive calls is just an assumption.

InvenTree/part/test_api.py
martonmiklos: Is there any way to get a coverage info from the test runs?
At the moment I as I see we can improve it (one case for part delete, one case for category delete and a third for move to parent for both), but in the case in the future the code changes it might loose coverage.

InvenTree/part/api.py
```diff
@@ -27,7 +27,8 @@
 from InvenTree.filters import InvenTreeOrderingFilter
 from InvenTree.helpers import (DownloadFile, increment_serial_number, isNull,
                                str2bool, str2int)
-from InvenTree.mixins import (CreateAPI, ListAPI, ListCreateAPI, RetrieveAPI,
+from InvenTree.mixins import (CreateAPI, CustomRetrieveUpdateDestroyAPI,
```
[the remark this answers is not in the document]
martonmiklos: Uhmm, I am just a python ditcher, so I am not 100%ly understand the question/problem. The RetrieveUpdateAPI and RetrieveUpdateAPI existed and were used before the PR, so I did not wanted to touch them. Both of them are used in this file so did not wanted to remove the import. Are you suggesting me to remove the RetrieveUpdateAPI and RetriveUpdateDestroyAPIs and replace them with the Custom variants (which passing the kwargs)? I am open to the later to reduce the code, however I would be doing that in a separate PR to keep things clean.

InvenTree/InvenTree/metadata.py
SchrodingersGat: The changes here as suggested by me should be acceptable

SchrodingersGat:
@matmair any feedback here for @martonmiklos ?

InvenTree/part/api.py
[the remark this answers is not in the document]
matmair: Let's leave it like this for now

InvenTree/part/test_api.py
matmair: @martonmiklos you can look into the check runs. There is a check from coveralls with the coverage and a link.  Alternatively, you can check the SQLite + coverage check. there is a step for uploading coverage, in the logs, you will find a link to the coverage results.

I think one case less would be sufficient to cover the changes but let's leave the tests as is. In the future, a more explanatory docstring about the test function - especially with such a clever duplication reduction mechanism - would probably speed up reviewing.

matmair:
Sorry @martonmiklos @SchrodingersGat did not realise there was something open from me

SchrodingersGat:
@martonmiklos as far as I can see, this is good to go? Just need to fix the conflicts in `test_api.py`. Anything from you @matmair ?

matmair:
@SchrodingersGat should be good to go from my side

SchrodingersGat:
Fantastic! Just a small tweak from you then @martonmiklos and we are good to go

martonmiklos:
> Fantastic! Just a small tweak from you then @martonmiklos and we are good to go

Sorry guys for the late response, I am traveling this week, but I will try to put together my dev setup to sort out the outstanding issues.

martonmiklos:
> If this is resolved, is there also a plan for integrating all that also to the stock locations? Should I create an issue for that?

I can handle that after the current week.

SchrodingersGat:
@martonmiklos thanks for the great work here, a fantastic contribution :) Look forward to the same for stock locations!

Issue #3438, opened 2022-07-30
[BUG] Treegrid is loading an eternity for huge amounts of data


luwol03:
**Problem**
If you have deeply nested storage locations and viewing them via the tree view has two disadvantages. 
1. It needs a very long time to load
2. I'ts not very readable.
![image](https://user-images.githubusercontent.com/60048565/181996528-df8e070f-609e-4f69-98c3-8d7ce7a95bf9.png)
(These naming makes no sense of an actual naming, but this was just the result of some demo data from me. In my real storage its not nested like that, its more like Room/rack/shelve/box/box location (one box is devided into front/middle/back). So my tree is nested very deeply and each branch contains a lot of items. And imagine a storage with 4 rooms each 12 racks each 32 boxes each splitted into 3 sections)

**Suggested solution**
Add a max tree depth and then lazy load the items if we try to open that specific StorageLocation.

**Describe alternatives you've considered**
-

**Examples of other systems**
-

**Do you want to develop this?**
If this is really wanted I would need some hints where to start.

SchrodingersGat:
@luwol03 how long is the API request taking to load all this data?

luwol03:
It's taking about 10 minutes for me. I generated the data with this script:
```py
def generate_nested(parent, loc):
    if len(loc) == 0:
        return

    cnt, name = loc.pop(0)
    print(parent.pathstring)
    for i in range(cnt):
        s = StockLocation.create(api, data={"parent": parent.pk, "name": f"{name} {i}"})
        generate_nested(s, loc.copy())


data = [
    (4, "Room"),
    (12, "Rack"),
    (32, "Box"),
    (3, "Section")
]

generate_nested(StockLocation(api, 305), data)
```

### Timing:
Interesting, there are several requests with different timings.
- `GET /api/stock/location/tree/` - `0,294s`
- `GET /api/stock/location/?parent=305&cascade=true` - `23.608s`

Also interesting is, that the page needs about 10 minutes to load the tree. So there must be also some client side overload.

SchrodingersGat:
10 minutes is an eternity. How many locations in that dataset?

luwol03:
You can calculate it via 4*12*32*3=4608.

SchrodingersGat:
> Add a max tree depth

Do you mean that items below a certain "depth" in the tree would not be displayed in the breadcrumb tree?

luwol03:
[...]

PR #3443, opened 2022-07-31
Tree query improvements

SchrodingersGat:
https://github.com/inventree/InvenTree/issues/3438 focused my attention on the `PartCategory` and `StockLocation` list API endpoints.

Currently they are quite inefficient! 

Taking the *PartCategory* list endpoint as an example: `/api/part/category/?limit=<x>`

| Number of Results | Number of Queries | Time (s) |
| --- | --- | --- |
| 10 | 24 | 0.1 |
| 100 | 203 | 0.3 |
| 500 | 1000 | 1.0 |
| 1000 | 1994 | 1.9 |
| 5000 | 8910 | 8.6 |

~~This is very slow! (The *StockLocation* endpoint has similar issues)~~

**After fixes applied**

| Number of Results | Number of Queries | Time (s) |
| --- | --- | --- |
| 10 | 6 | 0.1 |
| 100 | 6 | 0.1 |
| 500 | 6 | 0.25 |
| 1000 | 6 | 0.45 |
| 5000 | 6 | 2.5 |

There are two culprits here causing *1 + N* query problems.

## Part Count

Each category object has a field "parts" which is a count of the number of parts in the category (or any subcategories)

### Solution

- [x] Use a query annotation (with a tree-based subquery) to annotate the queryset and prevent any additional database lookups.

*This has been tested, and works very well - query count has been cut in half!*

## Pathstring

The `pathstring` field is calculated "on the fly" for each part category. One DB query is required for each item in the results. For thousands of results, this means thousands of queries!

*This currently presents a more difficult challenge to fix!*

Options to pursue here:

### Caching

We could use the "cache" framework to save and load pathstring values, so we can load from memory instead of the database. This works (and has been tested), reducing the problem down to O(1) instead of O(n).

*However* the default caching framework can only handle a certain number of keys before it runs out. This solution works up to a point, but is always going to be limited by the maximum number of cache entries allowed by the system.

### Annotation

We could (maybe) work out a queryset annotation to fetch the required data from the database and reduce the number of DB hits to a constant number.

*However* this is a harder problem than any existing queryset annotations. Might be not possible as we need to return multiple results per row, which is not generally possible with django's annotation framework.

### Pre-Calculate and Store

We could pre-calculate the `pathstring` field and store in the database. Whenever the model is saved, re-compute and save the "pathstring" again. As these don't change very often, this would be not too bad.

*However* we would be limited by the available number of characters in the char field. This might be OK if we decide on a mechanism to handle "very long" path strings, maybe something like:

`/some/path/which/is/super/long/....../middle/is/cut/out/and/replaced/with-dots/`

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3443"></a>


matmair (approved):
Looks good for now. Prerendering the path might be the best solution.

Issue #2789, opened 2022-03-29
[FR] Switch UI to React


matmair:
**Is your feature request the result of a bug?**
No - More or a vision

**Problem**
Currently we build a lot of html templates and js. As there is a disconnect between the server and the frontend we do not really have tests for the frontend. Also there is a lot more data transmitted than needed.

**Suggested solution**
- Switch to React (or a framework that uses React as basis like remix or next.js)
- Add UI testing with [cypress](https://www.cypress.io)
- Run test via [gh-actions](https://github.com/cypress-io/github-action)

**Describe alternatives you've considered**
Current solution works fine but this FR would be cool.

**Examples of other systems**
A lot seems to run on this - an example in our field would be https://partsbox.com/

**Do you want to develop this?**
I would love to help but this would be a team effort.

matmair:
@SchrodingersGat can we get a 1.0 milestone for that? @eeintech

matmair:
@eeintech huge adoption, not very opinionated, good toolset ecosystem and a lot of examples that this sort of application can be done in react.
The tooling is a good fit for what we already do with js. A lot of code should be adaptable rather easy. And there are not as much heavy breaking changes as in vue. Got burned by that too often now.

eeintech:
Sounds good thanks!

matmair:
@eeintech always open for feedback but I think react is a good compromise.

I will do a prototype for how I would tackle it so we can discuss the architecture, todos etc. in the coming months. It is a lot but not as much one would think - we can reuse all the API forms from what I can see.

4jag:
I dislike React. I'm working on a project, and at the end of every day I want to go out to the garage and hang myself. Please stay clear of React.

matmair:
@4jag we are open for other suggestions.
I have worked and am working on multiple projects with next.js and find it to be much better than bare js + html. Could you elaborate on the problems you see / which framework would be better suited for this project?

matmair:
Tutorial for Vue: https://realpython.com/python-django-blog/#step-6-set-up-vue-router

phuicy:
Geniunely, why? Your front end seems clear, simple, and capable. You could acheive more UI testing with your current frontend. You could even use this sort of system: https://playwright.dev/

matmair:
Have you worked with the front-end code? A lot of bugs are because of missing typing and lazy linking of important parts via jquery. It also needs full reloads for a lot of actions.

PR #3451, opened 2022-08-01
Fix: Treegrid is loading an eternity for huge amounts of data

### LINKED ISSUE #3438: [BUG] Treegrid is loading an eternity for huge amounts of data

luwol03:
Loading the tree grid with huge amounts of data can take more than 5 minutes as described in #3438. I tried to add a max depth in this PR and lazy load the rest as its requested by the user.

### Screenshots

https://user-images.githubusercontent.com/60048565/182368321-cf4b4193-67a2-4c5b-9636-997ccec277a5.mov

image

#### 1. Tree does not keep the collapsed items collapsed after new items are added to the tree

This seems like an upstream issue. I opened an issue on their side: wenzhixin/bootstrap-table#6304

#### 2. The spinning animations is not shown correctly

I think this is caused by the browser, because if more data is loading and processed by the browser the thread is blocked and nothing else can happen, so the animation would start after loading, but then the loading state is removed, so we don't see any spinning animation.
--> Seems like this is working if they are very few items in the table

Let me know what you think of this solution and if there is another tree grid which should be handled similar. The categories?

fixes #3438

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3451"></a>

### DIFF
--- a/InvenTree/part/api.py
+++ b/InvenTree/part/api.py
@@ -85,6 +85,8 @@ class CategoryList(ListCreateAPI):
 
         cascade = str2bool(params.get('cascade', False))
 
+        depth = params.get("depth", None)
+
         # Do not filter by category
         if cat_id is None:
             pass
@@ -94,12 +96,18 @@ class CategoryList(ListCreateAPI):
             if not cascade:
                 queryset = queryset.filter(parent=None)
 
+            if depth is not None:
+                queryset = queryset.filter(level__lte=int(depth))
+
         else:
             try:
                 category = PartCategory.objects.get(pk=cat_id)
 
                 if cascade:
                     parents = category.get_descendants(include_self=True)
+                    if depth is not None:
+                        parents = parents.filter(level__lte=category.level + int(depth))
+
                     parent_ids = [p.id for p in parents]
 
                     queryset = queryset.filter(parent__in=parent_ids)
--- a/InvenTree/part/api.py
+++ b/InvenTree/part/api.py
@@ -97,7 +97,7 @@ class CategoryList(ListCreateAPI):
             if not cascade:
                 queryset = queryset.filter(parent=None)
 
-            if depth is not None:
+            if cascade and depth is not None:
                 queryset = queryset.filter(level__lte=depth)
 
         else:
--- a/InvenTree/part/api.py
+++ b/InvenTree/part/api.py
@@ -25,7 +25,8 @@ from company.models import Company, ManufacturerPart, SupplierPart
 from InvenTree.api import (APIDownloadMixin, AttachmentMixin,
                            ListCreateDestroyAPIView)
 from InvenTree.filters import InvenTreeOrderingFilter
-from InvenTree.helpers import DownloadFile, increment, isNull, str2bool
+from InvenTree.helpers import (DownloadFile, increment, isNull, str2bool,
+                               str2int)
 from InvenTree.mixins import (CreateAPI, ListAPI, ListCreateAPI, RetrieveAPI,
                               RetrieveUpdateAPI, RetrieveUpdateDestroyAPI,
                               UpdateAPI)
@@ -85,7 +86,7 @@ class CategoryList(ListCreateAPI):
 
         cascade = str2bool(params.get('cascade', False))
 
-        depth = params.get("depth", None)
+        depth = str2int(params.get('depth', None))
 
         # Do not filter by category
         if cat_id is None:
@@ -97,7 +98,7 @@ class CategoryList(ListCreateAPI):
                 queryset = queryset.filter(parent=None)
 
             if depth is not None:
-                queryset = queryset.filter(level__lte=int(depth))
+                queryset = queryset.filter(level__lte=depth)
 
         else:
             try:
@@ -106,7 +107,7 @@ class CategoryList(ListCreateAPI):
                 if cascade:
                     parents = category.get_descendants(include_self=True)
                     if depth is not None:
-                        parents = parents.filter(level__lte=category.level + int(depth))
+                        parents = parents.filter(level__lte=category.level + depth)
 
                     parent_ids = [p.id for p in parents]
 
--- a/InvenTree/stock/api.py
+++ b/InvenTree/stock/api.py
@@ -252,8 +252,8 @@ class StockLocationList(ListCreateAPI):
             # If we allow "cascade" at the top-level, this essentially means *all* locations
             if not cascade:
                 queryset = queryset.filter(parent=None)
-                
-            if depth != None:
+
+            if depth is not None:
                 queryset = queryset.filter(level__lte=int(depth))
 
         else:
@@ -264,9 +264,9 @@ class StockLocationList(ListCreateAPI):
                 # All sub-locations to be returned too?
                 if cascade:
                     parents = location.get_descendants(include_self=True)
-                    if depth != None:
+                    if depth is not None:
                         parents = parents.filter(level__lte=location.level + int(depth))
-                        
+
                     parent_ids = [p.id for p in parents]
                     queryset = queryset.filter(parent__in=parent_ids)
 
--- a/InvenTree/stock/api.py
+++ b/InvenTree/stock/api.py
@@ -253,7 +253,7 @@ class StockLocationList(ListCreateAPI):
             if not cascade:
                 queryset = queryset.filter(parent=None)
 
-            if depth is not None:
+            if cascade and depth is not None:
                 queryset = queryset.filter(level__lte=depth)
 
         else:
--- a/InvenTree/stock/api.py
+++ b/InvenTree/stock/api.py
@@ -26,7 +26,7 @@ from InvenTree.api import (APIDownloadMixin, AttachmentMixin,
                            ListCreateDestroyAPIView)
 from InvenTree.filters import InvenTreeOrderingFilter
 from InvenTree.helpers import (DownloadFile, extract_serial_numbers, isNull,
-                               str2bool)
+                               str2bool, str2int)
 from InvenTree.mixins import (CreateAPI, ListAPI, ListCreateAPI, RetrieveAPI,
                               RetrieveUpdateAPI, RetrieveUpdateDestroyAPI)
 from order.models import PurchaseOrder, SalesOrder, SalesOrderAllocation
@@ -241,7 +241,7 @@ class StockLocationList(ListCreateAPI):
 
         cascade = str2bool(params.get('cascade', False))
 
-        depth = params.get("depth", None)
+        depth = str2int(params.get('depth', None))
 
         # Do not filter by location
         if loc_id is None:
@@ -254,7 +254,7 @@ class StockLocationList(ListCreateAPI):
                 queryset = queryset.filter(parent=None)
 
             if depth is not None:
-                queryset = queryset.filter(level__lte=int(depth))
+                queryset = queryset.filter(level__lte=depth)
 
         else:
 
@@ -265,7 +265,7 @@ class StockLocationList(ListCreateAPI):
                 if cascade:
                     parents = location.get_descendants(include_self=True)
                     if depth is not None:
-                        parents = parents.filter(level__lte=location.level + int(depth))
+                        parents = parents.filter(level__lte=location.level + depth)
 
                     parent_ids = [p.id for p in parents]
                     queryset = queryset.filter(parent__in=parent_ids)
--- a/InvenTree/stock/api.py
+++ b/InvenTree/stock/api.py
@@ -241,6 +241,8 @@ class StockLocationList(ListCreateAPI):
 
         cascade = str2bool(params.get('cascade', False))
 
+        depth = params.get("depth", None)
+
         # Do not filter by location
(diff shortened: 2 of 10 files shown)


InvenTree/templates/js/translated/stock.js
```diff
@@ -2226,6 +2226,7 @@ function loadStockLocationTable(table, options) {
 
     if (tree_view) {
         params.cascade = true;
+        params.depth = MAX_DEPTH;
```
SchrodingersGat, line 2229: I would also say that if we are in tree view, we *must* set the `cascade=true` parameter. Otherwise the tree-view part doesn't work :)

SchrodingersGat:
@luwol03 I really like your implementation here. Clean, and "on brand" with how we have written the rest of the code.

> Let me know what you think of this solution and if there is another tree grid which should be handled similar. The categories?

Yep, the part-category table would benefit from this also.

#### API Version

As we have added a new parameter to the API (`depth`) can you please increment the version number in `InvenTree/api_version.py` and add a note in the changelog in that file.

luwol03:
@SchrodingersGat Thank you for your feedback. I'll add that also to the part category later.

Are there also some documentations where I have to add that? The coreapi browser?

matmair:
@luwol03 I can not answer everything but I will what I can.
1. The API self-documents all arguments. If you want to add anything custom there you can just change the docstring of the affected api view. The coreapi browser pulls the information from there.
2. I think @SchrodingersGat should decide if this should be a setting. That would have to be added into `/common/models.py` into the settings / usersetting definition. These are then exposed to the .js context
3. The translations are handled automatically. Once your changes are in `master` they will be pulled into a translation branch and uploaded to crowdin. Translations are merged into `master` from there and then prerendered on each run of `inv update`

Changing the codebase fundamentally is not on the roadmap - we are happy for PRs anway ;-). We are planning to switch to some js-framework for 1.0 (either react or vue seem to have the strongest community #2789) but this low priority rn as there is a lot more functional stuff that we are working on. If you have knowledge in this space your opinion would be very valued - we are all non-CS enigneers around here.

luwol03:
@matmair Thank you for your answers.

1. I thought I need to add the new depth param somewhere so that I have a new input available in the coreapi browser. I'll add that.

I read about the proposed migration to react from you a few days ago as I explored the open issues. I can also fully understand that a fundermental change (eg use const let and a more modern es6 syntax) is not a choice but I suggest writing new logic with a more modern syntax like I tried. Let me know if I should change my variables to var to fit to the rest of the code.

I'm no professional Frontend developer, but have some experience with it. From my perspective I can recommend react because it has a huge ecosystem and there is also a big company behind it which tries to make it better every day.

SchrodingersGat:
@luwol03 I very much appreciation your advice regarding best practice for JS and your points are well received. My background is in embedded software (low level C) and I'm not up-to-date with the latest web-dev stuff. Any contributions and improvements you can make here would be very welcome!

SchrodingersGat:
> Are there also some documentations where I have to add that? The coreapi browser?

As @matmair says the API is mostly self-describing. Most fields are detected by default, however if you add custom code in the "filter_queryset" method (as we are doing here) then the API documentation cannot introspect that. 

To have the API automatically extract and document the available API fields there are two ways:

1. Add the field to the `filterset_fields` list:

https://github.com/inventree/InvenTree/blob/9ca0d6b0a95561ee8c71a6e3c4e815a4b0a6daeb/InvenTree/part/api.py#L324

This allows simple name based filtering against that field.

2. Create a custom filterset for the API endpoint

https://github.com/inventree/InvenTree/blob/9ca0d6b0a95561ee8c71a6e3c4e815a4b0a6daeb/InvenTree/part/api.py#L797

This allows you to create much more complex filters which *will* be described in the documentation. The filters you construct here can be much more complicated.

3. Add docstring

Finally (and this is the "least good" option) if the fields do not auto-describe you can add an entry to the docstring for the API endpoint. This is obviously not idea.

#### Notes

**Multi Variable Filters**: In the instance here you want to construct a filter based on the provided "location" and also the provided "depth". I do not know of a nice way to make a complex filter based on two different parameters which fits into one of the self-describing functions as outlined above. I *imagine* that you could build it within a custom FilterSet class. Worth looking into.

**Legacy Code**: A lot of the API is a bit "dated" - you will find plenty of instanced where custom filtering is hacked into either the `get_queryset` or `filter_queryset` methods. This is *bad* and a result of me learning as I go. This means that some more complex API functions are currently undocumented because they are not exposed to the `api-doc` endpoint. I am starting to migrate these to use the `FilterSet` class whenever I touch old code. And any modern code I try to use the `FilterSet` class.

It is probably worth moving your code here to use the `FilterSet` approach if possible

SchrodingersGat:
> Is there a globals file or should we add a setting for the MAX_DEPTH, for better UX?

It would be worth making this a user-configurable setting, I think. @matmair has outlined this above.

In addition to adding a new setting in `common/models.py` you will need to add an entry into the "settings" page template - there are plenty of examples of how this is done currently.

e.g. https://github.com/inventree/InvenTree/blob/master/InvenTree/templates/InvenTree/settings/stock.html

luwol03:
Thank you for your long response. 

> Multi Variable Filters: In the instance here you want to construct a filter based on the provided "location" and also the provided "depth". I do not know of a nice way to make a complex filter based on two different parameters which fits into one of the self-describing functions as outlined above. I imagine that you could build it within a custom FilterSet class. Worth looking into
Sorry, but after reading through the [Django filters](https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html#adding-a-filterset-with-filterset-class) documentation I have no idea how to implement that. I would appreciate it if you can implement that for me on one resource (StockLocation or PartCategory).

luwol03:
I'm now looking into the settings. In which category should I add the setting? Should it be globally (-> Serverconfiguration) or should there be two individual settings one in the categories and one in the stock page?

SchrodingersGat:
I think a global setting makes sense here:

image

luwol03:
@SchrodingersGat I implemented the setting now. I hope that this [comment](https://github.com/inventree/InvenTree/pull/3451#issuecomment-1203611842) is not sunk in the comments. I would really appreciate your help.

InvenTree/part/api.py
```diff
@@ -85,6 +85,8 @@ def filter_queryset(self, queryset):
 
         cascade = str2bool(params.get('cascade', False))
 
+        depth = params.get("depth", None)
```
SchrodingersGat, line 88: I would suggest that you try parsing `depth` to an int here, and set it to None if the integer coerce fails. 

- Won't throw an error if you pass something like `?depth=yes`
- You don't have to cast to int multiple times later on.

InvenTree/stock/api.py
```diff
@@ -241,6 +241,8 @@ def filter_queryset(self, queryset):
 
         cascade = str2bool(params.get('cascade', False))
 
+        depth = params.get("depth", None)
```
SchrodingersGat, line 244: Same as /part/api.py - try / catch integer conversion

SchrodingersGat:
> Thank you for your long response.
> 
> > Multi Variable Filters: In the instance here you want to construct a filter based on the provided "location" and also the provided "depth". I do not know of a nice way to make a complex filter based on two different parameters which fits into one of the self-describing functions as outlined above. I imagine that you could build it within a custom FilterSet class. Worth looking into
> 
> Sorry, but after reading through the [Django filters](https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html#adding-a-filterset-with-filterset-class) documentation I have no idea how to implement that. I would appreciate it if you can implement that for me on one resource (StockLocation or PartCategory).

Reading through the django-filters docs, I also cannot see a nice way to make a filter which is dependent on two variables simultaneously. The [method filter](https://django-filter.readthedocs.io/en/latest/ref/filters.html#filter-method) provides only limited information via the arguments.

Maybe we just continue with the implementation you have.

luwol03:
> Maybe we just continue with the implementation you have.

Ok.

@SchrodingersGat I now refactored the int conversion into its own helper function. I hope you also like that implementation. From my point, I think this PR is ready to merge. Let me know if there is something else to improve.

SchrodingersGat:
@luwol03 this is looking very nice! The last thing we should add is some unit tests to ensure that:

1. The "depth" filter is working as intended
2. Any future changes to not cause a regression

so e.g. for the part category you can add a new test to this class:

https://github.com/inventree/InvenTree/blob/a2c2d1d0a40d10882a0008dca6f18ae6bc21c161/InvenTree/part/test_api.py#L22

which tests that the 'depth' filter works as intended under various conditions. Some examples I can think of:

- Without a 'category' parameter passed in
- With a 'category' parameter passed in 
- Different "depth" levels specified

In each case, test that the returned results are what you would expect!

matmair:
Hi there @luwol03 could you please stop force pushing? The code comment notifications break with that and we are squashing PRs anyway before merging.

luwol03:
> Hi there @luwol03 could you please stop force pushing? The code comment notifications break with that and we are squashing PRs anyway before merging.

@matmair Sorry to do that. Good to know that you will squash merge anyway. I just wasn't sure about it and thought correcting the typo in the message is a good idea and amending the files I forget. But as I now know you'll squash merge, ill stop force pushing.

luwol03:
@SchrodingersGat I added some tests now, however I have not much experience with testing. I would appreciate some feedback.

matmair (approved):
Test look good @luwol03 ! Best is to always to comment why the tests assert something so that we know what broke once they fail - which they surely will at some point.

matmair:
Once the CI finishes (takes currently around 25 min) the checks will contain a link to the coverage results.

luwol03:
> Best is to always to comment why the tests assert something so that we know what broke once they fail - which they surely will at some point.

@matmair Should I change something?

InvenTree/stock/test_api.py
```diff
@@ -54,11 +54,44 @@ def setUp(self):
...
+            ({'cascade': False, 'depth': 10}, 8, 'Cascade with no parent, depth=0'),
+            ({'parent': 'null', 'cascade': True, 'depth': 0}, 7, 'Cascade with null parent, depth=0'),
+            ({'parent': 'null', 'cascade': True, 'depth': 10}, 8, 'Cascade with null parent and bigger depth'),
+            ({'parent': 'null', 'cascade': False, 'depth': 10}, 3, 'No cascade even with depth specified with null parent'),
+            ({'parent': 1, 'cascade': False, 'depth': 0}, 2, 'Dont cascade with depth=0 and parent'),
+            ({'parent': 1, 'cascade': True, 'depth': 0}, 2, 'Cascade with depth=0 and parent'),
+            ({'parent': 1, 'cascade': False, 'depth': 1}, 2, 'Dont cascade even with depth=1 specified with parent'),
+            ({'parent': 1, 'cascade': True, 'depth': 1}, 2, 'Cascade with depth=1 with parent'),
+            ({'parent': 1, 'cascade': True, 'depth': 'abcdefg'}, 2, 'Cascade with invalid depth and parent'),
+        ]
```
matmair, line 72: @luwol03 this seems to be a duplicate of the changes in `InvenTree/part/test_api.py` - could we reduce that with a helper function?

InvenTree/stock/test_api.py
```diff
@@ -54,11 +54,44 @@ def setUp(self):
...
+            'items',
+            'pathstring',
+            'owner',
+            'url'
+        ]
+
+        response = self.get(self.list_url, expected_code=200)
+        for result in response.data:
+            for f in fields:
+                self.assertIn(f, result)
```
matmair, line 94: This assert could provide a message what went wrong so that the error message is a bit more verbose

InvenTree/stock/test_api.py
luwol03: Not really, because they differ in the response length.

matmair: did not see that, sorry for the fuss

SchrodingersGat:
@luwol03 thanks for the tests, they look great!

luwol03:
no, I added some more tests :cry: I was to slow.

luwol03:
@SchrodingersGat should I open a new PR for that?

SchrodingersGat:
Sorry I should have confirmed with you. You can push to the same branch and then open a new PR.

luwol03:
> Sorry I should have confirmed with you. You can push to the same branch and then open a new PR.

no, because squash merge, I have to rebase. One sec

PR #3472, opened 2022-08-04
Depenency updates

matmair:
This PR:
- Updates all depenecies - at least the django one should be done ASAP - there is a critical security flaw in that
- Adds a workflow to daily compile the updates

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3472"></a>

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,0 +1,22 @@
+name: Update dependency files regularly
+
+on:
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
Will this make a PR every day or only when there are changes?

matmair:
@SchrodingersGat I will update the branch each day. We can pull that into master whenever we want.

PR #3473, opened 2022-08-04
Tests/stocklocation and partcategory

luwol03:
This PR adds the missed tests from #3451.

@SchrodingersGat Sorry, I didn't thought you're already awake, as you normally act like 3am to my time (UTC+2 or CEST).

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3473"></a>

PR #3474, opened 2022-08-04
vscode devcontainer

luwol03:
I already shared this in this in https://github.com/inventree/InvenTree/pull/3429#issuecomment-1205407752 . I used this setup for my latest two pr #3451 , #3473 and it works great.

@SchrodingersGat what do you think of that? What IDE are you using locally?

I will create this PR as draft and if the team likes this, I'll do some improvements I have in mind.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3474"></a>


SchrodingersGat:
@luwol03 I might need to get my head around this.

I use vscode for dev, with InvenTree running in docker-desktop:

This means that I can shell into the running containers via vscode, it's very clean! It also means that I am always running using the development docker environment.

luwol03:
@SchrodingersGat I improved it even a step further, maybe you will also like the debugger and the available syntax highlighting. I recorded a little video to show the features. Note, that the first time start will take some longer time, as the requirements needs to be fetched and the docker file be build.

https://user-images.githubusercontent.com/60048565/183087281-7df51a40-4743-41d6-b1f5-fbf19ad37309.mp4

image

.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
...
+    {
+      "name": "Python: Django",
+      "type": "python",
+      "request": "launch",
+      "program": "${workspaceFolder}/InvenTree/manage.py",
+      "args": [
+        "runserver"
+      ],
+      "django": true,
+      "justMyCode": true
```
matmair, line 16: Suggestion: Add a second config with justMyCode disabled - we tend to call a lot of 3rd party packages, debugging is nearly useless without looking into those.

.pre-commit-config.yaml
```diff
@@ -15,7 +15,7 @@ repos:
     -   id: check-added-large-files
     -   id: mixed-line-ending
 -   repo: https://github.com/pycqa/flake8
-    rev: '5.0.0'
+    rev: '5.0.4'
     hooks:
```
matmair, line 19: Please Change: This is a dependency update and might lead to problems further down the line - please submit this in a new PR so we can label and reproduce it easier.

.gitignore
```diff
@@ -66,9 +66,16 @@ secret_key.txt
...
 *.code-workspace
-.vscode/
 .bash_history
 
+# https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore
+.vscode/*
+#!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
```
matmair, line 76: Question: Won't this lead to a lot of changes in git? I have 50+ extensions in my full InvenTree dev enviroment

.devcontainer/devcontainer.json
```diff
@@ -33,26 +33,19 @@
...
+      "extensions": ["ms-python.python", "ms-python.vscode-pylance","batisteo.vscode-django"]
     }
   },
 
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
   "forwardPorts": [8000],
 
   // Use 'postCreateCommand' to run commands after the container is created.
-  "postCreateCommand": "python3 -m venv venv && . venv/bin/activate && cd /workspaces/InvenTree && pip install invoke && pip install -r requirements.txt && mkdir -p dev && inv setup-dev && inv migrate && inv server",
+  "postCreateCommand": "python3 -m venv venv && . venv/bin/activate && cd /workspaces/InvenTree && pip install invoke && pip install -r requirements.txt && mkdir -p dev && inv setup-dev && inv update",
```
matmair, line 48: Suggestion: Nothing from the requirements should be needed to call inv update, if you turn around the order we code safe one action probably:
```suggestion
  "postCreateCommand": "python3 -m venv venv && . venv/bin/activate && cd /workspaces/InvenTree && pip install invoke && inv update && mkdir -p dev && inv setup-dev",
```

.devcontainer/devcontainer.json
```diff
@@ -33,26 +33,19 @@
...
+        "python.linting.pylintPath": "/usr/local/py-utils/bin/pylint"
       },
 
       // Add the IDs of extensions you want installed when the container is created.
-      "extensions": [
-        "ms-python.python",
-        "ms-python.vscode-pylance",
-        "ms-python.black-formatter"
-      ]
+      "extensions": ["ms-python.python", "ms-python.vscode-pylance","batisteo.vscode-django"]
```
matmair, line 40: Suggestion: I use a coverage extension to quickly check if new tests cover all new code @luwol03 @SchrodingersGat would that be a reasonable addition? We would need to add `coverage.py` to the dev-reqs then probably.

matmair (changes requested):
Looks good @luwol03 ; Please move the dependency update out the rest will be a great improvement.

luwol03:
@matmair thank you for your review. Ill take a deeper look later.

Good idea, please suggest useful extensions (with their identifier) we should add.

.gitignore
luwol03: I just used the template linked above, but we can comment this and uncomment as we need it. I guess in `extensions.json` we can specify extensions for usage without the devcontainer, as they are already specified in the `devcontainer.json`.

.pre-commit-config.yaml
luwol03: sorry, forgot to remove that file from committing. It was changed by some of the auto update invokes.

.devcontainer/devcontainer.json
luwol03: Do you know how to improve the testing setup as its pretty bad marked in https://github.com/inventree/InvenTree/issues/3384#issuecomment-1206799712 ? I tried to setup the test extension, but didn't got it working, so I sticked with the CLI and prints.

matmair (approved):

luwol03:
@matmair can you share the identifiers of the extensions you recommend?

.devcontainer/devcontainer.json
matmair: @luwol03 we are using the default solution for Django. It is the same as in most projects.
The test extension normally auto-discovers test if you chose the right framework at the start (unittest or pytest).

I really do not care how the code gets developed as long as the prints do not make into production code. My suggestion would be to use the debug console or start the Django shell and develop the tests in there. Most test consist of short one-liners so that usually works fine.

luwol03: Which is the correct one? unites or pytest?

matmair:
@luwol03 Github codespaces are actually pretty expensive compared to gitpod in my experience.

matmair:
@luwol03 the changes to enable my testing setup in devcontainers turned out to be quite involved.
I think it is better to add some docs to CONTRIBUTING for this changeset and merge it before this PR gets out of hand. I will make a separate PR once I figure out how to streamline the whole thing (would currently need 5+ extensions, a lot of setup and an old node.js version)

.devcontainer/devcontainer.json
matmair: @luwol03 they support the same call API. While we use unittest, both will work.

luwol03:
> @luwol03 Github codespaces are actually pretty expensive compared to gitpod in my experience.

Yeah, just wanted to mention that one if somebody will need it.


> @luwol03 the changes to enable my testing setup in devcontainers turned out to be quite involved.
> I think it is better to add some docs to CONTRIBUTING for this changeset and merge it before this PR gets out of hand. I will make a separate PR once I figure out how to streamline the whole thing (would currently need 5+ extensions, a lot of setup and an old node.js version)

So this PR can be marked as ready to review for your side and you'll add your extensions in another PR?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/local/share/nvm/nvm.sh && nvm install ${NODE_VERSION} 2>&1"; fi
+
+# [Optional] If your pip requirements rarely change, uncomment this section to add them to the image.
+# COPY requirements.txt /tmp/pip-tmp/
+# RUN pip3 --disable-pip-version-check --no-cache-dir install -r /tmp/pip-tmp/requirements.txt \
+#    && rm -rf /tmp/pip-tmp
+
+# [Optional] Uncomment this section to install additional OS packages.
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive && \
+    apt-get -y install --no-install-recommends gettext libpangocairo-1.0-0
```
SchrodingersGat, line 18: For a fully compatible setup you should install all required system packages 

https://github.com/inventree/InvenTree/blob/e1d3392f3743f7813c0aba0705223f1442526ca7/Dockerfile#L61

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,71 @@
...
+  // Comment out to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
+  "remoteUser": "vscode",
+  "features": {
+    "git": "os-provided"
+  },
+
+  "containerEnv": {
+    "INVENTREE_DB_ENGINE": "sqlite3",
+    "INVENTREE_DB_NAME": "/workspaces/InvenTree/dev/database.sqlite3",
+    "INVENTREE_MEDIA_ROOT": "/workspaces/InvenTree/inventree-data/media",
```
SchrodingersGat, line 63: I'd prefer to keep all these directories under the same directory. Currently you have them split between `./dev/` and `./inventree-data/`.

For the development docker image we use `./data/media/` `./data/static/` (etc)

SchrodingersGat:
@luwol03 while I am hesitant to have "too many" ways to achieve the same thing (in this case a development setup) this does seem like a great way to get up and running very quickly.

There are some downsides I can see:

- Limited to sqlite (fine for simple dev)
- Doesn't run the background worker process (again, fine for simple dev)

I'd be happy to see this as a "get a quick dev environment setup" tutorial (documented of course), with a lead-in to the full development setup e.g. "for a more fully featured development setup, refer to the development docker guide" or something like that.

@matmair what do you think about this? How do you see this vscode integration fitting in with our current setup guide (we have 4 currently!)

.devcontainer/devcontainer.json
luwol03: I copied this from the gitpod file. But you're right.

luwol03:
@SchrodingersGat Thank you for your feedback.

You're not limited to sqlite. You can also configure a docker compose stack in the devcontainer.json and attach to a specified service. This would be helpful to setup a database, redis, ...

### Questions

1. How do I start the work?
2. How do we make a select which of the featured levels you'd like to start? Because I guess for first time contributor who has no idea about the setup, the current state is just fine, but for you as core developer, you need more different things like different database, worker, redis, ...

SchrodingersGat:
I'm not sure how *useful* it is to fully flesh out with multiple containers etc. Even having the background worker *not* running is a good thing for development as it forces the processes to run in the server thread - great for debugging.

As for "which features should you include"... I don't know. Every new setup is another thing we need to document and support. 

I'm happy to promote this as a "developer setup" guide, and anything beyond is left to the user. 

The current setup guides do not really help with new developers who want to get stuck into coding. This would be a great addition in this regard. We should add a new setup guide to the docs specifically for developers. Your video would work well there I think

luwol03:
If we want to use that video, I would like to make a new one which shows more available features. Maybe we should also wait until @matmair added some more extensions?

luwol03:
@SchrodingersGat Please let me know what the next steps will be.

SchrodingersGat:
@luwol03 I am currently running through the process of setting up on my system to replace my dev setup.

One thing I see is that we already *have* a dockerfile which you are essentially copying (the first part of at least).

Is it possible to inherit from the Dockerfile at the top level directory? That way if we ever have to update our docker file, at least we won't have to change two files?

SchrodingersGat:
I ran into this error during the initial setup process:

image

Have you seen this?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+# Update pip
+RUN pip install --upgrade pip
+
+# preserve command history between container starts
+# Ref: https://code.visualstudio.com/remote/advancedcontainers/persist-bash-history
+# Folder will be created in 'postCreateCommand' in devcontainer.json as it's not preserved due to the bind mount
+RUN echo "export PROMPT_COMMAND='history -a' && export HISTFILE=/workspaces/InvenTree/dev/commandhistory/.bash_history" >> "/home/vscode/.bashrc"
+
+WORKDIR /workspaces/InvenTree
```
SchrodingersGat, line 43: Does `/workspaces/InvenTree` have to be hardcoded here? Will this still work if the user calls the workspace something else? Or *must* the user load the workspace as `InvenTree`?

SchrodingersGat:
@luwol03 I made the following patch with some small additions for a "cleaner" initial setup.

Please apply the patch and push the changes.

[devcontainer.diff.txt](https://github.com/inventree/InvenTree/files/9276193/devcontainer.diff.txt)

.devcontainer/Dockerfile
luwol03: This is the path in the container. If you want to change that you also have to specify something in the devcontainer.json which is not available in codespaces, but this should be the default.

luwol03:
> Is it possible to inherit from the Dockerfile at the top level directory? That way if we ever have to update our docker file, at least we won't have to change two files?

I don't know of any way, because we need the vscode image as base. I'd be happy if you know any ways.


> @luwol03 I made the following patch with some small additions for a "cleaner" initial setup.
> Please apply the patch and push the changes.

Thank you for the additions. They look good. I'll apply it for you, but you can also push yourself as a maintainer to my branch. You need to checkout the refs/pulls/3348 ref I guess, but this is simpler with the gh cli, it's just gh pr checkout 3348 and then you're able to push to my branch.

luwol03:
@SchrodingersGat There is one thing I have to notice here,
```diff
   // Use 'postCreateCommand' to run commands after the container is created.
-  "postCreateCommand": "mkdir -p /workspaces/InvenTree/dev/commandhistory && python3 -m venv venv && . venv/bin/activate && cd /workspaces/InvenTree && pip install invoke && inv update && mkdir -p dev && inv setup-dev",
+  "postCreateCommand": "mkdir -p /workspaces/InvenTree/dev/commandhistory && cd /workspaces/InvenTree && python3 -m venv dev/venv && . dev/venv/bin/activate && pip install invoke && inv update && inv setup-test --ignore-update --path dev/demo-dataset",
```

If we do it like this, we should make sure that `setup-test` don't run if already data exists in the database, because this command runs every time you remove you recreate your devcontainer (e.g. rebuild, delete all exited containers and then start the devcontainer again, ...). But I prefer it as a task (Use cmd pallet type `> Run task` press enter, select `setup-test` press ok, wait until complete)

luwol03:
We can share this video if you want.

https://user-images.githubusercontent.com/60048565/183285495-77d9a32c-300f-4e23-b764-1678cfeb5e51.mp4

SchrodingersGat:
> If we do it like this, we should make sure that setup-test don't run if already data exists in the database, because this command runs every time you remove you recreate your devcontainer (e.g. rebuild, delete all exited containers and then start the devcontainer again,

Ah good point. I guess it makes more sense to run the "demo data" setup as a separate action. 

The `setup-dev` command failed locally for me, I'll have to look into that some more.

SchrodingersGat:
@luwol03 I'll have another look at this hopefully tomorrow :)

luwol03:
@SchrodingersGat Can you please share the error you had with setup-dev?

matmair:
Looks good. Don't wait for the extra extensions @luwol03 , I have to rewrite several tasks for it to work seamlessly.

luwol03:
@matmair no, I'm not waiting. I already posted the video. Currently I'm waiting for @SchrodingersGat to answer my notices from his patch in https://github.com/inventree/InvenTree/pull/3474#issuecomment-1207351858 . If that's all resolved, I think we can merge this from my side.

SchrodingersGat:
> Can you please share the error you had with setup-dev?

Looks like a permission error. Still see it when running the task.

image

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,14 @@
...
+cd /workspaces/InvenTree
+
+# create venv
+python3 -m venv dev/venv
+. dev/venv/bin/activate
+
+# setup inventree server
+pip install invoke
+inv update
+inv setup-test --ignore-update --path dev/demo-dataset
```
SchrodingersGat, line 14: You're right that this "setup-test" should not be performed "every time" the container is built. We should suggest it as a one-time step after setup.

Can you revert this to the "setup-dev" command?

SchrodingersGat:
Getting pretty close @luwol03 

Regarding the demo video, I would suggest a simplification to the workflow. There should be only three steps to get "up and running":

### Download / Build

Download (clone) code and build the devcontainer

### Demo Dataset

Install demo dataset with the `setup-test` task

### Debug Server

Launch the server in debug mode

luwol03:
And what is with showing the needed extension?

SchrodingersGat:
> And what is with showing the needed extension?

Sorry I don't understand what you are asking

luwol03:
@SchrodingersGat The remote containers extension is needed for all that to work. I showed that at the beginning after cloning the repo in my last video.

SchrodingersGat:
Ah understood. Yes that absolutely needs to be in there. I really liked the structure of your video. There are a number of smaller things that I guess I did not specify in my "simple three step plan" above :)

luwol03:
Can you describe exactly what you want to be in the video? And another idea would be to add litte texts to the video like Requirements, building, debugging, ... (the chapters)

SchrodingersGat:
@luwol03 I have opened a new issue on the docs site - https://github.com/inventree/inventree-docs/issues/350 - where we should discuss the documentation aspect on this. Let's keep this PR for the actual feature :)

SchrodingersGat:
I think the only outstanding item here is to work out the `setup-dev` bug which was a permission issue. @luwol03  do you have any idea why it might have failed on my end?

I'm running vscode under windows 10 if that makes any difference.

luwol03:
@SchrodingersGat Regarding to your bug, have you checked the file permissions in the container? The vscode user needs at least execute and read. Maybe find out which exact command that is and try running it directly instead through the post task.

I can only test on macos, and cannot reproduce that. I'll try that one more time later.

SchrodingersGat:
@luwol03 I'll have a look into that further. I'm happy to accept this as is, might just be my setup.

Can you have a look at https://github.com/inventree/inventree-docs/issues/350 when you have some time?

luwol03:
Oh nice, that this is merged. Yes, I already had a look at the other issue on the docs site. Great team btw.

Issue #3483, opened 2022-08-05
[BUG] Notification link to error message fails


SchrodingersGat:
If an error occurs, a notification is provided to staff users.

However the URL reverse fails and causes a 500 error:

![image](https://user-images.githubusercontent.com/10080325/183224005-c87db580-2218-418a-8030-e268e2adc679.png)

Ref: https://github.com/inventree/InvenTree/pull/3431

Note: This seems to be related to https://github.com/inventree/InvenTree/issues/3355 - only saw this in the condition where the simpleactionplugin was enabled.

The proposed fix in https://github.com/inventree/InvenTree/pull/3484 also seems to address this issue.

matmair:
@SchrodingersGat I have no idea why this stuff is all resulting from the app order; probably has to do with overwritten urls.

Issue #3355, opened 2022-07-18
[BUG] Internal Server Error


luwol03:
**Describe the bug**
I get an Internal Server error after start and don't know why.

**Steps to Reproduce**

1. Start the server
2. Watch logs
3. Navigate to `/admin`

**Expected behavior**
Should show the Django admin

![](https://user-images.githubusercontent.com/60048565/179567211-ee9312ce-17bd-46b1-9523-12cc2e612230.png)

**Deployment Method**
- [x] Docker
- [ ] Bare Metal

**Version Information**

```
# Version Information:
InvenTree-Version: 0.7.5
Django Version: 3.2.13


Database: postgresql
Debug-Mode: False
Deployed using Docker: True
```

Log:
https://gist.github.com/luwol03/45794e822bfd00730f9edcb259028915

matmair:
I can not reproduce this with a fresh install @luwol03 anything else you changed? Do you have any custom plugins activated?

SchrodingersGat:
@luwol03 this is a new one for me too. Looks like the postgres database container is having some issues. Please let us know if you made any changes to the `docker-compose` or `.env` files

luwol03:
Hello, the first lines until line 826 are cannons connect to database, because the database is not listening, but then the connection is successful but an SQL statement is executed 3 times concurrently which causes an error. And I guess this has something to to with that the 500 on /admin. I more and more think, that this is caused because I run it via macos docker. (Also the git safepermissions error). I try to export the database and try on Linux. Thank you for taking a look.

SchrodingersGat:
@luwol03 please let us know if you find anything further :)

luwol03:
@SchrodingersGat I now copied everything to a linux machine and started the containers, open `/admin` in browser and see error. I have no idea why. I would appreciate if you could take a look. I bundled everything into a targz archive:[inventree-mac.tar.zip](https://github.com/inventree/InvenTree/files/9142308/inventree-mac.tar.zip)

### Steps to reproduce:
1. extract the tar from the zip (gh only allowed me to upload zip)
2. `tar xfzv inventree-mac.tar`
3. `cd inventree-mac`
4. `docker-compose up -d && docker-compose logs -f`
5. wait for start and see error
6. browse `http://localhost:8000/admin` (if login is needed, there is an admin user: `admin` with password: `Passw0rd`)

luwol03:
Hey @SchrodingersGat, I analyzed it a bit further and I did the following:
[...]

PR #3484, opened 2022-08-06
load admin first

SchrodingersGat:
- Fixes https://github.com/inventree/InvenTree/issues/3355
- Fixes https://github.com/inventree/InvenTree/issues/3483

Placing 'django.contrib.admin' at the start of `INSTALLED_APPS` seems to address this issue, although it is a very unsatisfying result. Still no idea why the simpleactionplugin causes the admin page to error out.

@matmair any thoughts as to negative consequences of putting 'django.contrib.admin' at the start of the list?

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3484"></a>

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -135,6 +135,8 @@ MEDIA_URL = '/media/'
 # Application definition
 
 INSTALLED_APPS = [
+    # Admin site integration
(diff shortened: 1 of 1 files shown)


matmair:
@SchrodingersGat nothing negativ comes to mind. Plugins still work so should be fine.

Issue #3487, opened 2022-08-06
[FR] Specify custom background image for login screen.


snorkrat:
**Problem**
Can not specify custom background image for login screen.

**Suggested solution**
Similar to how I can specify a custom logo in the .env or config.yaml, there could be an option to specify `custom_bg` and `custom_bg_large` to replace `paper_splash.jpg` and `paper_splash_large.jpg`

_However, as noted in https://github.com/inventree/InvenTree/issues/3486 these would need to be in the` /static ` folder or somewhere other than the `/media` folder as by default files in the media folder require user authentication, so the splash background would not be seen as a user has not logged in._ 

**Describe alternatives you've considered**
I have manually replace the `paper_splash.jpg` and `paper_splash_large.jpg` files in the `static/img` directory.

matmair:
@snorkrat do you want to develop this? We are currently focusing on other parts of the project so it might take a while for anybody to get to this.

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/3489 - closely related issue which would help with implementing this feature.

snorkrat:
> @snorkrat do you want to develop this? We are currently focusing on other parts of the project so it might take a while for anybody to get to this.

I don't really have any experience or know how to develop this sorry!

PR #3499, opened 2022-08-08
Custom splash screen

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/3487

Adds the ability to specify a custom splash screen for the various login screens

e.g.

image


InvenTree/templates/account/base.html
```diff
@@ -49,10 +49,8 @@
 {% endblock %}
 </head>
 
-<body class='login-screen'>
-    <!--
-        Background Image Attribution: https://unsplash.com/photos/Ixvv3YZkd7w
```
matmair, line 54: Suggestion: Maybe we should add this attribution to the credits in the docs?

InvenTree/InvenTree/settings.py
```diff
@@ -839,8 +839,18 @@
...
 
+# Check for a custom splash screen
+CUSTOM_SPLASH = get_setting('INVENTREE_CUSTOM_SPLASH', 'customize.splash', None)
+
+if CUSTOM_SPLASH:
+    if static_storage.exists(CUSTOM_SPLASH):
+        logger.info(f"Loading custom splash screen from static directory: {CUSTOM_SPLASH}")
+    else:
+        logger.warning(f"The custom splash screen '{CUSTOM_SPLASH}' could not be found in the static directory")
+        CUSTOM_SPLASH = False
```
matmair, line 850: Suggestion: Maybe this section could be refactored together with CUSTOM_LOGO to remove duplication?

matmair (approved):

InvenTree/templates/account/base.html
SchrodingersGat: Yeah docs PR incoming

InvenTree/InvenTree/settings.py
SchrodingersGat: Could do, I'll leave that for a future PR though

Issue #3500, opened 2022-08-08
[BUG] "&" gets translated to "&amp;" when creating new company


eeintech:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### Describe the bug*

See title

### Steps to Reproduce

1. Create new company
2. Use "&" symbol in the name
3. Click submit
4. Name shows "&amp;" in the name instead of "&"

### Expected behavior

"&" should stay as-is

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information
InvenTree-Version: 0.9.0 dev
Django Version: 3.2.15
Commit Hash: 964d674
Commit Date: 2022-08-08
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Active plugins: False

matmair:
@eeintech this is a security measure to protect against CSRF/XSS as the symbol can be used in a scripting context. I am afraid it will have to stay that way as changing it would make us vulnerable again.
Ref: https://github.com/inventree/InvenTree/security/advisories/GHSA-rm89-9g65-4ffr

matmair:
Also please add the version information to all bug reports. The demo is a moving target

eeintech:
I see... not idea, some company names are using this symbol (and maybe others).

I don't understand how would it make our instances vulnerable for this kind of attack as most (I hope all) are behind firewalls/VPN, can you explain?

A work around would be to display "&" instead of "&amp;" using the translation files?
Can an API GET return "&" while POST keep the security fix?

SchrodingersGat:
@eeintech I get the frustration, the current solution is not ideal, but is a compromise to prevent a malicious user from inserting [XSS attack](https://owasp.org/www-community/attacks/xss/) into another users browser. 

Data gets sanitized on the server and then again when rendering. This double sanitization step makes the "chain of data" something like:

`Black & Decker` -> `Black &amp; Decker` -> `Black &amp;amp; Decker`

There is *probably* something that can be done about this, would require a bit of a deep dive to understand all the edge cases. There are multiple places where data gets rendered (in different ways):

- HTML templates
- Via javascript (e.g. API forms)
- tables (using bootstrap-table)

Each would need separate consideration.

SchrodingersGat:
@matmair I can see the appeal of keeping data "as entered" by the user wherever possible. Having e.g. `Black &amp; Decker` in the database is *wrong* and what happens when the user exports data e.g to excel?

[...]

PR #3503, opened 2022-08-09
Update bleach clean function

SchrodingersGat:
- Invalid tags are stripped out
- & > < characters are accepted

Ref: https://github.com/inventree/InvenTree/issues/3500

@matmair this is a proposed fix for #3500 which does not replace special chars like `&` `<` `>` from database fields.

Thoughts? There may be another way to handle this.

The major caveat here is that any `<html><tags>` will be removed completely, rather than escaped.

So something like:

```Black & Decker ```

becomes

```Black & Decker```

Essentially anything which looks like a HTML tag will be removed rather than escaped?

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3503"></a>

### TODO

- [ ] ~~Fix rendering in modal forms~~
- [ ] ~~Confirm correct rendering in bootstrap-table~~
- [ ] ~~Confirm correct rendering in HTML templates~~
- [ ] ~~Fix existing unit tests~~
- [ ] ~~Add more unit tests~~

### DIFF
--- a/InvenTree/InvenTree/mixins.py
+++ b/InvenTree/InvenTree/mixins.py
@@ -1,15 +1,18 @@
 """Mixins for (API) views in the whole project."""
 
+from django.utils.translation import gettext_lazy as _
+
 from bleach import clean
 from rest_framework import generics, status
+from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
 
 
 class CleanMixin():
-    """Model mixin class which cleans inputs."""
+    """Model mixin class which cleans inputs using the Mozilla bleach tools."""
 
-    # Define a map of fields avaialble for import
-    SAFE_FIELDS = {}
+    # Define a list of field names which will *not* be cleaned
+    SAFE_FIELDS = []
 
     def create(self, request, *args, **kwargs):
         """Override to clean data before processing it."""
@@ -34,7 +37,7 @@ class CleanMixin():
 
         return Response(serializer.data)
 
-    def clean_string(self, data: str) -> str:
+    def clean_string(self, field: str, data: str) -> str:
         """Clean / sanitize a single input string.
 
         Note that this function will *allow* orphaned <>& characters,
@@ -57,6 +60,12 @@ class CleanMixin():
         for o, r in replacements.items():
             cleaned = cleaned.replace(o, r)
 
+        # If the length changed, it means that HTML tags were removed!
+        if len(cleaned) != len(data):
+            raise ValidationError({
+                field: [_("Remove HTML tags from this value")]
+            })
+
         return cleaned
 
     def clean_data(self, data: dict) -> dict:
@@ -77,8 +86,11 @@ class CleanMixin():
         clean_data = {}
 
         for k, v in data.items():
-            if isinstance(v, str):
-                ret = self.clean_string(v)
+
+            if k in self.SAFE_FIELDS:
+                ret = v
+            elif isinstance(v, str):
+                ret = self.clean_string(k, v)
             elif isinstance(v, dict):
                 ret = self.clean_data(v)
             else:
--- a/InvenTree/InvenTree/mixins.py
+++ b/InvenTree/InvenTree/mixins.py
@@ -34,6 +34,31 @@ class CleanMixin():
 
         return Response(serializer.data)
 
+    def clean_string(self, data: str) -> str:
+        """Clean / sanitize a single input string.
+
+        Note that this function will *allow* orphaned <>& characters,
+        which would normally be escaped by bleach.
+
+        Nominally, the only thing that will be "cleaned" will be HTML tags
+
+        Ref: https://github.com/mozilla/bleach/issues/192
+        """
+
+        cleaned = clean(data, strip=True)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: InvenTree/InvenTree/mixins.py [hunk 1]
```diff
@@ -34,6 +34,31 @@ def update(self, request, *args, **kwargs):
...
+
+        Note that this function will *allow* orphaned <>& characters,
+        which would normally be escaped by bleach.
+
+        Nominally, the only thing that will be "cleaned" will be HTML tags
+
+        Ref: https://github.com/mozilla/bleach/issues/192
+        """
+
+        cleaned = clean(data, strip=True)
```
matmair, line 48: Question: This adds additional stripping - won't that remove spaces that were formally left in? Are we creating a new situation like the current one with this in the future when someone uses spaces to signal *something*?

InvenTree/InvenTree/mixins.py
```diff
@@ -46,17 +71,21 @@ def clean_data(self, data: dict) -> dict:
...
         Returns:
-            dict: Profided data sanatized; still in the same order.
+            dict: Provided data sanatized; still in the same order.
         """
+
         clean_data = {}
+
         for k, v in data.items():
             if isinstance(v, str):
-                ret = clean(v)
```
matmair, line 54: Suggestion: From a readability standpoint I would prefer to run `clean` here and then run a renamed `clean_string` with the output. Namesuggestion: clean_escape

matmair (commented):
Looks good so far - I still have to run some functional tests

matmair:
@SchrodingersGat tbh I am not really happy with this solution but it will probably 'solve' #3500 - which in my eyes is not a bug but a minor inconvenience that could be solved by just forbidding a bunch of characters - ampersand included.

InvenTree/InvenTree/mixins.py
SchrodingersGat: I think that `strip` here is culling bad tags, not removing spaces.

e.g. something like:

| Original | strip=False | strip=True |
| --- | --- | --- |
| ```hello world ``` | ```hello world &lt;img src='hax0r.png'&gt;``` | ```hello world ``` |

SchrodingersGat:
> which in my eyes is not a bug but a minor inconvenience that could be solved by just forbidding a bunch of characters - ampersand included.

Forbidding the characters entirely is a different approach, and "cleaner" too. But these characters are pretty darn useful and appear often in real datasets.

For example I just discovered that any "link" field will replace `&` with `&amp;` - breaking any links to URLs which contain & characters.

Any solution here is a comprimise:

### No Validation

#### Good

- No forbidden characters
- Users get what they expect
- Simple

#### Bad

- Attack vectors everywhere
- Hackers rejoice

### Existing Solution (bleach)

#### Good

- Safe against XSS etc

#### Bad

- User data gets corrupted unexpectedly
- Rendered data does not match input data
- Unable to truly represent some data

### Illegal Chars

#### Good

- Simple to implement
- Pretty safe in terms of XSS etc

#### Bad

- Some markdown elements won't work
- Some links will break and users will complain
- Data might not be representative

### This proposal

Bleach input and remove any html tags, and then put `&` `<` and `>` chars back in

#### Good

- Any html tags are removed, so should be safe against XSS
- User data is preserved (except for anything looking like a html tag)
- Data is stored and rendered "as is" (no unexpected rendering)

#### Bad

- Anything looking like HTML data is silently dropped

matmair:
@SchrodingersGat maybe we should add validation step that runs clean and errors out so that stuff is not just happening magically in the background?
Also urls could just be validated and do not really need to run though clean()

SchrodingersGat:
I like the idea of validation. Throwing an error something like "html tags detected"?

eeintech:
> ### This proposal
> Bleach input and remove any html tags, and then put `&` `<` and `>` chars back in
> 
> #### Good
> * Any html tags are removed, so should be safe against XSS
> * User data is preserved (except for anything looking like a html tag)
> * Data is stored and rendered "as is" (no unexpected rendering)
> 
> #### Bad
> * Anything looking like HTML data is silently dropped

Looks the best trade-off to me!

SchrodingersGat:
@matmair LMK what you think of where it's at now. Should be a reasonable compromise I think.

matmair (approved):
Looks good @SchrodingersGat - there are still a bunch of todos in the header, are they resolved?

SchrodingersGat:
> Looks good @SchrodingersGat - there are still a bunch of todos in the header, are they resolved?

Yes, they no longer apply as we went with a different approach.

eeintech:
A big THANK YOU! :partying_face:

MR-DOS:
Hi, does this also correct the bug which causes replacements in Part Parameter Value field? For example if I insert "3-4...Vcc for VCC<30V" into part parameter value, then the "<" gets replaced by "&lt;". Then if you edit the field and save again, the "&lt;" will then show as this code and not as "<".
image

SchrodingersGat:
@MR-DOS should fix that also.

PR #3517, opened 2022-08-10
Improve plugin testing

matmair:
Adds testing for #3516 
This PR:
- Refactors `collect_plugins` to foley collect plugins (needed for simpler testing)
- Makes `get_module_meta` more resilient in changing env
- Removes several safety catches from coverage
- Fixes several issues with custom plugin dirs
- Adds tests for helpers and registry functions

Closes #3408

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3517"></a>

PR #3535, opened 2022-08-14
Plugin framework updates

### LINKED ISSUE #3478: [FR] Plugins compatible version in metadata

Might be if no other solution could work, bug: #3355

**Problem**

During major upgrades some plugins might break or will be unsupported, this is normal. WordPress, Firefox, Chrome, and major SaaS providers also define versions for each plugin/extension/addon to ensure consistency across versions.

There should be version control for plugins so that when an author creates a plugin, users could ensure such a plugin works for their build and won't break it.

**Suggested solution**

Define system supported version in metadata, if the plugin support < system version, disable the plugin by default, and when the user wants to enable it create a warning that this plugin was not designed for the newest version which might result in unexpected behavior.

None

**Examples of other systems**

WordPress, Firefox, Chrome etc ... all have versions for their plugins and compatibility list according to platform version.

### LINKED ISSUE #3520: Disable plugin signatures by default
Plugin signatures lead to confusion. While I think it is very important for users that are concious of their supply-chain security most do not need it.

This should be hidden by default.
I want to dev this.

@matmair Thank you very much, now it's working.
But one thing is not clear to me, what is the purpose of this signature field? And how can I get my plugin a signature? 
image

_Originally posted by @wolflu05 in https://github.com/inventree/InvenTree/discussions/3506#discussioncomment-3373793_

### LINKED ISSUE #3534: [FR] Make AppMixin plugin loading more flexibel
# Issue
AppMixin plugins have to be contained in a package named the same as the `NAME` of the plugin.

# Solution
Make it dynamically register the packagename and use that to load the AppConfig - Maybe even make the name of the AppConfig configurable in the plugin.

# Ref
@wolflu05 I checked out the issue and the behaviour for packages is hard-coded right now.
That has no technical reason - it just reflects that I build plugins around my workflows - in which `NAME` and package-name are atomically the same.

We can only change it in `0.9.9` as it will be breaking but I think we can change it till then. There are at least a few changes lined up for plugins so it does not hurt.

_Originally posted by @matmair in https://github.com/inventree/InvenTree/discussions/3523#discussioncomment-3391097_

matmair:
This PR:
- Adds typing to parts of the plugin ecosystem
- Adds a list with all plugin entities to the registry
- Show if plugin is a sample in admin
- Breaking: Changes `path` and `file` to class methods to enable access even if the plugin is not initialised
- Makes display of signatures optional - turned off by default #3520
- Breaking: Changes the AppMixin loading mechanism to always use the packagename instead of `NAME` #3534
- Add version checker for plugins - disables incompatibel apps by default #3478

This PR inlcudes https://github.com/inventree/InvenTree/pull/3517

Closes #3478
Closes #3520
Closes #3534 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3535"></a>


matmair:
Had to force push due to a polluted branch

matmair:
This is ready for review and merge @SchrodingersGat

SchrodingersGat:
@matmair please address conflicts and I'll have a look :)

matmair:
@SchrodingersGat should be good to go now!

SchrodingersGat (approved):

wolflu05:
After this changes, my plugin doesn't load with the following error in the plugin error stack:
```
discovery | django | Model class inventree_bulk.models.BulkCreationTemplate doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

I guess this has something todo with a wrong configuration, but I cannot see where. This is my plugin [repo](https://github.com/wolflu05/inventree-bulk-plugin/tree/dev).

wolflu05:
@matmair if I add a `Meta` class with an app label, this error is gone and the plugin loads without issues, but how do I get it working without that extra meta class like it was working before?

PR #3536, opened 2022-08-14
refactor custom file lookup

matmair:


SchrodingersGat:
Looks good, except that (unless there's a reason) I'd rather it be called `get_custom_file` over `get_cstm_file`?

matmair:
@SchrodingersGat no problem, I will change this

SchrodingersGat:
Thanks for the refactor :)

SchrodingersGat:
@matmair bad news. This has broken *stuff*..

https://github.com/inventree/InvenTree/runs/7837878419?check_suite_focus=true

Basically, the `invoke` script makes use of this file and must run *before* django is installed. Classic chicken and egg.

I think we might have to revert this one? LMK if you can see a cleaner solution otherwise I'll rewind this.

matmair:
@SchrodingersGat I will create a patch

Issue #3528, opened 2022-08-12
Label text looks weird if multiple labels are printed.


miggland:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### Describe the bug*

When printing multiple labels to a PDF, the text includes spacing. When printing a single PDF, this does not occur.

Example with the built-in part label template `Barcode Part Label - Simple part label with Code128 barcode`:
One label created:
![image](https://user-images.githubusercontent.com/1314853/184400791-5dba0cf4-28e1-43ac-8b8f-38f926eb9124.png)

Two labels created:
![image](https://user-images.githubusercontent.com/1314853/184400846-7de74142-ea1b-4f05-aaca-97671625a3bc.png)

Example with a stock item template I made. These use the same template, but in one I call:

Output with one single label:
![image](https://user-images.githubusercontent.com/1314853/184399658-e9f4caad-ca77-44c4-a08a-2510f4e4f8bb.png)

Output with multiple labels:
![image](https://user-images.githubusercontent.com/1314853/184399778-4381fa89-da76-43d8-9c4b-ea4f8723920d.png)

The fact that some lines are missing is not a problem, that is due to the template. However, where is the spacing from?

### Steps to Reproduce

1. Print label with text for one part.
2. Print label with text for several parts at once.
3. Compare, and see that it looks weird in one case.

### Expected behavior

The output should look the same regardless of whether I create multiple labels at once or not.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.9.0 dev
Django Version: 3.2.15
Commit Hash: 139e059a4
Commit Date: 2022-08-10
Database: postgresql
Debug-Mode: True
Deployed using Docker: True
Active plugins: []

miggland:
I tried seeing what happens by going to debug mode, but that unfortunately puts the second label on top of the second label so doesn't really help..

nwns:
I am seeing the same sort of issue with 0.8.1 running on bare metal (Debian 
Bullseye (11)).
![image (2)](https://user-images.githubusercontent.com/10690911/184453457-9317ac1f-64e3-454b-b7f8-514a2f0624d6.png)
![image (1)](https://user-images.githubusercontent.com/10690911/184453461-1af2245e-5205-4110-bd5c-576e63a6fdeb.png)
[...]

PR #3537, opened 2022-08-14
WIP: a solution for multi label printing

nwns:
A hack and slash solution to multi-label printing (note that this is against stable not master, though it looks like it will auto merge against master as well).

A proposed fix for #3528 

TODO:
- [ ] Determine how this affects the plugin system.  Will require understanding what the plugin system is accomplishing in the first place.
- [ ] Template debug mode is ignored.  Labels are always rendered as pdf.
- [ ] PDF inline is ignored.  File is always downloaded.

A, likely breaking, refactor of the label templates would probably be required to clean this up.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3537"></a>

### DIFF
--- a/InvenTree/label/api.py
+++ b/InvenTree/label/api.py
@@ -1,5 +1,6 @@
 """API functionality for the 'label' app"""
 
+import datetime
 from django.conf import settings
 from django.core.exceptions import FieldError, ValidationError
 from django.http import HttpResponse, JsonResponse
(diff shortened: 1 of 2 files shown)


matmair:
While I applaude the effort this breaks nearly every feature of labels and sets us back a year in that area. @SchrodingersGat any insights into reports/how that could be done while keeping the current featureset?

nwns:
@matmair I am not sure I follow how it breaks nearly everything?

There will be breaking changes for the templates themselves, but everything in the [documentation](https://inventree.readthedocs.io/en/latest/report/labels/) should be possible how I proposed it.

Alternatively, assuming I understand what the BarcodePlugins are supposed to do, we always use the plugin path and convert every label to a png first, then using my new template, create the pdf for the user from the list of images as a single pdf document.

SchrodingersGat:
@nwns this is actually an issue with the [weasyprint](https://github.com/Kozea/WeasyPrint) library we use for PDF generation. I've tracked it down to a known issue: https://github.com/Kozea/WeasyPrint/issues/1517

I'll pin the library to a version that works and that will address the issue.

nwns:
@SchrodingersGat awesome.  Downgrading weasyprint fixed it for me.

SchrodingersGat:
Good to hear!

PR #3545, opened 2022-08-15
Fix bug in exporting records

miggland:
Introduced in #3392

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3545"></a>

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -267,7 +267,7 @@ def export_records(c, filename='data.json', overwrite=False, include_permissions
 
     print(f"Exporting database records to file '{filename}'")
 
-    if filename.exists() and overwrite is False:
+    if Path(filename).is_file() and overwrite is False:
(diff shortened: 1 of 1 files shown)


matmair (approved):

matmair:
This does not seem to be covered by tests acuratly. @miggland do you mind assing a small test that triggers for the error case? Coverage for labels/reports is kind of low.

@SchrodingersGat this might be worth a backport to stable

miggland:
> This does not seem to be covered by tests acuratly. @miggland do you mind assing a small test that triggers for the error case? Coverage for labels/reports is kind of low.

Can you point me to where the `invoke` scripts are tested? I haven't found that file (and if there are none, your request becomes quite a bit more difficult..)

matmair:
@miggland ok I will implement the test in a seperate PR

PR #3547, opened 2022-08-15
moves imports to enable startup without django beeing loaded

### LINKED ISSUE #3546: Django not available on startup
@matmair bad news. This has broken *stuff*..

https://github.com/inventree/InvenTree/runs/7837878419?check_suite_focus=true

Basically, the `invoke` script makes use of this file and must run *before* django is installed. Classic chicken and egg.

I think we might have to revert this one? LMK if you can see a cleaner solution otherwise I'll rewind this.

_Originally posted by @SchrodingersGat in https://github.com/inventree/InvenTree/issues/3536#issuecomment-1214999583_

matmair:
Fixes #3546

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3547"></a>

### DIFF
--- a/InvenTree/InvenTree/config.py
+++ b/InvenTree/InvenTree/config.py
@@ -7,9 +7,6 @@ import shutil
 import string
 from pathlib import Path
 
-from django.contrib.staticfiles.storage import StaticFilesStorage
-from django.core.files.storage import default_storage
-
 import yaml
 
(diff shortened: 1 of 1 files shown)


matmair:
@SchrodingersGat this should fix the error introduced in #3536

SchrodingersGat:
Thanks @matmair looks like that should do it

Issue #3575, opened 2022-08-18
Decimal point in label template


wolflu05:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### Describe the bug*

Part templates are not rendered correctly in size due to localized interpolation of `width` and `height` in https://github.com/inventree/InvenTree/blob/da715d738163f09afb3fbdf897a7ea1b3ceae6ab/InvenTree/label/templates/label/label_base.html#L6-L11 because in https://github.com/inventree/InvenTree/blob/da715d738163f09afb3fbdf897a7ea1b3ceae6ab/InvenTree/label/models.py#L178-L179 `width` and `height` are passed in as floats which Django templating handles localized and interpolates with `,` instead of `.` as we can see in the weazyprint error that height and width are invalid values:
```
Ignored `width: 50,1mm` at 36:5, invalid value.
Ignored `left: 50,1mm` at 43:5, invalid value.
```

### Steps to Reproduce

1. have German as language (I don't know where to set this)
2. Try to print label to pdf
3. See errors in console and invalid size of pdf

### Expected behavior

PDF should be in the size like selected for the template.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

```
# Version Information:
InvenTree-Version: 0.9.0 dev
Django Version: 3.2.15
Commit Hash: ebdb1c707
Commit Date: 2022-08-18
Database: sqlite3
Debug-Mode: True
Deployed using Docker: False
Active plugins: [{'name': 'inventree_bulk', 'slug': 'bulkaction', 'version': '0.1'}, {'name': 'CupsLabels', 'slug': 'cups', 'version': '0.1.0'}]
```

SchrodingersGat:
That's interesting, localisation bugs pop up in the strangest of places.

PR #3578, opened 2022-08-19
Fix dimensions for label templates

SchrodingersGat:
- Disable localization in certain areas
- Different localization settings could mess with label generation

Closes https://github.com/inventree/InvenTree/issues/3575

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3578"></a>


wolflu05:
Thank you for that quick fix

PR #3591, opened 2022-08-23
Sanitize search input

SchrodingersGat:
Remove ASCII and unicode control characters from search input text.

Copying and pasting text to search with may include hidden control characters which can either:

- Cause incorrect / unexpected search results
- Throw errors if the database collation does not allow extended unicode characters

A simple fix here is to strip such characters out from the search text

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3591"></a>

### DIFF
--- a/InvenTree/templates/js/translated/helpers.js
+++ b/InvenTree/templates/js/translated/helpers.js
@@ -341,10 +341,10 @@ function sanitizeInputString(s, options={}) {
     s = s.trim();
 
     // Remove ASCII control characters
-    s = s.replace(/[^\x20-\x7E]+/g, "");
+    s = s.replace(/[^\x20-\x7E]+/g, '');
 
     // Remove type-setting charactrs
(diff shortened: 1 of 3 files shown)


InvenTree/templates/js/translated/helpers.js
```diff
@@ -326,3 +327,24 @@ function setupNotesField(element, url, options={}) {
...
+ *
+ * - Remove leading / trailing whitespace
+ * - Remove hidden control characters
+ */
+function sanitizeInputString(s, options={}) {
+
+    s = s.trim();
+
+    // Remove ASCII control characters
+    s = s.replace(/[^\x20-\x7E]+/g, '');
```
matmair, line 344: Suggestion: @SchrodingersGat  should we maybe just reduce to utf8?

matmair (approved):

InvenTree/templates/js/translated/helpers.js
SchrodingersGat: Worked out a much simpler regex for removing non-printable characters

PR #3597, opened 2022-08-24
Remove hidden characters from form fields

SchrodingersGat:
- Remove control and non-printable characters from form fields (server side)
- Fix regex on client side (javascript)

Ref: https://github.com/inventree/InvenTree/pull/3591

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3597"></a>

### DIFF
--- a/InvenTree/InvenTree/mixins.py
+++ b/InvenTree/InvenTree/mixins.py
@@ -1,9 +1,8 @@
 """Mixins for (API) views in the whole project."""
 
-import re
-
 from django.utils.translation import gettext_lazy as _
 
+import regex
 from bleach import clean
 from rest_framework import generics, status
 from rest_framework.exceptions import ValidationError
@@ -73,11 +72,11 @@ class CleanMixin():
                 field: [_("Remove HTML tags from this value")]
             })
 
-        # Remove control characters from the provided string
-        cleaned = re.sub('[\x01-\x1F]+', '', cleaned)
+        # Remove ASCII control characters
+        cleaned = regex.sub(u'[\x01-\x1F]+', '', cleaned)
 
-        # Remove non-printable characters from the provided string
-        cleaned = re.sub('[^ -~]+', '', cleaned)
+        # Remove Unicode control characters
+        cleaned = regex.sub(u'[^\P{C}]+', '', cleaned)
 
         return cleaned
 
--- a/InvenTree/InvenTree/mixins.py
+++ b/InvenTree/InvenTree/mixins.py
@@ -1,5 +1,7 @@
 """Mixins for (API) views in the whole project."""
 
+import re
+
 from django.utils.translation import gettext_lazy as _
 
 from bleach import clean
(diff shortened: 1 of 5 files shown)


MR-DOS:
@SchrodingersGat 
Hi, by removing all "other" (C) unicode characters, you are forbidding 15/16 of Unicode table, because "other" also includes non-UTF-8 "surrogate" characters. So, once someone tries to use UTF-16, certain characters may be unavailable.
Yes, most of the upper part of Unicode table is just gibberish, but there are still some usable characters - Kana extensions for Japanese users, CJK unified ideograms for Chinese/Japanese/Korean users.
It should be no problem as long as users stick to UTF-8/32, but UTF-16 may be somewhat limited by the "C" group. Is this inclusion of way too many characters intentional?
Also, ASCII DEL character 0x7F is still allowed. So is 0x00 NUL character. What happens if I try to insert this character?

matmair:
This was done mostly for security reasons, there were several attacks using non-printable characters so this decision is very unlikely to be rolled back / changed

MR-DOS:
Cs and Co blocks are printable characters.
Cf and Cc (and maybe Cn) are non-printable.
C (which is used in the new commits) includes all 5 of these, thus also removing printable characters.

The reason why I'm bothering you with this is because until 0.8.4, it was impossible to use characters like _-()*^ in parameter names. For security reasons. And it really made it difficult to make readable parameter names. So I just wanted to say that this may become an issue which someone may run into later.

SchrodingersGat:
@MR-DOS can you provide a reference for these `Cs` and `Co` character blocks? Would not be too hard make an adjustment here

MR-DOS:
@SchrodingersGat My reference was this:
https://javascript.info/regexp-unicode

Cs should be almost anything above 0xFFFF - those characters do not fit into UTF-16 and have to use surrogate characters which take 4 bytes and have to be a specific combination of "low surrogate" (0xD800-0xDBFF) and "high surrogate" (0xDBFF-0xDFFF).

Co (0xE000-0xF8FF, 0xF0000-0xFFFFD and 0x100000–0x10FFFD and in UTF-16 also 0xDB80-0xDBFF) are "private use" meaning they are reserved and unassigned. Those can be used for special characters not present in Unicode should someone need them. Currently, those are publicly used mostly for medieval scripts, phonetics, Tibetan ligatures, some combined letters, east asian symbols, some mathematical symbols, Tolkien script. More notably, Apple uses several hundred characters in this range. I haven't found any sign of it being used for control characters.
https://en.wikipedia.org/wiki/Private_Use_Areas

SchrodingersGat:
> Also, ASCII DEL character 0x7F is still allowed. So is 0x00 NUL character. What happens if I try to insert this character?

Good catch, I'll look into this.

Regarding Cs / Co - do you have a use-case where you have needed these characters, or purely hypothetical at this stage?

MR-DOS:
Right now, hypothetical.

PR #3598, opened 2022-08-24
Bug fix for loading asset files in reports (#3596)

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/3596

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3598"></a>

### DIFF
--- a/InvenTree/report/templatetags/report.py
+++ b/InvenTree/report/templatetags/report.py
@@ -5,7 +5,7 @@ import os
 
 from django import template
 from django.conf import settings
-from django.utils.safestring import mark_safe
+from django.utils.safestring import SafeString, mark_safe
 
 import InvenTree.helpers
 from common.models import InvenTreeSetting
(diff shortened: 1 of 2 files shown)


InvenTree/report/templatetags/report.py
```diff
@@ -28,11 +28,15 @@ def asset(filename):
     Raises:
         FileNotFoundError if file does not exist
     """
+    if type(filename) is SafeString:
+        # Prepend an empty string to enforce 'stringiness'
+        filename = '' + filename
```
matmair, line 33: question: wouldn't `str()` be a cleaner solution or is that not working?

SchrodingersGat: Nope, tried that, the addition of an empty string was called out on the django bug tracker as the "right" solution :p

matmair: Interesting... does not look pythonic but if upstream thinks it is a good solution who am I to judge?

Issue #3595, opened 2022-08-24
[FR] Open PDFs in new window


SchrodingersGat:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

Ref: https://github.com/inventree/InvenTree/discussions/3594

### Suggested solution

Allow labels / reports to be opened in a new browser window

### Describe alternatives you've considered

Leave existing solution which opens PDFs in the same window.

### Do you want to develop this?

- [ ] I want to develop this.

PR #3602, opened 2022-08-24
Open labels and reports in a new window

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/3595

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3602"></a>


matmair:
@SchrodingersGat weren't there security concerns with opening a new tab with attachments?

SchrodingersGat:
That was specific to HTML files opening up in the context of the same tab.

We already support opening PDFs "in browser", this simply opens in a new tab.

PR #3633, opened 2022-09-02
Add decorator to not require auth

matmair:
This PR adds the option to not require auth for certain sites.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3633"></a>


SchrodingersGat:
@matmair what is the intent here? Are there any existing views you would wrap in this function?

matmair:
@SchrodingersGat no. This would be useful for plugins requiring oAuth flows for API access (eg. digikey).

SchrodingersGat:
Is there a simple way you can think to unit test this decorator?

PR #3637, opened 2022-09-02
Plugin API refactor

matmair:
This PR:
- includes a small refactor: Moves the checks if a plugin ref is valid into a separate function
- adds a check if the requested plugin is active
- fixes the lookup for settings. Currently, we are looking at `SETTINGS` - which refers to the static definition of settings. A plugin can also define settings at runtime so we need to lookup `settings`.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3637"></a>


InvenTree/plugin/api.py
```diff
@@ -164,18 +197,10 @@ def get_object(self):
...
-        plugin = registry.get_plugin(plugin_slug)
-
-        if plugin is None:
-            # This only occurs if the plugin mechanism broke
-            raise NotFound(detail=f"Plugin '{plugin_slug}' not found")  # pragma: no cover
+        # Look up plugin
+        plugin = check_plugin(plugin_slug)
 
-        settings = getattr(plugin, 'SETTINGS', {})
+        settings = getattr(plugin, 'settings', {})
```
SchrodingersGat, line 203: @matmair are you sure this won't break existing code?

SchrodingersGat:
Sorry @matmair this one slipped under my radar. Please see my review comment

InvenTree/plugin/api.py
matmair: @SchrodingersGat It might but the current behavior is wrong and inconsistent.
SETTINGS should be a static constant following python naming conventions while settings is a variable. In the plugin-loading process settings is set from SETTINGS but could be changed later (either directly or by modifying the loading method) to dynamically change the provided settings. This functionality is present in multiple mixins and also documented in the docstrings. So the current usage of SETTINGS is just wrong.

SchrodingersGat:
@matmair fair enough. Please address failing CI and I'll merge

SchrodingersGat:
@matmair CI is passing but conflicts in `tests.py`?

matmair:
> @matmair CI is passing but conflicts in `tests.py`?

Those are new; will take a look now

matmair:
@SchrodingersGat fixed the conflict; no real code changes so should go through

Issue #1435, opened 2021-04-02
Support different unit of measure (UoM) for each supplier


cdwilson:
I'm not exactly sure the correct term for this, but I'm wondering if InvenTree supports different unit of measure (UoM) for each `SupplierPart`.  For example, when buying screws from McMaster-Carr, the UoM can be specified in "packs", where the a quantity of "1 pack" equals 100 screws:

![image](https://user-images.githubusercontent.com/14631/113455184-dc0f6d80-93be-11eb-92f1-a1ccd4ddf0fd.png)

If I'm building an assembly that requires 4 screws per assembly, and I'm building 25 assemblies in total, I would need to generate a PO for McMaster for qty 1 (pack) of screws.  However, if have multiple suppliers for the same `Part`, each supplier may have a different UoM for their SKU which will determine the Qty to be used on the PO.  Another supplier for the same part may sell the screws in "bags" where "1 bag" equals 25 screws (and the PO for this other supplier would require Qty 4).

I'm curious if InvenTree has any way to handle this?

SchrodingersGat:
Sounds like this is a very similar issue to:

- https://github.com/inventree/InvenTree/issues/1126
- https://github.com/inventree/InvenTree/issues/1192

Currently InvenTree does not support this, other than adding quantity "100" to the order..

I'd be interested in suggestions as to how to best handle this...

matmair:
@SchrodingersGat @cdwilson I would like to implement this:
- each SupplierPart can be assigned a number of contained parts (one part-definition per SupplierPart - either Part, ManufacturerPart or SupplierPart)
- on receiving the item it gets split automatically in the screen - so if 100 are in the SKU then you can assign 20 to location A and 80 to location B in this step
- multi-part SKUs should be done as a separate FR I think (see #1192)

Let me know what you think - I am thinking about imlementing it in a few weeks, should be more contained then other things.

SchrodingersGat:
I also think it makes sense to implement these things separately. Please let me know when you've got space to implement this and we can discuss!

cdwilson:
@matmair FYI for when you start looking into this, I ran across these articles which give a good overview of UoM and some additional examples:

- https://www.aligni.com/doc/guides/units-of-measure/
- https://www.aligni.com/doc/basics/units/

> each SupplierPart can be assigned a number of contained parts (one part-definition per SupplierPart - either Part, ManufacturerPart or SupplierPart)

[...]

PR #3644, opened 2022-09-05
Feature: Supplier part pack size

SchrodingersGat:
Adds the concept of "pack size" for the supplier part

- e.g. buy a single "bag" which contains 10 parts
- e.g. buy a single "tube" which contains 30mL vs a larger tube which contains 100mL

Closes https://github.com/inventree/InvenTree/issues/1435

### TODO

- [x] Add documentation for new feature - https://github.com/inventree/inventree-docs/issues/365
- [x] Add extra unit testing for 'on_order' quantity
- [x] Add unit testing for receiving items with pack_size != 1
- [x] Update demo dataset with supplier parts which have larger pack size - https://github.com/inventree/demo-dataset/pull/32
- [x] Update queryset annotation for `on_order` quantity (must include calculation of pack_size)
- [x] Update 'on_order' method for Part model
- [x] Update "on order" display for particular part page
- [x] Display pack size when adding lines to order via normal "Add Line Item" form
- [x] Display pack size when adding lines to order via "auto calculate" form
- [x] Ensure "units" are displayed appropriately for part quantity values in HTML views and forms

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3644"></a>


SchrodingersGat:
@eeintech @matmair do you think that "pack_size" is the right nomenclature here?

matmair (approved):
Looks good. Am I right that the new field is only as a data point but without buisne3ss logic for splitting?

matmair:
> @eeintech @matmair do you think that "pack_size" is the right nomenclature here? 

I think package_unit would be better.

SchrodingersGat:
> Looks good. Am I right that the new field is only as a data point but without buisne3ss logic for splitting?

There will have to be some business logic added. Thinking about it, it becomes quite complex but should be attainable. For example, (and the use case I am trying to implement specifically for) say you have a part "thermal adhesive" which your units are "cc". You can buy it from one supplier in a tube of 30cc, and another in a tub of 500cc. The supplierpart quantity is "1" in each case, but the amount of "part" you are getting is markedly different.

So there needs to be logic in place to convert the incoming stock quantity from "supplier part units" to "part units". And then update every place where the "on order" quantity is reported to take this new pack quantity into account...

matmair:
@SchrodingersGat sounds a lot like https://github.com/inventree/InvenTree/issues/1435

derekwolfson:
This is nice -- can you add this as a context variable to reports?  Would be nice to show total quantity and total packs/sets on purchase orders.

matmair:
@derekwolfson the PR is merged half a year. Please open new feature requests for new requests.

PR #3689, opened 2022-09-18
Update config_template.yaml

IncompleteIntegers:
I removed the space characters preceding 'static_root:' in line 132 and 'media_root:' in line 129 as they cause a YAML parsing error if not removed when uncommenting these options.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3689"></a>


SchrodingersGat (approved):

SchrodingersGat:
@IncompleteIntegers good catch, thanks for the patch.

PR #3742, opened 2022-10-03
Remove broken action

matmair:
This action broke recently - seems like it (the actions repo) is abandoned in general so I would propose to just outright remove it.


matmair:
@SchrodingersGat this seems to block ppl from time to time... I think it is not worth the hassle.

SchrodingersGat:
Thanks @matmair

PR #3743, opened 2022-10-04
Add package-based install method

matmair:
This PR adds installing inventree bare-metal via the native package managers. This will reduce installation requirements by a lot. We can define dependencies, package the code and ensure safe setups.

The whole thing is powered by [Packager](https://packager.io), I got the idea from the smooth experience offered by Zammad.
For now only Debian and Ubuntu are supported. SLES, CentOS, Rocky Linux should also be possible. I have not used those in a while so I refrained from trying to implement them.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3743"></a>


matmair:
@SchrodingersGat this will need to be squashed on merge pls 😇

matmair:
The link in the readme is already set to the path once this is merged;
@SchrodingersGat either you or I need to enable packaging once this is merged.

matmair:
For testing, you need a VM or docker running Ubuntu LTS (18.-20.) or a modern Debian (11 or 11). Run the following command as a user with sudo rights:
```bash
curl https://raw.githubusercontent.com/matmair/InvenTree/deploy-test/contrib/install.sh | sh
```
After 3-15min you will have a fully set up InvenTree instance running Nginx and SQLite as a database.
The configs are located in `/etc/inventree`, the app code is in `/opt/inventree` and data is in `/opt/inventree/data`.

SchrodingersGat:
1. @matmair so essentially this would be replacing the exiting "manual" bare-metal install process?

2. Can you make a PR to the docs, in parallel, before merging this in?

> either you or I need to enable packaging once this is merged.

Can you elaborate on this for me?

matmair:
@SchrodingersGat 
1: Yes
2: Yes; will get to work on this
3: The packaging itself and distributing is done via https://packager.io/ - free for open source

matmair:
Ok this might need a day or two more; While writing the docs I had some improvement ideas to make the whole thing more flexible and reduce steps even more

matmair:
@SchrodingersGat the docs PR is now in https://github.com/inventree/inventree-docs/pull/368
To finish this you will need to sign up [here](https://packager.io/users/auth/github) using the InvenTree org account so we get the namespace `inventree` - ping me if you need assistance.

Procfile
```diff
@@ -0,0 +1,2 @@
+web: env/bin/python InvenTree/manage.py runserver 0.0.0.0:$PORT
```
SchrodingersGat, line 1: Should we be pushing a "proper" server (e.g. gunicorn) by default? My understanding is that the django server is not particularly performant?

matmair: @SchrodingersGat oh this is still the test setup; give me a few minutes

matmair: @SchrodingersGat now using gunicorn

matmair:
Finally all database backends, envs etc. work. I will leave it at this as I keep adding and adding more niche features for no real gain.

matmair:
@SchrodingersGat any thing missing here?

SchrodingersGat (approved):

SchrodingersGat:
> @SchrodingersGat any thing missing here?

Looking pretty good :)

SchrodingersGat:
@matmair thanks for the great work here :)

PR #3782, opened 2022-10-12
Update deps to newest releases

matmair:
Dependency update.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3782"></a>


matmair:
@SchrodingersGat at least Django and oauthlib are pressing updates; there are CVEs fixed in them

PR #3790, opened 2022-10-16
Switch to registry for tasks

### LINKED ISSUE #3787: [FR] Make registry for scheduled tasks

### Problem statement

Currently, we are registering tasks with a function and static names as strings.

### Suggested solution

A decorator for the functions might be better suited and reduce code duplication.

We could switch to function references instead of strings.

### Examples of other systems

_No response_

- [X] I want to develop this.

matmair:
Adds a registry for tasks and a wrapper to quickly register scheduled tasks.

Closes #3787 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3790"></a>


matmair:
@SchrodingersGat this way it might be simpler to add new tasks - it enables registering tasks with a wraper.

SchrodingersGat (approved):

SchrodingersGat:
Nice, this does make things as lot cleaner :)

SchrodingersGat:
Hmm as per [the documentation](https://django-q.readthedocs.io/en/latest/schedules.html) scheduled tasks can *only* be specified using dotted string notation - you cannot pass an actual function reference:

image

Looks like we'll have to revert this one

matmair:
> Hmm as per [the documentation](https://django-q.readthedocs.io/en/latest/schedules.html) scheduled tasks can _only_ be specified using dotted string notation - you cannot pass an actual function reference:
> 
> 
> Looks like we'll have to revert this one

That can be worked around rather easy, I'll create a patch

PR #3791, opened 2022-10-16
Supplier integration

### LINKED ISSUE #3261: [FR] Supplier integration
**Problem**
There are a bunch of integrations for different vendors. Most implement the same logic.

**Suggested solution**
Create models and APIs to make integration easy. Create them in a way that TOSs are not borken.
- [x] Model + API for API keys and connections settings
- [ ] Model + API for requests + answers (+ cleanup)
- [x] Plugin mixin to enable easy search integration
- [ ] API + python integration to enable third parties proxied calls without keys on the system (or knowledge of them at all).

A lot of small implementations work right now.

Yes - I have some code that will be seriously reduced by that.

I really would like to get some feedback on this and maybe make a list of all vendors that should be considered:
- [x] DigiKey
- [ ] Mouser
- [ ] SnapEDA #566
- [ ] OcotPartAPI #541
- [ ] LSCS https://github.com/inventree/InvenTree/issues/677#issuecomment-612674739

matmair:
This PR:
- GenericWebTransaction model
- transitions WebhookMessage to GenericWebTransaction
- Adds WebConnection (including admin, API)
- Adds WebConnectionTransaction  inheriting from GenericWebTransaction)
- Adds ConnectionSetting (including API, admin)
- Adds PluginSearch view (in the API) to search through connections and adds it to search
- Adds SupplierMixin
- Adds get_api_url method to PluginConfig model
- Adds plugin_connections tag

Open:
- [x] tests for webconnections
- [ ] tests for search API
- [x] API to add part from search result
- [x] tests for part add API
- [x] UI for adding part from search result

Before merge:
- [ ] Move OAuth checks to mixin
- [ ] Move requests middleware to mixin
- [ ] Move `WebConnectionData` to plugin namespace
- [ ] Move `SearchRunResult` to plugin namespace

Fixes parts of #3261


matmair:
Ok only tests left before this can be tried out

matmair:
Ok, this is ready for testing.

## Install
1.1 Use this branch
1.2 activate plugins
1.3 install and activate `https://github.com/matmair/inventree-digikey/tree/v1` (use v1 branch!)
1.4 make sure all integrations are active
## One-time setup
2.1 You need to register a production app at [DigiKey](https://developer.digikey.com/).
2.2 Activate all product information endpoints on the app
2.3 Go to the settings page for the plugin
2.4 Add a new connection under 'web connections'
2.5 Add your production key and secret to the connection
2.6 Click on the `setup` URL link on the plugin page
2.7 Log in on the Digikey auth page
## Test instructions
3.1 Open the start page of your instance
3.2 Click on 'Supplier API' navigation tab
3.3 Click on 'Import supplier part'
3.4 Fill out the popup. Select the connection you added in 2.4, add a digikey reference (eg. 1528-2280-ND) and select the category the part should be created in
3.5 The new part should open
## Caveats
4.1 THIS IS NOT PRODUCTION READY
4.2 The current setup flow is not how it is intended to be in the final stage
4.3 The digikey sample plugin is very naive and has limited options
4.4 Currently only search and import APIs are integrated (the sample plugin only contains import API for now)
4.5 The plugin does not do oauth refreshes for now - so you will need to press the setup button from time to time - THIS IS NOT PRODUCTION READY

SchrodingersGat:
@matmair this is a lot to comprehend! I'm going to have to set aside some time to look at this :)

matmair:
@SchrodingersGat I still have to do a bunch of refactoring, waiting till all tasks are done is probably more efficient.

matmair:
A tip on how the test is working on sqlite but not postgres would be highly appreciated though.

InvenTree/plugin/base/supplier/models.py
```diff
@@ -0,0 +1,75 @@
...
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+import common.models
+from plugin import registry
+from plugin.models import PluginConfig
+from plugin.plugin import InvenTreePlugin
+
+
+class ConnectionSetting(common.models.BaseInvenTreeSetting):
```
wolflu05, line 11: Note: This class needs some rework after merging master in (Ref: #4834 )

matmair: probably, a lot has changed in this area

matmair:
Still in the WIP pipeline

SchrodingersGat:
@matmair I'm still very keen to see this implemented - do you have motivation on this still too? Does this in your mind require the react integration, or can we look at getting it merged in now?

matmair:
I will move this to a plugin; it requires either the react integration or shipping the UI - which I will do with the plugin. Development velocity is also much easier to handle in a plugin.

30350n:
How is that going to work then? Would this plugin basically enable the use of other supplier plugins? Or would it just be one big plugin supporting multiple suppliers?

Tbh I haven't really had time to look at this in depth yet, but I feel like it'd integrate well with what I've been working on in #5509 and the backend part should probably be in core ... buuut I really have no clue about the frontend requirements here. Could this maybe be divided into multiple smaller features/PRs?

Also, regarding actual supplier support: I've seen that you already have a WIP digikey plugin. I've been working on my own part importer script (basically a simplified/improved version of Ki-nTree, with proper support for nested categories). It's working quite well for my purposes, but I still feel like it shouldn't really be a script, but a plugin aswell, thus useable directly from the inventree interface. So I'd be interested in contributing that into core/your plugin aswell.

matmair:
There will be a plugin that provides a basic UI and all that is required to handle the typical flows like search, data gathering, and creation of new parts/stock. I will not continue development with core as a target.

This plugin can be used as a target for vendor-specific APIs.

nikolai2111:
Is there a better / recent setup guide than this https://github.com/inventree/InvenTree/pull/3791#issuecomment-1439224710?

In my current setup there is only one setting available and it's called "Supplier - The Supplier which acts as 'DigiKey'".

matmair:
This or is closed, development of this was shifted and none of the linked components are finished. I refactored basically the whole thing, some of the decisions made here were not scalable enough.

30350n:
> In my current setup there is only one setting available and it's called "Supplier - The Supplier which acts as 'DigiKey'".

That's from the supplier barcode plugins (introduced in #5509).

nikolai2111:
Ah I see, I thought it was the same thing.
So there won't be a native DigiKey plugin?

Thanks for the info.

PR #3792, opened 2022-10-16
Raise error if version fails

### LINKED ISSUE #3789: [Plugin] Raise error message if plugin fails version check
If a particular plugin fails version check during loading, it currently silently loads (but is inactive).

It would be helpful to add an error message to alert the user that the plugin has failed to load, due to a version incompatibility.

matmair:
Raise an error message if the plugin fails the version check.

Closes #3789

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3792"></a>


matmair:
@SchrodingersGat does this cover what you had in mind?

SchrodingersGat:
@matmair I was actually thinking of a simpler display message here:

image

But I like your approach too! Perhaps it makes sense to have admin users notified on any plugin load crash?

matmair:
@SchrodingersGat did not even think of the error stack, I only use that during development. Will add a message there too.

Doing admin notifications might be good too. At least in dev environments that might create a lot of messages - a bunch of sample plugins raise errors.

matmair:
@SchrodingersGat aaaand I just remembered why there is a simple error stack: database calls while initialising plugins are problematic and also slow. I switched back to the simple message;

SchrodingersGat (approved):

SchrodingersGat:
Looks like CI needs some attention here

SchrodingersGat:
@matmair ready to merge?

matmair:
@SchrodingersGat yes

PR #3803, opened 2022-10-17
Revert "Switch to registry for tasks"

SchrodingersGat:
Reverts inventree/InvenTree#3790

Reason: scheduled tasks cannot be registered using the function reference - only using 'dotted string' notation:

image


SchrodingersGat:
@matmair we will need a new PR to perform daily backups as the task scheduling has been removed for the `run_backup` task

PR #3805, opened 2022-10-17
Fix task register

matmair:
This PR fixes the wrong call that was reverted in #3803
In detail:
- it enables multiple tasks lists (optional)
- uses function name instead of ref
- adds a step to collect tasks from apps

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3805"></a>

### DIFF
--- a/InvenTree/InvenTree/apps.py
+++ b/InvenTree/InvenTree/apps.py
@@ -25,7 +25,7 @@ class InvenTreeConfig(AppConfig):
 
     def ready(self):
         """Setup background tasks and update exchange rates."""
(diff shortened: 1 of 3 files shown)


matmair:
@SchrodingersGat I tested it manually - should work as expected now. I looked into testing this e2e but that seems quite involved with the multiple servers etc.

matmair:
@SchrodingersGat this should be ready to go

SchrodingersGat:
@matmair just testing locally now :)

SchrodingersGat:
Fingers crossed, but it looks good on my end.

Issue #3836, opened 2022-10-22
[FR] Add package support for CentOS, SLE, Ubuntu 22.04


matmair:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

We currently do not package for [all possible](https://github.com/crohr/pkgr/wiki/Supported-distributions) operating systems. Missing are:
- [ ] Ubuntu 22.04
- [ ] CentOS
- [ ] Suse Linux Enterprise Server

### Suggested solution

Add support. That involves (per OS):
- [ ] extend the packaging configs
- [ ] check that all packaging scripts run through
- [ ] test installing on a local machine
- [ ] test updates
- [ ] test full plugin support
- [ ] test setting config values
- [ ] check all logs for errors
- [ ] check that logs work
- [ ] extend the installer
- [ ] full run-through with the installer on a new machine (DO, AWS, GCP, Azure, ...)
- [ ] check the set permissions, opened ports and added 3rd party software configs following security-best-practices
- [ ] extend docs with support

### Describe alternatives you've considered

The current support covers major LTSs and bare metal works on everything else

### Do you want to develop this?

- [ ] I want to develop this.

matmair:
Some more infos for anybody willing to tackle this:
- we use pkgr / packager.io for packaging
- the current state can be found on the [dashboard](https://packager.io/gh/inventree/InvenTree)
- the per-os settings are in this file: https://github.com/inventree/InvenTree/blob/master/.pkgr.yml
- packaging scripts are in this dir: https://github.com/inventree/InvenTree/tree/master/contrib/packager.io
- we only build the branches master/stable
- for development I recommend creating a fork, initiating packager.io for this fork (only enable the branch you use for packager work); this only works for public repos

matmair:
Not stale, just nobody working on it

SchrodingersGat:
Not stale  @github-actions

matmair:
I think there is no visible intrest by users @SchrodingersGat can we close this?

SchrodingersGat:
I'm happy to close it out if you think it's not worthwhile

artizirk:
To get Inventree installed on Debian 13 Trixie I did the following


1. Create fake packages for python3.11 python3.11-dev python3.11-venv using `equivs` and install the resulting deb
	https://askubuntu.com/a/656153
[...]

PR #3841, opened 2022-10-24
Packager add 22.04 support

matmair:
WIP

Part of https://github.com/inventree/InvenTree/issues/3836

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3841"></a>


matmair:
Currently running into a pretty big show-stopper: 
grafik
the SSL module is not available in the buildpack so pip installs are not working

matmair:
This is a somewhat known issue; I would prefer to not build openssl on each packaging run: https://stackoverflow.com/questions/72133316/ubuntu-22-04-libssl-so-1-1-cannot-open-shared-object-file-no-such-file-or-di

matmair:
ok we need to get this package onto the build server: https://packages.ubuntu.com/focal/libssl1.1

matmair:
Resting this for now; the build is pretty much not working rn; as we are rebuilding the env in the postinstall we might be able to just disable the compile step but that would not be very future proof.

SchrodingersGat:
@matmair did you want this in 0.9.0?

matmair:
@SchrodingersGat this is actually not needed anymore as the 20.04 package seems to run fine on 22.04

PR #3854, opened 2022-10-25
Make PartCategories to be filterable by name and description

martonmiklos:
As we discussed here:
https://github.com/inventree/inventree-python/issues/143

There are cases when an user would like to make an exact search to category by name or description.
This PR adds these two fields to the filterable fields.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3854"></a>

### DIFF
--- a/InvenTree/InvenTree/api_version.py
+++ b/InvenTree/InvenTree/api_version.py
@@ -2,10 +2,13 @@
 
 
 # InvenTree API version
-INVENTREE_API_VERSION = 77
+INVENTREE_API_VERSION = 78
 
 """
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@martonmiklos looks great! Only addition I would request is to increment the API version here:

https://github.com/inventree/InvenTree/blob/824a44a4869d6b8c15ee96c790e1887fa3412598/InvenTree/InvenTree/api_version.py#L5

And add a brief description of the change.

SchrodingersGat:
Thanks for the contrubition :)

PR #3876, opened 2022-10-28
Allow PartCategories to be filtered by parent_id

martonmiklos:
I apologize for not including this field in #3854, but now I worked on pimping Kin-Tree, and there would be a need for querying child categories.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3876"></a>

### DIFF
--- a/InvenTree/InvenTree/api_version.py
+++ b/InvenTree/InvenTree/api_version.py
@@ -2,10 +2,13 @@
 
 
 # InvenTree API version
-INVENTREE_API_VERSION = 78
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@martonmiklos this is already possible:

https://github.com/inventree/InvenTree/blob/e6461929b0c3f96a192e4385a756c7a8f97b84fc/InvenTree/part/api.py#L116

Otherwise, the "part category" table would not work properly:

image

So, you can just filter like `?parent=<x>`

Issue #3886, opened 2022-11-01
[FR] in a placed PO: edit function for extra line item 


gooutbefree:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

In a placed purchase order, an order item can be edited (or deleted) e.g. updating the price. This function is missing for an extra line item

### Suggested solution

I propose to add this functions also to an extra line item because also these parameters can change e.g. shipping costs ..., parts, that are in the same order but not part of the standard company stock ...

Further I think it's needed to cancel a single PO line item. Up to now you can only cancel the whole PO or delete a line item (information lost)

### Describe alternatives you've considered

no alternatives

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
See https://github.com/inventree/InvenTree/pull/3888

PR #3888, opened 2022-11-01
Allow editing of extra line items in order table

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/3886

Allows "extra line items" to be edited even if an order is placed / completed / etc

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3888"></a>


matmair:
@SchrodingersGat this change seems problematic to me. It makes extra lines not reliable as they always can be changed later and there is no record of it. I would strongly suggest reconsidering this.

SchrodingersGat:
Would you be happy to make it an option?

matmair:
Either that or a clear warning for the release notes

PR #3931, opened 2022-11-14
Support structural part categories

### LINKED ISSUE #3897: [FR]Support "structural" categories

### Problem statement

It would be great if there would be an option to create part categories which would not allow direct part creation.
I.e. I would not like that lazy people in my team would add parts directly under the documentation category, they should either use the proper categories or create a new one.

The same might be useful for the Stock locations as well. 

### Suggested solution

Implement it ;)

-

### Examples of other systems

Arena PLM has this option, they call Structural categories

- [X] I want to develop this.

martonmiklos:
Added a new 'structural' property to the art categories.
No parts can be assigned to structural categories only sub category creation is allowed. 

Fixes #3897

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3931"></a>


wolflu05:
Good work and idea. I can also see this behavior for stock locations.

matmair:
@martonmiklos the branch was missing a migration, I added it

wolflu05:
@matmair This is duplicate, your migration just Alters just a ' in the help Text near by it's. Can't these two be just squashed into one?

martonmiklos:
> and idea. 

Idea was came from the Arena PLM 😆 

> I can also see this behavior for stock locations.

That's going to be the next thing to plumb if this got merged.

martonmiklos:
> @martonmiklos the branch was missing a migration, I added it

Uhm, the branch had the migration (just I altered the help text in the python code after generation with an apostrophe), just it failed applying it with something similar error:

```
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0089_auto_20221112_0128, 0089_partcategory_structural in part).

```
I have squashed the two migrations and force pushed it (yay force of habits), but tests dies the same way now.
Any help on the reason would be welcome (I do not have any 0089_auto_20221112_0128 locally so I assume it is generated).

SchrodingersGat:
The migration file conflict is because there's another migration `0089` - https://github.com/inventree/InvenTree/blob/master/InvenTree/part/migrations/0089_auto_20221112_0128.py

What you should do to fix this is:

1. Merge `master` branch back into your branch 
2. Change your migration to `0090_xxxx.py`
3. Edit your migration file so that the "dependencies" points to `0089_xxx` not `0088_xxx`

InvenTree/part/test_api.py
```diff
@@ -403,6 +404,54 @@ class Target(IntEnum):
...
+        )
+
+        part_count_before = Part.objects.count()
+        try:
+            part = Part.objects.create(
+                name="Part which shall not be created",
+                description="-",
+                category=structural_category
+            )
+        except ValidationError:
```
matmair, line 428: Suggestion: A better way to make sure the right error is thrown is with wrapping the whole thing in an assertion instead of an try/except

```
with self.assertRaises(ValidationError):
    …
```

InvenTree/part/test_api.py
```diff
@@ -403,6 +404,54 @@ class Target(IntEnum):
...
+            description="-",
+            category=non_structural_category
+        )
+
+        part.category = non_structural_category
+
+        try:
+            part.save()
+        except ValidationError:
+            pass
```
matmair, line 451: Question(blocking): this Error is not raised during the tests (see test coverage); why is it here? If it should be raises I would suggest using self.assertRaises and to remove it if that is not the case

matmair (changes requested):

SchrodingersGat:
@martonmiklos this looks really good. A couple of points I'd like to see:

### API Filtering

Allow the *PartCategory* API to be filtered by the "structural" parameter. This allows you to return results which are only structural, or only not structural, etc.

### Part Edit Form Update

Update the "part fields" method (in `part.js`) to filter the "category" field by `structural: false` - this filters out structural categories when creating or editing a part

### API Version Bump

Increment the API version in `api_version.py`

martonmiklos:
> @martonmiklos this looks really good. A couple of points I'd like to see:
> ### API Filtering
> 
> Allow the _PartCategory_ API to be filtered by the "structural" parameter. This allows you to return results which are only structural, or only not structural, etc.
> ### Part Edit Form Update
> 
> Update the "part fields" method (in `part.js`) to filter the "category" field by `structural: false` - this filters out structural categories when creating or editing a part
> ### API Version Bump
> 
> Increment the API version in `api_version.py`

One thing came into my mind when fixing the issues above:
There is a button in the part add dialog which could be used to add directly a new category.
It might make sense to hide the Structural form input in the case if the new Part category form is invoked from there.

SchrodingersGat (approved):

SchrodingersGat:
@matmair just waiting on your approval for the changes :)

matmair (approved):
Looks good

PR #3966, opened 2022-11-20
Fix typing

matmair:
This PR adds a small typing fix that makes installs with python 3.8 easier

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3966"></a>


SchrodingersGat:
Can we revert our min version requirement to 3.8 then?

matmair:
@SchrodingersGat I would rather not do that. It works with the standard tests on 3.8 but I have not extensively tested InvenTree with 3.8 and we do not have full tests coverage.
This mainly fixes an error that I discovered replicating #3965

PR #3986, opened 2022-11-21
Add unit test for deleting part which has pricing information

SchrodingersGat:
Attempting to delete a part (either via API or admin interface) raises an integrity error on the PartPricing table:

image

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/3986"></a>


SchrodingersGat:
Cannot reliably reproduce this issue.

matmair:
@SchrodingersGat might this be something with caching?

SchrodingersGat:
Interesting idea, I do not think so though. I found that deleting a Part via the shell worked without issue. Deleting via the API or the admin interface threw an error on occasion, but after observing the initial issue I was not able to get it to happen again.

SchrodingersGat:
@matmair can you try to replicate on your end? I think I have narrowed it down to the following:

If pricing data exists (or has existed for a part), deleting via the API or admin interface throws an IntegrityError. However deleting via the shell works normally.

matmair:
@SchrodingersGat can reproduce with this infomation set

matmair:
@SchrodingersGat it could be a missing key. Do we have any recently new FKs on part

SchrodingersGat:
> Do we have any recently new FKs on part 

Only in the PartPricing model (a OneToOne relationship with Part)

SchrodingersGat:
Strangely today I cannot reproduce the issue...

matmair:
@SchrodingersGat I can still replicate. But not with tests. The test does not fail - the manual action does

SchrodingersGat:
@matmair I think I've finally nailed it:

For reference: https://stackoverflow.com/questions/37313300/django-post-delete-is-it-safe-to-use-reverse-relation

SchrodingersGat:
For reference there was a somewhat circular relationship caused by the `post_delete` receiver on a number of models. The solution is to catch these errors before they propagate too far.

References:

- https://stackoverflow.com/questions/59074908/django-cascade-and-post-delete-interaction
- https://stackoverflow.com/questions/30074733/how-is-post-delete-firing-before-delete-in-django
- https://stackoverflow.com/questions/37313300/django-post-delete-is-it-safe-to-use-reverse-relation

PR #4007, opened 2022-11-29
Add 1D barcode support to browser webcam scanner

### LINKED ISSUE #3997: [FR] 1D Barcode Search

### Problem statement

Search for Parts by scanning a 1D barcode (e.g. a GS1 GTIN barcode in UPC and/or EAN format).

Currently the browser barcode scanner doesn't seem to detect 1D barcodes, while the android app scanner does.
Additionally, the app scanner can read existing label information encoded as a 1D barcode (e.g. `{"part":1}` encoded as a code-128 barcode).

I've tried looking through the interface and documentation for the appropriate place to designate a barcode for a part, but there doesn't seem to be one available.

### Suggested solution

Upgrade existing barcode and/or search functions to search for parts which have a designated barcode value.
Additionally, the browser barcode scanner will need 1D barcode recognition to be implemented if not already available.

Implementing this feature may require the Part model to be updated with a `barcode` field.

As this feature deviates from electrical components and leads toward general inventory management, it may make sense to add an option to the global barcode settings menu to make it optional.

Generic barcode searching could also be added to the normal search bar (i.e. not the barcode search menu) so any information encoded as a barcode could be used as a search input.

### Examples of other systems

_No response_

- [ ] I want to develop this.

maksimstojkovic:
Fixes #3997 

Changed barcode scanning library to [html5-qrcode](https://github.com/mebjas/html5-qrcode).

This scanner performs very well on phone browsers, especially on Android, which implement the [BarcodeDetector API](https://github.com/mebjas/html5-qrcode/blob/master/experimental.md). However on desktop browsers, performance is a little slower.

Note that while the current implementation as of this PR submission is functional, it seems that the performance of the scanner is severely impacted by the size of the containing div. The offending section will be flagged in the code for further review.

I have also created a [CodePen](https://codepen.io/maksimstojkovic/pen/vYravXz) demonstrating this behaviour and will submit an issue to the library repo to see if there are any simple fixes. It is very likely that there is a CSS-related fix that could be implemented which allows the full webcam resolution to be used while shrinking the size of the rendered video.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4007"></a>


matmair:
@maksimstojkovic any chance to look at performance again for this? In the current state it is barely usable on iPhone with safari. The old thing has no problems in this config.

maksimstojkovic:
@matmair Will boot up an iPhone and have a look on Safari. Have you tried any other mobile browsers?

Also the issue mentioned above has been raised here: mebjas/html5-qrcode#631

maksimstojkovic:
@matmair Will boot up an iPhone and have a look on Safari. Have you tried any other mobile browsers?

maksimstojkovic:
@matmair I've added some CSS that fixes how the video feed is displayed, so it is now responsive across displays.
I have also tried the scanner in Safari and it seems to work. I'm just in the middle of an iOS update to make sure it continues to work on the latest version, and will give an update once I do.

Also, what would be the best way to run JS style checks locally so I can fix them up before committing?

matmair:
@maksimstojkovic seems to work fine now.

SchrodingersGat:
@martonmiklos I have checked with QR and barcode and both seem to work great! This is a fantastic improvement.

We should remove the old javascript files: `/static/script/qr-scanner-*`

maksimstojkovic:
@SchrodingersGat Done

maksimstojkovic:
Just a heads up, the maintainer for [html5-qrcode](https://github.com/mebjas/html5-qrcode) took interest in the solution I proposed so the library might be updated in the future such that CSS isn't necessary to improve performance. Hopefully it is just a matter of setting an attribute on the `<div id="barcode_scan_video"></div>` element.

Probably worth keeping an eye on the following issue for any updates in the future:

[https://github.com/mebjas/html5-qrcode/issues/631](https://github.com/mebjas/html5-qrcode/issues/631)

SchrodingersGat:
Thanks for the contribution

PR #4035, opened 2022-12-09
gunicorn: revert support for gevent

SchrodingersGat:
- For some reason this causes production docker image to fail to build
- but only in GitHub CI ??

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4035"></a>


SchrodingersGat:
The production image kept failing in ci, I'm more keen to get the next release out and then we can potentially revisit

PR #4067, opened 2022-12-15
Add Date input for table filter

Bbillyben:
Add a 'date' field to be processed in table filter.

This will report a naive date to the server. This needs to be 'timezoned' before being processed into queryset.

table filter definition would be like : 
```
date_field:{
                title: '{% trans "Date Field" %}',
                type: 'date'
            },
```

Any though on this? Any evident drawbacks I missed ?

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4067"></a>


SchrodingersGat:
This looks nice! Do you have any particular fields in mind to query with this?

Bbillyben:
Not in the core, even if I'm pretty we could find some utility,   but I'm implenting it in my plugins.
The goal is to filter on specific timeframe or date, to export directly relevant data regarding our quality processes (deeply stringent).

Bbillyben:
I've added date filtering on expiry date in stock item table in stocklocation tab.
=> I haven't found a way to get the user timezone properly

pre-commit issue reference the first 'import datetime' on line 3 as unused, I had to remove it to passe QC.

wolflu05:
> => I haven't found a way to get the user timezone properly

Are you searching for a way to get the client's timezone in javascript? This should work like described here: https://stackoverflow.com/a/1091399

matmair:
@Bbillyben I would suggest doing all client timezone things on the client - that is really messy/error-prone serverside.

Bbillyben:
Ok, got it, 

I moved to store timezoned timestamp from js and process the timestamp in py.

Is that better?

Bbillyben:
As far as I can see it's working without raising any warnings. (so i turn check it ready for review)

I haven't go through build, purchase order and sales there should be something interesting to implement with the different dates (dates, target date, completion dates, ....)

SchrodingersGat:
@Bbillyben can you explain the reasoning behind using integer dates rather than say an ISO datestring i.e. YYYY-MM-DD?

For example look at how we do filtering for the calendar display views.

Also, as the API has been updated, please bump the version in `InvenTree/api_version.py` :)

InvenTree/stock/api.py
```diff
@@ -830,6 +830,16 @@ def filter_queryset(self, queryset):
                     queryset = queryset.filter(StockItem.EXPIRED_FILTER)
                 else:
                     queryset = queryset.exclude(StockItem.EXPIRED_FILTER)
+            # Filter by 'expiry date'
+            expired_date_lte = params.get('expiry_date_lte', None)
+            if expired_date_lte is not None:
+                date_lte = datetime.fromtimestamp(int(expired_date_lte) / 1e3)
```
matmair, line 836: I would very much prefer passing UTC ISO dates here instead of timestamps. Requires less computation on the client side

InvenTree/stock/api.py
```diff
@@ -830,6 +830,16 @@ def filter_queryset(self, queryset):
...
+                queryset = queryset.filter(expiry_date__lte=date_lte)
+
+            expiry_date_gte = params.get('expiry_date_gte', None)
+            if expiry_date_gte is not None:
+                date_gte = datetime.fromtimestamp(int(expiry_date_gte) / 1e3)
```
matmair, line 841: Same as above

matmair (changes requested):

Bbillyben:
Thanks for your feed back. 

That was a mistake attempting to store a timezoned timestamp. I was assuming that integer are always better than string !

I've seen https://github.com/inventree/InvenTree/blob/8a095f00cf07dc8af5603643a658aed0a8cf71e1/InvenTree/build/models.py#L114
from build filter_queryset
https://github.com/inventree/InvenTree/blob/8a095f00cf07dc8af5603643a658aed0a8cf71e1/InvenTree/build/api.py#L137

Is that what you have pointed out @SchrodingersGat  ?

So as far as I can understand, dates are used (/stored?) in simple non extended ISO 8601 (aka YYYY-MM-DD), without any information on user timezone, inferring de facto a UTC date ?

I will the get back to my original commit (without the in-python timezoning) and add an exception's catching.

matmair (approved):

SchrodingersGat (approved):

SchrodingersGat:
Thanks for the contribution @Bbillyben

PR #4077, opened 2022-12-17
Load plugins for collectstatic

wolflu05:
Hello,

this PR changes the loading mechanism of the plugins so that they are loaded if they are activated and `collectstatic` is executed. Otherwise plugins are not able to expose static files. This was already added for database migrations in #3762 .

Ref: #4020 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4077"></a>


SchrodingersGat:
@wolflu05 thanks for the fix :)

Issue #4087, opened 2022-12-20
[BUG] Can't make storage location structural if sub storage location contains parts


bloemp:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### Describe the bug*

After adding a part to a sub stock location, the parent stock location no longer can be switched to structural

### Steps to Reproduce

1. Create a storage location "test1" and don't make it structural
2. Create a storage location beneath "test1" called "test2"
3. Add a part to storage location "test2"
4. Now try to make "test1" structural... if fails with message " You cannot - make this stock location structural because some stock items are already located into it!"

### Expected behavior

Storage location "test1" should be able to be changed to structural

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

InvenTree Version 	0.10.0 dev
Commit Hash	fb3440947

bloemp:
After reporting the bug i've started searching the source code... I'm proficient in many programming languages, but python isn't one of it (yet)... but it seems that changing the following line of code maybe the solution

On line 155 of file InvenTree/InvenTree/stock/models.py change line
        if self.pk and self.structural and self.item_count > 0:
to
        if self.pk and self.structural and self.stock_item_count(False) > 0:

seems to fix this problem.
As mentioned i'm a python and inventree noob... so i'm very interested if this would be the correct solution, or why it should be solved in another way.

SchrodingersGat:
@bloemp looks like you have successfully worked out where the error is coming from. Would you be willing to submit a PR against this?

bloemp:
@SchrodingersGat as requested, I've made pull request #4089 containing the fix

PR #4089, opened 2022-12-21
Fix stock location structural check

bloemp:
Exclude sub stock location items from preventing that a stock location can be switched to structural.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4089"></a>

Closes https://github.com/inventree/InvenTree/issues/4087

### DIFF
--- a/InvenTree/part/models.py
+++ b/InvenTree/part/models.py
@@ -148,7 +148,7 @@ class PartCategory(MetadataMixin, InvenTreeTree):
 
         - Ensure that the structural parameter cannot get set if products already assigned to the category
         """
-        if self.pk and self.structural and self.item_count > 0:
+        if self.pk and self.structural and self.partcount(False, False) > 0:
             raise ValidationError(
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@bloemp thanks! Can you also make the same fix to the part category model:

https://github.com/inventree/InvenTree/blob/d9e99e38d2313ec0673986e2765915af37c1c36e/InvenTree/part/models.py#L151

bloemp:
@SchrodingersGat modified the pull request to include a fix for the part categorie.
`        if self.pk and self.structural and self.partcount(False, False) > 0:
`
I've added both parameters to the partcount call, while not strictly needed... it seemed future proof (if anyone ever changes the default, the check still functions). But as a newbie to python, is this the best way to go?

SchrodingersGat:
@bloemp looks good :)

SchrodingersGat:
Thanks for your contribution @bloemp

PR #4101, opened 2022-12-22
Various fixes to the "Download image from URL" functionality

bloemp:
This commit adds an user-agent string to the download request wich prevents some sites from returning a 403 (forbidden) status.
Clicking the download button on the thumbnail, also showed the thumbnail because the event wasn't stopped. The parts thumbnail didn't update after a succesfull download of the image, reloadImage was never called.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4101"></a>

### DIFF
--- a/InvenTree/InvenTree/helpers.py
+++ b/InvenTree/InvenTree/helpers.py
@@ -151,7 +151,6 @@ def download_image_from_url(remote_url, timeout=2.5):
             timeout=timeout,
             allow_redirects=True,
             stream=True,
-            headers={"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64)"},
         )
         # Throw an error if anything goes wrong
         response.raise_for_status()
--- a/InvenTree/InvenTree/helpers.py
+++ b/InvenTree/InvenTree/helpers.py
@@ -151,6 +151,7 @@ def download_image_from_url(remote_url, timeout=2.5):
             timeout=timeout,
             allow_redirects=True,
             stream=True,
+            headers={"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64)"},
         )
         # Throw an error if anything goes wrong
         response.raise_for_status()
--- a/InvenTree/InvenTree/helpers.py
+++ b/InvenTree/InvenTree/helpers.py
@@ -145,12 +145,20 @@ def download_image_from_url(remote_url, timeout=2.5):
     # Calculate maximum allowable image size (in bytes)
     max_size = int(InvenTreeSetting.get_setting('INVENTREE_DOWNLOAD_IMAGE_MAX_SIZE')) * 1024 * 1024
 
+    # Add user specified user-agent to request (if specified)
+    user_agent = str(InvenTreeSetting.get_setting('INVENTREE_DOWNLOAD_FROM_URL_USER_AGENT'))
+    if user_agent:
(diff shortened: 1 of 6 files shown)


InvenTree/InvenTree/helpers.py
```diff
@@ -151,6 +151,7 @@ def download_image_from_url(remote_url, timeout=2.5):
             timeout=timeout,
             allow_redirects=True,
             stream=True,
+            headers={"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64)"},
```
matmair, line 154: I am not sure it is in good faith to immitate a user agent. I would rather not do that by default.

matmair (changes requested):

InvenTree/InvenTree/helpers.py
bloemp: I can see your point, the change has been reverted.

wolflu05: What about adding this as an option in the settings?

bloemp: Added a setting to allow the default user-agent to be overridden.

InvenTree/common/models.py
```diff
@@ -875,6 +875,12 @@ def save(self, *args, **kwargs):
             ]
         },
 
+        'INVENTREE_DOWNLOAD_FROM_URL_USER_AGENT': {
+            'name': _('User-agent used to download from URL'),
+            'description': _('Allow to override the user-agent used to download mages and files from external URL (leave blank for the default)'),
```
matmair, line 880: Missing letter for image

matmair (changes requested):

InvenTree/common/models.py
bloemp: @matmair  It seems typing is hard this morning... but I've added the missing letter

InvenTree/InvenTree/helpers.py
```diff
@@ -145,12 +145,20 @@ def download_image_from_url(remote_url, timeout=2.5):
     # Calculate maximum allowable image size (in bytes)
     max_size = int(InvenTreeSetting.get_setting('INVENTREE_DOWNLOAD_IMAGE_MAX_SIZE')) * 1024 * 1024
 
+    # Add user specified user-agent to request (if specified)
+    user_agent = str(InvenTreeSetting.get_setting('INVENTREE_DOWNLOAD_FROM_URL_USER_AGENT'))
```
SchrodingersGat, line 149: I think you do not want to cast this to a string. If `get_setting` returns None, you'll actually get "None"

bloemp:
I've also found that the thumbnail delete button is not showing after an upload. Would it be okay for me to throw these changes in with this PR? Or should I make a separate one?

SchrodingersGat:
@bloemp seems related enough for it to be in the same PR

SchrodingersGat:
All good from my end, requires sign off from @matmair

matmair (approved):
looks good, thanks @bloemp !

bloemp:
Glad to be of some help while trying to create an iventory for my electronics hobby and learning python and django along the way.

Merry Christmas!

SchrodingersGat:
@bloemp thanks for your contributions! Everyone starts somewhere :) I hope we can see more contributions from you. Have a great holiday

Issue #4050, opened 2022-12-12
PurchaseOrder Import Wizard appears to renders wrong template


Nerade:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find similar issue

### Describe the bug*

When selecting a *.xlsx file for upload the next form step seems to render the form of a different wizard step. It is not possible to proceed to the next step.

![2022-12-12 13_50_30-InvenTree _ Purchase Order](https://user-images.githubusercontent.com/3959229/207050016-3680c020-b85d-4e9e-aaa4-3217f1f01c38.png)

### Steps to Reproduce

1) Create PurchaseOrder
2) Select Upload File
3) Select *.xlsx file
4) Press upload

Example file:
[PO-import-error.xlsx](https://github.com/inventree/InvenTree/files/10208112/PO-import-error.xlsx)

### Expected behavior

According to some screenshots the form wizard step "Matching fields" should look more "helpful"?

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

Running v0.8.4 (stable branch)

SchrodingersGat:
Certainly doesn't look right. Can you check if the same behaviour is exhibited on the [demo server](https://demo.inventree.org)

Nerade:
Tested this right now with the provided sample file. I can confirm the same erroenous behavior on demo server instance.

SchrodingersGat:
Alright, thanks for checking. We will try to include this as part of https://github.com/inventree/InvenTree/issues/3101

Nerade:
Do you have by any chance any further information on this issue? I have seen (however too late, sorry) the FR regarding the migration to a new importer and the expected timeframe. I would like to dig a little bit deeper to the underlying problem with the aim to understand whats the actual issue. I have already done some research on that but have not come to much of a conclusion on what to change/fix in code.

SchrodingersGat:
@Nerade are you looking for assistance in where to look so you can potentially fix this issue?

Nerade:
Yes, exactly. Sorry if I was not clear on that. :)

SchrodingersGat:
Starting point is the `PurchaseOrderUpload` class: 

https://github.com/inventree/InvenTree/blob/2082455c4e14e0e53e90aed38e66d1aa97bf51b7/InvenTree/order/views.py#L70

Then, look at the templates which are being rendered in each step.

Nerade:
[...]

PR #4111, opened 2022-12-26
Part import wizard fixes

bloemp:
This PR tries to fix the part importer, it was getting stuck on the second page without any hope of importing.
After reading up on PR #4050 I did almost the same thing and it started working (sort of).

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4111"></a>


bloemp:
@SchrodingersGat 
During the creation of this PR, I came across the problem that it is possible to create a Part without a Category. 
There are no checks (I would have thought that the DB would balk at this). But if you make a Part without Category, InvenTree stops showing parts all together... 

My guess is that a Part without a category shouldn't be possible and that the DB should have some NONNULL attribute added to the field.... but my knowledge of python, django and even the intended behaviour of InvenTree is lacking.

Issue #4073, opened 2022-12-17
[FR] load custom template in plugins settings page


wolflu05:
## Feature request description

Plugins should have the ability to define a custom html and js template which is loaded into their own settings page.

## Reference

[...] Currently I have my settings load as a custom panel on the StockIndex view. But is there an option for my plugin to add some custom html and js over here like categories and parts have? 
![image](https://user-images.githubusercontent.com/76838159/208187189-64396235-f727-46f7-9282-a7e00f89c3e8.png)

I know that there is a settings mixin which already adds some text fields to this page, but my settings are way more complex than just simple text fields or booleans.

_Originally posted by @wolflu05 in https://github.com/inventree/InvenTree/discussions/4071_

> There is no way currently to add completely custom information to the plugin settings page. It would make sense to add this as a feature, at the very least each plugin could (optionally) define a custom rendering method for its own plugin configuration tab.

_Originally posted by @SchrodingersGat in https://github.com/inventree/InvenTree/discussions/4071_

wolflu05:
What is the current status of this issue? Afaik there is a SettingsContentMixin currently which supports custom html. Is it planned to expand this to js templates as well or should that issue be closed as completed?

matmair:
I have no need for it and I am not sure how we would debug if the JS code works/is ever called so my interest is pretty low. Not against it in any form, just no need for it.

wolflu05:
We could actually also use script tags inside the returned html so there is no need for an extra js template. So from my side we can close this as already implemented.

matmair:
If you are ok with that workaround/solution I would suggest closing as not planned

wolflu05:
The scope of this issue was actually both. Closing as already implemented.

Issue #3901, opened 2022-11-05
Epic: React


matmair:
This is an epic for https://github.com/inventree/InvenTree/issues/2789

Techstack Devtools:
- Node 18
- Yarn
- Vite

Techstack UI:
- React
- [Mantine](https://mantine.dev) UI, Hooks, Core, Date
- [mantine datatable](https://github.com/icflorescu/mantine-datatable) for tables
- React Router
- React Query for Api fetching
- Zustand for local state management
- [html5-qrcode](https://www.npmjs.com/package/html5-qrcode) for scanning
- [React Grid Layout](https://github.com/react-grid-layout/react-grid-layout) for Dashboard
- [LingUI JS](https://lingui.js.org/index.html) for translations

Tasks:
- [x] Create a base file structure
- [x] Crate basic UI
- [x] Create Dashboard UI
- [ ] Adapt server API endpoints to include all needed info
       #3978 
       #3979
- [x] Create Renderer for Form endpoints
- [ ] Create Settings UI
- [x] Create Part UI
- [ ] Create Stock UI
- [ ] Create Build UI
- [ ] Create Order UI
- [ ] Create Company UI

Optional:
- [ ] Create a tour with https://github.com/gilbarbara/react-joyride

wolflu05:
Im definitely interested in this. I'm already experimenting with preact on my plugin as vanilla js is bad for reactive UIs.

wolflu05:
Is there also a plan for what tooling inventree will use (e.g. nextjs, vite, ...)? If you would go with create-react-app, I would definitely advise you to not do that, [more infos](https://youtu.be/7m14f0ZzMyY). My first go would be nextjs or bundling with vite, which uses esbuild which is extremely fast.

If inventree won't use nextjs which has routing already implemented, I would consider using a much smaller package called [react location](https://react-location.tanstack.com).

matmair:
@wolflu05 I have access to a sizeable private codebase that runs on the proposed stack; they used plain CRA, WebPack and nothing SSR so my plan is/was to use that as I know that works. Do you have production experience with the proposed packages?

The jsx-world seems to be very fast moving but I would prefer to target mature, battle-proven packages. I do not really fancy switching core parts out twice a year. We are essentially 2 devs right now.

wolflu05:
@matmair Currently I'm on the same point of making a decision what tooling I should use for my new app. I was always using cra in many of my latest projects. 

### Why not to use CRA
[...]

PR #4114, opened 2022-12-28
[FR] Slack notifications & notification method updates

### LINKED ISSUE #3843: [FR] Slack notification

### Problem statement

I'd like to have slack as an alternative notification delivery method to email, with the ability to specify which notifications trigger a message, and which channels they should send the message to. Ideally this could all be configured through the settings page.

### Suggested solution

It looks like this would involve adding a new plugin to handle the message send logic. If I understand correctly it seems that I can use the `CoreNotificationsPlugin` as a template. Do I also need to use the `EventMixin`? 

Sending messages to slack channels is relatively straightforward with [webhooks](https://api.slack.com/messaging/webhooks) and the Slack API.

I'm relatively new to this so would appreciate any suggestions/advice on how best to proceed.

Open to other suggestions.

### Examples of other systems

Prometheus/Alertmanager has nice integration with slack webhooks: https://prometheus.io/docs/alerting/latest/configuration/#slack_config

- [X] I want to develop this.

matmair:
This PR:
- [x] Adds a slack sending method
- [x] Adds a mixin for adding settings conent
- [x] Adds a central panel for global notification settings
~- [ ] Extends notification methods to include registering~

Fixes #3843

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4114"></a>

### DIFF
--- a/InvenTree/common/notifications.py
+++ b/InvenTree/common/notifications.py
@@ -192,6 +192,7 @@ class MethodStorageClass:
         for item in current_method:
             plugin = item.get_plugin(item)
             ref = f'{plugin.package_path}_{item.METHOD_NAME}' if plugin else item.METHOD_NAME
+            item.plugin = plugin
             filtered_list[ref] = item
 
         storage.liste = list(filtered_list.values())
--- a/InvenTree/common/notifications.py
+++ b/InvenTree/common/notifications.py
@@ -192,7 +192,7 @@ class MethodStorageClass:
         for item in current_method:
             plugin = item.get_plugin(item)
             ref = f'{plugin.package_path}_{item.METHOD_NAME}' if plugin else item.METHOD_NAME
-            item.plugin = plugin
+            item.plugin = plugin() if plugin else None
             filtered_list[ref] = item
 
         storage.liste = list(filtered_list.values())
--- a/InvenTree/plugin/base/integration/mixins.py
+++ b/InvenTree/plugin/base/integration/mixins.py
@@ -742,3 +742,24 @@ class PanelMixin:
             panels.append(panel)
 
         return panels
+
+
+class SettingsContentMixin:
+    """Mixin which allows integration of custom HTML content into a plugins settings page.
+
+    The 'get_settings_content' method must return the HTML content to appear in the section
+    """
+
+    class MixinMeta:
+        """Meta for mixin."""
+
+        MIXIN_NAME = 'SettingsContent'
+
+    def __init__(self):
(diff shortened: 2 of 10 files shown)


miggland:
@matmair: I just stumbled on this package again. Have used it earlier (CLI only) for Gotify. However, it may give the possibility of having general notifications, without explicitly adding every service independently..
https://github.com/caronc/apprise

matmair:
@miggland looks cool, I created a new FR for it. While it is easy to send simple notifications I prefer to use native integrations as they allow for the use of platform-specific features (like Slack Block Kit for example).

SchrodingersGat:
@matmair nice work this looks really good.

Do you think it also closes out https://github.com/inventree/InvenTree/issues/4073 ?

matmair:
@SchrodingersGat I would have loved a simple one-click add button for setup but that would require me publishing a central app for slack (and therefore everyone trusting me not screwing with that access).

Not sure if everything Lukas envisioned is covered, I think we should let him decide. @wolflu05

wolflu05:
@matmair Sorry, didn't get the context why you mentioned me. Or do you mean the new settings content mixin which gets added by this pr which is not included in the title? Form my side, I would preferre an interface more like the panels mixin which also adds the ability to render custom js templates directly and also render html templates via filename. I know this could be made already via some custom logic, but I think I'm not the only one which will try to add custom templates instead of html strings.

matmair:
@wolflu05 yeah you were mentioned for that / #4073 
@SchrodingersGat so this won't close that. I do not intend to extend the current PR as the whole custom rendering is not really working well with #3901 and would add another work site for the transition.

SchrodingersGat:
@matmair nice work. A small docs PR would be great if you have a chance

wolflu05:
Yeah, the whole Django rendering part with templates, plugins provide will be tricky in the react ui. Maybe it can be just embedded via iframes?

wolflu05:
@matmair from what limitations are you talking?

matmair:
@wolflu05 some APIs can not be called by js code in an iframe.

wolflu05:
@matmair And what about using [dangerouslysetinnerhtml](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) for legacy plugins, and current plugins can just provide an bundle file which has a default export which is included via runtime import() and rendered as component?

matmair:
@wolflu05 I think we should take this conversation into #3901 or the subrepo - it does not really have anything to do with this issue.

PR #4118, opened 2022-12-30
docker venv fix

SchrodingersGat:
- Run once with --upgrade-deps
- Required to ensure that setuptools is automatically upgraded

This change is an addition to recent changes required to get the docker CI build to run. 

Turns out a similar issue prevents docker dev build from running under macOS.

When creating a venv, an older version of setuptools is bundled in (and overrides the hard-coded newer version in system python).

Ref: https://stackoverflow.com/questions/19115391/difference-in-version-of-setuptools-with-virtualenv

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4118"></a>

### DIFF
--- a/docker/init.sh
+++ b/docker/init.sh
@@ -35,7 +35,7 @@ if [[ -n "$INVENTREE_PY_ENV" ]]; then
         # venv already exists
         echo "Using Python virtual environment: ${INVENTREE_PY_ENV}"
     else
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@matmair FYI this looks to be the reason that we were seeing an older version of setup tools for the failing docker builds

PR #4154, opened 2023-01-04
Catch potential error when refreshing part pricing

SchrodingersGat:
Sentry reported a recurring error on the demo instance, need to catch a DoesNotExist exception

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4154"></a>


matmair:
@SchrodingersGat do we know why this is raised on a part that has a pk? Maybe we need some check there.

Issue #4155, opened 2023-01-05
CI - docker workflow overwrites newer images 


wolflu05:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### Describe the bug*

If the docker workflow needs a bit longer than a workflow which was started after that workflow, the docker image on docker hub will be an old image because it gets overritten by the running job on an older commit.

![image](https://user-images.githubusercontent.com/76838159/210777176-feb29432-6928-4da0-8ec6-39b9383ea66e.png)

![image](https://user-images.githubusercontent.com/76838159/210778710-9e6b59ed-42db-4a8c-b49f-d2322db1799f.png)

### Steps to Reproduce

-

### Expected behavior

There are two possibilities to fix this issue:

1. Longer running older jobs should check before a push if there was already pushed something newer
2. Newer jobs should cancel older running docker jobs

I would prefer the second fix, because pushing also can take some time where something newer was already pushed after the check.

Can you please retrigger the docker pipeline for now so I can run my test ci script to see if it works?

### Deployment Method

- [ ] Docker
- [ ] Bare metal

### Version Information

-

matmair:
@SchrodingersGat maybe we should set [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency)?

wolflu05:
Thank you very much. This a pretty powerful option. Didn't know this exists. Thought you would need to manually check for running jobs and cancel them.

matmair:
@wolflu05 the syntax is quite wide, you can do a lot with it. GitHub Action grew to get a quite potent system, second only to Azure DevOps Pipelines in my opinion.

wolflu05:
My knowledge of the github actions syntax is quite a bit deprecated, as I wasn't doing something with it the last year, where it got a major update with powerful features.

wolflu05:
May you can just retrigger the newest job for now so that latest is the actual latest so I can test out my ci pipeline?

matmair:
> May you can just rerun the newest docker job for now so that latest is the actual latest so I can test out my ci pipeline?

I restarted the run. Will take probably around an hour - our build times are quite long.

PR #4160, opened 2023-01-05
Adds concurrency config for docker CI workflow

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/4155

Prevents docker builds from overwriting each other "out of sequence".

Running a new docker build (for a particular type of event) will cancel any other pending builds, but a "latest" won't interfere with "stable" or "tag" release.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4160"></a>

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -29,6 +29,9 @@ jobs:
 
   # Build the docker image
   build:
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.event_name }}
+      cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


matmair (approved):
looks good - should we maybe include the branch name too in the ref?

SchrodingersGat:
I think that the current implementation is sufficient

matmair:
Then good to merge I would say!

SchrodingersGat:
Looks like it is working as expected:

image

PR #4170, opened 2023-01-07
Initialize target date for new PO line and SO line

bloemp:
Set the initial target date for new production order line or sale order line to the target date of the order it 
belongs to (if  set)

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4170"></a>

### DIFF
--- a/InvenTree/order/templates/order/purchase_order_detail.html
+++ b/InvenTree/order/templates/order/purchase_order_detail.html
@@ -179,6 +179,9 @@ $('#new-po-line').click(function() {
         currency: '{{ order.supplier.currency }}',
         {% endif %}
         {% endif %}
+        {% if order.target_date %}
+        target_date: '{{ order.target_date|date:'Y-m-d' }}',
+        {% endif %}
         onSuccess: function() {
             $('#po-line-table').bootstrapTable('refresh');
         }
--- a/InvenTree/order/templates/order/purchase_order_detail.html
+++ b/InvenTree/order/templates/order/purchase_order_detail.html
@@ -180,7 +180,7 @@ $('#new-po-line').click(function() {
         {% endif %}
         {% endif %}
         {% if order.target_date %}
-        target_date: '{{ order.target_date|date:'Y-m-d' }}',
+        target_date: '{% render_date order.target_date %}',
         {% endif %}
         onSuccess: function() {
             $('#po-line-table').bootstrapTable('refresh');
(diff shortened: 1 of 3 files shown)


InvenTree/order/templates/order/purchase_order_detail.html
```diff
@@ -179,6 +179,9 @@ <h4>{% trans "Order Notes" %}</h4>
         currency: '{{ order.supplier.currency }}',
         {% endif %}
         {% endif %}
+        {% if order.target_date %}
+        target_date: '{{ order.target_date|date:'Y-m-d' }}',
```
SchrodingersGat, line 183: Use the "render_date" template tag here - it will automatically render dates according to the user's preference.

bloemp: Good to know, I've updated the code

matmair (approved):

SchrodingersGat:
@bloemp actually sorry to be difficult - I missed a subtlety here in that the dates are not being rendered, but passed to the forms interface. So they should not be passed through the "render date" function (I believe)

matmair (approved):

SchrodingersGat:
@bloemp I've come back to this after realizing that this new behaviour is not really desirable. The `target_date` field on the order line (either sales or purchase order) is used to specify if the expected delivery date *of that particular line item* differs from the expected delivery date of the parent order.

That is, if there is a target date associated with the order itself, then the individual line items should leave this field **blank** - to indicate that the order date is to be used. Thus, if the user changes the target date for the entire order, this new date is automatically used for the individual line items.

The `target_date` field on each line item is for when an individual line item has a *different* target date value.

So my suggestion would be that we revert this PR, and leave the `target_date` field blank by default.

Any thoughts on this?

bloemp:
@SchrodingersGat personally I don't see a mayor difference to leaving it blank or having the orderline date being the same as that of the parent order. My initial thoughts for the PR where based on filling out the orderline entry form as complete as possible without having to specify this date. It was (at least for me) unclear that this is an alternative date to that of the order.

If you think that leaving the 'target_date' of the order line blank to indicate the use of the orders 'target_date' is a better way to go, or at least the intended behavior. I'm okay with that. 

But I feel that in that case, we should document this behavior somewhere (at least in the code). Maybe we should even chenge the label text/help text of the orderline entry form, making it clear that the field is an "alternative target date"

SchrodingersGat:
I'm happy to improve the field description strings. Any suggestions?

bloemp:
I was afraid you would ask me that ;) but maybe something like:

"Alternative target shipping date for this line item, in case it differs from the orders target date"

PR #4169, opened 2023-01-07
SSO tweaks

SchrodingersGat:
Some relatively minor tweaks for SSO integration

- Improve basic account templates, more pleasing to the eye
- Catch and handle some potential errors
- Provide more information to the user / admin

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4169"></a>

PR #4171, opened 2023-01-07
[SSO] Separate settings for local and social signup

### LINKED ISSUE #4166: [SSO] Separate settings for local and social signup
The "self registration" setting controls self-signup of both local (username/password) and social (SSO) accounts.

Ideally, we would have separate control for allowing (or preventing) self-reg of local accounts and / or SSO accounts.

For example, prevent users from self-registration of local accounts, but allow self-registration of SSO accounts through an external app which already restricts which user accounts have access.

Currently, enabling "self registration" (required for SSO self reg) also lets local accounts be created without restriction.

matmair:
This PR separates the settings for local and SSO signup.

Fixes #4166

### DIFF
--- a/InvenTree/InvenTree/forms.py
+++ b/InvenTree/InvenTree/forms.py
@@ -207,10 +207,11 @@ class CustomSignupForm(SignupForm):
 
 class RegistratonMixin:
     """Mixin to check if registration should be enabled."""
+    REGISTRATION_SETTING = 'LOGIN_ENABLE_REG'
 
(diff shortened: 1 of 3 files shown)


matmair:
@SchrodingersGat the interface might need some adjustments once you are done with #4169 - please only merge after

matmair:
Ready for merge

SchrodingersGat:
@matmair nice work, have validated that it works as expected

PR #4178, opened 2023-01-08
[FR] Update to OpenAPI from CoreAPI

### LINKED ISSUE #3226: [FR] Update to OpenAPI from CoreAPI
DRF has deprecated CoreAPI - move to OpenAPI
https://www.django-rest-framework.org/coreapi/schemas/

matmair:
This PR:
- Adds an endpoint to get an OpenAPI schema of the API
- Switches to use modern ReDoc
- Adds a task to export the current schema

Fixes #3226

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4178"></a>


matmair:
@SchrodingersGat would you rather keep shipping self-contained API docs or hyperlink to the needed resources from a CDN?

SchrodingersGat:
Can we package them as part of docs.inventree.org? That would be my preference, keep all the documentation together.

wolflu05:
But how does this work with CORS and own instances (I guess @matmair is speaking about something like swagger docs). I like this idea, but think that this should also be distributed on each inventree instance using only self-contained resources and not some CDN.

matmair:
@SchrodingersGat we can too, I was more thinking of the self-packaged API docs we ship today.
We could make serving them optional.

matmair:
@wolflu05 this version already includes local assets

wolflu05:
If I get this correct, your action will keep committing for each push that contains a diff in the api schema. And if the changes are removed from the pr the comment is still there?

matmair:
@wolflu05 I removed that part - we want to release more often and figuring this out will probably take good time.
I have got to get all my API changes in this release so we can ship the React prototype with the next.

wolflu05:
@matmair Ok. Sounds like a good plan. You want to ship the react ui with v0.10.0 parallel to the current ui in experimental mode to get some feedback?

matmair:
No, I want to prove out the APIs first and enable testing with the beta frontend while I find a solution for plugin templates. Still working out problems with shipping the compiled data in an efficient way without node, yarn etc.

wolflu05:
> shipping the compiled data in an efficient way without node, yarn etc.

You mean for non-docker installs? For docker it would be easy to use multi stage builds for that.

matmair:
@wolflu05 mainly, we have to get it running for deb packages and bare metal

matmair:
@SchrodingersGat what is your opinion on this PR? Should this be in 0.10.0 or pushed back?

SchrodingersGat:
I think let's push it back to next release - I'd like to get `0.10.0` out sooner rather than later, as there are lot of important changes. This one is not as vital

matmair:
@SchrodingersGat this would be ready for a review now. This changes the API docs (2 versions are included. Redoc on /api-doc and Rapidoc on /api-doc2)

SchrodingersGat:
@matmair looks reasonable, although do we need two different version?

InvenTree/InvenTree/settings.py
```diff
@@ -367,6 +369,15 @@
     # Enable browsable API if in DEBUG mode
     REST_FRAMEWORK['DEFAULT_RENDERER_CLASSES'].append('rest_framework.renderers.BrowsableAPIRenderer')
 
+SPECTACULAR_SETTINGS = {
+    'TITLE': 'InvenTree API',
+    'DESCRIPTION': 'API for InvenTree - the intuitive open source inventroy management system',
```
martonmiklos, line 374: inventroy - inventory typo I think

matmair:
@SchrodingersGat redoc is pretty similar to the current docs, rapidoc is quite different.

redoc
grafik
rapidoc
grafik
grafik

SchrodingersGat:
I quite like the redoc one, can we just use that?

matmair:
Yes, I will adjust the PR

SchrodingersGat:
Excellent work as always, thanks for implementing this @matmair

matmair:
I will release rapidoc as a plugin https://github.com/matmair/inventree-rapidoc/

PR #4186, opened 2023-01-08
[FR] Add API endpoint to activate plugins 

### LINKED ISSUE #4182: [FR] :checkered_flag: Add API endpoint to activate plugins

### Problem statement

There is currently no endpoint to activate a plugin by a key

### Suggested solution

Add an endpoint `/api/plugin/activate` that is supplied with the key/slug of the plugin to enable them

You could run a patch on the detail endpoint but a separate endpoint would be nice

### Examples of other systems

_No response_

- [X] I want to develop this.

matmair:
This PR:
- integrate general plugin patterns into the plugin patterns 
- BREAKING moves all plugin management endpoints to `/api/plugins` instead of `/api/plugin`
- adds an endpoint to activate a plugin

Fixes #4182

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4186"></a>


SchrodingersGat:
Is the breaking change ("plugins" vs "plugin") necessary here? This will cause at the least the app to break.

Also please bump the API version in `api_version.py`

matmair:
@SchrodingersGat I think it has strong benefits as the current naming schema prevents some slugs from being used for plugin URLs and the number of 'reserved' slugs will only grow with time.
A clear separation between `/api/plugin` for plugin-provided URLs and `/api/plugins` for meta fields also makes sense to prevent accidentally overriding core functions.

SchrodingersGat:
@matmair please increment the API version so external services (e.g. app) can determine which endpoint to call :)

matmair:
@SchrodingersGat API version is bumped!

Issue #4167, opened 2023-01-07
[FR] Allow label printing plugins to handle printing in bulk


wolflu05:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

Currently it is not possible to bulk print multiple labels in a single job. This is useful if your labels are not printed with a label printer, instead a normal paper printer. There you don't want to waste much paper, so printing out each label onto a singe A4/Letter page is wasting paper.

### Suggested solution

The current solution calls the print function of the label printing plugin once for each label. https://github.com/inventree/InvenTree/blob/09e02755ca5e934509af31d3ebf5ecc0ecb702c7/InvenTree/label/api.py#L126-L140

Maybe there could be a second function which can be implemented by the label plugin for print_labels (function gets list of labels). It would be also great to implement a printing queue directly into inventree and add a quick access button to print the whole queue, but I am not sure how many people would benefit from this.

### Describe alternatives you've considered

The `print_label` function of a plugin could queue labels itself until a page is full or ... and then print out all labels, but if the page is not filled and you want to have your labels printed you would have to add a custom url to your plugin which can trigger an "early print". This API endpoint could be called by a custom navigation button.

### Examples of other systems

don't know any other systems

### Do you want to develop this?

- [ ] I want to develop this.

matmair:
This is probably a bit more complex than it sounds when implemented right but I like the idea.

In general, I think it would be good to add a layer of attraction wherever we interface with 'real' machines, printers etc. So that we can have some meta info for the devices, maybe connect multiple ones, scope them to locations etc.

As for the print queue: If we need/want this fast it could just be moved into the `print_label` function of a plugin and add navigation to get to a page where settings and a 'print' button are.

wolflu05:
> So that we can have some meta info for the devices, maybe connect multiple ones, scope them to locations etc.

This is actually a feature which I haven't thought of, but I like the solution of scoping printers to a location. And for meta info, you mean plugins can expose if wired/wireless printers are "alive" or not reachable and also some kind of indicator for remaining pages/labels, ...?

> As for the print queue: If we need/want this fast it could just be moved into the print_label function of a plugin and add navigation to get to a page where settings and a 'print' button are.

But doesn't that mean it would break existing plugins? 


I would even go a step further and say, that label printing plugins can also expose multiple printers. E.g. the one I own has two label printers in it. One for big spool and one for a tapes. So it would be nice if plugins can expose multiple printers?

[...]

PR #4189, opened 2023-01-09
Change default value for auto-backup setting

SchrodingersGat:
This PR changes the "auto backup" scheduled task to be off by default (for new installs or upgrades).

The idea here is to not silently introduce a significant new feature without user interaction. The docs will be updated to reflect this.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4189"></a>


SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/4167

Issue #4180, opened 2023-01-08
Inventree 0.9+ cannot be started on low-end servers due to infinite loop of restarting timeouting workers


MR-DOS:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### Describe the bug*

It is impossible to start Inventree 0.9+ on low-end servers, at least after update from 0.8.3. The problem is that by default Inventree spawns 4 workers. After startup of Inventree, these workers get a task to make a backup of database. This task fails because the workers time out (the server gets loaded to 100% on 2c/4t Atom N2800) and get restarted. Also, Inventree tries to create 3 consecutive backups with time period of 2 minutes which is guaranteed to fail on an Atom server with HDD (currently my DB is about 373 kB and media 343 MB).
This forces Inventree startup into infinite loop of backup/fail and web interface never starts.
When Gunicorn servers and background workers are limited both to 1, it is possible to start Inventree.
Also, with 0.8.x it was perfectly fine. And it's not due to low memory - the server has 4 GB which don't get full when the problem arises, Docker has no artificial limits set. It may also be linked to HDD speed.

### Steps to Reproduce

1. Install Inventree 0.9+ using docker-compose on a slow machine and with a large media store.
2. Start up Inventree in non-detached mode.
3. Observe the repeated restarts of Inventree workers in a loop while trying to make a backup.
4. Shut down Inventree and add this to your .env file
```
INVENTREE_GUNICORN_WORKERS=1
INVENTREE_BACKGROUND_WORKERS=1
```
and also possibly this to config.yaml
```
background:
  workers: 1
  timeout: 90
```
1. Start Inventree. Eventually, it starts and from now on works normally.

Example of behaviour before modification of workers.
![image](https://user-images.githubusercontent.com/29711626/211215225-66e9b1d3-1d62-4b30-bbe2-40269b2c7a3a.png)

### Expected behavior

Backups shouldn't be run at the very moment of starting a server. These should be postponed until Inventree reaches some uptime (at least a few minutes).

Maybe, if workers periodically fail in a loop, try reducing the amount of workers. If they fail because the HW gets overwhelmed by them, killing and respawning all of them doesn't solve any issue, it just halts Inventree.

Maybe Inventree shouldn't be spawning all workers at once when it starts. Some delay could be added before each of them is spawned.

Inventree shouldn't be trying to make several backups in a row (it did so 3x even after it successfully started).

Discussion thread:
[...]

PR #4190, opened 2023-01-09
Configuration options for dbbackup

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/4180

This PR adds some setup / configuration options for database backup.

Additionally, backup is no longer "forced" during the "invoke update" step. This allows the update step to be run without backup, in situations where the user might not want this, or where backup is misconfigured.

On some (lower end) systems the "backup" task can stall for a long period of time which can cause worker timeout.

Ref: https://github.com/inventree/InvenTree/discussions/4179

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4190"></a>

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -149,7 +149,7 @@ MEDIA_URL = '/media/'
 
 # Database backup options
 # Ref: https://django-dbbackup.readthedocs.io/en/master/configuration.html
-DBBACKUP_SEND_EMAIL = get_boolean_setting('INVENTREE_BACKUP_EMAIL', 'backup_email', False)
+DBBACKUP_SEND_EMAIL = False
 DBBACKUP_STORAGE = get_setting('INVENTREE_BACKUP_STORAGE', 'backup_storage', 'django.core.files.storage.FileSystemStorage')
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@matmair due to the change to the `invoke update` step, (which no longer calls "backup" by default) does the installer script need to be adjusted?

matmair (approved):
I would not recommend running migrations without a backup as that might destroy an instance permanently. You are the maintainer, if you think this is in the best intrest of the user/project I wont stand against it.

matmair:
> @matmair due to the change to the `invoke update` step, (which no longer calls "backup" by default) does the installer script need to be adjusted?

The installer just calls the invoke task so no changes required. Maybe a migration that adds the backup option for exsisting installs would be good. Or a warning that this default changed could be sent to all superusers.

MR-DOS:
Hi, the problem was not during "invoke update", but on regular startup of Inventree server. Updating went completely OK.
With these changes, I could theoretically start up the server with backups disabled with otherwise default settings. Then if I wanted to enable them, I would have to restart the server and run into the same deadlock issue again, have to decrease amount of background workers, start up the server, wait for it to get satisfied with creating backups, shut it down and restart again with full amount of bkg workers. Or am I mistaken?

PR #4208, opened 2023-01-13
Fix for returning stock from the customer to a specified location

### LINKED ISSUE #4205: [BUG] :checkered_flag: Parts not marked as 'available' after action "Return to stock"

### Describe the bug*

See discussion https://github.com/inventree/InvenTree/discussions/4200

Returned stock should be marked as available

### Steps to Reproduce

1. use demo server
2. Check stock of part 108, filter by available: "Blue chair": 14 pcs (https://demo.inventree.org/part/108/)
3. Open the stock item sold&shipped in SO0001: https://demo.inventree.org/stock/item/994/
4. Use "Return to stock" action on in top right menu.
5. Choose a location, eg "Room 101" (same as other 14 chairs), add note "Testing return to stock"
6. Check stock of part 108, filter by available. Expect 17 pcs -> Only 14 pcs available.

Before:
image

After:
image

All stock items after change, without filtering:
image

### Expected behavior

Returned stock should be counted/shown as part of available stock, without extra actions.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.10.0 dev
Django Version: 3.2.16
Commit Hash: e730b5c
Commit Date: 2023-01-10
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Active plugins: False

### Relevant log output

_No response_

bloemp:
When returning stock from the customer, the allocations of that stock are freed. If the new stock location is the same as the parent stock, merge the stock back into the parent.

This is an attempt to fix #4205 

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4208"></a>

### DIFF
--- a/InvenTree/stock/tests.py
+++ b/InvenTree/stock/tests.py
@@ -498,7 +498,7 @@ class StockTest(StockTestBase):
 
         it = StockItem.objects.get(pk=2)
 
-        # First establish total stock for thid part
+        # First establish total stock for this part
         allstock_before = StockItem.objects.filter(part=it.part).aggregate(Sum("quantity"))["quantity__sum"]
 
         n = it.quantity
@@ -532,7 +532,7 @@ class StockTest(StockTestBase):
         self.assertEqual(track.tracking_type, StockHistoryCode.RETURNED_FROM_CUSTOMER)
         self.assertIn('Stock removed from customer', track.notes)
 
-        # Establish total stock for part after remove from customer to check that we stille have the correct quantity in stock
+        # Establish total stock for the part after remove from customer to check that we still have the correct quantity in stock
         allstock_after = StockItem.objects.filter(part=it.part).aggregate(Sum("quantity"))["quantity__sum"]
         self.assertEqual(allstock_before, allstock_after)
 
--- a/InvenTree/stock/tests.py
+++ b/InvenTree/stock/tests.py
@@ -8,8 +8,10 @@ from django.test import override_settings
 
 from build.models import Build
 from common.models import InvenTreeSetting
+from company.models import Company
 from InvenTree.helpers import InvenTreeTestCase
 from InvenTree.status_codes import StockHistoryCode
+from order.models import SalesOrder
 from part.models import Part
 
 from .models import (StockItem, StockItemTestResult, StockItemTracking,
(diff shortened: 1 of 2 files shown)


miggland:
I started a dev server with your branch - in my quick tests it seems to work.

bloemp:
That is good to hear, but stock handling is a complex thing and I'm still very new to python/django and InvenTree... so be careful and hopefully this patch gets through the QA process of @matmair  and @SchrodingersGat

matmair (approved):
Looks great @bloemp !

We won't merge for a few days as that is the maintainers prerogative. So if you want you could add a test for the merging of stock as a cherry on top.

bloemp:
Let me see if we can get that cherry, haven't written any tests yet... so this is another nice challenge.

matmair:
We would especially look for:
Carrying over of the notes
Removal of the old associations
An error if there is no parent - with the atomic change rolling back correctly

bloemp:
@matmair I think i've created tests for both allocateToCustomer and return_from_customer. I'm not quite sure how to implement "An error if there is no parent - with the atomic change rolling back correctly".
Any pointers you could give me? The return_from_customer function already checks for the parent being None so I'm not completely sure this test will work

miggland:
@bloemp: Not sure if it adds much, but should the original reason for my bug report be tested? Ie, check that the returned stock becomes available again?

bloemp:
@miggland that should already be tested by the added test. We are testing 2 situations:

1) The returned stock is returned to the original locatation (aka parent location) then it will be merged with the parent. We check if the amount of the parent is increased and the previous allocated stock record is removed. 
Because of the increased amount, the stock will be available

2) The returned stock is de-allocated to another stock location. The test checks to see if the references to customer, salesorder and ownedby fields are cleared and the new location is set. We also test that a stock track item is added with the correct status.
Because this stock record is no longer owned or allocated to a customer/salesorder it will be available as new stock.

So I think we have your original issue covered, but if you can think of more ways to check I will try to add it to the test.

miggland:
Especially your first point pretty much covers what I had in mind. I was thinking along the lines of counting the total available part stock before and after return, and to make sure the numbers make sense. However, I can't say that I've checked the logic in detail here so maybe that's a bit overkill/covered elsewhere.

matmair (approved):
Looks good to me, full coverage and sensible tests. I am delighted at this point. Your recent contributions are great and noticed @bloemp !
Testing full stock numbers might give some extra assurance as @miggland suggested.

bloemp:
Adding a total stock check is definitely a good catch, that why I've added a total stock check that compares the stock at the start and at the end of the remove_from_customer test.
This will hopefully ensure that all stock is returned.

InvenTree/stock/tests.py
```diff
@@ -465,6 +467,75 @@ def test_add_stock(self):
...
+        # Assigned sales order should be None
+        self.assertFalse(ait.sales_order)
+
+        # Check that a tracking item was added
+        track = StockItemTracking.objects.filter(item=ait).latest('id')
+
+        self.assertEqual(track.tracking_type, StockHistoryCode.RETURNED_FROM_CUSTOMER)
+        self.assertIn('Stock removed from customer', track.notes)
+
+        # Establish total stock for part after remove from customer to check that we stille have the correct quantity in stock
```
matmair, line 535:
```suggestion
        # Establish total stock for the part after remove from customer to check that we still have the correct quantity in stock
```
typo :-)

InvenTree/stock/tests.py
```diff
@@ -465,6 +467,75 @@ def test_add_stock(self):
...
+
+        self.assertEqual(track.tracking_type, StockHistoryCode.SENT_TO_CUSTOMER)
+        self.assertIn('Allocated some stock', track.notes)
+
+    def test_return_from_customer(self):
+        """Test removing previous allocated stock from customer"""
+
+        it = StockItem.objects.get(pk=2)
+
+        # First establish total stock for thid part
```
matmair, line 501:
```suggestion
        # First establish the total stock for this part
```

matmair (changes requested):
small typo

InvenTree/stock/tests.py
bloemp: Fixed

matmair (approved):

matmair:

SchrodingersGat (approved):

SchrodingersGat:
@bloemp many thanks :)

exp:
This fix effectively corrupts existing databases by removing items the Sales Order. These sales orders appear as incoherent, showing a greater number of items shipped than are listed.


I believe in fact that this means there is now absolutely no record anywhere of the original product shipped to a customer. You can see individual items were shipped, but the only record is which customer, not which sales order.


I think this is really bad, I'm going to have to recover from a database backup to restore this tracking information, and for many production installs out there they may have completely lost this information now as the stock tracking deltas do not include the appropriate IDs.


Could someone please confirm my findings are accurate? I apologise for not contributing in the last year but we have indeed adopted Inventree on a large scale and I would like to be of use. Fixing this is now very high priority for us.

matmair:
The item is returned from the customer so it is merged back into stock. Not sure I see the problem.

exp:
Actually in the last day a couple of the problems have been fixed by the Return Orders merge. Stock tracking deltas now do include sales orders, but I can't see that this has been historically populated. That may pose a problem later on.

In addition, I believe the original Sales Order is still modified so there is no record a stock item was ever included (that I can find). I can't think of any reason a historical Sales Order should ever be modified. This is a real pain for auditing purposes and especially if you deal with many departments at a customer and now have no idea what you shipped to whom and when.

The way I would re-engineer this (and am happy to, I will also try and fill in a bunch of other issues ASAP) is to instead use the latest entry in the StockItemTracking to determine if something is assigned to an order, has been shipped to a customer, is in the process of RMA etc. I can't see that this has been done as part of the latest merge, but perhaps I missed it?

SchrodingersGat:
@exp I can see the issue here, I think I have a way to address this which also makes the salesorder process much simpler. Will get onto this ASAP, as returning via RMA does remove the item from the salesorder, which is a critical bug!

exp:
@SchrodingersGat Thank you. If there's anything I can do to assist please let me know. The Return Orders merge was incredibly timely as this problem was discovered in the last few days and then you immediately almost completely fixed it!

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/4541

PR #4214, opened 2023-01-15
[FR] Add last updated column for supplier parts

### LINKED ISSUE #3327: [FR] :checkered_flag: Add last updated column for supplier parts

matmair:
This PR:
- Moves the different updated model fields to a mixin
- Adds `updated` column for supplier parts
- Add this column to the tables where it is needed

Fixes #3327

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4214"></a>


matmair:
@SchrodingersGat updated the api version and text

PR #4219, opened 2023-01-16
Reload purchase order table after part order

bloemp:
When on the part page, going to the "Purchase order" sidebar tab. You can order the part by pressing the "Order Part" button. After a successful order, the form closes but the table on the page is never refreshed.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4219"></a>


matmair (approved):

SchrodingersGat:
@bloemp thanks for the contribution :)

PR #4240, opened 2023-01-21
Make sure that print label dialog has a submit button.

bloemp:
Steps to reproduce problem:
Go to a screen that has a modal without a submit button like "Show QR code" of the Sublocation overview. Close the QR code modal and after that go to the "Print label" option. It will no longer show a submit button.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4240"></a>


matmair (approved):
Thanks!

PR #4244, opened 2023-01-22
[BUG] Fix ownership

### LINKED ISSUE #4229: [BUG] :checkered_flag: Stock item ownership results in stock item being read-only

### Describe the bug*

This is a followup for #4228.

When a user is a member of more than one group, stock items owned by one of these groups are read-only despite the group having VACD rights on stock items.

### Steps to Reproduce

1. Create two groups with VACD rights on stock items.
2. Add a normal (non-admin) user account to both groups.
3. Have a stock item being owned by one of these groups.
4. Notice how the stock item is read-only when logged in as said user.

### UPDATE 2022-01-20 ###

After further investigation it seems stock item ownership is broken in general. Even when the stock item is owned by the current user it appears read-only.

### Expected behavior

The stock item is modifiable by the user.

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.10.0 dev
Django Version: 3.2.16
Commit Hash: 28ec6b9d1
Commit Date: 2022-12-25
Database: postgresql
Debug-Mode: False
Deployed using Docker: False
Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}]

matmair:
This PR:
- adds the function is_user_allowed to Owner - it checks if a user is allowed to access a resource that belongs to an owner
- patches get_owner to also work with lazy objects
- adds a ton of tests regarding stock ownership

Fixes #4229

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4244"></a>

### DIFF
--- a/InvenTree/stock/models.py
+++ b/InvenTree/stock/models.py
@@ -146,7 +146,7 @@ class StockLocation(InvenTreeBarcodeMixin, MetadataMixin, InvenTreeTree):
             # So, no ownership checks to perform!
             return True
 
-        return user in owner.get_related_owners(include_group=True)
+        return owner.is_user_allowed(user, include_group=True)
 
     def clean(self):
         """Custom clean action for the StockLocation model:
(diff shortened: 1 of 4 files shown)


SchrodingersGat:
Cheers @matmair

Issue #4199, opened 2023-01-11
[BUG] Calculation of total price is wrong with mixed currencies


miggland:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### Describe the bug*

When adding line items with different currencies, the total price calculated in the line items overview does not take into account different currencies per line.

This happens in Purchase and in Sales orders. 10 yen are only about 8 cents - 20 USD is way off the correct value.

![image](https://user-images.githubusercontent.com/1314853/211769348-c9c537dd-4ffb-4429-a619-0a13a6aa76b5.png)
![image](https://user-images.githubusercontent.com/1314853/211770368-3a091be9-27ea-4654-83ae-5114e360f0a9.png)

These were both created on the demo server. The total cost, calculated at the top of the order, is correct in both cases (10.08 USD)

### Steps to Reproduce

1. Create sales|purchase order
2. Add line items with different currencies
3. Check bottom of table summary - find wrong sum

### Expected behavior

One could argue that it makes no sense to use different currencies for each line item. I was trying to use this in a sales order to calculate a total cost of an idea, with parts purchased from different vendors, in different countries.

Regardless of which, it is currently possible to do this, and the sum is wrong. Something must be changed. Options:

1. Make the calculation take into account the currencies, as with the total cost.
2. Remove the sum from the table - better no info than wrong info, possibly on if different currencies available.
3. Make it impossible to use different currencies in line items in an order. This would require setting a currency for an order, and sticking to that.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.10.0 dev
Django Version: 3.2.16
Commit Hash: e730b5c
Commit Date: 2023-01-10
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Active plugins: False

matmair:
@miggland Thanks for the report.
Seems like we will have to disable totals for now as the frontend does not know the rate.

miggland:
I'm definitely not very good at how the frond-end works. I tried finding where that line is created, but no luck.

Can you point me to the right place, @matmair ?

matmair:
@miggland I am also not really good with the current thing - although I am currently rebuilding it.

bloemp:
@miggland you can disable the total price column footer of the purchase/sale order tables by commenting out the footer-formatter code:

[...]

PR #4245, opened 2023-01-22
Calculate the sub-totals of the po/so in the default currency

bloemp:
Instead of summing all prices in the table which could be in different currencies, load them through a callback and the API. See issue #4199

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4245"></a>

### DIFF
--- a/InvenTree/templates/js/translated/order.js
+++ b/InvenTree/templates/js/translated/order.js
@@ -2626,10 +2626,7 @@ function loadPurchaseOrderExtraLineTable(table, options={}) {
                         fields: fields,
                         data: data,
                         title: '{% trans "Duplicate Line" %}',
-                        onSuccess: function(response) {
-                            $(table).bootstrapTable('refresh');
-                            reloadTotal();
-                        }
+                        onSuccess: reloadTable
                     });
                 }
             });
@@ -2990,7 +2987,6 @@ function loadSalesOrderShipmentTable(table, options={}) {
                 method: 'DELETE',
                 onSuccess: function() {
                     $(table).bootstrapTable('refresh');
-                    reloadTotal();
                 }
             });
         });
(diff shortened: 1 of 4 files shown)


InvenTree/templates/js/translated/order.js
```diff
@@ -2371,17 +2384,7 @@ function loadPurchaseOrderLineItemTable(table, options={}) {
...
-                    }).reduce(function(sum, i) {
-                        return sum + i;
-                    }, 0);
-
-                    var currency = (data.slice(-1)[0] && data.slice(-1)[0].purchase_price_currency) || 'USD';
-
-                    return formatCurrency(total, {
-                        currency: currency
-                    });
+                    return '<div id="sub_total_items_price">';
```
matmair, line 2391: @bloemp  couldn't you just use the data field to look up the total price in the data field?

matmair (approved):

InvenTree/templates/js/translated/order.js
bloemp: At the moment the data field doesn't contain the sub_total fields. Therefore I'm trying to get these fields in the output of the non-api get request. But I haven't won that struggle with Django yet
Also the Style JS check is failing, but I see the same construct in other parts of the file which aren't flagged. Changing the quotes to single ones makes the code fail...

bloemp: The code is now changed significantly so this should no longer be an issue

matmair:
@SchrodingersGat do you have an idea how we could remove the additional API calls and set the total from within the table refresh?

SchrodingersGat:
@bloemp @matmair sorry I haven't had a chance to look at this one yet. I am getting through my backlog and will have some time for this one pretty soon :)

Issue #3785, opened 2022-10-13
[FR] Add interacive options to installer


matmair:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

The installer currently is non-interactive. It can be modified with environment variables but it would be cool to have an interactive guide to pick a database backend and maybe set up Redis ... there are options.

### Suggested solution

Create an interactive installer:
- [ ] 'no input' switch (default on)
- [ ] database selection (sqlite, postgres, MySQL)
- [ ] add cache (redis)
- [ ] (optional) create optional installer pingback

### Describe alternatives you've considered

The new installer in #3743 is a great improvement (IMO, I wrote it) but could be more interactive.

### Examples of other systems

The installer for [Coolify](https://docs.coollabs.io/coolify/installation) is great, something like it could be used as a blueprint.

### Do you want to develop this?

- [X] I want to develop this.

ScaredyCat:
I agree, particularly on the backend selection. At the moment sqllite is NOT recommended and yet it's the default with no option to select another. If you've used the script installer you can't "invoke update" to define the schema even if you've changed the engine.

matmair:
You can absolutely run invoke update

ScaredyCat:
Sorry, I meant "invoke migrate", to generate the db schema. 

Well, you can run the command. It will fail, but you can run it. 

The "builds and installs everything for you" script doesn't install mysqlclient 

I would proffer that 90% of those using it don't know enough to fix the issue which is why they're using a script to do it for them.  Quick fix is to install it regardless (same for pqsql lib too I suppose) but it'd be much better to just ask which backend the user wanted.

matmair:
@ScaredyCat You are welcome to submit a PR that works across all supported OS, database and shell variants dynamically and without input; here is the installer code: https://github.com/inventree/InvenTree/blob/master/contrib/installer/src/root_command.sh

matmair:
Added 2 more requrements. I imagine a system with main commands and then interactive requests or flags for automation
commands:
- install
- reinstall
- add_pkg_pip
- add_pkg_native
- switch

matmair:
thinking more about this: having interactive options in an apt install is very annoying for SysOp / SysAdm so no longer plan to work on this

PR #4274, opened 2023-01-28
Add cli to package installs

matmair:
This PR adds an optional call to the command line to run invoke commands easily.
You can now run management commands (invoke tasks) using `sudo inventree run cli` - so eg. `sudo inventree run cli migrate`.

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4274"></a>

### DIFF
--- a/.pkgr.yml
+++ b/.pkgr.yml
@@ -29,6 +29,7 @@ dependencies:
   - gettext
   - nginx
   - jq
+  - libffi7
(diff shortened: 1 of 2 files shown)


matmair:
Related to #3785

SchrodingersGat:
Nice, seems like a good addition. Worth adding some docs?

PR #4276, opened 2023-01-29
Force builtin plugins to be active

SchrodingersGat:
- Not setting the "active" flag can cause ambiguous errors
- e.g. some settings don't work if the plugin is inactive
- By design, builtin plugins are always "active" anyway
- This PR just forces the active field to True

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4276"></a>

### DIFF
--- a/InvenTree/plugin/test_api.py
+++ b/InvenTree/plugin/test_api.py
@@ -193,8 +193,3 @@ class PluginDetailAPITest(PluginMixin, InvenTreeAPITestCase):
         with self.assertRaises(NotFound) as exc:
             check_plugin(plugin_slug=None, plugin_pk='123')
         self.assertEqual(str(exc.exception.detail), "Plugin '123' not installed")
-
-        # Not active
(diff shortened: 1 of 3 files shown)


InvenTree/plugin/test_api.py
```diff
@@ -193,8 +193,3 @@ def test_check_plugin(self):
         with self.assertRaises(NotFound) as exc:
             check_plugin(plugin_slug=None, plugin_pk='123')
         self.assertEqual(str(exc.exception.detail), "Plugin '123' not installed")
-
-        # Not active
-        with self.assertRaises(NotFound) as exc:
```
matmair, line 198: @SchrodingersGat was this test not working or why was it removed? I think it would make sense to check for this.

SchrodingersGat: As this plugin is now always "active" (because it is a builtin plugin) the test was no longer working

Issue #4279, opened 2023-01-31
[BUG/ UI Improvement] New Part form: Input fields for supplier/manufacturer part can be filled although toggle is OFF


simonkuehling:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When adding a new part, the form area "Supplier Options" is visible and can be filled out although the toggle switch "Create initial supplier data for this part" is not active.
In case I fail to activate the toggle, the data provided in that section of the form is ignored and no supplier parts or manufacturer parts are created - which is hard to catch, especially when adding multiple parts in a row

### Steps to Reproduce

1. Go to "Parts"
2. open "New Part" modal
3. find the "Supplier Options" section at the bottom of the form

### Expected behaviour

in order to avoid that mistake I would propose to activate/deactivate the four input fields for supplier/manufacturer part based on the toggle switch - or, even better, actually show/hide the input fields based on the toggle.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.10.0 dev
Django Version: 3.2.16
Commit Hash: 26f71b3
Commit Date: 2023-01-26
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Active plugins: [{'name': 'IPNAutoGenerator', 'slug': 'ipnautogenerator', 'version': None}, {'name': 'Brother Labels', 'slug': 'brother', 'version': '0.6.0'}]

SchrodingersGat:
Related: https://github.com/inventree/InvenTree/issues/4266

SchrodingersGat:
@simonkuehling thanks for adding this report - I'm currently working on a fix for #4266 so I'll ensure this gets fixed also.

simonkuehling:
Your speed and dedication to this project are mind boggling :-) awesome!

Issue #4266, opened 2023-01-27
[BUG] Supplier SKU not validated against length in "New Part" form


simonkuehling:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When adding a supplier part directly in the ```[New Part]``` modal form, there is no length validation run against the ```Supplier SKU``` field. When submitting the form with a longer SKU (more than 100 characters), there is an internal error thrown in the notification box and no part is created.

### Steps to Reproduce

1. open the [New Part] form
2. add a part and also input a supplier and supplier part SKU
3. choose a SKU with >100 characters
4. submit form

### Expected behaviour

run a form input validation for length in the SKU field

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.10.0 dev
Django Version: 3.2.16
Commit Hash: 26f71b3
Commit Date: 2023-01-26
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Active plugins: [{'name': 'IPNAutoGenerator', 'slug': 'ipnautogenerator', 'version': None}, {'name': 'Brother Labels', 'slug': 'brother', 'version': '0.6.0'}]

### Relevant log output

```shell
Traceback (most recent call last):

File "/root/.local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
return self.cursor.execute(sql, params)

psycopg2.errors.StringDataRightTruncation: value too long for type character varying(100)



The above exception was the direct cause of the following exception:


Traceback (most recent call last):

File "/root/.local/lib/python3.9/site-packages/rest_framework/views.py", line 506, in dispatch
response = handler(request, *args, **kwargs)

File "/root/.local/lib/python3.9/site-packages/rest_framework/generics.py", line 242, in post
return self.create(request, *args, **kwargs)

File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
return func(*args, **kwds)

File "/home/inventree/InvenTree/part/api.py", line 1358, in create
SupplierPart.objects.create(

File "/root/.local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
return getattr(self.get_queryset(), name)(*args, **kwargs)

File "/root/.local/lib/python3.9/site-packages/django/db/models/query.py", line 453, in create
obj.save(force_insert=True, using=self.db)

[...]

PR #4281, opened 2023-01-31
Improvements for part creation API endpoint

SchrodingersGat:
This started as a simple PR to fix https://github.com/inventree/InvenTree/issues/4266

However there were some issues which cropped up, so fixing those also. 

The main thrust of this PR is bringing the extra "part creation" fields into the DRF serializer, rather than hacking them in manually. This has multiple advantages:

- Exposes the fields to the API auto-doc
- Runs validators on the fields
- Exposes the fields to the self-describing metadata framework

------------

- Fixes https://github.com/inventree/InvenTree/issues/4266
- Fixes https://github.com/inventree/InvenTree/issues/4279

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4281"></a>

### DIFF
--- a/InvenTree/InvenTree/api_version.py
+++ b/InvenTree/InvenTree/api_version.py
@@ -2,17 +2,24 @@
 
 
 # InvenTree API version
-INVENTREE_API_VERSION = 90
+INVENTREE_API_VERSION = 91
 
 """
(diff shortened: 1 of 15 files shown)


matmair (approved):
Very nice and makes it much easier to reuse too.

PR #4293, opened 2023-02-02
Client side currency conversion

SchrodingersGat:
Exposes currency exchange rates via the API, allowing currency conversion to be performed efficiently by the client.

This allows tables, overviews, etc, to be calculated "on the fly" without expensive DB hits. Also is a bit more flexible for any future developments.

Fixes https://github.com/inventree/InvenTree/issues/4199
Closes https://github.com/inventree/InvenTree/pull/4245

Pricing calculation has been added for:

- [x] Purchase Order Line
- [x] Purchase Order Extra Line
- [x] Sales Order Line
- [x] Sales Order Extra Line

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4293"></a>

### DIFF
--- a/InvenTree/common/api.py
+++ b/InvenTree/common/api.py
@@ -104,7 +104,10 @@ class WebhookView(CsrfExemptMixin, APIView):
 
 
 class CurrencyExchangeView(APIView):
-    """API endpoint for displaying currency information"""
+    """API endpoint for displaying currency information
+
+    TODO: Add a POST hook to refresh / update the currency exchange data
+    """
 
     permission_classes = [
         permissions.IsAuthenticated,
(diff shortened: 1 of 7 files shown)


SchrodingersGat:
@matmair @miggland @bloemp what do you guys think of this approach? Rather than calculating on the back-end, all currency conversion is performed by the client. Should be a lot simpler (and quicker)!

InvenTree/common/api.py
```diff
@@ -102,6 +103,36 @@ def _get_webhook(self, endpoint, request, *args, **kwargs):
             raise NotFound()
 
 
+class CurrencyExchangeView(APIView):
+    """API endpoint for displaying currency information
+
+    TODO: Add a POST hook to refresh / update the currency exchange data
```
matmair, line 109: Question: Is this TODO for this PR or later? Should we maybe create a FR so we do not foget?

matmair (approved):

InvenTree/common/api.py
SchrodingersGat: Yes, I'll add a FR for this. Currently we do this with a web-based view (which should be removed)

Issue #4294, opened 2023-02-02
[FR] Move currency update view to API


SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/4293#pullrequestreview-1280517073

Currently the manual currency conversion update(s) are handled by a (legacy) web-view. This should be moved to the API (before the move to react)

PR #4300, opened 2023-02-03
Currency API Updates

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/4294

- Adds API endpoint for manually updating currency exchange rates
- Adds extra information to existing currency endpoint

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4300"></a>

### DIFF
--- a/InvenTree/common/api.py
+++ b/InvenTree/common/api.py
@@ -104,10 +104,7 @@ class WebhookView(CsrfExemptMixin, APIView):
 
 
 class CurrencyExchangeView(APIView):
-    """API endpoint for displaying currency information
-
-    TODO: Add a POST hook to refresh / update the currency exchange data
-    """
+    """API endpoint for displaying currency information"""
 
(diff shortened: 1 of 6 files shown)


InvenTree/common/api.py
```diff
@@ -133,6 +138,29 @@ def get(self, request, format=None):
...
+        permissions.IsAuthenticated,
+        permissions.IsAdminUser,
+    ]
+
+    def post(self, request, *args, **kwargs):
+        """Performing a POST request will update currency exchange rates"""
+
+        from InvenTree.tasks import update_exchange_rates
+
+        update_exchange_rates
```
matmair, line 157: Question @SchrodingersGat : should this call the function or just reference it?

SchrodingersGat: Uh, definitely call. I'll fix that.

PR #4303, opened 2023-02-03
Currency update fix

SchrodingersGat:
Fixes small bug introduced in https://github.com/inventree/InvenTree/pull/4300

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4303"></a>

### DIFF
--- a/InvenTree/common/api.py
+++ b/InvenTree/common/api.py
@@ -154,7 +154,7 @@ class CurrencyRefreshView(APIView):
 
         from InvenTree.tasks import update_exchange_rates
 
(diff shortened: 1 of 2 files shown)


matmair (approved):
Suggestion: Should we check some how that calling the endpoint actually results in currencies being updated? Seeing that we already had a problem with it once.

SchrodingersGat:
@matmair yep I was just trying to figure out what the test would look like :)

matmair (approved):
Lgtm

PR #4306, opened 2023-02-04
Switch to ManifestStorage

### LINKED ISSUE #3816: [FR] Remove need to refresh browser cache after update

### Describe the bug*

It seems that if Inventree is updated (in my case from 0.7.6 to 0.8.3), previously active user sessions still work but certain unexpected things may break.
I noticed that in the Parts-Categories view, you cannot see the amount of parts in categories. Also, "About Inventree" button does not work.
Happens on admin/superuser account as this is the only one I use on my computer, cannot tell for other users.
Clearing cookies resolves the problem.
No errors are thrown, at least not visible under the admin panel.

### Steps to Reproduce

1. Start up 0.7.6 Inventree from Docker.
2. Log in, create some categories and parts. You should be able to see amount of parts in a given category in tree view. Do not log out. My computer with this session was hibernated while I was upgrading the server.
3. Stop the server and upgrade image to 0.8.3, invoke update.
4. Start the server. (At this step, for me, a lot of errors from database about some not existing columns were thrown. It looked like it was mostly if not all about user session keys. As usual - I'm a user who does not understand what happens under the lid.)
5. Open the Parts-Categories view. Instead of numbers of the parts, all you can see, is "-". The error persists until cookies are cleared manually.
6. Try pressing the "About Inventree" button. Nothing happens. The error persists until cookies are cleared manually.

### Expected behavior

After server update, Inventree should detect that incompatible cookies are stored on user computer and they should be corrected/invalidated.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.8.3
Django Version: 3.2.15
Commit Hash: f9d2b14
Commit Date: 2022-09-23
Database: postgresql
Debug-Mode: False
Deployed using Docker: True

### Relevant log output

_No response_

matmair:
This PR changes static storage to ManifestStorage - providing cache busting and removing the need to force refresh the browser after updating.

Closes #3816

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4306"></a>


matmair:
Works on my machine now - let's see what the CI says.

SchrodingersGat:
Having issues with the splash screen image, it appears?

matmair:
Static file discovery in general is creating problems. We could also just version our rendered javascript files, that could also solve the problem.

SchrodingersGat:
Sounds like this needs some further thought - I'm going to bump this to the next release

SchrodingersGat:
> Static file discovery in general is creating problems. We could also just version our rendered javascript files, that could also solve the problem.

What would "versioning" look like here?

matmair:
@SchrodingersGat appending a hash of the file to the filename probably. It bugs me that ManifestStorage is not working, we would basically rebuild the base featureset of that.

SchrodingersGat:
@matmair with the move to a modern front end framework do we need this anymore?

matmair:
Not for the new components.

Issue #4317, opened 2023-02-07
Email URL not complete


method1020:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

The URL for linking to purchase orders (or other sections when the relevent email is sent) is not complete.

![Screenshot 2023-02-07 100756](https://user-images.githubusercontent.com/3278824/217283431-c2f368e8-2877-4532-bbbd-bfc911496a87.jpg)

### Steps to Reproduce

Late purchase order example:

1. Receive a late purchase order notification, try to click the `Click on the following link to view this order:` link. The host information is not part of the URL.

(you can see this in the mouse over of the link).

### Expected behaviour

Expected to link to the full URL.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.9.2
Django Version: 3.2.16
Commit Hash: 236ca36
Commit Date: 2023-01-04
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Active plugins: []

matmair:
@method1020 have you checked if `INVENTREE_BASE_URL` is set in the global settings? Will not work without it.

matmair:
@method1020 have you had a chance to look into this?

matmair:
Closing due to lack of response. Please feel free to reopen if the problem reappears.

method1020:
That fixes it. Thank you

PR #4325, opened 2023-02-09
Update documentation URLs

SchrodingersGat:


matmair:
@SchrodingersGat that wont fix #4317 ? I am still not sure what is causing the issue there.

SchrodingersGat:
No, this is not to address #4317 - I just noticed that the URLs had not been updated since we moved to inventree.org

matmair:
Ah great; still not sure what is causing the issue there. I thought I replicated it but it seems to work on current master.

matmair (approved):

PR #4336, opened 2023-02-11
Add barcode detection with onScan.js

### LINKED ISSUE #1087: Direct barcode entry
PartKeepr did this neat thing: you could assign a key combination that caused the following text to be piped directly into the barcode decoding module. If your wedge was fancy enough, you could assign that key combo as a prefix and a CR/LF as the suffix and simply scan a barcode into any page without focusing on a text entry field. Open webpage, scan barcode, StockItem appears.

Is this the sort of thing that Django inherently supports, such that this is a super-simple "hey, great idea!" kind of feature suggestion, or more of a "No, we would have to re-write Django ground-up, take a hike," kind of thing?

ldet:
Resolves #1087 

Uses: https://github.com/axenox/onscan.js

Basic functionallity already works and is hidden behind both the barcode server setting and a new user setting. Some things are still missing:
- [ ] Prefixes
- [ ] Settings in localstorage 
        As settings are heavily dependent on hardware attached to device, user settings are not the ideal place to store that information
- [ ] Currently detection is disabled when an `<input>` field has focus, evaluate if that is enough or maybe even too much
- [ ] Testing with different browsers, locales (keyboard layouts) and scanners

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4336"></a>


SchrodingersGat:
Hey @ldet thanks for the contribution :) Do you need any assistance with this?

ldet:
Thanks for your comment, I am hoping to continue work on this next weekend. 
I am not quite sure what the best approach with the setting for prefix and suffix is (keys that are sent before/after the actual barcode data), because the settings for those require key codes, which is not something the average user knows how to set up. My current idea is to provide a simple known barcode on screen that the user scans and then try to detect prefix and suffix keys from that.

SchrodingersGat:
@ldet a simple generated qr-code or barcode with some known sample text in there is a great idea!

SchrodingersGat:
@ldet how are you progressing with this one?

ldet:
@SchrodingersGat Hey, sorry for my slow response. I ran into some issues with the onscan.js prefix/suffix handling. I think its best to skip that for now and only support Scanners configured without Prefix and a pretty standard Enter/Tab suffix. That part is already working in the PR as is.
Something else I need some feedback about is the local settings (i.e. per Device settings in localstorage). Currently i have implemented them similar to the other settings on the server-side and inject the keys and default values via the template engine. This means that the settings pages can be rendered the same as the other ones, but editing requires special handling as there is no API endpoint associated with the local settings. I am not quite sure if I like my current solution, as there is really no need for the Server to know about the client side settings at all, but implementing all that custom rendering on the client side might not make sense. Especially if that plan to switch the UI to React still exists.
Thanks for your help and I hope to continue work on this soon.

InvenTree/InvenTree/static/script/inventree/inventree.js
```diff
@@ -251,6 +253,38 @@ function inventreeLoad(name, defaultValue) {
     }
 }
 
+/**
```
SchrodingersGat, line 256: Why re-invent these functions? These alrady exists here:

https://github.com/inventree/InvenTree/blob/4006ccfa0d1a8d076c3a6987399b36c0de8fe6a1/InvenTree/InvenTree/static/script/inventree/inventree.js#L228

InvenTree/InvenTree/urls.py
```diff
@@ -117,6 +117,7 @@
     re_path(r'^tables.js', DynamicJsView.as_view(template_name='js/translated/tables.js'), name='tables.js'),
     re_path(r'^table_filters.js', DynamicJsView.as_view(template_name='js/translated/table_filters.js'), name='table_filters.js'),
     re_path(r'^notification.js', DynamicJsView.as_view(template_name='js/translated/notification.js'), name='notification.js'),
+    re_path(r'^barcode_detection.js', DynamicJsView.as_view(template_name='js/translated/barcode_detection.js'), name='barcode_detection.js'),
```
SchrodingersGat, line 120: I would suggest that you bundle your new functionality into the existing `barcode.js` file!

InvenTree/templates/InvenTree/settings/device_barcode_scanner.html
```diff
@@ -0,0 +1,24 @@
...
+{% block heading %}
+{% trans "Barcode Scanner" %}
+{% endblock %}
+
+{% block content %}
+
+<div class='row'>
+    <table class='table table-striped table-condensed'>
+        <tbody>
+            {% include "InvenTree/settings/setting.html" with key="BARCODE_DETECTION_ENABLED" icon='fa-tag' local=True %}
```
SchrodingersGat, line 17: I think that adding "local" settings is going to complicate things a *lot* - very hard to debug too!

Unless the extra overhead can be justified, I'd rather see these be settings stored on the server

SchrodingersGat (changes requested):

SchrodingersGat:
@ldet thanks for getting back to me. I think that the "local settings" approach adds a significant amount of complexity and should be avoided as much as possible. Settings should either be added as a global or user setting...

Most of the PR seems to be associated with managing per-device settings, which I'd really rather not entertain

SchrodingersGat:
@ldet see https://github.com/inventree/InvenTree/issues/4608 for some more barcode related stuff - in particular another argument for browser-specific or device-specific settings. While browser side settings do make things more complex, perhaps it *is* the right idea here. Are you still working on this PR?

diodenschein:
In my use case particularly having input settings stored in the browser only sounds more work. (for me) You’d have to set it in every OS user’s browser settings and re-set it after cleaning browser data. 

Is it possible to detect the input device and have a rule if that has something in it that tells you what scanner is being used?

Alternatively: Most barcode scanners let you pre and ap-pend extra characters to the barcode data that could let you determine that it came from a particular type of scanner. 

I’ll have a look around how other software solves this and use our scanners a bit more. Maybe it becomes obvious.

PR #4340, opened 2023-02-13
[FR] Refactor plugin registry

### LINKED ISSUE #4184: [FR] Refactor plugin registry/mixins -> move activation sequences to mixin code
The current plugin mixin system does not allow for shipping global (de)activation sequences on (re)load. To fix this - and to clean up the registry of not really related code -  the sequencing should be moved to the mixins meta function itself.

matmair:
This PR moves the (de)activation sequences for plugins out of the registry and into the mixins themselves. It also factors out the impacted plugins into their own files.

The main advantage of this is that mixin-specific parts are isolated in that mixins code. This makes tests more stable and possibly enables dynamic loading of mixins ie. mixins as plugins. The larger goal is to make core plugin code slimmer and simpler.

Fixes #4184

<a href="https://gitpod.io/#https://github.com/inventree/InvenTree/pull/4340"></a>


matmair:
@SchrodingersGat this would be ready for review. No change in functionality, just a refactor.

SchrodingersGat:
Should we break *all* mixins out into their own file, if we are doing some of them?

matmair:
@SchrodingersGat I only broke out the complex ones but I can do all integration mixins if you think it is better. Trying to keep my PRs more manageable in size ;-)

SchrodingersGat:
> Trying to keep my PRs more manageable in size ;-)

Appreciated, it can be hard to keep up with larger PRs as the codebase evolves! I'll merge this in as-is for now, we can factor out others later as required. Thanks for the continued hard work :)

PR #4354, opened 2023-02-16
Support podman and selinux for dev setup

p-fruck:
This PR includes the changes required to get the InvenTree dev setup to work on my local machine which requires podman and SELinux. The following changes are included in this PR:

- SELinux labels are required in order to mount the volumes into the dev containers.
- The build config has been added to the dev-worker to prevent podman from trying to pull from docker hub

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
             - POSTGRES_DB=inventree
         volumes:
           # Map 'data' volume such that postgres database is stored externally
-          - ./data:/var/lib/postgresql/data
+          - ./data:/var/lib/postgresql/data:z
         restart: unless-stopped
 
     # InvenTree web server service
@@ -40,7 +40,7 @@ services:
         container_name: inventree-dev-server
         depends_on:
           - inventree-dev-db
-        build:
+        build: &build_config
             context: .
             target: dev
         # Cache the built image to be used by the inventree-dev-worker process
@@ -50,7 +50,7 @@ services:
             - 8000:8000
         volumes:
             # Mount local source directory to /home/inventree
-            - ./:/home/inventree
+            - ./:/home/inventree:z
         env_file:
             - docker.dev.env
         restart: unless-stopped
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@p-fruck what happens to the volumes in a "regular" docker setup? Are the labels ignored?

p-fruck:
I just added the labels from pure routine but just noticed that in fact there is no difference for SELinux labels. Do you configure them somewhere or is this an improvement to podman that I missed? Previously I had to define this label in order to be able to access my files inside the container

p-fruck:
Ok, my working directory was already configured with the correct SELinux labels. After setting up InvenTree in a new directory with default SELinux labels I got the following error:
image

This is because the container requires the `container_file_t` label which is automatically added using the `:z` flag. Note that `:z` is used instead of `:Z`  because the InvenTree directory is shared between server and worker container and the data directory is shared between worker, server and db.

p-fruck:
@SchrodingersGat is there any further information required? I'd be glad to upstream the compose file for further local development :)

matmair:
@p-fruck do I understand the above statements correctly in that it will break the docker workflow by default? Because I do not think we will merge something that breaks docker just to make podman work.

p-fruck:
No, this is not what those changes do. Note that the only podman related part is the `build` config that has been added to the `dev-worker`. The `:z` labels are added for SELinux users, which are currently not able to deploy InvenTree by simply running `docker-compose up` because of the permission issue shown above. Adding the `:z` options to volume mounts makes it possible for SELinux users to deploy InvenTree just like regular users do without having to manually configure the SELinux context of the given directories previously. This will not have any negative effect for non-SELinux users

SchrodingersGat:
https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

matmair:
Thanks for clearing that up for me; I mistook the screenshot. Should be fine @SchrodingersGat

matmair:
@p-fruck are you aware of https://github.com/moby/moby/issues/32579 like it occurred in vscode for example https://github.com/microsoft/vscode-docker/issues/3656

p-fruck:
@matmair No I wasn't, but isn't this only for docker stack? Also I think this is mainly due to the following concern:

> Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand.

While this might be an issue when People do use vscode-docker to open their home directory for whatever reason, this is basically impossible with the InvenTree compose file as all bind mounts are located inside the InvenTree folder and we do not try to modify any files outside this project directory which might have unexpected consequences to our users.

SchrodingersGat:
Thanks for the contribution @p-fruck :)

PR #4355, opened 2023-02-16
Configure CORS and allowed_hosts using env vars

p-fruck:
As an administrator I want to able to utilize InvenTree containers in a more stateless way and configure CORS URLs and allowed_hosts using the containers environment variables instead of actively maintaining the config.yaml file


matmair (approved):
Looks good @p-fruck ! Thanks for the PR.
@SchrodingersGat the docs look good too - should be ready once the CI runs through.

SchrodingersGat:

PR #4357, opened 2023-02-17
Client side QR Codes

SchrodingersGat:
Currently, QR codes are displayed via a custom view, which is requested and rendered by the server.

This is not required, a simpler approach is to use a client-side library to render the QR codes.

This PR implements that, and removes a bunch of server side views.

Additionally, support for client-side QR code rendering will be required for https://github.com/inventree/InvenTree/pull/4336


matmair:
@SchrodingersGat we have dependencies for generating the QRcode. Should we remove them or add a general purpose APIs route for rendering out an image of any qrcode?

SchrodingersGat:
Those dependencies are still needed for generating reports / labels / etc on the server side. But for client side rendering I think this is a cleaner approach.

PR #4362, opened 2023-02-17
add user setting to select a default label printer (#4349)

### LINKED ISSUE #4349: [FR] Setting to define a default label printer

### Problem statement

It would be very convenient to have an option to define a default label printer in the settings. I am using the `inventree-brother-plugin` heavily for all label making where it is a minor hassle to always have to manually change the selection from `Export as PDF` to `Brother Labels - Brother Label Printer` in the label printing modal form.

### Suggested solution

Some kind of configuration setting would be nice - could be a simple "make this printer the default selection" toggle somewhere.
It is unclear to me where to put this, as I am unsure where and how the brother plugin registers itself. Would the main `Label Printing` server configuration page be aware of the available printers? A drop-down to select the default printer seems logical to me.

Or is it a setting that the plugin should enforce somehow? (But that would potentially interfere with multiple plugins that each provide printers).

I am happy to give it a go myself with some pointers regarding these questions...

This could be a user-specific setting as well - but I think since this is more of an "infrastructure/hardware" setup and should be in the server settings...

### Examples of other systems

_No response_

- [X] I want to develop this.

simonkuehling:
This is my initial implementation for the "default label printer" user setting - it is working like intended so far already, but I'm open for comments...

Closes #4349

### DIFF
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -46,6 +46,7 @@ import InvenTree.ready
 import InvenTree.tasks
 import InvenTree.validators
 import order.validators
+from plugin import registry
 
 logger = logging.getLogger('inventree')
 
@@ -1628,6 +1629,15 @@ class InvenTreeSetting(BaseInvenTreeSetting):
             return False
 
 
+def label_printer_options():
+    """Build a list of available label printer options."""
+    printers = [('', _('No Printer (Export to PDF)'))]
+    label_printer_plugins = registry.with_mixin('labels')
+    if label_printer_plugins:
+        printers.extend([(p.slug, p.name + ' - ' + p.human_name) for p in label_printer_plugins])
+    return printers
+
+
 class InvenTreeUserSetting(BaseInvenTreeSetting):
     """An InvenTreeSetting object with a usercontext."""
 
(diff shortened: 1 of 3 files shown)


matmair (approved):

simonkuehling:
I have run a bit of "monkey testing" the edge cases like deactivating a plugin after it was configured as default printer and such, and everything is behaving correctly - so I would see this as finished from my side...

InvenTree/common/models.py
```diff
@@ -1628,6 +1629,15 @@ def requires_restart(self):
             return False
 
 
+def label_printer_options():
+    """Build a list of available label printer options."""
+    printers = [('', _('no printer (Export to PDF)'))]
```
SchrodingersGat, line 1634: Small thing, please capitalize "No Printer"

SchrodingersGat (changes requested):

SchrodingersGat:
@simonkuehling this works really well, just a small change to the text :)

InvenTree/common/models.py
simonkuehling: good catch - that would be terribly inconsistent :facepalm: - fixed it!

wolflu05:
Btw. you don't need to force push, as they anyway merge via squash so history will be clean in the end.

SchrodingersGat (approved):

matmair (approved):
Lgtm

SchrodingersGat:
Thanks again @simonkuehling

Issue #4359, opened 2023-02-17
[FR] Duplicate part modal: clear IPN field if PART_ALLOW_DUPLICATE_IPN = False


simonkuehling:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

When `PART_ALLOW_DUPLICATE_IPN = False` option is set for the server, the `Duplicate Part` modal form should open with an empty `IPN` field instead of copying the original over.

### Suggested solution

Check for `PART_ALLOW_DUPLICATE_IPN = False` when generating the form data.

This should only need an additional API call to `/api/settings/global/PART_ALLOW_DUPLICATE_IPN/` in `/InvenTree/templates/js/translated/part.js` to get the configuration value and issue a `fields.ipn = ''` if needed, right?

### Describe alternatives you've considered

-

### Do you want to develop this?

- [X] I want to develop this.

matmair:
@simonkuehling sounds fine. There is the array `global_settings` in settings.js with all settings values,

SchrodingersGat:
Yep all the settings are already loaded statically into a javascript file as @matmair says :)

simonkuehling:
sweet, a quick fix is in via PR https://github.com/inventree/InvenTree/pull/4365

PR #4365, opened 2023-02-18
Clear IPN field when duplicating part and PART_ALLOW_DUPLICATE_IPN is False

simonkuehling:
implementation for #4359

### DIFF
--- a/InvenTree/templates/js/translated/part.js
+++ b/InvenTree/templates/js/translated/part.js
@@ -414,6 +414,11 @@ function duplicatePart(pk, options={}) {
                 data.is_template = false;
             }
 
+            // Clear IPN field if PART_ALLOW_DUPLICATE_IPN is set to False
+            if (!global_settings['PART_ALLOW_DUPLICATE_IPN']) {
+                data.IPN = '';
+            }
+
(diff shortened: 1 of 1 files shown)


matmair (approved):

simonkuehling:
@matmair one of the QC checks (pre-commit) failed on this one - from the logs it looks like an error in the process run itself rather then my code...

matmair:
Not really sure what is going on there. Did you branch off at the current master or were there some commits after that? We had a similar issue a few days ago.

matmair:
@simonkuehling I have found the problem - it needs to be addressed upstream; I will submit a temporary fix to master shortly so we can keep working for now.

SchrodingersGat (approved):

SchrodingersGat:
@simonkuehling the PR itself looks good, but some bugs in CI are causing this to fail. Can you merge the latest master branch into this PR, and re-push? That should fix it and get these tests to run properly.

simonkuehling:
Looking good to go now :+1:

matmair:
Thanks for the contribution! You are crushing it lately!

PR #4368, opened 2023-02-18
Update dependencies

matmair:
There are some updates we should include soon.


matmair:
@SchrodingersGat this one is kind of urgent due to multiple high CVEs in Django

SchrodingersGat:
Thanks @matmair :)

Issue #4353, opened 2023-02-16
Error while dublicating part with category parameter templates and part templates


SergeoLacruz:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

I get a server error when I try to duplicate a part in case the category has parameter templates and the part has pemplates and both switches are on. 

### Steps to Reproduce

Add part category parameter template to a category (e.g. CAP) in the admin interface (e.g. Capacitance)
Select a part in the normal site from that category that already has the parameter Capacitance
Select part duplicate
Fill name IPN and description with proper values
Leave all switches in their default position specially: Copy Category Parameters ON and Copy Parameters ON
Duplicate fill fail with message 
An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.

### Expected behaviour

Duplicate should work.
Somehow it is logical that this cannot work. We need to set a priority here. I would expect that the Copy Category Parameters is ignored in that case, or warning please switch OFF. 

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

Development server 
DEBUG on

InvenTree Version 	[0.11.0 dev]
Commit Hash	216fb8d15
Commit Date	2023-02-16

### Relevant log output

```shell
Traceback (most recent call last):

File "/home/konrad/.local/lib/python3.10/site-packages/rest_framework/views.py", line 506, in dispatch
response = handler(request, *args, **kwargs)

File "/home/konrad/.local/lib/python3.10/site-packages/rest_framework/generics.py", line 242, in post
return self.create(request, *args, **kwargs)

File "/usr/lib/python3.10/contextlib.py", line 79, in inner
return func(*args, **kwds)

File "/home/konrad/src/InvenTree/part/api.py", line 1231, in create
part.save(**{'add_category_templates': copy_templates})

File "/home/konrad/src/InvenTree/part/models.py", line 483, in save
for template in category.get_parameter_templates():

File "/home/konrad/.local/lib/python3.10/site-packages/django/db/models/query.py", line 280, in __iter__
self._fetch_all()

File "/home/konrad/.local/lib/python3.10/site-packages/django/db/models/query.py", line 1324, in _fetch_all
self._result_cache = list(self._iterable_class(self))

File "/home/konrad/.local/lib/python3.10/site-packages/django/db/models/query.py", line 51, in __iter__
results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)

[...]

PR #4370, opened 2023-02-19
Part duplicate bug fix

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/4353

This PR fixes a bug which exists when duplicating a part with existing parameters.

Not only would the old code throw an error, it would not have worked anyway...

### DIFF
--- a/InvenTree/part/api.py
+++ b/InvenTree/part/api.py
@@ -2,7 +2,6 @@
 
 import functools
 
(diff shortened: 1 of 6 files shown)


matmair (approved):
Should there be a test that it works too? Seems like something we might break in the future inadvertently.

Issue #2960, opened 2022-05-09
[FR] "External" locations


SchrodingersGat:
This one requires some thought, and the concept is not fully fleshed out

It would be handy to be able to mark certain locations as "external", for example if stock is located at an external supplier, board assembler, etc

Being able to differentiate between "internal" and "external" stock might be useful.

matmair:
@SchrodingersGat do you imagine this as an informal field or to influence business logic?
Also would you like to expose these external locations on a webinterface to external parties?

SchrodingersGat:
I would imagine it influences the business logic, somehow.

gooutbefree:
At least a boolean to mark external locations and available over API would help. 
We are thinking of using sell orders to move components in an external stock (virtual stock to track assembly company's stock) and have the benefit of all shipping documents etc.
In a first step we can use a plugin to do that

PR #4367, opened 2023-02-18
Add 'Tag' management

matmair:
This PR adds tag, based on django-taggit, to several models:
- Part, SupplierPart, ManufacturerPart
- StockItem, StockLocation

Tags are also added to the (full) serializers for those models. They are not meant for frontend usage but to add taints/tags to parts for usage with plugins. That way objects can be marked and filtered for actions for eg.

Fixes #83

PR #4377, opened 2023-02-20
Add "external" field to StockLocation

SchrodingersGat:
- New model for field
- Database migrations
- Serializer / API updates
- Front end tables / filters

Closes https://github.com/inventree/InvenTree/issues/2960


wolflu05:
Do you plan to add some custom business logic in the future? If not, would that be a good usecase for #4367 ?

SchrodingersGat:
> Do you plan to add some custom business logic in the future? If not, would that be a good usecase for #4367 ?

Likely yes, but for now just the field

Issue #4312, opened 2023-02-07
[BUG] price calculation modal uses price per packing-unit, not single-piece price


simonkuehling:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When adding supplier parts with packing-units of more than 1 (for example length of 1000mm of steel tube), the single-unit price is correctly calculated in the `Pricing` tab. When using the price calculator modal (through the `$` icon in the top right), the prices for the full packing units are shown as price range - and calculations of multiples are way off through this, of course...

### Steps to Reproduce

Example part with pricing history:
![Bildschirmfoto vom 2023-02-07 11-08-45](https://user-images.githubusercontent.com/296454/217216687-3d8c59b9-2204-48a7-8616-21a20bf86f6c.png)

Supplier Parts for this Part:
![Bildschirmfoto vom 2023-02-07 11-09-29](https://user-images.githubusercontent.com/296454/217216882-2ae47e2f-335a-487e-8f82-6cf0c8e34fc1.png)

Pricing calculator for quantity of 1:
![Bildschirmfoto vom 2023-02-07 11-10-07](https://user-images.githubusercontent.com/296454/217217005-28340cdc-5d3b-4b99-a114-8822ac62f876.png)

Pricing calculator for multiple quantities:
![Bildschirmfoto vom 2023-02-07 11-10-18](https://user-images.githubusercontent.com/296454/217217089-639e873f-0ff5-4925-87a8-c77427dceb41.png)

### Expected behaviour

use price per single unit (in my example: per `mm`) in the calculator modal

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.10.0 dev
Django Version: 3.2.16
Commit Hash: ce3dabe
Commit Date: 2023-02-03
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}, {'name': 'IPNAutoGenerator', 'slug': 'ipnautogenerator', 'version': None}, {'name': 'Brother Labels', 'slug': 'brother', 'version': '0.6.0'}]

SchrodingersGat:
[...]

Issue #4042, opened 2022-12-11
Refactor LineItemPricing view (or remove)


SchrodingersGat:
After recent pricing refactor, need to re-address how "LineItemPricing" view is used.

Issue #4009, opened 2022-11-30
Cache "has_complete_bom_price"


SchrodingersGat:
When calculating (and caching) BOM pricing, we can determine if an assembly has a "complete BOM price" (pricing available for each line in the BOM).

Store this as a boolean value, can be retrieved and used when displaying BOM data.

- An assembly must have pricing data for each line in the BOM
- Each subassembly must also have complete pricing data
- So, "has complete pricing" is only true if there is a complete "top to bottom" set of pricing data

matmair:
@SchrodingersGat is this still valid? I think pricing is cached now?

SchrodingersGat:
Only some parts are cached, this calculation is still made on the fly, but should be stored in the db

Issue #4008, opened 2022-11-29
Refactor / remove "get_bom_price_range"


SchrodingersGat:
- Method is too expensive to calculate
- Migrate to use new "cached" pricing structure

PR #4382, opened 2023-02-21
Pricing form updates

SchrodingersGat:
This PR makes some significant changes to pricing, removing some old (*IMO outdated*) pricing calculations which are:

- Not cached
- Potentially recursive
- Blocking
- Don't add much useful information

The intent here is to move all pricing functionality to the new (cached) scheme, where pricing is available as a *range* which is calculated and updated by the background worker. These old pricing calculations are now a major sticking point for moving to a more API-driven / reactive user interface.

If there are particular features here that get removed, and users have been relying on them, we will need to work out how to make them fit with the new approach.

- Closes https://github.com/inventree/InvenTree/issues/4042
- Closes https://github.com/inventree/InvenTree/issues/4008
- Closes https://github.com/inventree/InvenTree/issues/4009
- Closes https://github.com/inventree/InvenTree/issues/4312

### Tasks

- [x] remove old sales order line item pricing view (super expensive)
- [x] Remove old part-pricing view (modal, slow)
- [x] Remove `get_supplier_price_range`  method
- [x] Remove `get_bom_price_range` method
- [x] Remove `get_price_range` method
- [x] Remove `get_internal_price_range` method
- [x] Remove `get_price_info` method
- [x] Add cache field for `has_complete_bom_price` - https://github.com/inventree/InvenTree/issues/4009
- [ ] Implement business logic for caching 'complete_bom_price' field
- [ ] Data migration for 'complete_bom_price' field
- [ ] Update BOM tables / etc to show complete_bom_price information
- [ ] Add new sales order line pricing (via API)


InvenTree/part/models.py
```diff
@@ -1628,31 +1627,6 @@ def clear_bom(self):
         """
         self.bom_items.all().delete()
 
-    def getRequiredParts(self, recursive=False, parts=None):
```
matmair, line 1631: Question: Why is this removed? I use it in multiple plugins, is there something wrong with it?

InvenTree/part/models.py
```diff
@@ -1692,147 +1666,6 @@ def schedule_pricing_update(self):
...
-
-        if min_price == max_price:
-            return min_price
-
-        min_price = normalize(min_price)
-        max_price = normalize(max_price)
-
-        return "{a} - {b}".format(a=min_price, b=max_price)
-
-    def get_supplier_price_range(self, quantity=1):
```
matmair, line 1719: Question: What replaces this? A API call?

InvenTree/part/models.py
```diff
@@ -1871,10 +1704,6 @@ def add_price_break(self, quantity, price):
             price=price
         )
 
-    def get_internal_price(self, quantity, moq=True, multiples=True, currency=None):
```
matmair, line 1874: Comment: I am using this in plugins

InvenTree/part/models.py
```diff
@@ -1885,23 +1714,6 @@ def internal_price_breaks(self):
         """Return the associated price breaks in the correct order."""
         return self.internalpricebreaks.order_by('quantity').all()
 
-    def get_purchase_price(self, quantity):
```
matmair, line 1888: Comment: I am using this in plugins

matmair (commented):
A pretty big PR; Removes a lot of internal functions I am currently using - how do you imagine them being replaced?
Also, the part pricing modal and line pricing functions are - in my view - integral parts of the pricing function. Having no way to get pricing ranges fast (even if it is blocking) for newly added BOMs removes the main usage for the prototyping guys.

SchrodingersGat:
@matmair  yep a big one. Still trying to work out how to re-introduce those functions in a more "efficient" manner.

It would be easier if we could ignore "quantity" pricing - this is very difficult to cache. And I would argue that because of all the external factors in calculating "quantity" pricing, that there's not much value in InvenTree doing this for the user? I would much rather present a somewhat "fuzzy" range of pricing (based on min/max data) and let the user infer what they will from that.

Thoughts? Probably we have to have a more general conversation about this before I progress this any further.

matmair:
@SchrodingersGat at that point InvenTree becomes less usable than a spreadsheet for pricing. If someone asks how much a part will cost to build at a quantity and they get back a range the tool is not useful, you can not make decisions on that. What sense do price breaks or any price calculations at that point?

SchrodingersGat:
While I don't use it for that purpose, I see your point. However when I do try to use the price-break calculations, it can take tens of seconds to calculate. I'd like (bare minimum) this to be reduced to a more user friendly time. How can we do this do you think?

matmair:
@SchrodingersGat I think with multiple parts there is not much we can do; maybe optimise the database calls? I have no problem moving it to some dedicated page with warning labels or two calculation strategies but outright removing it kneecaps the software IMO.

SchrodingersGat:
@matmair at the moment I cannot think of a good way of making this less expensive. I'll close this for now, maybe revisit at some point.

Issue #4385, opened 2023-02-21
[FR] Improve stock status badges for parts in forms and search results


simonkuehling:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

When creating Child Build Orders or Purchase Orders from Build Orders in the [Allocate Stock] tab, there is no sufficient indicator available which quantity of a part is available/beeing built/ordered/in demand right now.

- If stock is available, there is no info about `on order` oder `building`
- There is no info how much demand from build orders there are for this part (a number that is available in the part view through its `Allocated to Build Orders` metric)

This can quickly lead to bad planning in how many parts need to be built or ordered to serve all demands from current Build Orders.

### Suggested solution

I propose to introduce a `Demand: X` info and combine all relevant numbers into one badge. The color scheme for the badge should then be chosen based on overall "stock healthyness" (all factors considered). The content logic for the stock badge should follow these rules:

- `No stock` or `No stock available` or `Available: X/Y` --> always shown
- `Building: X` --> if beeing built right now
- `On Order: X` --> if beeing on order right now
- `Demand: X` --> if beeing in demand but not yet allocated from current Build Orders

Example
![Bildschirmfoto vom 2023-02-21 22-38-02](https://user-images.githubusercontent.com/296454/220465518-f2bfb957-0128-4e62-9c8d-e84a20342827.png)

The same badge could then also be used in the `Allocate Stock` tab in Build Orders to give a full overview/projection if stock needs to be increased by additional build orders or purchase orders, or if enough is incoming or in production already to be allocated later.

### Describe alternatives you've considered

.

### Do you want to develop this?

- [X] I want to develop this.

matmair:
Not sure if this would not crowd the UI too much - could you mock up some examples (should be simple with the page inspection tool) and add them to the first post? I understand the need and see the added utility.

I think this one requires input from more ppl from @inventree/triage before any action is done.

SchrodingersGat:
I agree with @matmair some example screenshots would be nice here

matmair:
@simonkuehling you will need to ping us / inventree/triage again to get the changes reliable into our mailboxes.

eeintech:
Looks fine to me :+1:

matmair:
@simonkuehling interfaces are hard - I think your solution looks good. I am curios how that will look with translations

SchrodingersGat:
[...]

PR #4399, opened 2023-02-23
fix stock health calculation to account for allocated stock (#4385)

simonkuehling:
The calculation of the `stock_health` variable did not account for allocated stock - which created too optimistic color codes in some cases.

### DIFF
--- a/InvenTree/templates/js/translated/part.js
+++ b/InvenTree/templates/js/translated/part.js
@@ -644,7 +644,7 @@ function partStockLabel(part, options={}) {
     }
 
     // Determine badge color based on overall stock health
-    var stock_health = part.in_stock + part.building + part.ordering - part.minimum_stock - demand;
+    var stock_health = part.in_stock + part.building + part.ordering - part.allocated_to_build_orders - part.minimum_stock - demand;
     var bg_class = '';
     if (stock_health < 0) {
         // Unsatisfied demand and/or below minimum stock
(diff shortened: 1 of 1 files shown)


simonkuehling:
@SchrodingersGat sure - the `stock_health` line is written like this to make the logic easy to follow:

```
POSITIVE side:
All stock that will be available without further action needed
(just waiting for shipment or production outputs)
    + part.in_stock        --> physically available
    + part.building        --> future stock that will come in from production
    + part.ordering        --> future stock that will come in from purchasing

NEGATIVE side:
All stock that is used or is planned to be used already
    - part.allocated_to_build_orders
    - part.minimum_stock
    - demand
```

I could certainly change the line to this I guess:
```python
// Determine badge color based on overall stock health
    var stock_health = part.in_stock + part.building + part.ordering - part.minimum_stock - required;
```

Regarding Sales Orders: you're right - I totally missed that!

simonkuehling:
I have added the required stock from sales orders now as well...

simonkuehling:
The logic for `stock_health` now is


```
POSITIVE side:
All stock that will be available without further action needed
(just waiting for shipment or production outputs)
    + part.in_stock        --> physically available
    + part.building        --> future stock that will come in from production
    + part.ordering        --> future stock that will come in from purchasing

NEGATIVE side:
All stock that is used or is planned to be used already
    - part.minimum_stock
    - required_build_order_quantity
    - required_sales_order_quantity
```

which is even cleaner to comprehend I think.

matmair (approved):

SchrodingersGat:
Thanks for the improvement!

Issue #4404, opened 2023-02-23
Backups stopped working and fail with a error


StormTrooper:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

I've had intree running for a few days. The backup schedule is enabled, set to backup every day. 

For the first few days the backup run successfully but has now stopped working. 

### Steps to Reproduce

In settings backup is enabled and set to run every 1 day. 

### Expected behaviour

Backups should be performed every day. 

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.11.0 dev
Django Version: 3.2.16
Commit Hash: e78161bd0
Commit Date: 18/02/2023
Database: mysql
Debug-Mode: True
Deployed using Docker: False
Active plugins: False

### Relevant log output

```shell
11:33:15 [Q] INFO Processed [quebec-crazy-four-zebra]
11:33:15 [Q] INFO Process-1:7 processing [yankee-item-don-kentucky]
11:33:16 [Q] INFO Enqueued 3574
11:33:16 [Q] INFO Processed [sad-london-uranus-pizza]
11:33:16 [Q] ERROR Failed [quiet-carpet-carbon-sodium] - can't compare offset-naive and offset-aware datetimes : Traceback (most recent call last):
  File "/home/inventree/env/lib/python3.10/site-packages/django_q/cluster.py", line 432, in worker
    res = f(*task["args"], **task["kwargs"])
  File "/home/inventree/src/InvenTree/InvenTree/tasks.py", line 484, in run_backup
    if last_success > threshold:
TypeError: can't compare offset-naive and offset-aware datetimes

11:33:16 [Q] INFO Process-1:9 processing [december-blue-summer-april]
11:33:16 [Q] INFO Enqueued 3575
11:33:16 [Q] INFO Process-1:8 processing [yellow-south-quebec-high]

I've checked and it's not a permissions issue. Have also tried restarting the server

root@inventree:/home/inventree/data/backup# ls -la
total 14336
drwxrwxr-x 2 inventree inventree     4096 Feb 20 11:33 .
drwxrwxr-x 5 inventree inventree     4096 Feb 18 11:09 ..
-rw-r--r-- 1 inventree inventree      501 Feb 18 11:17 default-inventree-2023-02-18-111759.dump.gz
[...]

PR #4406, opened 2023-02-24
Bugfix for auto-backup task

SchrodingersGat:
- Dont' mix datetime with timezone

Fixes https://github.com/inventree/InvenTree/issues/4404


matmair (approved):
@SchrodingersGat will we need to change all internal settings?

SchrodingersGat:
@matmair I think I got them all - ping me if you think there are any other internal ones

PR #4408, opened 2023-02-24
Filter build orders by responsible owner (#4383)

### LINKED ISSUE #4383: [FR] Filter Build Orders by Assigned User/Group

### Problem statement

I am using Groups to assign build orders to different manufacturing departments (like "CNC", "Composites", "3D Printing", "Electronics" etc). Currently I don't see a filter option to only show build orders that are assigned to a specific user or group other than yourself.

### Suggested solution

It would be very useful to have an additional filter in the build orders list to filter by `assigned to:` where users and groups can be selected.
Or maybe split into two filters: `assigned to user:` and `assigned to group:` - otherwise the options drop-down would be confusing and possibly crowded.

I think this would live here, right?
https://github.com/inventree/InvenTree/blob/75ca325892a09b6836250b863399649234d95d66/InvenTree/templates/js/translated/table_filters.js#L339-L359

is there a global variable already to access a list of users and groups? In case not, where should that be created best? Right there in `table_filters.js` from an api call, or maybe somewhere more central?

none that I can think of really - other than sorting the table by `Assigned to` column. But that is not so helpful with huge projects and many active build orders.

### Examples of other systems

_No response_

- [X] I want to develop this.

simonkuehling:
I have implemented the API+UI filters for filtering build orders by their responsible users/groups as discussed in #4383.

Filtering by a user also shows build orders assigned to groups he is part of - like the `assigned_to_me` filter does.
Along the way I added a user/group icon to the `Responsible` column, which is taken from the sales orders table.

Full table including the icons:

Bildschirmfoto vom 2023-02-24 12-40-39

Filtered by a responsible group:

Bildschirmfoto vom 2023-02-24 12-41-04

Filtered by a responsible user:

Bildschirmfoto vom 2023-02-24 12-41-30

Closes #4383

### DIFF
--- a/InvenTree/build/api.py
+++ b/InvenTree/build/api.py
@@ -2,6 +2,7 @@
 
 from django.urls import include, re_path
 from django.utils.translation import gettext_lazy as _
+from django.contrib.auth.models import Group, User
 
 from rest_framework import filters
 from rest_framework.exceptions import ValidationError
@@ -62,6 +63,23 @@ class BuildFilter(rest_filters.FilterSet):
             queryset = queryset.filter(responsible__in=owners)
         else:
             queryset = queryset.exclude(responsible__in=owners)
+
+        return queryset
+
+    assigned_to = rest_filters.CharFilter(label='assigned_to', method='filter_assigned_to')
+
+    def filter_assigned_to(self, queryset, name, value):
+        """Filter by orders which are assigned to the specified user or group."""
+        user_or_group = User.objects.get(username=value)
+        owners = []
+        owners.append(Owner.get_owner(value))
+        print(owners)
+
+        # if we query by a user name, also find all ownerships through group memberships
+        if type(owners[0]) is User:
+            owners.append(Owner.get_owners_matching_user(owners[0]))
+
+        queryset = queryset.filter(responsible__in=owners)
 
         return queryset
 
--- a/InvenTree/build/api.py
+++ b/InvenTree/build/api.py
@@ -66,18 +66,16 @@ class BuildFilter(rest_filters.FilterSet):
 
         return queryset
 
-    assigned_to = rest_filters.CharFilter(label='assigned_to', method='filter_assigned_to')
+    assigned_to = rest_filters.NumberFilter(label='assigned_to', method='filter_assigned_to')
 
     def filter_assigned_to(self, queryset, name, value):
-        """Filter by orders which are assigned to the specified user or group."""
-        user_or_group = User.objects.get(username=value)
+        """Filter by orders which are assigned to the specified owner."""
         owners = []
-        owners.append(Owner.get_owner(value))
-        print(owners)
+        owners.append(Owner.objects.get(pk=value))
 
-        # if we query by a user name, also find all ownerships through group memberships
-        if type(owners[0]) is User:
-            owners.append(Owner.get_owners_matching_user(owners[0]))
+        # if we query by a user, also find all ownerships through group memberships
+        if owners[0].label() == 'user':
+            owners = Owner.get_owners_matching_user(User.objects.get(pk=owners[0].owner_id))
 
         queryset = queryset.filter(responsible__in=owners)
 
--- a/InvenTree/build/api.py
+++ b/InvenTree/build/api.py
@@ -66,9 +66,9 @@ class BuildFilter(rest_filters.FilterSet):
 
         return queryset
 
-    assigned_to = rest_filters.NumberFilter(label='assigned_to', method='filter_assigned_to')
+    assigned_to = rest_filters.NumberFilter(label='responsible', method='filter_responsible')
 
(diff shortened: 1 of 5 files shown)


InvenTree/build/api.py
```diff
@@ -65,6 +66,21 @@ def filter_assigned_to_me(self, queryset, name, value):
 
         return queryset
 
+    assigned_to = rest_filters.NumberFilter(label='responsible', method='filter_responsible')
+
+    def filter_responsible(self, queryset, name, value):
+        """Filter by orders which are assigned to the specified owner."""
+        owners = []
+        owners.append(Owner.objects.get(pk=value))
```
SchrodingersGat, line 74: You should use `Owner.objects.filter()` - `get()` can raise an exception if the matching PK does not exist

```python
owners = list(Owner.objects.filter(pk=value))
if len(owners) > 0 and owners[0].label() == 'user':
    ...
```

InvenTree/templates/js/translated/build.js
```diff
@@ -2694,11 +2694,19 @@ function loadBuildTable(table, options) {
...
-                    if (value) {
-                        return row.responsible_detail.name;
-                    } else {
+                    if (!row.responsible_detail) {
                         return '-';
                     }
+
+                    var html = row.responsible_detail.name;
+
+                    if (row.responsible_detail.label == 'group') {
```
SchrodingersGat, line 2703: Nice addition :)

InvenTree/templates/js/translated/table_filters.js
```diff
@@ -355,6 +368,10 @@ function getAvailableTableFilters(tableKey) {
                 type: 'bool',
                 title: '{% trans "Assigned to me" %}',
             },
+            assigned_to: {
+                title: '{% trans "Responsible" %}',
+                options: ownersList,
```
SchrodingersGat, line 373: This is a nice approach, but the API request is going to be made every time the table is rendered (even when not necessary).

I'd say that we want to be able to dynamically fill out the "options" attribute for a table filter when the filter is added. This way that API request only fires when the user selects "owner" from the drop-down menu of available filters. Will need to extend the functionality of `table_filters.js` to achieve this.

SchrodingersGat:
Also please ensure that `InvenTree/api_version.py` gets updated accordingly :)

InvenTree/build/api.py
simonkuehling: thanks for the tip!

But using `User.objects.get()` should be fine, as it is already verified to exist by the first owners query, right?
Or is it generally favorable to use `.filter()` here as well?

SchrodingersGat: >  as it is already verified to exist by the first owners query, right?

I'm not 100% sure about that - you should check by trying to filter with a bad ID. If it does do that check for you, it'd be fine to use get

InvenTree/templates/js/translated/build.js
simonkuehling: thanks!

InvenTree/templates/js/translated/table_filters.js
simonkuehling: well, I thought about that, too... We could change the value for this selection option into a function callback in `table_filter.js --> getAvailableTableFilters()`, like so:

```js
[...]
           assigned_to: {
              title: '{% trans "Responsible" %}',
              options: function () { 
                  var ownersList = {};
                  inventreeGet(`/api/user/owner/`, {}, {
                      async: false,
                      success: function(response) {
                          for (key in response) {
                              var owner = response[key];
                              ownersList[owner.pk] = {
                                  key: owner.pk,
                                  value: `${owner.name} (${owner.label})`,
                              };
                          }
                      }
                  });
                  return ownersList;
              },
          },
[...]
```
and then extend the html generation in `filter.js --> setupFilterList()` to run that callback like this:
```js
        if (options instanceof Function){
            // run callback function for filter options
            options = options();
        }
```
Now we can use either a `list` as before, or a `function()` for the filter options.
I tried that and it works flawlessly - should I commit this?

InvenTree/build/api.py
simonkuehling: looking good: any bad ID is successfully discarded by the first `Owner.objects.filter(pk=value)` already, so I'm happy with a `get()` here...

InvenTree/templates/js/translated/table_filters.js
simonkuehling: I added the proposed solution as a commit for better overview as https://github.com/inventree/InvenTree/pull/4408/commits/acc31a20e1fed4dba11726a91760ab505047fb3e - let me know if you like it or not, I'm open for alternatives as well...

SchrodingersGat:
@simonkuehling nice, seems to work as desired. Thanks for this contribution :)

Issue #4202, opened 2023-01-11
[FR] All Models should have Metadata attached


miggland:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

I  have some ideas of things I would like to implement in some plugin/using my own API scripts. However, the central models are missing the Metadata extension in several cases.

### Suggested solution

Add the Metadata plugin to the Models missing it. See PR #3895 as a reference.

* List updated *

Already present:
- [x] Company
- [x] PO
- [x] SO
- [X] Part categories
- [x] Part
- [x] Stock item
- [x] Stock location
- [x] Manufacturer part
- [x] Supplier part
- [x] Part labels
- [x] Stock item labels
- [x] Stock location labels
- [x] PO lines
- [x] PO extra lines
- [x] SO lines
- [x] SO extra lines
- [x] SO shipments
- [x] Build Orders
- [x] Build order items
- [x] BOM items
- [x] Reports BOM
- [x] Reports Build
- [x] Reports PO
- [x] Reports SO
- [x] Reports test
- [x] Part parameter template

Not present:
- [ ] Part test templates
- [ ] Part parameters
- [ ] Part category parameter templates
- [ ] BOM item substitute
- [ ] Part relateds
- [ ] Stock item test result

New models:
- [ ] Return order
- [ ] Return order lines
- [ ] Return order extra lines
- [ ] Company contacts

No action on these:
- ~~Users~~
- ~~Part pricings~~
- ~~various price breaks~~ 
- ~~various attachments~~
Input welcome - which ones make sense? Did I miss any which have had metadata models added since 0.9.2?

### Describe alternatives you've considered

Alternative is to not do this at all :)

### Do you want to develop this?

- [X] I want to develop this.

matmair:
I would not change the Users model, which can unintended effects in Django.
I would also prefer to not touch pricings/price breaks as we had big refactors there in the latest releases and I would like to wait for a few releases till we are sure nothing unobvious broke.

miggland:
@matmair Makes sense, thanks for the input. I've reorganised the text above to make it a bit clearer on this, and what's already there.

matmair:
Seems fine. If you create a PR please make sure that you reduce the number of migrations to as low as possible.

SchrodingersGat:
Not stale

miggland:
@SchrodingersGat I've updated the list above according to #4410 - so there are still a few models missing.

[...]

PR #4410, opened 2023-02-24
Add Metadata to further models

miggland:
Adds the metadata mixin, and thus the ability to add metadata to the following models, as mentioned in https://github.com/inventree/InvenTree/issues/4202 (List here is slightly different):

- [X] Manufacturer part
- [X] Supplier part
- [X] Part labels
- [X] Stock item labels
- [X] Stock location labels
- [X] PO lines
- [X] PO extra lines
- [X] SO lines
- [X] SO extra lines
- [X] SO shipments
- [X] Build Orders
- [X] Build order items
- [X] BOM items
- [X] Part parameter templates
- [X] Reports BOM
- [X] Reports Build
- [X] Reports PO
- [X] Reports SO
- [X] Reports stock item test

To do:
- [x] Migrations
- [x] Tests
- [x] Update API version


miggland:
@SchrodingersGat @matmair: After adding metadata to a BOM, it seems there is an extra column included in the BOM export file, which means this test fails:

```
FAIL: test_export_csv (part.test_bom_export.BomExportTest)
Test BOM download in CSV format.
----------------------------------------------------------------------
  File "/home/inventree/InvenTree/part/test_bom_export.py", line 136, in test_export_csv
    self.assertIn(header, expected)
AssertionError: 'metadata' not found in ['BOM Level', 'BOM Item ID', 'Parent ID', 'Parent IPN', 'Parent Name', 'Part ID', 'Part IPN', 'Part Name', 'Description', 'Assembly', 'Quantity', 'optional', 'consumable', 'overage', 'Reference', 'Note', 'inherited', 'allow_variants', 'Default Location', 'Total Stock', 'Available Stock', 'On Order']
```

What's you take on this - should metadata be included in BOM export or should metadata be excluded from the export somehow?

InvenTree/report/api.py
```diff
@@ -16,11 +16,12 @@
...
 import part.models
-from InvenTree.mixins import ListAPI, RetrieveAPI, RetrieveUpdateDestroyAPI
+import report.models
+from InvenTree.mixins import (ListAPI, RetrieveAPI, RetrieveUpdateAPI,
+                              RetrieveUpdateDestroyAPI)
+from plugin.serializers import MetadataSerializer
 from stock.models import StockItem, StockItemAttachment
 
-from .models import (BillOfMaterialsReport, BuildReport, PurchaseOrderReport,
-                     SalesOrderReport, TestReport)
```
matmair, line 23: Question: Why this rewrite?

InvenTree/report/api.py
```diff
@@ -295,13 +296,13 @@ def filter_queryset(self, queryset):
 
             valid_report_ids = set()
 
-            for report in queryset.all():
+            for qreport in queryset.all():
```
matmair, line 299: Comment: Without changing the imports this variable rename would be unnecesarry

InvenTree/report/api.py
```diff
@@ -683,13 +684,36 @@ def get(self, request, *args, **kwargs):
...
 
 
+class ReportMetadata(RetrieveUpdateAPI):
+    """API endpoint for viewing / updating Report metadata."""
+
+    def get_serializer(self, *args, **kwargs):
+        """Return correct MetadataSerializer instance depending on which model is requested"""
+        # Get type of report, make sure its one of the allowed values
+        reporttype = self.kwargs.get('reporttype', 'PurchaseOrder')
+        assert reporttype in ['PurchaseOrder', 'SalesOrder', 'Build', 'BillOfMaterials', 'Test']
```
matmair, line 694: Please change: asserts are not garunteed to be followed in python, please raise explicitly if the condition is not met.

InvenTree/report/api.py
```diff
@@ -683,13 +684,36 @@ def get(self, request, *args, **kwargs):
...
+        # Return corresponding Serializer
+        UseModel = getattr(report.models, f"{reporttype}Report")
+        return MetadataSerializer(UseModel, *args, **kwargs)
+
+    def get_queryset(self, *args, **kwargs):
+        """Return correct queryset depending on which model is requested"""
+        # Get type of report, make sure its one of the allowed values
+        reporttype = self.kwargs.get('reporttype', 'PurchaseOrder')
+        assert reporttype in ['PurchaseOrder', 'SalesOrder', 'Build', 'BillOfMaterials', 'Test']
```
matmair, line 703: Please change: See above

InvenTree/report/api.py
```diff
@@ -683,13 +684,36 @@ def get(self, request, *args, **kwargs):
         return self.print(request, orders)
 
 
+class ReportMetadata(RetrieveUpdateAPI):
+    """API endpoint for viewing / updating Report metadata."""
+
+    def get_serializer(self, *args, **kwargs):
+        """Return correct MetadataSerializer instance depending on which model is requested"""
+        # Get type of report, make sure its one of the allowed values
+        reporttype = self.kwargs.get('reporttype', 'PurchaseOrder')
```
matmair, line 693: Question: Is it intended that no argument = PO? Is there a need for a default value or could we raise it if no kwarg is defined? To make sure it is better debuggable.

matmair (changes requested):
Looks good in general @miggland! Please take look at the suggestions/change requests.

matmair:
> @SchrodingersGat @matmair: After adding metadata to a BOM, it seems there is an extra column included in the BOM export file, which means this test fails:
> 
> ```
> FAIL: test_export_csv (part.test_bom_export.BomExportTest)
> Test BOM download in CSV format.
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/inventree/InvenTree/part/test_bom_export.py", line 136, in test_export_csv
>     self.assertIn(header, expected)
> AssertionError: 'metadata' not found in ['BOM Level', 'BOM Item ID', 'Parent ID', 'Parent IPN', 'Parent Name', 'Part ID', 'Part IPN', 'Part Name', 'Description', 'Assembly', 'Quantity', 'optional', 'consumable', 'overage', 'Reference', 'Note', 'inherited', 'allow_variants', 'Default Location', 'Total Stock', 'Available Stock', 'On Order']
> ```
> 
> What's you take on this - should metadata be included in BOM export or should metadata be excluded from the export somehow?

Metadata should not be included, it is to be used by plugins - not the user.

InvenTree/report/api.py
miggland: The reason is that I made a combined method for all Report metadata - where I  use getattr based on the argument.

`UseModel = getattr(report.models, f"{reporttype}Report")`

This is not possible (without some other, sketchy solutions using `sys`) if the modules are imported in the main namespace.

If you have a suggestion of how to change the `getattr` method, please let me know.

InvenTree/report/api.py
miggland: Correct. But, see above.

InvenTree/report/api.py
matmair: I would be keener on a strongly typed reference. Names might change, this way refactoring will not work automatically. I will type a suggestion up.

miggland: OK, please share if you have another solution! However, looking at the other imports let's the one I removed stand out as different:
```
import build.models
import common.models
import InvenTree.helpers
import order.models
import part.models
```

matmair:
@miggland I have made a commit to [this branch](https://github.com/matmair/InvenTree/tree/miggland/metadata-hardreferences) that rolls back most of the import changes and uses hot refs for the view

InvenTree/report/api.py
miggland: > @miggland I have made a commit to [this branch](https://github.com/matmair/InvenTree/tree/miggland/metadata-hardreferences) that rolls back most of the import changes and uses hot refs for the view

Ah, good solution! I'll try to get back to this and incorporate your ideas in the next few days.

miggland:
Anybody have an idea what causes this?

```
5.7124s test_package_loading (plugin.test_plugin.RegistryTests)
5.5662s test_printing_endpoints (plugin.base.label.test_label_mixin.LabelMixinTests)
5.3070s test_printing_process (plugin.base.label.test_label_mixin.LabelMixinTests)
Exception ignored in: <function Local.__del__ at 0x7fc079d0ddc0>
  File "/home/inventree/data/env/lib/python3.9/site-packages/asgiref/local.py", line 89, in __del__
    for context_obj in self._context_refs:
  File "/home/inventree/data/env/lib/python3.9/site-packages/asgiref/local.py", line 100, in __getattr__
    with self._thread_lock:
  File "/home/inventree/data/env/lib/python3.9/site-packages/asgiref/local.py", line 100, in __getattr__
    with self._thread_lock:
  File "/home/inventree/data/env/lib/python3.9/site-packages/asgiref/local.py", line 100, in __getattr__
    with self._thread_lock:
  [Previous line repeated 988 more times]
RecursionError: maximum recursion depth exceeded
```

matmair:
Seems like it could be related to plugins/app.ready() - I will take a look in the next days.

miggland:
The test is still failing on my test VM, with the error given above - even though it's working here. Odd.

It is somehow related to `slowtests`- if I disable those (see change below), then the tests run fine.

I'm also seeing extra migration files created constantly, with the following content. Strange, anything I need to worry about?
```
# Generated by Django 3.2.18 on 2023-03-17 11:06

from django.db import migrations


class Migration(migrations.Migration):

    dependencies = [
        ('InvenTree', '0001_initial'),
    ]

    operations = [
    ]
```
```
# Generated by Django 3.2.18 on 2023-03-17 12:00

from django.db import migrations


class Migration(migrations.Migration):

    dependencies = [
        ('InvenTree', '0002_auto_20230317_1106'),
    ]

    operations = [
    ]

```


Change to slowtests:
```
diff --git a/InvenTree/InvenTree/settings.py b/InvenTree/InvenTree/settings.py
index 1aab68d5d..6f8234917 100644
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -38,8 +38,8 @@ if TESTING:
     PASSWORD_HASHERS = ['django.contrib.auth.hashers.MD5PasswordHasher',]
 
     # Enable slow-test-runner
-    TEST_RUNNER = 'django_slowtests.testrunner.DiscoverSlowestTestsRunner'
-    NUM_SLOW_TESTS = 25
+    #TEST_RUNNER = 'django_slowtests.testrunner.DiscoverSlowestTestsRunner'
+    #NUM_SLOW_TESTS = 25
 
     # Note: The following fix is "required" for docker build workflow
     # Note: 2022-12-12 still unsure why...
diff --git a/tasks.py b/tasks.py
index 85501e024..3788b5f29 100644
--- a/tasks.py
+++ b/tasks.py
@@ -559,7 +559,7 @@ def test(c, disable_pty=False):
     pty = not disable_pty
 
     # Run coverage tests
-    manage(c, 'test --slowreport', pty=pty)
+    manage(c, 'test', pty=pty)
 
 
 @task(help={'dev': 'Set up development environment at the end'})
```

InvenTree/report/api.py
```diff
@@ -418,13 +420,39 @@ class SalesOrderReportPrint(SalesOrderReportMixin, ReportPrintMixin, RetrieveAPI
...
+
+        if reportmodel not in [PurchaseOrderReport, SalesOrderReport, BuildReport, BillOfMaterialsReport, TestReport]:
+            raise ValidationError("Invalid report model")
+        return reportmodel
+
+        # Return corresponding Serializer
+    def get_serializer(self, *args, **kwargs):
+        """Return correct MetadataSerializer instance depending on which model is requested"""
+        # Get type of report, make sure its one of the allowed values
+        UseModel = self._get_model(*args, **kwargs)
```
SchrodingersGat, line 439: Can't we use a simpler approach here:

```python
UseModel = self.Meta.model
```

(or similar)?

The endpoint already has either a model or queryset defined, righ?

SchrodingersGat:
@miggland looking good, but see my comment above

@matmair any thoughts on this?

InvenTree/report/api.py
miggland: This question is for @matmair, he helped out and contributed this part of the code. I'm not sure of the implications / the details of your question..

InvenTree/report/api.py
```diff
@@ -292,7 +294,7 @@ class StockItemTestReportPrint(StockItemTestReportMixin, ReportPrintMixin, Retri
     def report_callback(self, item, report, request):
         """Callback to (optionally) save a copy of the generated report"""
 
-        if common.models.InvenTreeSetting.get_setting('REPORT_ATTACH_TEST_REPORT', cache=False):
+        if common.models.InvenTreeSetting.get_setting('REPORT_ATTACH_TEST_REPORT'):
```
matmair, line 297: @miggland Question: why was this changed?

matmair (approved):
Lgtm.
There is a caching change I am unsure of - the rest looks great!

InvenTree/report/api.py
matmair: @SchrodingersGat not in this case, the model is provided in the url definition as an argument. Reduced the boilerplate as everything is the same for these 4 models. We can type it out if you prefer that. This mainly cuts down on the LoC.

InvenTree/report/api.py
miggland: Well spotted. That was not intentional, must have occurred in a merge operation. I've changed it back.

miggland:
Fixed as above, and merged latest master into the branch.

SchrodingersGat:
@miggland many thanks for the great work here :)

PR #4425, opened 2023-02-27
Add unallocated quantity to StockItem renderer

simonkuehling:
In the "Allocate Stock to Build Orders" workflow I noticed that the StockItems offered in the allocation modal only present their total quantity, not the quantity that is actually available for allocation. That can lead to confusion to why the quantity-field for allocation takes a smaller value than expected if much of that StockItem was already allocated before.

[EDITED]

Therefor I canged the `renderStockItem()` output to show the available stock instead of total stock in the allocation modal:

With changes:

Bildschirmfoto vom 2023-02-27 21-44-23

Before:

Bildschirmfoto vom 2023-02-27 20-50-13

### DIFF
--- a/InvenTree/templates/js/translated/build.js
+++ b/InvenTree/templates/js/translated/build.js
@@ -2296,6 +2296,7 @@ function allocateStockToBuild(build_id, part_id, bom_items, options={}) {
                         render_part_detail: true,
                         render_location_detail: true,
                         render_pk: false,
+                        render_available_quantity: true,
                         auto_fill: true,
                         auto_fill_filters: auto_fill_filters,
                         onSelect: function(data, field, opts) {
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
Looks nice @simonkuehling - do you plan on re-introducing the other changes that you previously had screenshotted here?

simonkuehling:
First I statically modified the renderer to always output `Available: X/Y` instead of the original `Quantity: Y,` which would have been a compromise for for both use cases (allocate parts form, search results). Only later I found that I can just add a parameter to display different information based on what is needed in the specific context.

For the search results, allocation data isn‘t of much value I think. And for allocating stock to build orders the total quantity is irrelevant - here only the available quantity counts.

SchrodingersGat:
Thanks for the continued contributions :)

PR #4430, opened 2023-02-28
Fix some regressions in new "Responsible" filter for build orders

simonkuehling:
This fixes two issues introduced with the latest addition of filtering by responsible user/group in build orders:

- The active filter did not show the human readable value due to the introduction of dynamically loaded filter options
- The user/group icon was not correctly selected in translated interfaces

Fixed:

Bildschirmfoto vom 2023-02-28 16-52-32

Before:

Bildschirmfoto vom 2023-02-28 16-53-49


SchrodingersGat:
Thanks for the fix 🙂

Issue #4431, opened 2023-03-01
Cannot display correctly when browser's preferred language is Chinese


ResRipper:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

I found this issue with my own deployment and official demo.

When using Chinese as preferred language in browser, index page and setting page shows nothing:

![Language - Chinese](https://user-images.githubusercontent.com/16681844/222055308-3c556c43-0aa9-4b7a-b966-980c69b4f298.png)
![Preferred language setting](https://user-images.githubusercontent.com/16681844/222056086-02ccaf23-e036-47a8-8a5a-a7e0bc6a09c5.png)

By changing the language to English, things went normal:

![Language - English](https://user-images.githubusercontent.com/16681844/222056525-34a76426-e289-47f0-8e0b-3151549bf2b7.png)
![Preferred language setting](https://user-images.githubusercontent.com/16681844/222056556-96f2f429-ddc5-4afb-810a-84aabd71d5ed.png)

According to the console, it is trying to load files started with `static/i18n/zh-cn.*`, while the files should be `static/i18n/zh.*`, probably caused by [settings.py](https://github.com/inventree/InvenTree/blob/fdfc3e5e7eb930ca0f71d6c23157c099cc174770/InvenTree/InvenTree/settings.py#L721)? I also tried French, Greek and Korean and they don't have this issue.

![dev console](https://user-images.githubusercontent.com/16681844/222059552-6feaa937-5f92-4279-8e4f-6d710e375a61.png)
![static folder](https://user-images.githubusercontent.com/16681844/222062985-c30a96c5-a0b4-407a-b0dc-91a3bde87ec4.png)

### Steps to Reproduce

1. Use `Chinese` as browser's preferred language
2. Visit any InvenTree demo server listed in https://inventree.org/demo.html

### Expected behaviour

Index page can be display as normal

### Deployment Method

- [ ] Docker
- [ ] Bare metal

### Version Information

InvenTree demo server: 0.11.0 dev
My deployment: 0.10.1

Browser tested:
|MacOS 13.2|Android 13|
|---|---|
|Firefox 110.0|Duckduckgo 5.151.0|
|Safari 16.3||
|Edge 110.0.1587.57||

SchrodingersGat:
@ResRipper thanks for reporting, I can replicate this. I'll look into it!

SchrodingersGat:
[...]

PR #4433, opened 2023-03-01
Translation Fixes

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/4431

This PR fixes an issue relating to running translations in a production build.

Static javascript files are pre-rendered by running `invoke static` - and a copy produced for each supported language. However, there were some issues related to the requested language code not exactly matching the translated language code.

This PR serves to standardize the language code representations, and also provides a backup path if the requested javascript files are not found.


matmair (approved):
Interesting... I was under the impression Django does this automatically and checked when implementing with Spanish/Mexican Spanish.
The changes look good to me.

The language codes for the different Chinese languages are always a problem because there are old and new identifiers (the 2 letters being the old one). Some tools only know one of them - especially older translation tools.

PR #4435, opened 2023-03-01
Slow tests

SchrodingersGat:
Show which unit tests are running slow

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -32,15 +32,24 @@ INVENTREE_NEWS_URL = 'https://inventree.org/news/feed.atom'
 # Determine if we are running in "test" mode e.g. "manage.py test"
 TESTING = 'test' in sys.argv
 
-# Note: The following fix is "required" for docker build workflow
-# Note: 2022-12-12 still unsure why...
-if TESTING and os.getenv('INVENTREE_DOCKER'):
-    # Ensure that sys.path includes global python libs
-    site_packages = '/usr/local/lib/python3.9/site-packages'
+if TESTING:
 
-    if site_packages not in sys.path:
-        print("Adding missing site-packages path:", site_packages)
-        sys.path.append(site_packages)
+    # Use a weaker password hasher for testing (improves testing speed)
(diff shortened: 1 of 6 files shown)


matmair:
Very interesting results so far. I will try to get us hooked up to Azure DevOps over the weekend so we can get statistics and good alerting tools for tests / coverage.

InvenTree/InvenTree/settings.py
```diff
@@ -32,15 +32,24 @@
...
+
+    # Enable slow-test-runner
+    TEST_RUNNER = 'django_slowtests.testrunner.DiscoverSlowestTestsRunner'
+    NUM_SLOW_TESTS = 25
+
+    # Note: The following fix is "required" for docker build workflow
+    # Note: 2022-12-12 still unsure why...
+    if os.getenv('INVENTREE_DOCKER'):
+        # Ensure that sys.path includes global python libs
+        site_packages = '/usr/local/lib/python3.9/site-packages'
```
matmair, line 48: Note: This will break once the version gets bumped. Can we detect the python version and then set the path?

InvenTree/part/test_category.py
```diff
@@ -224,21 +224,23 @@ def test_category_tree(self):
...
+        C23 = PartCategory(name='C23', parent=B2)
 
-        C31 = PartCategory.objects.create(name='C31', parent=B3)
-        C32 = PartCategory.objects.create(name='C32', parent=B3)
-        C33 = PartCategory.objects.create(name='C33', parent=B3)
+        C31 = PartCategory(name='C31', parent=B3)
+        C32 = PartCategory(name='C32', parent=B3)
+        C33 = PartCategory(name='C33', parent=B3)
+
+        PartCategory.objects.bulk_create([B1, B2, B3, C11, C12, C13, C21, C22, C23, C31, C32, C33])
```
matmair, line 243: I do not think we can bulk_create the parent and children in the same action

matmair (approved):

SchrodingersGat:
Already knocks a good ~15 minutes off the tests. Some more work to be done, but now we have an obvious metric (worst performing tests) to guide us.

Issue #4442, opened 2023-03-03
[FR] Add "Total Cost" column to purchase order list table


simonkuehling:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

The purchase order table in `/order/purchase-order/` currently does not show the `total_price` value of each order as a `Total Cost` column in the list.

In order to quickly add that column as a sortable+switchable value, the API needs an additional field `total_price_currency` which is not included right now. `/api/order/po/{id}/` does come with a `total_price` value, but that is not very usable without the currency of course. Internally in python that field of the model is of type `Money`, which templates can extract the currency from - through the API this needs an extra field (I guess..?)

### Suggested solution

- Add `total_price_currency` to the `/api/order/po/{id}/` API endpoint
- Add a switchable and sortable extra column `Total Cost` to the table in `/order/purchase-order/`

### Describe alternatives you've considered

.

### Do you want to develop this?

- [X] I want to develop this.

SchrodingersGat:
Ordering this field becomes tricky when we take currency conversion into account!

We could either:

a) Calculate the `total_cost` for a purchase order whenever any of the lines item change
b) Annotate the queryset "on the fly" with a total-cost calculation

If we got for a) we can convert to a common currency (which would allow sorting). Or, we could have a separate *hidden* column which converts the total price to a common currency, and then silently sort on *that* field.

If we go for b) we potentially don't need any new columns, but sorting becomes complicated.

simonkuehling:
Hmm - isn't the `total_cost` already converted to a common currency right now?

If I create a PO with only `€` line items, the `Total Cost` is converted to the common default currency it seems:

Mixed PO items:
![Bildschirmfoto vom 2023-03-03 10-36-05](https://user-images.githubusercontent.com/296454/222688213-afedb1f9-540b-48ae-9400-39d1536bf545.png)

Only `€` PO items:
![Bildschirmfoto vom 2023-03-03 10-49-19](https://user-images.githubusercontent.com/296454/222688380-a5c36bde-10ad-4d8e-8ef6-5c61133d7308.png)


```
GET /api/order/po/

[
    {
        "pk": 11,
        "issue_date": null,
        "complete_date": null,
        "creation_date": "2023-03-03",
        "description": "Test EUR",
        "line_items": 2,               <--- these are all €
        "link": "",
        "overdue": false,
        "reference": "PO0011",
        "responsible": null,
[...]

PR #4445, opened 2023-03-03
Add total cost column to purchase order list

simonkuehling:
Implements #4442 and adds sorting by "responsible" column as well.


SchrodingersGat:
@simonkuehling thanks for the addition. Note that I'll be looking further into how the `total_price` is added to the queryset, I think it could be done a bit more efficiently.

PR #4486, opened 2023-03-14
Update BomItem 'validated' field

SchrodingersGat:
The 'validated' field on the BomItem API was previously calculated at run-time which is expensive, and also makes filtering by 'validated' field *extremely* slow.

This PR provides simple database caching for the 'validated' field, making API query filtering orders of magnitude faster.

On an example dataset, filtering against this field went from ~20s to ~200ms

Note: API interface / etc are unaffected - this is a backend-only change


matmair (commented):
Nice Speedup!

Issue #4304, opened 2023-02-04
[Roadmap] background worker process


SchrodingersGat:
We have been using [django-q](https://github.com/Koed00/django-q) for ages, but the project has not received any updates in a few years.

There are also some shortcomings - in particular the tasks seem to run on multiple workers - see https://github.com/inventree/InvenTree/issues/4290

Perhaps it is time to migrate to something like [celery](https://github.com/celery/celery):

- Better support (project is active)
- More comprehensive feature set
- Future proofing

I have not yet done the required reading to determine how difficult this might be, or if it would require using redis, etc.

matmair:
@SchrodingersGat the thing I like about django q is that it does not need a broker to run. To my knowledge, we will need redis to use celery. Thus making testing and small bare metal setups harder.

SchrodingersGat:
Indeed. I do like the simplicity of the current solution. Before switching I'd like to understand if there is a way around the current issues. The biggest one being the double execution of tasks

SchrodingersGat:
Ref: https://github.com/Koed00/django-q/issues/633
Ref: https://github.com/Koed00/django-q/issues/577
Ref: https://github.com/Koed00/django-q/issues/566

matmair:
@SchrodingersGat can we close this now as #5167 is merged

SchrodingersGat:
Yes, I think we can. Sticking with django-q is more attractive now that we've found an active fork

PR #4499, opened 2023-03-16
Bypass cache when displaying settings on "settings" page

SchrodingersGat:
- Sometimes caching issues can cause "old" values to be stored (depends on the worker)
- Until we have a shared cache, this is a problem
- Force the settings to be re-loaded from the database when displaying
- Further improvement would be to render them via the API

Note: We should really address having a proper caching system which is shared between the background workers and the front-end workers

Ref: https://docs.djangoproject.com/en/4.1/topics/cache/

### DIFF
--- a/InvenTree/label/api.py
+++ b/InvenTree/label/api.py
@@ -185,7 +185,7 @@ class LabelPrintMixin(LabelFilterMixin):
         outputs = []
 
         # In debug mode, generate single HTML output, rather than PDF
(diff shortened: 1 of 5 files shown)


matmair:
@SchrodingersGat making redis required would make a lot of stuff easier/faster. Would couple well with #4304

SchrodingersGat:
Would memcached be a viable alternative? Looks a bit simpler to setup?

matmair:
@SchrodingersGat I do not see big differences between redis and memchached; I am just used to redis as a lot of other projects in my cluster use it.

PR #4505, opened 2023-03-18
Search refactor

SchrodingersGat:
This PR refactors the "quick search" functionality, bringing significant speed and usability improvements.

Basically, we introduce a single API endpoint for global search, which simply consolidates search results into a single query. This means that only one API request is needed for *all* search results, and this means that all the web workers are no longer consumed with a single set of search requests.

### TODO

- [x] Reimplement links for search results
- [x] Unit testing
- [x] Unit testing without necessary permissions
- [x] Remove "no results" text when searching
- [x] Handle "no results found"


matmair (approved):
How does this handle missing privileges?

SchrodingersGat:
> How does this handle missing privileges?

Will add a unit test for this

Issue #3501, opened 2022-08-08
[FR] Purchase Order Approvals


cpattersonic2:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

We are excited to move from partkeepr to inventree at our small company, and really appreciate the work you have put into this piece of software. However, in our line of work a purchase must always have been approved by the Project Lead.

### Suggested solution

Would it be possible to add an approver field to the purchase order, and then show a time stamp when the approver/user has approved the PO for purchase. (e-mailing users when there is an outstanding approval would be cool, and when approved). Would this need to be done in the inventree core, or could this be added by us as a plugin?

Thanks for any feedback or help!

### Describe alternatives you've considered

We plan on moving over to inventree anyway, but we currently dislike our current purchasing software and would love to just use inventree for the whole 9 yards. Having inventory and purchasing integrated sounds like the bees knees.

### Do you want to develop this?

- [ ] I want to develop this.

matmair:
@cpattersonic2 invenTree is not a purchasing software. We do not handle any form of accounting so you will still need a solution for that even if this FR is implemented.
I would not hold my breath for this to be implemented in core as there are multiple things that would need to be adjusted and the core team is small.
This should work well as a plugin though.

SchrodingersGat:
@cpattersonic2 I can see the value here, although it would not be for everyone. if it was to be implemented as a first class feature rather than a plugin it would need (at minimum):

- Settings which control whether PO's require approval (default OFF)
- Add model fields to purchase order to indicate who approved (and when)
- User interface additions (i.e. display approval, buttons to approve, etc)
- Update API to allow approval / revoke approval of order
- Generation of notifications based on approval
- Unit testing

So this is a fair amount of work. If it is something you are keen to pursue, and don't want to develop yourself, you can reach out to us to sponsor development of this feature. Otherwise it will probably sit on the pile for some time (there's a *lot* of outstanding stuff).

cpattersonic2:
Thank you so much for your replies, and I understand that it could be a
good bit of development effort. I can run it up the ladder to ask if
development funding is available. Can you direct me towards the procedure
for funding this development so that I can present it for funds. Thanks!

SchrodingersGat:
@cpattersonic2 for such requests please send an email to support (at) inventree (dot) org

matmair:
[...]

PR #4543, opened 2023-03-30
Implement a generic API endpoint for enumeration of status codes

SchrodingersGat:
Adds a generic API endpoint for enumeration of status codes, which will allow external applications (e.g. the mobile app) to access available status codes, as well as translated labels

Implemented for:

- [x] Purchase Order
- [ ] Sales Order
- [ ] Return Order
- [ ] Return Order Line Item
- [ ] Stock Item
- [ ] Stock History
- [ ] Build Order


matmair:
Nice! This will tye nicely into #3501

PR #4554, opened 2023-04-02
Add option to use InvenTree as OIDC provider

p-fruck:
This draft utilizes the `django-oidc-provider` package to allow InvenTree to act as an OIDC provider to other services. This feature can be enabled by setting `INVENTREE_OIDC_PROVIDER_ENABLED` to true (default value is false). This allows InvenTree users to integrate other services with the InvenTree account management system.

I am currently working towards a PoC of federation between InvenTree instances, which I wanted to develop as a separate plugin initially. Using InvenTree as an identity provider is the first step towards this goal which I could not implement into a separate plugin because I required an unprotected route for the the `.well-known/openid-configuration`. The only way I could make this route unprotected was to serve it under the `accounts/` route.

If you would like to accept this PR I will also create some documentation to configure the OIDC provider.

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -842,6 +842,10 @@ ACCOUNT_FORMS = {
 SOCIALACCOUNT_ADAPTER = 'InvenTree.forms.CustomSocialAccountAdapter'
 ACCOUNT_ADAPTER = 'InvenTree.forms.CustomAccountAdapter'
 
+OIDC_PROVIDER_ENABLED = get_boolean_setting('INVENTREE_OIDC_PROVIDER_ENABLED', 'oidc_provider_enabled', False)
+if OIDC_PROVIDER_ENABLED:
+    INSTALLED_APPS.append('oidc_provider')
+
 # Markdownify configuration
 # Ref: https://django-markdownify.readthedocs.io/en/latest/settings.html
 
--- a/InvenTree/InvenTree/urls.py
+++ b/InvenTree/InvenTree/urls.py
@@ -195,6 +195,9 @@ urlpatterns = [
     re_path('', include(backendpatterns)),
 ]
 
+if settings.OIDC_PROVIDER_ENABLED:
+    urlpatterns.append(path('accounts/oidc-provider/', include('oidc_provider.urls', namespace='oidc_provider')))
+
 # Server running in "DEBUG" mode?
 if settings.DEBUG:
     # Static file access
--- a/requirements.in
+++ b/requirements.in
@@ -17,6 +17,7 @@ django-maintenance-mode                 # Shut down application while reloading 
 django-markdownify                      # Markdown rendering
 django-money<3.0.0                      # Django app for currency management  # FIXED 2022-06-26 to make sure py-moneyed is not conflicting
 django-mptt==0.11.0                     # Modified Preorder Tree Traversal
+django-oidc-provider                    # Allow to use InvenTree as an OIDC Proivder for other services
 django-redis>=5.0.0                     # Redis integration
 django-q                                # Background task scheduling
 django-q-sentry                         # sentry.io integration for django-q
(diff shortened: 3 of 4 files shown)


SchrodingersGat:
@p-fruck this is outside my area of knowledge - can you expand a bit more on how you would see something like this being used in practice?

matmair:
I am not sure InvenTree should be an OIDC provider. We have only minimal user and group management features. Which was left purposely that way when I made the MFA and social provider upgrades so as to not have to maintain a secure identity provider suite and instead be a consumer of one. Are you willing to maintain this feature in a secure state for the foreseeable future? Because the second we merge this InvenTree is an identity provider and will be used as such - making it a critical security asset and possibly a target.

I do not see why we would need this for federation, as long as you have unique IDs federation should work (eg. nextcloud uses the domain of the server and userid on the server). Can you lay out what you plan to implement in a feature request so we can discuss the planned architecture?

Also, the sub route under `/accounts` for wellknown is not really up to standard, please move it out into root. You can remove protections by wraping the view function as [exempt](https://github.com/inventree/InvenTree/blob/c28c991591c9bd2c29be4c0046d500e3271feda2/InvenTree/InvenTree/permissions.py#L80).

wolflu05:
I would actually be interested in what kind of federation features you have in mind?

p-fruck:
> I am not sure InvenTree should be an OIDC provider

I am neither. I wanted to implement it as a plugin at first, but was not able because of the protected routes.

> Also, the sub route under /accounts for wellknown is not really up to standard
> You can remove protections by wraping the view function as [exempt](https://github.com/inventree/InvenTree/blob/c28c991591c9bd2c29be4c0046d500e3271feda2/InvenTree/InvenTree/permissions.py#L80).

Is this also possible for extensions? I am fairly new to django and I really struggled integrating the OIDC provider into a separate extension

> Are you willing to maintain this feature in a secure state for the foreseeable future

I'd be, as I will explain in my usecase.


### Usecase

I am part of a local chaos community that interacts with other regional chaos communities on a regular basis. Since all of these communities have their own assets inside their hacker/maker spaces, we wanted to implement some kind of inventory management system and collectively agreed on using InvenTree which makes it a common standard among our communities. Since those communities are decentralized and run their own independent infrastructure, they also have different InvenTree instances and authentication providers.

Since we also share assets across communities regularly, we wanted to have some functionality to share items across instances, so we can always keep track of items shared with other communities. Since we trust the other communities, we also want to give them access to our InvenTree instance and manage the user permissions independently.

Since we do not want to force other communities or individuals to some specific identity provider, we thought it would make sense for InvenTree to become its own identity provider (using the established OIDC standard) so that even communities/individuals using the integrated InvenTree (Django) accounts could authenticate against our instance using their respective account credentials. For now, this PR would allow us to give users of trusted InvenTree instances access to our instance simply by adding the foreign InvenTree instance as an OIDC provider. The users of this instance would the be able to manipulate items in our InvenTree.

Based on this functionality, there should be features added in the future that allow for sharing resources like items and perhaps categories across instances using the InvenTree web GUI.

#### Questions

- What do you think of this idea in general?
- Would you like to have such functionality integrated into InvenTree at some point in the future or do you want to keep it separate (e.g. using a plugin, if this is even possible)
- Do you have additional ideas/advices for the concrete implementation? Does it make sense to use an OIDC provider for this use case or how would you establish trust between two federated instances?

matmair:
Ok, you got my attention with ccc/chaos. Are you ok with moving this discussion into an issue and postponing the decision on this PR until we have a better understanding of the plan for federation?

Knowing your target is chaos / the fediverse: why not use/build around ActivityPub and Mastodon? We could implement a notification provider with ActivityPub and sue Mastdons OAuth to do authentication.
If this is something more spaces will use I would like to move OIDC config into the admin UI to make it more portable.

On a side note: A bunch of users and devs are German, I bet we have a few ppl from chaos here already.

Issue #4510, opened 2023-03-20
[FR] make exact match search possible (again)


winzkigermany:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

I can't do an `exact match` search anywhere.

As I learned from the project history, the **search** field has intentionally been modified from "exact match" to a "partial string search", where I can enter partial strings. 
e.g. searching for `platine_` will find `platine_0` and `platine_0_20`

However, as a result, I can no longer do an exact search, if I really really mean to **only** find `platine_0`
For example, if I intend to export a stock list for only one part, I need to remove all the by-catch from the exported file later.

### Suggested solution

It would be great to have an **"exact match"** option next to the search field,
which will _not_ return results like `platine_0_1` along with the results for `platine_0`

This would work either by adding a checkbox called `exact matches only`, 
**or** by allowing to put the search term between double quotes, for example `"platine_0"`

I see that it may be a breaking change for users, who already need to search for double quotes,
so probably a checkbox for `exact match search` would be the least intrusive way to add this feature.

### Describe alternatives you've considered

For parts, instead of searching for the contents of the _IPN_ field, I could search for contents of the _Description_ field.

But this has many drawbacks, like:
a) all users would need to remember two identifiers per part, the IPN and the distinguishable part of the description
b) the description fields of parts with similar IPNs would need distinguishable Description texts that don't contain the same substrings
c) there would be an undocumented constraint about how the description needs to be authored, so that it's possible to return  exact matches
d) To get exact results on a description, probably 100% of the description string would need to go in my search string

### Examples of other systems

Google search knows search modifiers like quotes `""` to cut back search fuzziness

File system search uses modifiers like "*" for doing partial searches, because it's not by default "fuzzy".

I don't have an example for a "exact match only" checkbox at hand.

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
This is a similar request to https://github.com/inventree/InvenTree/issues/1623 - which was looking for regular expression matching in searches.

We could certainly look at adding search modifiers for the quick-search bar. Perhaps a "settings" icon in the location indicated below:

![image](https://user-images.githubusercontent.com/10080325/226474670-4e8fbb0f-ff34-43f3-bfe1-a0d57a743c2d.png)

SchrodingersGat:
[...]

PR #4566, opened 2023-04-03
Enable regex search

SchrodingersGat:
This PR enables "whole word search" and "regex search" when searching via the API.

Users can enable regex search functionality (for their own results)

image

image

This PR also serves as an entry point into further search result customization

Closes https://github.com/inventree/InvenTree/issues/4510

### DIFF
--- a/InvenTree/InvenTree/ready.py
+++ b/InvenTree/InvenTree/ready.py
@@ -13,7 +13,7 @@ def isImportingData():
     return 'loaddata' in sys.argv
 
 
-def canAppAccessDatabase(allow_test: bool = False, allow_plugins: bool = False):
+def canAppAccessDatabase(allow_test: bool = False, allow_plugins: bool = False, allow_shell: bool = False):
     """Returns True if the apps.py file can access database records.
 
     There are some circumstances where we don't want the ready function in apps.py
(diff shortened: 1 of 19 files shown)


InvenTree/InvenTree/ready.py
```diff
@@ -42,6 +41,9 @@ def canAppAccessDatabase(allow_test: bool = False, allow_plugins: bool = False):
         'mediarestore',
     ]
 
+    if not allow_shell:
```
matmair, line 44: @SchrodingersGat what has this to do with search?

InvenTree/order/api.py
```diff
@@ -1304,7 +1304,7 @@ def download_queryset(self, queryset, export_format):
 
     filter_backends = [
         rest_filters.DjangoFilterBackend,
-        filters.SearchFilter,
+        InvenTreeSearchFilter,
         filters.OrderingFilter,
     ]
```
matmair, line 1309: Suggestion: Seems like this block is used often - can de deduplicate that?

matmair (approved):

InvenTree/InvenTree/ready.py
SchrodingersGat: Nothing, but during testing for search I found that some functions were blocked entirely if running from the shell.

InvenTree/order/api.py
SchrodingersGat: Yeah, this would be a good candidate for a refactoring

InvenTree/InvenTree/ready.py
matmair: Ah ok, I was a bit surprised what effect it could have on search 👍

PR #4568, opened 2023-04-03
Cryptography version update

SchrodingersGat:
Major version update to cryptography library

This was pinned way back when we built our docker image from an "alpine" base - now we use "python:slim" and the original problem we were trying to fix no longer exists anyway

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,9 +19,6 @@ ARG commit_tag=""
 
 ENV PYTHONUNBUFFERED 1
 
-# Ref: https://github.com/pyca/cryptography/issues/5776
-ENV CRYPTOGRAPHY_DONT_BUILD_RUST 1
-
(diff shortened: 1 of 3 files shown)


matmair (approved):
Great, that reduces our exposure a lot. We should try to make a point release for 0.10.x soon as there are problems with OpenSSL in the previous versions of cryptography.

SchrodingersGat:
@matmair this is why we pinned cryptography lib: https://github.com/inventree/InvenTree/actions/runs/4601312659/jobs/8129025727

matmair:
I do not really get why it is not using the wheels...

SchrodingersGat:
The docker image builds locally fine. It may be related to the arm targets which don't have native wheels?

matmair:
ARM is a PITA to build for on debian

PR #4579, opened 2023-04-04
Attachment API refactor

SchrodingersGat:
Refactors repeated code to create a common attachment API endpoint management class.


matmair (approved):
Looks good. It would be good to have some tool to make sure non of the interfaces change with refactorings like this.

SchrodingersGat:
Abandoning this as it has introduced a *lot* of weird edge cases which the small amount of refactoring does not make up for.

However there are some interesting things here which can be used - i'll make a new PR with these changes cherry picked

PR #4583, opened 2023-04-04
Reduce duplication with filters

matmair:
Reduce duplicated filterset definitions


matmair:
Related to #4566

SchrodingersGat:
Nice, this is a good overall simplification :)

PR #4584, opened 2023-04-04
Bump to dj >4

### LINKED ISSUE #2853: Update to Django 4.x
As raised in #2849 update to Docker 4.x -> might be a lot of work.
This will break a lot of forks.

matmair:
This would close #2853; not tested manually.


matmair:
So far I found 2 problems:
1) django-error-report is very much unmaintained. I made upgrades needed for 4.0 a while ago for something else so I switched to my fork for testing. We should consider moving to something else.
2) As far as I see, MPTT is unmaintained and not ready for Django 4.0. See [their changelogs](https://django-mptt.readthedocs.io/en/latest/changelog.html#id2).

matmair:
Ok the dependencies are solved mainly; a few functions need to be rewritten because they were removed with the switch to 4.0

matmair:
And the next [breaking change](https://docs.djangoproject.com/en/4.1/releases/4.1/#reverse-foreign-key-changes-for-unsaved-model-instances) that needs to be included.

matmair:
Ok down to one error after updating `inventree-zapier` for Django 4.0

matmair:
We should be golden now. @SchrodingersGat I think we could take the plunge to Django 4.x with the LTS for 4.2 now released. Thoughts?

matmair:
Discovered a fresh Django bug with Postgres: https://code.djangoproject.com/ticket/34459

matmair:
We will need for 4.2.1 / probably May 1st for this to be mergeable

matmair:
Not stale

SchrodingersGat:
With django-mptt unmaintained, what is your opinion on moving forward to django 4?

matmair:
There multiple changes that would require rewriting and all replacements for mptt I looked at are less efficient so I would let this one go for now.

wolflu05:
Is there an replacement library for mptt?

matmair:
@wolflu05 there are a bunch of them with varying degree of maturity and user base. The query efficiency of mptt is just top-notch.

SchrodingersGat:
MPTT is a *very good* library - it is unfortunate that it is not receiving any active maintenance :'(

PR #4586, opened 2023-04-05
German translations for new search settings

SergeoLacruz:


wolflu05:
From my understanding you need to provide translations via crowdin, [see](https://github.com/inventree/InvenTree/blob/master/CONTRIBUTING.md#translations)

SergeoLacruz:
OK, thanks for the hint. Will do so,

PR #4595, opened 2023-04-10
Remove "required" attribute for order descriptions

### LINKED ISSUE #4588: [FR] Make adding a description optional through the settings page

### Problem statement

In a lot of places in Inventree there is a `description` field, like in the test templates. This description, currently, cannot be left blank.  While testing Inventree I found that lots of time adding a description would not be necessary, but Inventree forces to add text in the field before being able to submit the form.

### Suggested solution

It would be great if this would be a toggle setting under the server settings. For me it can be a global toggle to enable or disable the required description through the whole Iventree, but for other people / usecase it might be nice to, per module (like tests, parts, stock etc.) define if a description is required or optional.

I am now manually copy and pasting the name into the description field, but this is just a workaround.

### Examples of other systems

_No response_

- [X] I want to develop this.

standekker:
Resolves #4588. Originally #4588 only talked about removing the requirement for descriptions, but during testing I found that build, purchase and return orders also required a title which, in my use-case, is not required, thus I also removed the requirement for those titles.

Feel free to comment if the titles on build, purchase and return orders should remain a required field, I will then happily revert the changes on the `order` model and re-submit if needed!


SchrodingersGat:
@standekker you will need to commit the required migration files :)

Run `invoke migrate` to generate these files and then add them to your PR

standekker:
@SchrodingersGat Migration files have been committed!

SchrodingersGat:

PR #4636, opened 2023-04-19
Project code support

SchrodingersGat:
Adds support for "project codes" which can be used to track against orders

### TODO

- [x] Unit testing
- [x] Documentation - https://github.com/inventree/inventree-docs/issues/470


matmair:
@SchrodingersGat we do not have a `Project` model - could we just name it project and maybe throw in metadata support? Could be valuable for plugins.

PR #4642, opened 2023-04-19
Fix typo in Update Check Interval setting

LavissaWoW:
Change will create a new translation key
The following locales have added strings to the old key:

DE
Intervall für die Suche nach Updates

HU
Frissítés keresés gyakorisága

PT
Atualizar Verificação de Intervalo


SchrodingersGat:

PR #4658, opened 2023-04-22
Fix sorting by location in stock item table

sur5r:
Currently, when sorting the stock item table by location, it gets sorted by location id.

This changes the sorting criterion to pathstring.

### DIFF
--- a/InvenTree/stock/api.py
+++ b/InvenTree/stock/api.py
@@ -1003,6 +1003,7 @@ class StockList(APIDownloadMixin, ListCreateDestroyAPIView):
     filter_backends = SEARCH_ORDER_FILTER_ALIAS
 
     ordering_field_aliases = {
+        'location': 'location__pathstring',
         'SKU': 'supplier_part__SKU',
         'stock': ['quantity', 'serial_int', 'serial'],
     }


SchrodingersGat:
@sur5r I would have expected this to be tacked at the api level, not at the db level. Have you confirmed if this works for both ascending / descending ordering?

sur5r:
Yes it works in both directions. My first attempt was `[ 'level', 'name', ]` but this lead to wrong results.

I tried to add this at the API level but failed to understand how it works and whether there is a corresponding place at all.

SchrodingersGat:
@sur5r there is a place, specifically:

https://github.com/inventree/InvenTree/blob/2ffd2354eb2bec7a9c09956be1df29879e6ede03/InvenTree/stock/api.py#L1007

You could add a line as follows:

```python
"location": "location__pathstring",
```

to switch the ordering type for the location field

Changing this at the database level may have some unintended consequences,

sur5r:
Ah, that was non-obvious to me. I tried it and it looks good to me.

SchrodingersGat:
Yah the API functionality has a few subtleties - the reassignment of ordering fields is one of the more complex ones

SchrodingersGat (approved):

PR #4681, opened 2023-04-22
Fix missed translation of overallocation string

LavissaWoW:
* Added missing translation prefix on string for Overallocation REJECT choice


matmair (approved):
LGTM, thanks @LavissaWoW

SchrodingersGat:
Thanks @LavissaWoW , not sure why the docs are failing to build! Will have to look into that one separately.

PR #4692, opened 2023-04-25
bump deps

matmair:
Bump dependencies for python and js. Disablers a new check for HTML files that throws errors with django templates.


SchrodingersGat:
@matmair happy to merge?

matmair:
@SchrodingersGat yes!

PR #4697, opened 2023-04-25
Link changes and in-table clipboards

### LINKED ISSUE #4606: [FR] Quick-search suppl/mfg site for SKU/MPN

### Problem statement

Registering a supplier part involves adding both part information, descriptions and links in order to make ordering as easy as possible. There is as of yet no quick copy button for SKUs in neither PO line items nor supplier parts on part views. This option exists solely on the Supplier Part Detail view. When executing PO's, this produces a lot of extra clicks or awkward click-and-drag marking and copying of SKUs. 

### Suggested solution

As a default, expand the quick copy button/icon to line items and part details. Moving on, it's possible to setup a generated search link for the relevant supplier. When creating a supplier part, the SKU is set as a required field. With some small additions to the company model, you have everything you need to enable a quick-search button to appear behind the given SKU.

Diving a bit into the matter: 
Say you have Mouser as a supplier. Including a new field on the model for a "search link", you can input "https://mouser.com/c/?q=" as the URL prefix for searching. 
Given a supplier part from Mouser with i.e. the SKU "139-ZX60-P33ULN" you can then present a button (i.e. FontAwesome icon) which opens a new tab with the complete URL "https://mouser.com/c/?q=139-ZX60-P33ULN"

As a general priority list for which attribute to display:
1. Supplier part Link
2. Autogenerated search link based on SKU
3. Copy SKU button

This makes it so that the presence of a link will override the generated search link in all cases. This would be important in cases where a search does not send you to a unique product page. Like when your SKU has variations on the supplier site with prefixes/affixes and so on. 

Opening line items (PO) and supplier parts (Part details) in new tabs to copy SKU manually.

### Examples of other systems

_No response_

- [X] I want to develop this.

LavissaWoW:
Closes #4606 

* Clipboard icon added to tables for screens >1200px wide. Enables copying of SKU/MPN/IPN from table cells where these otherwise are hyperlinks

* External links now open in new tabs with noreferrer

### DIFF
--- a/InvenTree/build/templates/build/detail.html
+++ b/InvenTree/build/templates/build/detail.html
@@ -106,7 +106,7 @@
                 <tr>
                     <td><span class='fas fa-link'></span></td>
                     <td>{% trans "External Link" %}</td>
-                    <td><a href="{{ build.link }}" rel="noreferrer" target="_blank">{{ build.link }}</a>{% include "clip.html"%}</td>
+                    <td>{% include 'clip_link.html' with link=build.link %}</td>
                 </tr>
                 {% endif %}
                 {% if build.issued_by %}
--- a/InvenTree/build/templates/build/detail.html
+++ b/InvenTree/build/templates/build/detail.html
@@ -106,7 +106,7 @@
                 <tr>
                     <td><span class='fas fa-link'></span></td>
                     <td>{% trans "External Link" %}</td>
-                    <td><a href="{{ build.link }}">{{ build.link }}</a>{% include "clip.html"%}</td>
+                    <td><a href="{{ build.link }}" rel="noreferrer" target="_blank">{{ build.link }}</a>{% include "clip.html"%}</td>
                 </tr>
                 {% endif %}
                 {% if build.issued_by %}
--- a/InvenTree/InvenTree/static/css/inventree.css
+++ b/InvenTree/InvenTree/static/css/inventree.css
@@ -1095,3 +1095,9 @@ a {
     width:  100%;
     text-align: left;
 }
+
+.flex-cell {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+}
--- a/InvenTree/company/templates/company/manufacturer_part.html
+++ b/InvenTree/company/templates/company/manufacturer_part.html
@@ -105,7 +105,7 @@ src="{% static 'img/blank_image.png' %}"
     <tr>
         <td><span class='fas fa-link'></span></td>
         <td>{% trans "External Link" %}</td>
-        <td><a href="{{ part.link }}" rel="noreferrer" target="_blank">{{ part.link }}</a>{% include "clip.html"%}</td>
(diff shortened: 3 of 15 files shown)


LavissaWoW:
httpstat.us appears to be down atm, thus the failing tests. Just as an FYI.

InvenTree/build/templates/build/detail.html
```diff
@@ -106,7 +106,7 @@ <h4>{% trans "Build Details" %}</h4>
                 <tr>
                     <td><span class='fas fa-link'></span></td>
                     <td>{% trans "External Link" %}</td>
-                    <td><a href="{{ build.link }}">{{ build.link }}</a>{% include "clip.html"%}</td>
+                    <td><a href="{{ build.link }}" rel="noreferrer" target="_blank">{{ build.link }}</a>{% include "clip.html"%}</td>
```
SchrodingersGat, line 109: I think it would be worth breaking this out into a template e.g.

```html
{% include "clip_link.html" with link=build.link %}
```

And then the template includes the clipboard functionality also.

LavissaWoW: Yeah, I see what you mean.
Small braindump:
A general rule of thumb for clip_link.html will then be that it serves the URL as the string of the anchor tag. 
In addition, it'll be targeted at links to be opened in new tabs, mainly external ones. 
If a more generic approach would be ideal, more parameter for string value and external toggle could be added to the template. 
LMK your thoughts, and I'll update the PR to use a template

LavissaWoW:
@SchrodingersGat TBH, the idea itself has lessened somewhat in value in my mind after the above changes, and some of the modifications that were introduced with 0.11.0

My initial observation was that - in 0.10.0 - our purchaser was finding it tedious to find supplPart links from PO's. 
Additionally, in the cases that there was no link on the supplPart, there was no easy way to copy SKUs from PO lines in order to search the supplier's webshop to locate the item. Adding further, if there was a link, he'd just click it and effectively lose his InvenTree tab, as it didn't pop out.

Thus, the original thought was born to be able to do a product search/lookup on the supplier's site with a single click. 
Using a URL slug stored on a company entity, you'd have a function that simply does something along the lines of
`<a href="{{company.search_slug+SKU}}">Some FA icon</a>` and append it to the SKU fields where applicable

matmair:
So this is not necessary again @LavissaWoW? Another note: have you looked into part/order metadata? Might also be a solution.

LavissaWoW:
This PR itself, as it stands, I still see a lot of value in terms of workflow and link security. But the base suggestion from the issue, as explained in my previous comment, has been largely mitigated with this PR and some of the changes that's present in 0.11.

SchrodingersGat (approved):

SchrodingersGat:
@LavissaWoW sorry to be bothersome, but there's some weird CI issues currently. Can you please try merging inventree/master into your branch, and pushing again?

SchrodingersGat:

Issue #4611, opened 2023-04-15
Inventree not using the reports after 12 time upload


mabroens:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

Hi i am playing around with customizing reports. so i took a default report and started modifying this but after 12 uploads (The revision counter is 12 or higher) then it no longer takes my changes. when i rename the same report from test.html to test1.html then it outputs the report with the new markup.

i have testted this several times for buid orders (BOM reports) as well as sales invoices. and for each it was the same result.

Br Martijn

### Steps to Reproduce

1. create a template and upload
2. change the template and upload.
3. repeat step 2 over and over

### Expected behaviour

it should just load and execute

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.10.0
Django Version: 3.2.16

Database: sqlite3
Debug-Mode: False
Deployed using Docker: False
Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}]

SchrodingersGat:
Can you try restarting the server and see if the new report is loaded correctly? There have been reports of a potential cache issue with reports which I haven't been able to isolate yet. Lmk if restarting the server addresses this for you

mabroens:
Hi Oliver,

yes restarting the PC solves the issue. i have created a feature request to add a restart service button to the user pulldown menu (upper richt corner).

we can close this one or reach out if you need my assitence for testing

SchrodingersGat:
> yes restarting the PC solves the issue.

The entire PC or just the InvenTree application?

What proxy are you using to host / serve the static and media files?

mabroens:
sorry here i'm over asked. i do not know what you are looking for here.
im runnig inventree on a small linux (ubuntu 20 lts) machine. 

if you want me to type in any commands feel free to pass them on an i will.

br martijn

SchrodingersGat:
during the installation process you would have had to setup some sort of webserver to host the "static" and "media" files - which includes the uploaded report templates. Something like apache, or nginx, etc

LavissaWoW:
@SchrodingersGat I encountered this issue myself when iterating on a report template design.
[...]

PR #4706, opened 2023-04-27
Add custom template loader class

SchrodingersGat:
- Specifically to ignore caching for generated reports and labels
- Fixes https://github.com/inventree/InvenTree/issues/4611

This PR adds a custom template loader class, which explicitly skips cache loading for "report" or "label" templates. This is to ensure that generated reports and labels are freshly regenerated all the time

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -328,7 +328,7 @@ TEMPLATES = [
                 'InvenTree.context.user_roles',
             ],
             'loaders': [(
-                'django.template.loaders.cached.Loader', [
+                'InvenTree.template.InvenTreeTemplateLoader', [
                     'plugin.template.PluginTemplateLoader',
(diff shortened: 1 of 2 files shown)


matmair (commented):
We will have to look at how this impacts performance;

Another way to handle this would be to drop the cache every time a new report or label template gets uploaded - that would be faster.

SchrodingersGat:
> We will have to look at how this impacts performance;

I believe that the new implementation will only be potentially slower for reports and labels. There should be no impact on other commonly loaded templates.

> Another way to handle this would be to drop the cache every time a new report or label template gets uploaded - that would be faster

This was actually the previous implementation, but the issue is that the cache only gets cleaned for the particular worker. Other workers retain their own cache (as we don't have a shared cache solution)

matmair:
Let's see if reporting sticks out performance issues and deal with the caching then.

PR #4721, opened 2023-04-28
Make sure plugins are always on a new line

### LINKED ISSUE #4719: "Install Plugin" via GUI fails to add plugin to plugins.txt correctly for first plugin

### Describe the bug*

When trying to install a plugin via the GUI function "Install Plugin" without any prior installed plugins, the plugin is appended to the first line of the plugins.txt.

As the first line is already populated with `# InvenTree Plugins (uses PIP framework to install)` the result is e.g. `# InvenTree Plugins (uses PIP framework to install)inventree-zebra-plugin@git+https://github.com/SergeoLacruz/inventree-zebra-plugin  # Installed 2023-04-28 06:55:40.146423+00:00 by xyz` thus ignoring the plugin in a docker env restart because of the `#`.

Subsequently installed plugins are correctly placed in an empty line of the plugins.txt.

### Steps to Reproduce

1. Have a fresh install without installed plugins.
2. Go to "Settings" > "Plugin Settings" > "Plugins" > "Install Plugin"
3. Fill "Package Name" with `inventree-zebra-plugin`
4. Fill "Source URL" with `git+https://github.com/SergeoLacruz/inventree-zebra-plugin`
5. "Submit"
6. Check plugins.txt file

### Expected behaviour

The installed plugin should be added below the comment such that it actually will be parsed by the plugin management.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.11.0
Django Version: 3.2.18
Commit Hash: 593a716
Commit Date: 2023-04-18
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}]

### Relevant log output

_No response_

matmair:
Fixes #4719

### DIFF
--- a/InvenTree/InvenTree/config.py
+++ b/InvenTree/InvenTree/config.py
@@ -277,6 +277,7 @@ def get_plugin_file():
 
         # If opening the file fails (no write permission, for example), then this will throw an error
         plugin_file.write_text("# InvenTree Plugins (uses PIP framework to install)\n\n")
+        plugin_file.write_text("\n# Add your required plugins here\n\n")
 
     return plugin_file
 
--- a/InvenTree/InvenTree/config.py
+++ b/InvenTree/InvenTree/config.py
@@ -277,7 +277,6 @@ def get_plugin_file():
 
         # If opening the file fails (no write permission, for example), then this will throw an error
         plugin_file.write_text("# InvenTree Plugins (uses PIP framework to install)\n\n")
-        plugin_file.write_text("\n# Add your required plugins here\n\n")
 
     return plugin_file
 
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
Interesting that the doc checks fail every now and then, claiming that external links cannot be hit (but they can) due to primarily 503 errors. I wonder if there's a config option to attempt multiple times?

SchrodingersGat:
@matmair how does the extra line fix the linked problem here?

matmair:
@jannick-gb that explains why I can't reproduce the core issue. I do not think that adding a newline every time is the right solution, I will look into making that change.

matmair:
Ok I implemented a lookup to make sure the plugins are always written on the first line

PR #4732, opened 2023-04-30
[Feature] Company Addresses

### LINKED ISSUE #4695: [FR] Split Address in Street, ZIP, Country

### Problem statement

would it be possible to split the Address field for company such that it gets a sepperate entry for Straat, Zip, Country, this would greatly improve report making.

### Suggested solution

Add sepperate field for street, zip and country

Add sepperate field for street, zip and country

### Examples of other systems

_No response_

- [ ] I want to develop this.

LavissaWoW:
Implements and closes #4695


InvenTree/company/models.py
```diff
@@ -250,6 +250,78 @@ def get_api_url():
     role = models.CharField(max_length=100, blank=True)
 
 
+class Address(models.Model):
```
wolflu05, line 253: Id suggest renaming this to CompanyAddress, what do you think @matmair ? Can this be also used for other models?

LavissaWoW: It's part of the Company app, so figured Address would work fine

LavissaWoW:
Obviously won't pass a lot of checks atm. Just getting the PR set up :) 
On that note: Is there any way for me to only run the tests I want, or am I stuck watching the whole suite run?

InvenTree/company/models.py
matmair: I think Adress is fitting; no strong feeling about it though

matmair:
@LavissaWoW you can run `python manage.py test` followed by dot notation for the test you want. So ex. `python mange.py test plugin.test_plugin.PluginTagTests.test_tag_plugin_list`

matmair:
On another note: without adding migrations no test will run in the CI as that is checked first.

SchrodingersGat:
@LavissaWoW one thing to keep in mind (if you're not already considering it) is that this will need a data migration to bring acrossing existing address data, before deleting the "old" field from the model. If you haven't done this before, it can be a bit tricky to get right. Ping me if you'd like any pointers :)

LavissaWoW:
@SchrodingersGat I've added a preliminary data migration file. (0058) Feel free to look it over

InvenTree/company/tests.py
```diff
@@ -174,6 +176,63 @@ def test_delete(self):
...
+        """Test that there can only be one company-'primary=true' pair"""
+        c2 = Company.objects.create(name='Test Corp2.', description='We make stuff good')
+        Address.objects.create(company=self.c, primary=True)
+        Address.objects.create(company=self.c, primary=False)
+        self.assertEqual(Address.objects.count(), 2)
+
+        # Testing the constraint itself
+        # Intentionally throwing exceptions breaks unit tests unless performed in an atomic block
+        with transaction.atomic():
+            self.assertRaises(IntegrityError, Address.objects.create, company=self.c, primary=True)
```
matmair, line 208: Comment: you could also use the following; there is a whole range of those assertations available
```suggestion
        with self.assertRaises(IntegrityError):
            Address.objects.create(company=self.c, primary=True)
```

InvenTree/company/models.py
```diff
@@ -266,10 +266,19 @@ class Address(models.Model):
...
     """
 
+    def __str__(self):
+        """Defines string representation of address to supple a one-line to API calls"""
+        return ", ".join([self.line1,
+                          self.line2,
+                          self.postal_code,
+                          self.postal_city,
+                          self.province,
+                          self.country])
```
matmair, line 276: Question: How does this handle empty line2 etc?

matmair (approved):
Looks good @LavissaWoW

InvenTree/company/migrations/0058_move_address_field_to_address_model.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('company', '0057_auto_20230502_1956'),
+    ]
+
+    operations = [
+        migrations.RynPython(move_address_to_new_model, reverse_code=revert_address_move)
```
matmair, line 36: Please fix: Typo
```suggestion
        migrations.RunPython(move_address_to_new_model, reverse_code=revert_address_move)
```

InvenTree/company/tests.py
LavissaWoW: I tried your suggestion when working this out and it still made the test fail, as I do some DB transactions after that block.
According to my research (and testing in this case), deliberately triggering an exception within a "TestCase" method will break the method itself if you try to apply any other form of transaction within the same method afterwards. The workaround is to wrap it in the `transaction.atomic()` call as I've done here.

InvenTree/company/models.py
LavissaWoW: That's an edgecase I haven't explored yet. But as far as I see, forms will commit an empty field as "". Concat on an empty string will just result in an empty string entry. In this case, it might indeed be wise to concat line1 and line2 if `line2 != ""` and use the resulting string (or just line1) as input to the join.

InvenTree/company/migrations/0058_move_address_field_to_address_model.py
LavissaWoW: Oh my! Noted!

SchrodingersGat:
Additionally for this we will need;

- A serializer class for the new model
- API endpoint to create and list address instances (and filter by company)
- API endpoint to edit / delete instances
- Unit tests for API endpoints

LavissaWoW:
Thanks for the list, @SchrodingersGat 
There's still a sizeable amount of code to get in here before the feature is anywhere near done.
I wanted to get you guys' review on the data migration file, in case I missed some important details at this stage.

InvenTree/company/migrations/0058_move_address_field_to_address_model.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 3.2.18 on 2023-05-02 20:41
+
+from django.db import migrations
+from company.models import Address
```
SchrodingersGat, line 4: Don't import models this way in a migration file - https://stackoverflow.com/questions/37765307/why-use-apps-get-model-when-creating-a-data-migration

Instead, use apps.get_model (as you have below for Company) e.g.

```python
Address = apps.get_model('company', 'Address')
```

derekwolfson:
I'm just lurking -- this is so great -- love seeing the collaboration!!

matmair:
@LavissaWoW looks like the migrations got shot somehow.

LavissaWoW:
@matmair I forgot renaming a migration in a test :)
On a similar note, tips on how to expose str(address) as company.address for backwards compability? Similar code to look at etc.
That's where the testing here is failing now.

matmair:
@LavissaWoW you could add a property to the model named `address`.

LavissaWoW:
@matmair Thanks, that works. I must've messed something up when I tried the same approach earlier.

matmair:
@LavissaWoW sometimes you can't see the trees in the forests, that is normal when coding.

LavissaWoW:
Additionally, I'm wondering if I should integrate addresses with the 3 types of orders in this PR, or if we should make another one for that.

matmair (approved):

matmair:
@LavissaWoW I think it would be cool if it integrated with all types of orders; not sure if it makes sense in this PR though.

InvenTree/company/models.py
```diff
@@ -156,6 +152,22 @@ def get_api_url():
...
     )
 
+    @property
+    def address(self):
+        """Return the string representation for the primary address
+
+        This property exists for backwards compability
+        """
+
+        addr = Address.objects.filter(company=self.id).filter(primary=True).first()
```
SchrodingersGat, line 162: We can replace this line with `addr = self.primary_address` right?

SchrodingersGat (approved):

LavissaWoW:
Been holding back commits for a little while to get a stable codebase before merging master and sorting potential migration issues and other merge conflicts. Expect a more or less final version within a day or two

LavissaWoW:
address-example
 Any comments?

matmair:
@LavissaWoW you will have to pull in master again but then it should be fine.

.pre-commit-config.yaml
```diff
@@ -41,6 +41,6 @@ repos:
         args: [requirements.in, -o, requirements.txt]
         files: ^requirements\.(in|txt)$
 -   repo: https://github.com/Riverside-Healthcare/djLint
-    rev: v1.28.0
+    rev: v1.29.0
```
wolflu05: you forgot to remove that before committing.

matmair (changes requested):

LavissaWoW:
@matmair Given the following fields:
image
Does the form system support an order of operations like this:
1. Primary address = False
Confirm is hidden
2. Primary Address set to True
Confirm appears, set to False
Perhaps red or other error indicator
3. Confirm Marked as True
Error indicator disappears

One needs to be able to set a new primary address withput unticking the old one. But that needs some form of active confirmation so as to not inadvertently making mistakes

matmair:
@LavissaWoW I think it should be possible, we have somewhat similar things. I am not really a frontend wizz - maybe @SchrodingersGat can give some hints on a suggested order of action.

SchrodingersGat:
@LavissaWoW there are some options here for sure. You can add `confirm_primary` as a serializer field, which means the back-end does the rendering. The field can be hidden on the front-end and made visible when the `primary` field is adjusted, using the `onEdit:` field callback, e.g.

https://github.com/inventree/InvenTree/blob/e21a5e62b8fa49e99e11226152bed393f89abba9/InvenTree/templates/js/translated/purchase_order.js#L271

Does something like this work?

LavissaWoW:
@SchrodingersGat @matmair 
My own tests are going to fail here, but I felt it wise to get the code up to pose my question:
I'm trying to check the `confirm_primary` value in the `pre_save` function, but my tests don't seem to pass it through with the object that's being created. Am I missing something?

matmair:
Should normally be passed, nothing jumps out immediately. There are some missing import statements in the js that need to be fixed (in the header section of the files). The CI was fixed recently so they now started popping up again.

LavissaWoW:
@matmair @SchrodingersGat 
Migration tests are failing in apps I haven't touched. I can't really see why.
Are there still issues with migration ordering that's been mentioned in recent commits?

LavissaWoW:
As soon as I get the PR to pass CI, I'll release it for review and merge

SchrodingersGat:
@LavissaWoW looks like the same issue I have been facing in https://github.com/inventree/InvenTree/pull/4808

Still scratching my head over that one. I'll need to dedicate some time to that again, and additionally now that you are facing the same problem.

SchrodingersGat:
Sorry I don't have an immediate answer for you

LavissaWoW:
No worries. I'll dig my nose into it too since I have a codebase that's having the same symptoms.

SchrodingersGat:
@LavissaWoW I think I've fixed the errors now, however you'll need to update your PR here because there are some conflicting migrations with the master branch

LavissaWoW:
LMK if the Postgres test fail is a problem. 
Same test passes locally, and isn't related to my code.

SchrodingersGat:
@matmair [this CI failure](https://github.com/inventree/InvenTree/actions/runs/5259899543/jobs/9506168595?pr=4732) is the last in the current batch of weird ones. I'm seeing it on multiple PRs, about 50% of the time.

Here's the failing line:

https://github.com/inventree/InvenTree/blob/c8365ccd0c9371ea4d127fe616e0029f35b3c19c/InvenTree/plugin/samples/event/test_event_sample.py#L27

Sometimes, `cm.warning.args[0]` returns an "unclosed file object", rather than a string. Any ideas?

matmair:
@SchrodingersGat this could be related to dulwich - if that is the only error we can ignore it for this PR.

.pre-commit-config.yaml
```diff
@@ -41,7 +41,7 @@ repos:
         args: [requirements.in, -o, requirements.txt]
         files: ^requirements\.(in|txt)$
 -   repo: https://github.com/Riverside-Healthcare/djLint
-    rev: v1.29.0
+    rev: v1.30.2
```
matmair, line 44: Comment: Unrelated CI change

InvenTree/company/models.py
```diff
@@ -253,6 +265,135 @@ def get_api_url():
     role = models.CharField(max_length=100, blank=True)
 
 
+class Address(models.Model):
+    """An address represents a physical location where the company is located. It is possible for a company to have multiple locations
+
+    Attributes:
+        company: Company link for this address
+        title: human readable name for the address
```
matmair, line 273:
```suggestion
        title: Human-readable name for the address
```

InvenTree/company/models.py
```diff
@@ -253,6 +265,135 @@ def get_api_url():
...
+        primary: True if this is the company's primary address
+        line1: First line of address
+        line2: Optional line two for address
+        postal_code: postal code, city and state
```
matmair, line 277:
```suggestion
        postal_code: Postal code, city and state
```

docs/docs/order/company.md
```diff
@@ -43,6 +43,44 @@ The list of contacts associated with a particular company is available in the <s
...
 
+### Addresses
+
+A company can have multiple registered addresses for use with all types of orders.
+An address is broken down to internationally recognised elements that are designed to allow for formatting an address according to user needs.
+Addresses are composed differently across the world, and Inventree reflects this by splitting addresses into components:
+- Line 1: Main street address
+- Line 2: Extra street address line
+- Postal Code: Also known as ZIP code, this is normally a number 3-5 digits in length
+- City: The city tied to the postal code
```
matmair, line 54: Suggestion: In parts of Europe this field is sometimes referred to as Region

matmair (approved):
Well rounded changeset; Good work @LavissaWoW!

docs/docs/order/company.md
LavissaWoW: Referring to City or Province?

matmair: City

InvenTree/templates/js/translated/model_renderers.js
```diff
@@ -173,6 +176,18 @@ function renderContact(data, parameters={}) {
 }
 
 
+// Renderer for "Address" model
+function renderAddress(data, parameters={}) {
+    console.log(data)
```
SchrodingersGat, line 181: @LavissaWoW please remove this debug :)

SchrodingersGat:
@LavissaWoW please address the small outstanding issues and then I think we are good to go :)

InvenTree/templates/js/translated/model_renderers.js
LavissaWoW: Ooh, missed that one.

wolflu05:
Wouldn't it be good to squash all migrations related to this pr into one file?

LavissaWoW:
Not in my knowledge. There is an order of operations that needs to happen correctly.
0063: the new Address model needs to be created.
0064: Then a data migration to create address rows for all currently defined company addresses
0065: Only after this, we can then proceed to delete the address field of the company model.
0066: I'm just not aware of how to squash this one into 0063 tbh. `AlterModelOptions` is the one I'm not sure how to handle.

SchrodingersGat:
Yeah AFAIK data migrations like these explicitly should not be squashed.

wolflu05:
Ok, good to know. That's why i asked.

SchrodingersGat:
@LavissaWoW thanks for this huge contribution :) Please keep it coming, it is greatly appreciated!

PR #4735, opened 2023-05-01
Use coveralls github action

SchrodingersGat:
Attempting to fix recent unknown change which has broken coveralls workflow

Have already tried updating our versions of coveralls / coverage without success:

image


SchrodingersGat:
Ref: https://github.com/lemurheavy/coveralls-public/issues/1708

Issue #4310, opened 2023-02-07
[FR] Pack sizes not considered when adding stock manually


miggland:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

See https://github.com/inventree/InvenTree/discussions/4282

When adding stock manually (not through PO), the pack size of a supplier part is not taken into account.

### Suggested solution

Pack-size should be taken into account

### Describe alternatives you've considered

Minimum: clarify which unit is used

### Do you want to develop this?

- [x] I want to develop this.

miggland:
Fair point - I agree that the case you described is likely to occur.

Being able to choose which to use sounds like a perfect solution (but also more complicated to implement..)

Maybe a checkbox under/next to Quantity, to multiply it by pack size, or not, would solve this (and would also be compatible with the API, if sent as a flag)?

matmair:
still relevant

miggland:
Please assign this to me.

miggland:
This was closed by https://github.com/inventree/InvenTree/pull/4741.

PR #4741, opened 2023-05-01
Add flag to API which allows using pack size

miggland:
Intended for use when adding stock items manually, see #4310 

This adds a flag `use_pack_size` to the stock create API, which when set to True, multiplies the given quantity by the pack size of the supplier part before adding stock.

Additionally, a check that the defined SupplierPart exists is added.


SchrodingersGat:
@miggland I like this approach! The PR will need to address the following:

**Forms** - Ensure that the user is provided a 'use_pack_size' checkbox in forms *if a supplier part is specified*
**Unit Tests** - Will need some unit tests for the various edge cases / etc
**Documentation** - Some docs around this feature will be needed as it is quite subtle and might lead to some confusion

miggland:
@SchrodingersGat: Tests added, and a field is in the form. It's static though, and I need to check that it actually works properly still.

More coming :)

matmair (approved):

miggland:
@SchrodingersGat @matmair : Docs added, tests as well.

Regarding this: "checkbox in forms if a supplier part is specified"
I have added a checkbox after supplier. It is always there, but doesn't have any effect unless there is a pack size != 1 defined. Do you really want it to be dynamic? (I'm not sure how this would be best done - however, it requires more calls to the API, which IMO is unnecessary)

matmair:
@miggland would this PR be ready for review or is there something else missing?

SchrodingersGat:
@miggland sorry for the delayed review on this one.

Looks like the "purchase unit price" does not take pack size into account 

For example if you buy a reel of wire, with a pack size of 500m and a purchase price of $1000 the final unit purchase price for the wire should be $2/m and not $1000/m

Might need a slight tweak to divide purchase price through by pack quantity

matmair:
And maybe a unit test for this case

miggland:
You're right, the issue is fixed. No test (yet)

SchrodingersGat:
@miggland thanks, I'll merge once a simple unit test is added :)

miggland:
I've added tests now.

SchrodingersGat:
@miggland thanks :) Only remaining thing is to increment the API version

miggland:
> @miggland thanks :) Only remaining thing is to increment the API version

Of course, forgot about that :) Done

SchrodingersGat:
Thanks for the contribution @miggland

PR #4743, opened 2023-05-01
Update docker dev docs

LavissaWoW:
* Added section about reusing running containers

* Added information about container cleanup

Related to #4731


LavissaWoW:
Removed section regarding `invoke setup-dev` from CONTRIBUTING.md, as the command is baked into both bare-metal and docker setups now

CONTRIBUTING.md
```diff
@@ -18,23 +18,13 @@ pip install invoke && invoke setup-dev --tests
...
-### Setup Devtools
-
-Run the following command to set up all toolsets for development.
-
-```bash
-invoke setup-dev
-```
-
-*We recommend you run this command before starting to contribute. This will install and set up `pre-commit` to run some checks before each commit and help reduce the style errors.*
-
```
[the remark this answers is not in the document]
LavissaWoW: setup-dev is mentioned in the last step of the bare metal setup, so it seemed redundant now.

docs/docs/start/docker_dev.md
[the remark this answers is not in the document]
LavissaWoW: I've tested this locally, and the --rm command has in my uses only deleted the generated inventree-dev-server container. I welcome someone to verify this ofc.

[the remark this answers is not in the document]
wolflu05: It will delete data if you did something outside the volume mounted folders and copy the commands without thinking about it (one shouldn't do this 😀 )

CONTRIBUTING.md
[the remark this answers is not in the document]
matmair: I would leave it there; we get questions from time to time about this specifically. That way we have a direct link.

docs/docs/start/docker_dev.md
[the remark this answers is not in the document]
matmair: If you start development of a new plugin in /plugins your data will be removed. I would not change the docker instructions - I do not see a lot of added value but the potential for harm.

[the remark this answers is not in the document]
LavissaWoW: The docker server/worker containers mount their volume in the root of the git project. If plug-ins are defined in `../plugins` or any variant of this, then maybe. That's a usecase i haven't tested. 

As it stands, however, there is no way for a dev potentially unfamiliar with the inner workings of docker to comprehend why they suddenly have 35 exited instances of inventree-dev-server. If the `--rm` option isn't desired, I'd like to add a notice regarding this and a potential guide to perform some housekeeping if you're running a larger dev branch with many runs of test, migrate, update etc. Every instance of an image takes up disk space, so we need a define a way to clean up.

CONTRIBUTING.md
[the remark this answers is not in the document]
LavissaWoW: Aah, that makes sense!

docs/docs/start/docker_dev.md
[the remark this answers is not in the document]
matmair: I understand but I am really disliking the thought of ppl. losing hours of work.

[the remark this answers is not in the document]
wolflu05: Maybe I would more likely investigate how we can reuse the existing container via exec, @matmair thoughts?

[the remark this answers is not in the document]
LavissaWoW: I'm able to execute commands in the running container with this: 
`docker exec -it inventree-dev-server bash -c "[commands]"`
The catch is however that venv isn't activated, so you have to source first. The commands get quite long then. I.e:
`docker exec -it inventree-dev-server bash -c "source data/env/bin/activate && invoke migrate"`

However, this also means you can access the bash command line of a running container with: 
`docker exec -it inventree-dev-server bash`
For active development, that's probably the "cleanest" way to do things?
1. Open container via docker exec
2. Source into Venv
3. You now have a persistent CLI in the container to run invoke commands and whatnot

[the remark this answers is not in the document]
matmair: The question is if you need a venv with docker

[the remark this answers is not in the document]
LavissaWoW: Now, _that's_ a decent argument.
It would be a bigger task to remove venv from docker then modifying some doc commands, though I would actually agree that it shouldn't be necessary given the nature of a docker container is to be single-purpose.

SchrodingersGat:
@LavissaWoW you can run `docker container prune` to achieve the same thing - which I think would be a better approach?

Ref: https://docs.docker.com/engine/reference/commandline/container_prune/

LavissaWoW:
Idk. `prune` docs says it removes _every_ container that's not running. It's probably more dangerous to an unwitting user than `--rm` is. (Note, I'm not advocating for `--rm` here)
Depending on your docker state at the time, you might actually delete the "permanent" containers, leading to a need for `docker compose up -d` to fix it. 

I pruned 4 unused containers as a test just now, and it freed 35MB of disk space. Not a lot ofc, but these things pile up. 
I'm gonna revise this PR when I got the time to do some testing.

SchrodingersGat:
What if we just add a note to the docs, about how a user could prune containers if they need to, but without mandating is as patr of our setup procedure?

docs/docs/start/docker_dev.md
```diff
@@ -107,6 +99,17 @@ Now that the database has been created, migrations applied, and you have created
 docker compose up -d
 ```
 
+### Create Admin Account
+
+If you are creating the initial database, you need to create an admin (superuser) account for the database. Run the command below, and follow the prompts:
+
+!!! info "Containers must be running"
```
wolflu05, line 106: Doesn't dc Handels this for you if the containers are marked as depends_on?

LavissaWoW: I've been having a lot of issues getting a superuser if I run `invoke superuser` before the initial `docker compose up -d`
The command works perfectly after the `up` command, but is very flimsy before.

wolflu05: Oh yeah, could be that the db is not yet ready if they are started just before the invoke superuser

LavissaWoW: That's also the reason I moved the superuser command down below the up command. That way, the db will def be running before, if you follow the doc sequentially.

docs/docs/start/docker_dev.md
```diff
@@ -115,6 +118,49 @@ This command launches the remaining containers:
 !!! success "Check Connection"
     Check that the server is running at [http://localhost:8000](http://localhost:8000). The server may take a few minutes to be ready.
 
+## Running commands in the container
+
+Using `docker compoase run [...]` commands creates a new container to run this specific command.
```
matmair, line 123: Typo

docs/docs/start/docker_dev.md
```diff
@@ -115,6 +118,49 @@ This command launches the remaining containers:
...
+```
+You then run the following to access the virtualenv:
+```
+source data/env/bin/activate
+```
+
+This sets up a bash terminal where you can run `invoke` commands directly.
+
+!!! warning "Tests"
+    It is not recommended to run `invoke test` in your currently active inventree-dev-server container.
```
matmair, line 138: Why?

docs/docs/start/docker_dev.md
```diff
@@ -115,6 +118,49 @@ This command launches the remaining containers:
...
+4f8e1b25aeba   inventree-dev-image   "/bin/bash ./docker/…"   3 days ago    Up 3 days   0.0.0.0:8000->8000/tcp   inventree-dev-server
+d674d203ae87   postgres:13           "docker-entrypoint.s…"   4 weeks ago   Up 4 days   5432/tcp                 inventree-dev-db
+```
+
+You should have 1 inventree-dev-worker, 1 inventree-dev-server, and 1 inventree-dev-db in that list.
+If, however, you don't have these, refer to [Start Docker Containers](#Start Docker Containers) and then return here.
+
+```
+docker container prune --filter label="com.docker.compose.project=inventree" --filter label="com.docker.compose.oneoff=True"
+```
```
matmair: Will the recommended commands only touch development containers? Is there a risk that a prod instance is also removed?

CONTRIBUTING.md
```diff
@@ -18,13 +18,23 @@ pip install invoke && invoke setup-dev --tests
...
 
+### Setup Devtools
+
+Run the following command to set up all toolsets for development.
+
+```bash
+invoke setup-dev
+```
+
+*We recommend you run this command before starting to contribute. This will install and set up `pre-commit` to run some checks before each commit and help reduce the style errors.*
```
matmair: Suggestion: pre-commit highlights a lot more than style errors. Maxbe remove `style`

docs/docs/start/docker_dev.md
LavissaWoW: Severely increased test running times. I had one run that lasted 10k seconds 🙈😅

I'll add the reasoning to the doc.

docs/docs/start/docker_dev.md
LavissaWoW: In my testing, `com.docker.compose.oneoff` is only set to `True` on containers that are executed with `docker compose run [image] [cmd]`. The containers running off of `docker compose up -d` have this flas set to `False`

docs/docs/start/docker_dev.md
matmair: I still do not see the harm in it TBH. If it takes time it takes time, you can always kill the test suit.

docs/docs/start/docker_dev.md
matmair: So if somebody starts their prod instance with run (maybe a scheduled task that opens a shell for that - I have seen that) it is also gone? Can we filter for explicitly dev instances? Else there should be a big warning above the commands.

LavissaWoW: Are you able to get me an output from 'docker inspect [containerid]' from a prod setup?

docs/docs/start/docker_dev.md
LavissaWoW: Well, the text works as you'd like then. It's saying that it's not recommended, but we don't say "don't ever do this"

docs/docs/start/docker_dev.md
wolflu05: Im not sure if we can DeTeXt prod/dev 100%, and even if we could, I'm also not sure if we should build such a big oneliner beginners should just copy and paste. And also Dev setups can also contain data that is not in a volume because you tested something. So my recommendation would be to only add this command with a very big disclaimer what can happen.

docs/docs/start/docker_dev.md
matmair: I do not agree with the warning about it. Running a few tests in the container is far faster than waiting for the CI you can specify which command should be run so there is no need to run everything.

SchrodingersGat (approved):

SchrodingersGat:
I'm happy with these changes, @matmair any further comments?

matmair (approved):

SchrodingersGat:
@LavissaWoW ready to merge?

LavissaWoW:
@SchrodingersGat Yes
Just had a small proofread yesterday :)

SchrodingersGat:
Thanks for another great contribution 🙂

PR #4744, opened 2023-05-01
CI cleanup

matmair:
This PR:
- merges the document checks into the main QC file
- removes the social workflow as both vendors would charge for the API now
- rename the QC workflow from "QC checks" to "QC" to make rendering of the workflows better readable on mobile
- adds a check to make sure we only run server code checks if code in InvenTree changed


SchrodingersGat:
Nice addition, the paths filter looks useful

Issue #4670, opened 2023-04-22
Documentation for vscode setup


SchrodingersGat:
### Body of the issue

https://github.com/inventree/inventree-docs/issues/350

wolflu05:
Where should we place that documentation?

SchrodingersGat:
Under the setup / development space, a workflow for "setup a development environment in vscode" would make sense

matmair:
@SchrodingersGat @wolflu05 can we close this with the devcontainer docs now being included?

SchrodingersGat:
Yep, thought we had already closed it!

PR #4761, opened 2023-05-05
Fix various devcontainer issues

wolflu05:
With this changes I'm able to commit with gpg signing and push the changes all from within the devcontainer.

## Things I encountered while setting this up on macOS
- you need to enable full disk access for vscode, otherwise your sshkeys are not available inside the container (https://github.com/microsoft/vscode-remote-release/issues/4024#issuecomment-831671081)
- you need to have `gnupg` and `pinentry-mac` installed and set up correctly [see medium post](https://medium.com/@jma/setup-gpg-for-git-on-macos-4ad69e8d3733)

## Changes
- fix dubious git ownership during postCreateCommand
- fix gpg signing in container
- fix local gitconfig in container
- add gitlens extension to devcontainer
- enable plugins in devcontainer

## How to develop plugins with this setup
1. Mount your plugin repo into the workspace
   ```json
   "mounts": [
     "source=/path/to/your/local/inventree-plugin,target=/workspaces/inventree-plugin,type=bind,consistency=cached"
   ],
   ```
2. Add `/workspaces/inventree-plugin` to your vscode workspace folders
3. Install your plugin as pip editable install
   ```bash
   pip install -e /workspaces/inventree-plugin
   ```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -16,7 +16,7 @@ RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/
 # [Optional] Uncomment this section to install additional OS packages.
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive && \
     apt-get -y install --no-install-recommends \
-    git gcc g++ gettext gnupg libffi-dev \
+    git gcc g++ gettext gnupg2 libffi-dev \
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@wolflu05 thanks, seems pretty reasonable. 

Are you able to add some documentation for setup in devcontainer? Ref: https://github.com/inventree/InvenTree/issues/4670

wolflu05:
Maybe 😆 - I know we definitely need it. I always pushed it infront of me because I thought if we do that, that needs to be handled properly.

PR #4767, opened 2023-05-05
Add option to hide plugin setting from auto-generated plugin setting page

### LINKED ISSUE #4591: [FR] Allow to hide plugin settings

### Problem statement

I want to build my own select component on the settings page of my plugin and handle user input validation myself (Using the settings content mixin). This is already possible thought the `get_setting` and `set_setting` methods, but the input ist still shown on the automatically generated setting page for the plugin. (related to #4546 and wolflu05/inventree-cups-plugin#1)

### Suggested solution

Add an extra `HIDE` parameter to the settings definition in the plugin and check that before rendering here: https://github.com/inventree/InvenTree/blob/0b6e2ee592b885183ee6bf5f70b94a904b0caae3/InvenTree/templates/InvenTree/settings/mixins/settings.html#L12-L14 

*ignore that the input is still shown*

### Examples of other systems

_No response_

- [ ] I want to develop this.

wolflu05:
Please let me know if there is something else where we need to change something and if we should name the new `hide` setting more precise like `hideFromSettingPage`.

fixes #4591


SchrodingersGat:
I like the simple approach but would prefer `hidden` (attribute / adjective) rather than `hide` (action / verb).

SchrodingersGat (approved):

SchrodingersGat:
Nice. Can you find somewhere in the documentation to add a quick note about this feature? Otherwise it will be quite undiscoverable for plugin devs

docs/docs/extend/plugins/settings.md
```diff
@@ -52,6 +52,11 @@ class PluginWithSettings(SettingsMixin, InvenTreePlugin):
                 MaxValueValidator(25)
             ]
         },
+        'HIDDEN_SETTING': {
+            'name': _('Hidden Setting'),
+            'description': _('This setting is hidden from the automatically generated plugin settings page'),
+            'hidden': True,
+        }
```
wolflu05, line 59: @SchrodingersGat That's why I added this. You want me to add a small note somewhere else?

SchrodingersGat: Yeah a small callout near that would be good

```markdown
!!! tip "Hidden Settings"
    Plugin settings can be hidden from the settings page by marking them as 'hidden'
```

wolflu05: Added it. Do you know suggestions code blocks? That way we could just click on apply directly on GitHub.

SchrodingersGat: didn't know that, I'll keep it in mind :)

SchrodingersGat:

PR #4768, opened 2023-05-05
Improve documentation on currency rendering

SchrodingersGat:


.devcontainer/postCreateCommand.sh
```diff
@@ -17,6 +17,6 @@ inv update
 inv setup-dev
 
 # remove existing gitconfig created by "Avoiding Dubious Ownership" step
-# so that it gets copyied from host to the container to have your global 
+# so that it gets copyied from host to the container to have your global
```
wolflu05, line 20: Oh sorry 😃 totally overlooked that.

PR #4774, opened 2023-05-05
fix caluclations->calculations

hasecilu:
Discovered when translating


SchrodingersGat:
Thanks for the contribution 😄

PR #4775, opened 2023-05-05
Add list of pending shipment to home page, with corresponding setting

miggland:
This adds a list of shipments which are not shipped to the homepage.

There is a setting to disable if required.


matmair (approved):
Lgtm! Clean implementation.

SchrodingersGat:
@miggland thanks for the contribution :)

PR #4780, opened 2023-05-05
Restore shopping cart icon in BO/Allocate Parts

sur5r:
The shopping cart icon for parts already on order was originally implemented in 1b421fb59a but got broken in 27aa16d55d as the return value of `makeIconBadge` is discarded.

Additionally, the FontAwesome JS renderer doesn't seem to like non-empty content for this, so even when adding it back to `icons`, it didn't get rendered properly. Instead, the count has to be added to the title.


SchrodingersGat:
@sur5r thanks for the fix, nice catch :)

PR #4784, opened 2023-05-06
Fix metadatalookup

wolflu05:
I'm not really sure where this `Project-URL` comes from, but I get the following error raised without that changes loading my [inventree-bulk-plugin](https://github.com/wolflu05/inventree-bulk-plugin), however my [inventree-cups-plugin](https://github.com/wolflu05/inventree-cups-plugin) loads fine.

image
I investigated that issue by using the debugger and found out after going through the call stack that it gets thrown here.
image

Maybe we should know why my bulk plugin raises that and my cups plugin not, I cannot find any difference in setup.

### DIFF
--- a/InvenTree/plugin/plugin.py
+++ b/InvenTree/plugin/plugin.py
@@ -366,7 +366,7 @@ class InvenTreePlugin(VersionMixin, MixinBase, MetaBase):
 
         try:
             website = meta['Project-URL'].split(', ')[1]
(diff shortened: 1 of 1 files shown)


matmair (approved):

matmair:
@wolflu05 the field is part of the Python packaging standard. Should be populated by any modern packaging tool

Issue #4789, opened 2023-05-07
[FR] Support multiple Label printers within the same label plugin


wolflu05:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

The current label printing system is not very scalable in larger environments with multiple printers of the same brand because each plugin only has the ability to expose one printer.

### Suggested solution

The current system has a huge downside. Every plugin can only provide one printer. I thought very long about how we can fix that system without the downsides listed in alternatives I've already considered. This is my final concept:

## 💡 The idea

1. We add a new `PrintingDriverMixin` which lets the developer define needed settings for **each** (physical) printer. That should work somehow the same way as the current `SettingsMixin`, but is scoped to a specific (physical) printer
2. InvenTree settings get a new Tab `Printers` to manage printers. We should think about a printing plugin more like a provided printing driver than just a function. In that tab we can manage printers. A printer has the following attributes:
   - name
   - driver (select field - plugin that provides that driver)
   - defined config for that specific printer (specified in 1)
   - (scopes)
   - ...
3. We add printers by giving it a name, selecting a driver and fill out the driver specific settings
4. When printing we can choose between all created printers. The printing plugin then gets called with the pdf that needs to be printed and the Printer model which contains the printer specific settings.

This changes the way printing plugins work, so we need some migration period to allow the migrating of the old `LabelPrintingMixin` plugins. 

## 🚀 Future ideas
This enables us much more possibilities in the future:
- printers can be sorted based on priority for a specific label that need to be printed (e.g. that paper box labels are printed with a different printer than labels for magazine boxes)
- printers can be scoped to locations, categories, ... (you're most like not using a printer that is in Fab1 while you are in Fab2)

## 👍 Advantages

- ✅ Easily scaleable for large environments
- ✅ Not every developer has to think about handling multiple printers on themself

### Describe alternatives you've considered

1. Allow plugins to be instantiated multiple times with different settings (Ref: #3568) 
   - ❌ not very durable and can cause many bugs
2. Allow plugins to expose multiple printers
    - ❌ every printer plugin dev needs to handle multiple printers support on their own

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
On the face of it, this approach seems a bit "heavy handed". For discussion, I'd like to propose a simpler alternative.

[...]

Issue #4164, opened 2023-01-06
[FR] Add integration with OpenPnp / LumenPnP


matmair:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

When you run a pick-and-place machine it is hard to keep track of the inventory used by the process.

### Suggested solution

Integrate (likely via a plugin) OpenPnp, the OSS control software for PnPs, so that picked parts are automatically used from a build order.

### Describe alternatives you've considered

Add a plugin that is able to just reduce stock in a location.

### Do you want to develop this?

- [ ] I want to develop this.

matmair:
This is a moonshot, as I currently see some architectural challenges (on the OpenPnP/my knowledge of Java side). Mainly so I have something I can link to.

SchrodingersGat:
I think this would be a very nice plugin to have as a showcase - although I am not familiar with Lumen myself and have not used OpenPNP!

SchrodingersGat:
not stale

wolflu05:
Related to #4824

wolflu05:
Still in development (at least the plugin Mixin), see #4824

matmair:
Not stale, waiting for #4824

matmair:
waiting for machine integration

SchrodingersGat:
Not stale

groutoutlook:
please dont stale

EDIT: well it's fine. it's simply a wishful thinking of mine since it's right after my newly purchase `lumenpnp` that `inventree` come across my mind.
Maybe I need to get myself familiar with both first.

matmair:
Having looked deeper into openpnp I do not think this is going to happen. Either it is a hacky thingy build around openpnp plugins or someone has to bring an event backend into openpnp which seems difficult

PR #4790, opened 2023-05-07
Cleanup plugin mixin registry

matmair:
This PR cleans up the registry mechanisms a bit:
- Adds a central mixin registry
- Discovers mixins dynamically
- Checks all mixins for registration/deregistration functions
- Removes the option to change the built-in order - that part is not documented or used to my knowledge; I found it introducing unreliability when used in container production envs
- Cleans up the registry from vars needed for mixins

### DIFF
--- a/InvenTree/plugin/base/integration/test_mixins.py
+++ b/InvenTree/plugin/base/integration/test_mixins.py
@@ -24,9 +24,9 @@ class BaseMixinDefinition:
     def test_mixin_name(self):
         """Test that the mixin registers itseld correctly."""
         # mixin name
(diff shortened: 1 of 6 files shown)


SchrodingersGat:
@matmair can you expand some more on what problems you are addressing here, or what new features you are adding?

matmair:
@SchrodingersGat reduces the code needed in the registry itself to create new mixins. That way the registry needs to change less and should be more stable with a growing set of mixins.
It also enables the definition of mixins completely outside of InvenTree. That way special - case mixins (like for example my proposal for #4789 or #4164 ) could live outside of core. Distributed via pypi, just like every other plugin.

SchrodingersGat:
Cheers @matmair

PR #4805, opened 2023-05-13
Add option to print multiple labels to a single page

### LINKED ISSUE #4626: [FR]Print multiple labels on a sheet

### Problem statement

I used to print storage labels to pre-cut self-adhesive A4 pages (in 3x8 ordering).
I could print out labels with InvenTree in the necessary size, however they will be added as separate pages.
I know that it might be said that this feature should be the task of the PDF printer too, but in Okular only the CUPS provided options are set (1, 2, 4, 6, 9, and 16) which does not aligned to my needs. Som

### Suggested solution

I would add two parameters to the label templates: 
- number of rows per page
- number of columns per page
- padding in mm

Look for some PDF readers supporting this.

### Examples of other systems

_No response_

- [X] I want to develop this.

Here is a list of the ideas collected over the PR/issue:

- [x] Mention in the docs that fixed positioning might lead to improper positioning in the paginated labels.
- [x] Add option to eliminate adding redundant styles to each rendered label (see. item nr. 4 here: https://github.com/inventree/InvenTree/pull/4805#pullrequestreview-1444267023)
    - A) Add ability to upload the paginated template
    - B) Add option to specify "global style" per label template
- [ ] Add edgecuts marking border feature  (see. item nr. 6 here: https://github.com/inventree/InvenTree/pull/4805#pullrequestreview-1444267023)  
- [ ] Add option to skip first n label see #4929
- [ ] Add configurable border spacing, page margin etc. to support "Avery style labels" (see https://github.com/inventree/InvenTree/pull/4805#issuecomment-1567037541)

martonmiklos:
Closes #4626


InvenTree/label/api.py
```diff
@@ -201,21 +204,97 @@ def print(self, request, items_to_print):
...
+                           % (label.multipage_border, label.multipage_border, label.width, label.height)
+            html_output += '.main-table tr:first-child td {border-top: %s}' % label.multipage_border
+            html_output += '.main-table-row> td:last-child {border-right: %s}' % label.multipage_border
+            html_output += 'body {margin: 0mm !important;}'
+            html_output += '</style></head><table class="main-table"><tr class="main-table-row">'
+            col_counter = 0
+            row_counter = 0
+            item_counter = 0
+
+            for item in items_to_print:
```
matmair, line 247: Question: Wouldn't this be easier with an html template?

SchrodingersGat: Agreed, can you make a template and throw the page data into the template as a list of "pages"?

```html
{% for page in page %}
<!-- Render page -->
{% endfor %}
```

InvenTree/label/api.py
```diff
@@ -201,21 +204,97 @@ def print(self, request, items_to_print):
         label_names = []
         label_instances = []
 
-        # Merge one or more PDF files into a single download
-        for item in items_to_print:
+        if not self.get_object().multipage:
```
SchrodingersGat, line 207: This method is now getting very complicated to read with the different options:

- Plugin printing
- Multi-page printing
- Debug

I'd like to see it broken into separate methods to improve readability and maintainability

InvenTree/label/models.py
```diff
@@ -146,6 +146,37 @@ def __str__(self):
...
 
+    multipage = models.BooleanField(
+        default=False,
+        verbose_name=_('Print multiple labels to a single page'),
+    )
+
+    page_width = models.FloatField(
+        default=210,
+        verbose_name=_('Width of one page [mm]'),
+        help_text=_('The number of label columns per page will be automatically calculated from the label\'s width'),
```
SchrodingersGat, line 157: Using `\'` can make translation strings difficult - please enclose the entire string in double quotes to avoid the slash

InvenTree/label/api.py
wolflu05: Also machine printing will extend this - currently implementing it (ref: #4789 )

matmair (commented):
The code looks clean; there are missing migrations though; you can generate them by running 'inv migrate' on your instance.

martonmiklos:
> The code looks clean; there are missing migrations though; you can generate them by running 'inv migrate' on your instance.

Yeah, I still need to separate the print method as Oliver requested.

InvenTree/label/api.py
martonmiklos: @SchrodingersGat I separated the method to single and multiple labels per page. After some further meditation I still think that the pagination should be done in the plugins (if they are involved).

wolflu05:
Please don't force push, this will make review much harder, as references break. All prs will be squash merged anyway, so history doesn't matter.

.pre-commit-config.yaml
```diff
@@ -41,6 +41,6 @@ repos:
         args: [requirements.in, -o, requirements.txt]
         files: ^requirements\.(in|txt)$
 -   repo: https://github.com/Riverside-Healthcare/djLint
-    rev: v1.25.0
+    rev: v1.29.0
```
matmair: Request: Please change this back; the changed rulesets could introduce unintended consequences. This should be a separate PR.

matmair (changes requested):

matmair:
@martonmiklos why is absolute positioning not possible? Couldn't we wrap the labels into 0-width containers that provide absolute position within?

.pre-commit-config.yaml
wolflu05: Can't we do something to prevent this from auto updating the number when setting up the devenv?

martonmiklos: Sorry my bad, I should have take extra care before commit. To be honest I do not know how did this changed.

wolflu05: This automatically updates when you run the setup dev script (the pre commit hook get installed on your machine) - was mainly a question to @matmair if he is aware of any solution to prevent it from changing the version if we only install the pre commit hook. Also happens several times to me.

matmair: Not bumping the versions on the versions lead to problems with yanked releases in the past. I would expect a dev to check what they submit to git more easily then diagnose yanked releases.

wolflu05: Isn't this more like a CI task than a task that should run on every dev machine? I personally think this is more annoying than helpful. Also didn't quite understand your point.

martonmiklos:
> I think the key cas setting is position: relative for the wrappers, then the inner labels which are positioned absolut use the origin of the relative container.

Well that is true, I fixed it by adding a wrapper div:
kép

wolflu05:
Do we really need this extra div? Doesn't this also work with the class applied to the td's?

martonmiklos:
> Do we really need this extra div? Doesn't this also work with the class applied to the td's?

That's true, it looks to be working fine, code updated.

InvenTree/label/admin.py
```diff
@@ -1,15 +1,87 @@
...
+                  'page_width', 'page_height', 'multipage_border']
+
+    pagesize_preset = forms.ChoiceField(
+        choices=[
+            ('custom', _('Custom')),
+            ('210x297', _('A4')),
+            ('297x420', _('A3')),
+            ('148x210', _('A5')),
+        ],
+        label=_('Preset paper size'),
```
matmair, line 27: Question: I am using metric but does it make sense to add 1 or 2 common American sheet sizes?

matmair (approved):
LGTM, great work!

InvenTree/label/admin.py
martonmiklos: Sure we can, what do you think which one would make the most sense? Letter+legal?

matmair: I am EU-based and have never used American formats but a quick search suggests letter, legal and maybe tabloid might make sense.

SchrodingersGat: Certainly letter and legal would be good options to add in.

InvenTree/label/templates/label/multipage_label_base.html
```diff
@@ -0,0 +1,49 @@
...
+
+<head>
+    <style>
+        {% localize off %}
+        @page {
+            size: {{ page_width }}mm {{ page_height }}mm;
+            margin-top: 0mm;
+            margin-bottom: 0mm;
+            margin-left: 0mm;
+            margin-right: 0mm;
```
wolflu05, line 11: Is there a reason why you provide them individually?

InvenTree/label/models.py
```diff
@@ -146,6 +146,37 @@ def __str__(self):
...
+    page_height = models.FloatField(
+        default=297,
+        verbose_name=_('Height of one page [mm]'),
+        help_text=_("The number of label rows per page will be automatically calculated from the label's height"),
+        validators=[MinValueValidator(2)]
+    )
+
+    multipage_border = models.CharField(
+        choices=[
+            ('0.5mm solid #000;', _('Solid')),
```
wolflu05, line 170: isn't the `;` at the end to much, as you also have it in the css template?

wolflu05 (changes requested):
Hello @martonmiklos ,
thank you for your work here. This is a very good addition and I also like that this is implemented directly in inventree core. This opens much doors also with machine integration (#4824). But I found a few bugs/had a few ideas while testing. Sorry that this list is so long. I really appreciate your work.

1. I think ci test failure is related 
2. The `.qr` css class in all preshipped templates contains a `position: fixed` setting which does not work as fixed ignores the relative positioning.
   - `InvenTree/label/templates/label/part/part_label_code128.html`
   - `InvenTree/label/templates/label/part/part_label.html`
   - `InvenTree/label/templates/label/stockitem/qr.html`
   - `InvenTree/label/templates/label/stocklocation/qr_and_text.html`
   - `InvenTree/label/templates/label/stocklocation/qr.html`
   - `InvenTree/report/templates/report/inventree_report_base.html`
3. I actually discovered, that `label_base.html` already adds a content wrapper with relative positioning to the label which causes some errors. You need to at least at the following to the `.content` class: 
   ```css
            top: 0;
            left: 0;
            height: 100%;
   ```
   otherwise labels look like this: 
image
4. I'm not sure if each label should have its own same style rendered, this is very weird for debugging in browser devtools. What do you think @matmair ? See: 
image
5. It would be good to have some way of controlling the border spacing. It seems like this is actually not possible with borders, but doable with linear gradient backgrounds. See [this post](https://stackoverflow.com/questions/6250394/how-to-increase-space-between-dotted-border-dots)
6. What about an edgecuts marker feature. This would only add small lines on the paper edges where to cut, instead of the whole paper. That way label would look cleaner on the edges as there is no border.
7. It would be cool to have some way of defining some kind of min spacing, so that labels aren't printed to close to the edges of the paper
8. I discovered an issue if label printing is not set to debug mode, so weasyprint is used to generate a PDF. The `@page` selector in the `label_base.html` template is interfering with weasyprint and the complete page is only that big. But even if I comment that part out temporarily, it still does not work properly with absolute positioning, as the `div.content` from the `label_base.html` is somehow misplaced even the changes added from 3). See (I decreased the size of the QR codes in this picture): 
image

matmair:
good work @martonmiklos and good review @wolflu05; my 2 cents:
4) moving it to a class makes more sense IMO. But there might be a reason this way was chosen
5) feels like a hack. This would need to be explained in docs or comment as it is really not self-explaining
6/7) good idea but possibly a future PR?
8) there is a border margin option in weasyprint (like requested in 7) - might that be it?

InvenTree/label/migrations/0010_auto_20230502_1646.py
```diff
@@ -0,0 +1,76 @@
+# Generated by Django 3.2.19 on 2023-05-24 20:24
+
+import django.core.validators
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    replaces = [('label', '0010_auto_20230502_1646'), ('label', '0011_auto_20230510_1939'), ('label', '0012_auto_20230523_2151')]
```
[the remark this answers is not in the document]
matmair: Seems to be resolved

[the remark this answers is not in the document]
martonmiklos: Yeah, I did a migration squash which produced this line. I fixed it since then and it made the QC pass.

InvenTree/label/api.py
```diff
@@ -271,14 +324,34 @@ def print(self, request, items_to_print):
...
+        debug_mode = common.models.InvenTreeSetting.get_setting('REPORT_DEBUG_MODE', cache=False)
+
+        """ When using print plugin and multiple labels per page option the
+            label data passed the same way as the single page per label to the plugin
+            to allow the "pagination" to be performed by the plugin itself (based on the supplied page&label geometry)
+        """
+        if self.get_object().multipage and plugin is None:
+            return self.print_multiple_per_page(request, items_to_print, debug_mode)
+        else:
+            return self.print_single_label_per_page(request, items_to_print, debug_mode, plugin)
```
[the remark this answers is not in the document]
martonmiklos: Well my idea/motivation between with this was the following: the plugins could have better understanding how the pagination can be done on the printer they are operating. For e.g. we have a Brady label printer which is capable to print  to per-perforated cable marking vinyls. But maybe I am wrong and passing paginated html/PDF is a better approach in most cases.

InvenTree/label/migrations/0010_auto_20230502_1646.py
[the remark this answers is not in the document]
matmair: Pretty straightforward; The migrations were quashed and then the referenced files were deleted but the squash command was left in this file,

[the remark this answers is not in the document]
wolflu05: @matmair i checked, none of these files were ever in my django_migrations table. I didn't cloned before. As far as I understand this doesn't run the migration of those other migration already run, or do I misunderstand something?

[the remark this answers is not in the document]
matmair: I think there is a misunderstanding about how migrations work. Django builds up a migration path from the code and then checks with the database what was already done. If there is an error in the migration structure it affects you - whether you ran them before or not; you never reach that point.
You can read more [about migration files here](https://docs.djangoproject.com/en/4.2/topics/migrations/#migration-files) and [squashing here](https://docs.djangoproject.com/en/4.2/topics/migrations/#squashing-migrations).

SchrodingersGat:
@martonmiklos does the current version work for you? I just pulled latest commit and it only prints a single lable instance, even if multiple are selected. It also does *not* print to full-page (i.e. A4)

SchrodingersGat:
Additionally I think we may need some more parameters to be able to support sheets of label stickers e.g. https://www.averyproducts.com.au/labels/shape/rectangle

Note the many and varied border sizes available: 
image

SchrodingersGat:
As per suggestion by @wolflu05 I removed the `@page` spec from base label template and now it seems to be working:

image

SchrodingersGat:
With regard to the "avery" labels above (and there are many others):

Should we expose even more configurable parameters to the label templates, to allow for such complexity? Or, should we consider how we could expose multi-label-printing to the plugin interface? That way, someone could write an "avery label plugin" for InvenTree which handles all their templates, and a different plugin for a different label manufacturer.

This way, we only need to consider the w/h dimensions of the individual labels themselves, and not the interactions between labels and pages.

We could implement what you have done above as a "built-in" plugin which supports multiple labels. Then, in the "print labels" dialog, you select "multi page printing" as the printer plugin target. Instead of throwing the labels at an external printer it returns a compiled .pdf object.

Thoughts?

matmair:
I think the prudent option is to get a minimal MVP going and mergend and then implement a way to use these labels.

martonmiklos:
> With regard to the "avery" labels above (and there are many others):
> 
> Should we expose even more configurable parameters to the label templates, to allow for such complexity? Or, should we consider how we could expose multi-label-printing to the plugin interface? That way, someone could write an "avery label plugin" for InvenTree which handles all their templates, and a different plugin for a different label manufacturer.

Well my use case was to print a bunch of stock location labels to a self-adhesive A4 sheet, cut with scissors and call it a day.

But as you an others  highlighted above (here and in the #4626 comments) there are many use cases and features which could make sense to implement.

I think the multi page label printing  should be quite common use case in the user base so properly supporting this with the different use cases (for e.g. "Avery style" label) would add to the traits of the InvenTree.

But yeah it might be useful to negotiate on a baseline feature set, review and merge it, then collect all the ideas into a new issue/issues properly discuss them and implement them in a second round.

SchrodingersGat:
@martonmiklos @matmair fair points, let's get this merged and then work on next round. I was going to look at the label / report printing API again soon anyway, but was holding off for this PR :)

wolflu05:
@SchrodingersGat Please also consider that the machine integration will build on the label system in sometime.

What is your opinion on https://github.com/inventree/InvenTree/pull/4805#discussion_r1205794950 ? Would need that for my machine integration.

SchrodingersGat:
> What is your opinion on https://github.com/inventree/InvenTree/pull/4805#discussion_r1205794950 ? Would need that for my machine integration.

I *think* it makes sense, but I'm concerned that the machine integration is becoming.. complex..

wolflu05:
Yeah, machine integration is definitely not that simple, but I think its a good approach to handle all refactors in separate PRs so it will be as easy as possible.

InvenTree/label/templates/label/multipage_label_base.html
martonmiklos: No reason, I am just dumb, I am going to fix this.

roman-dvorak:
This is a great feature! Thanks for implementing it.


I didn't see it mentioned earlier or in commits, but when you are printing labels on a sheet of stickers, it's useful to have the option to skip the first few labels. When I need to print only, let's say, 5 labels out of 40 (whole sheet), it results in large waste. If I can't reuse the rest of the sheet. So it would be great if the label generation process allows skipping the first x labels, and use rest of the sheet.

matmair:
@roman-dvorak Thanks for your input! Could you please open a new issue for this (using the feature request template)?

We are pretty heavy in the feature-creep phase now. I think it is much more efficient to get things asap into master and have more PRs. @wolflu05 I agree that multiple refactor PRs are the way to go

InvenTree/label/models.py
martonmiklos: I put the border styling between if - endif so it is not even going to be rendered. I also changed the db default values.

InvenTree/label/models.py
```diff
@@ -146,6 +146,37 @@ def __str__(self):
...
+            ('0.5mm dotted #000;', _('Dotted')),
+            ('', _('No border')),
+        ],
+        default='border: 0.5mm solid #000;',
```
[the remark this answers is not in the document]
wolflu05: Actually I meant the border: prefix.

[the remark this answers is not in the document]
martonmiklos: Dauh, I have not seen from my eyes the problem ;)

martonmiklos:
> 4) moving it to a class makes more sense IMO. But there might be a reason this way was chosen

I also noticed this, the HTML indeed looks ugly this way, but I have not found better way to overcome this.

I was thinking about adding a "global style" (or something) similar field to each label type which would be rendered once into the multipage_label_base.html but I think it might overcomplicate the thing. (I mean for a paginated label printing you would need to provide a template, a global stylesheet, dimensions etc...)

martonmiklos:
> As per suggestion by @wolflu05 I removed the `@page` spec from base label template and now it seems to be working:
> 
> image

Now I feel that one of the biggest issue with the current state of the PR is the fact that it does not work with the "factory supplied templates".

How should we deal with this?

A) document these limitations maybe providing a sample template which supports the paginated printing?
B) figure out something to extract/alter the label templates when paginated print enabled?

I know B is more attractive, but at the moment I do not have good idea how to parse/alter CSS embedded into HTML.

roman-dvorak:
> > 4. moving it to a class makes more sense IMO. But there might be a reason this way was chosen
> 
> I also noticed this, the HTML indeed looks ugly this way, but I have not found better way to overcome this.
> 
> I was thinking about adding a "global style" (or something) similar field to each label type which would be rendered once into the multipage_label_base.html but I think it might overcomplicate the thing. (I mean for a paginated label printing you would need to provide a template, a global stylesheet, dimensions etc...)

For inspiration... Our labels currently look like this:


This design looks very effective. We use an A4 sheet with 40 labels.
image

From the implementation of these labels, I have noticed that it is better to implement each dimension (label type) separately because layout is a bit different to show all necessary/required data. Simply adjusting the "width/height" setting of the label is not enought.

matmair:
I appreciate the effort but in the future, I think it would be better to create a new issue instead of changing the scope of the original one that is already being developed. This way it becomes unclear to new readers what the original intent was.

Content-wise I think everything is there.
If other ideas pop up they should be filed as a separate issue. The more we pack into this FR the longer it takes to get it merged - which means merging upstream and fixing migrations and API docs another few rounds.

wolflu05:
What is actually the current plan with moving this into a plugin? Was this just an idea or some planned refactor in the future?

martonmiklos:
> I appreciate the effort but in the future, I think it would be better to create a new issue instead of changing the scope of the original one that is already being developed. This way it becomes unclear to new readers what the original intent was.
> 
> Content-wise I think everything is there. If other ideas pop up they should be filed as a separate issue. The more we pack into this FR the longer it takes to get it merged - which means merging upstream and fixing migrations and API docs another few rounds.


I do not want to change the scope of this PR I just wanted to capture all the ideas which came up as comments/review comments in one place first etc. We can separate them easily to separate issues and remove that section.

martonmiklos:
> What is actually the current plan with moving this into a plugin? Was this just an idea or some planned refactor in the future?

Well it was one of the first questions/concerns:
https://github.com/inventree/InvenTree/issues/4626#issuecomment-1512235509
 and I decided it to go directly into InvenTree.

I generally like the plugins approach, however I think it could be a quite common use case to print labels with standard printers on standard paper sizes, so supporting it out of the box would add much to the traits of InvenTree.

wolflu05:
Actually I was more referring to: https://github.com/inventree/InvenTree/pull/4805?notification_referrer_id=NT_kwDOBJR1D7M2NDQ4NTU2NDg1Ojc2ODM4MTU5#issuecomment-1567063885

InvenTree/label/api.py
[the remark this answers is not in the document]
wolflu05: @SchrodingersGat @matmair what is your opinion on this thread? And what is the current state of this PR, I would need it for #4824 .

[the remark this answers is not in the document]
martonmiklos: > And what is the current state of this PR

My right hand is now in rail (not in cast) so I have limited typing ability. I rebased this PR onto master, added support for the Buildline labels (introduced in the meantime). I also improved the docs about avoiding the position: fixed in multipage labels. I think if we get agreement on the way we should handle the plugins then we could move forward with this.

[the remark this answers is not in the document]
SchrodingersGat: @martonmiklos good to hear you are on your way to recovery! There has been some discussion about cleaning up the current implementation to make this (and other plugin / machine features) more streamlined. I am currently working on a label printing architecture update which will impact this PR. Please hold off on any work on this until I've got this up and running (should be under a week). Keep resting :)

SchrodingersGat:
@martonmiklos please check out the refactor in https://github.com/inventree/InvenTree/pull/5251

The idea with the new plugin system is to radically simplify the API, and make *everything* handled by the plugin. The new system defines a "builtin" plugin for label printing, which provides the out-of-the-box PDF generation.

If you want to print multiple plugins to a single page, it will be now very easy. Implement as a plugin, and override the `print_labels` method. The plugin mixing has a number of helper methods which would make this really simple now..

If / when the refactor gets merged in, let's circle back and have a chat about how this will make your life much easier here. We could implement your process as an official plugin quite easily.

SchrodingersGat:
@martonmiklos https://github.com/inventree/InvenTree/pull/5251 is merged in now, so your goals here should be *significantly* easier to implement. Sorry to throw a spanner in the works, but LMK if you want any pointers regarding working within the new framework.

martonmiklos:
> @martonmiklos #5251 is merged in now, so your goals here should be _significantly_ easier to implement. Sorry to throw a spanner in the works, but LMK if you want any pointers regarding working within the new framework.

Thanks for the heads up, I will start rebasing and reworking the PR.

martonmiklos:
>  LMK if you want any pointers regarding working within the new framework.

I started to work on it, butt got stuck a bit: the InvenTreeLabel plugin looks to be loaded fine:
kép

However when I try to print a label the plugin parameter got undefined:

URL: /api/label/location/1/print/?location=87&plugin=undefined

How can I make the new builtin plugin default?

SchrodingersGat:
You should get the option when selecting the labels to print:

image

And you should also be able to set the default printer in the user settings page:

image

martonmiklos:
Dah, I had the plugins_enabled set to False in the setup.cfg.


Maybe it would be useful to change the builtin plugins active mode this case?
kép

SchrodingersGat:
> Maybe it would be useful to change the builtin plugins active mode this case? 

Builtin plugins are always enabled:

image

wolflu05:
Any updates on this? What was the last state?

SchrodingersGat:
@wolflu05 @martonmiklos sorry I have been very swamped with work / life - hoping to get back to this soon.

SchrodingersGat:
@martonmiklos apologies for leaving this one hanging. There has been a *lot* of activity which has distracted me from this, but I do want to get this implemented for our next major release (`0.13.0`).

After discussion offline with @matmair and @wolflu05 I would like to change the direction of this a bit. With some of the recent refactoring of the label printing functionality, we have now an opportunity to leverage the great basework you have done here, but basically implement it as a plugin. I'll explain in some further detail:

The "nuts and bolts" of what you hav implemented here is fantastic. However rather than placing all of the extra code right in the core "label printing" ecosystem I would like to see if become a plugin. The workflow would be as follows:

1. The user selects a number of items to print labels for
2. The user selects a label template, which defines how each *individual* label is printed
3. The user selects a label "printer" - in this case, the "multi label page plugin"
4. The new plugin takes all the labels, renders them against the template, and then combines them onto a single page

The recent feature by @wolflu05 to allow custom "printing options" to be specified at run-time by plugins helps us a lot here too - the user can select page size, number of labels per sheet, which ones to skip, etc.

The advantage here is that the base label printing mechanism does not need to know anything about how to combine labels into a single page. It remains simply a mechanism for rendering a simple template for printing a single label. The logic for "combining" is purely handled by the plugin.

We can make this a "builtin" plugin which provides some generic array functionality as you have here - most of your code would be able to be migrated to this approach I think.  Further down the line, similar plugins could even be written which provide the correct array geometry for rendering against third-party sticker sheets, etc.

Again, sorry for the huge delay and then change of direction!!

SchrodingersGat:
Replaced by https://github.com/inventree/InvenTree/pull/5883

PR #4804, opened 2023-05-13
Add 'project_code' field to Build model

SchrodingersGat:
Replaces https://github.com/inventree/InvenTree/pull/4646

Adds a "project code" reference to the build order model

PR #4808, opened 2023-05-14
Add ProjectCode support to build orders 

SchrodingersGat:
(take 3)

replaces https://github.com/inventree/InvenTree/pull/4804


SchrodingersGat:
```
django.db.utils.OperationalError: table "common_currency" already exists
```

@matmair any idea why this error keeps popping up? Three different iterations of the PR and all comes back to this error message.

matmair:
Interesting error. I had that myself a time ago. Maybe strip down the PR to the smallest possible diff, look if it works, and go from there? The diff is so large that it is probably hard to debug.

SchrodingersGat:
> Interesting error. I had that myself a time ago. Maybe strip down the PR to the smallest possible diff, look if it works, and go from there? The diff is so large that it is probably hard to debug.

Ref: https://github.com/inventree/InvenTree/pull/4814

Issue #3012, opened 2022-05-16
[FR] Allocate "untracked stock" to build order


SchrodingersGat:
### Current Behaviour

When "untracked" stock items (e.g. fungible parts such as capacitors, connectors, screws, etc) are allocated to a build order, the required stock quantity is depleted when the build is completed. The removed quantity is currently *deleted* from the database, and there is no way of tracking which batches of stock were consumed to fulfil a particular BuildOrder

### Proposed Change

When a BuildOrder is "completed", any allocated stock items which are not *tracked* are marked as "consumed by" the build. The stock item data remains in the database, but those stock items are not available for further use.

This allows us to go back over time and look at stock consumption information for a particular build order.

### Considerations

- Adds a new field on a StockItem to indicate it has been consumed by a BuildOrder
- When the BuildOrder is deleted, any `consumed_by` items must also be deleted
- Add unit testing

matmair:
@SchrodingersGat I think an option to 'recycle' already consumed parts would be also good.

mattbnz:
How about just using the existing StockItemTracking model with enhanced metadata for this, as is already done with purchase orders? 

E.g. putting a reference to the build order into the tracking entry that is created when the item is consumed, that seems more flexible than a field on the stock item itself, since it allows the stock item to be partially consumed by multiple builds - I think if you added a field on the stockitem itself, you'd have to split the item each time a build order consumes some of it, etc.

mattbnz:
See what you think of the commit above - I won't open a pull request yet, just an experiment for now... but if you like the approach I can  tidy it up and open a request.

matmair:
@mattbnz that pretty much matches the intent with the metadata in that model. Looks great!

Feel free to just open draft PRs if you are working on stuff. That way everyone knows and we can use gitpod.io to easily test the system in a production like environment.

SchrodingersGat:
@mattbnz I like the idea and your approach is clean, and I think that it forms part of what is needed here.

> I think if you added a field on the stockitem itself, you'd have to split the item each time a build order consumes some of it, etc.

Yep, you absolutely would have to do this. The idea here does present more "work" for the system, but it allows better traceability through the system. Say you need to find out where a reel of dodgy capacitors ended up.. Currently, the stock is just invisibly "consumed" (once it is used up, it is deleted). The new approach does not delete the stock, it splits into a new database entry which is assigned to the build.

mattbnz:
[...]

PR #4817, opened 2023-05-15
Build consume stock

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/3012

Untracked stock items are now "consumed" by a build order, rather than being "deleted" from the database.

### TODO

- [x] Unit testing
- [x] Documentation


docs/docs/terminology.md
```diff
@@ -6,23 +6,27 @@ title: Terminology
 
 There are different systems in the industry for the management of getting, storing and making parts. An overview what they are for and what the acronyms mean.
 
+**InvenTree** is mainly focused on [**IMS**](#inventory-management-system-ims) and [**PLM**](#part-library-management-plm) functionality.
```
matmair, line 9: Comment: The scope is growing ;-)

InvenTree/InvenTree/tasks.py
```diff
@@ -563,6 +563,16 @@ def send_email(subject, body, recipients, from_email=None, html_message=None):
...
 
+    import InvenTree.ready
+
+    if InvenTree.ready.isImportingData():
+        # If we are importing data, don't send emails
+        return
+
+    if InvenTree.ready.isInTestMode():
+        # If we are running in test mode, don't send emails
+        return
```
matmair, line 574: Question: Should we perhaps test if email delivery is working somewhere?

InvenTree/build/templates/build/detail.html
```diff
@@ -337,11 +360,9 @@ <h4>{% trans "Build Notes" %}</h4>
             build: {{ build.id }},
             is_building: false,
         },
-        groupByField: 'location',
```
matmair, line 340: Question: Will this change the default behavior and is that intended?
Suggestion: If yes: Should this PR be marked as breaking?

InvenTree/stock/models.py
```diff
@@ -755,6 +756,14 @@ def get_part_name(self):
         related_name='build_outputs',
     )
 
+    consumed_by = models.ForeignKey(
+        'build.Build', on_delete=models.CASCADE,
```
matmair, line 760: Question: Is it intended to delete the stock items completely when the build is deleted?

matmair (commented):
Nice addition; Does it make sense to update the selection dropdowns with a separate field for consumed items?

InvenTree/InvenTree/tasks.py
SchrodingersGat: I'm going to revert this particular line anyway, it's breaking CI :)

InvenTree/build/templates/build/detail.html
SchrodingersGat: No, stock-item grouping has long been removed! This is just a cleanup of legacy code

InvenTree/stock/models.py
SchrodingersGat: Yes, that's correct.

The logic is:

1. Currently, untracked items are removed from the database (deleted) once consumed by a build
2. New behaviour is that the untracked items are marked as "consumed_by" - but kept in the database
3. The "consumed_by" field is what tells the database that these items are unavailable for consumption
4. When the build is deleted, I think the "cleanest" thing to do is to remove those items (as opposed to returning them to stock)

Open to alternative suggestions though :)

SchrodingersGat:
> Nice addition; Does it make sense to update the selection dropdowns with a separate field for consumed items?

Do you mean the filters in the stock tables? Or something else?

SchrodingersGat:
I don't think that the user should manually mark the stock items as "consumed by", or have the ability to do this via the front-end modals.

The stock items can always be manually edited in the admin interface

Issue #4209, opened 2023-01-13
[FR] Introduce optional physical units to parameter input


matmair:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

Currently, you have to enter the value of a parameter always in the same unit, which is mostly okay - except when someone in your team uses non-SI measurement systems. Then you start calculating.

### Suggested solution

Integrate [pinet](https://github.com/hgrecco/pint) to facilitate a unit conversion. This would also enable setting metric or imperial in the user settings.

- [ ] API endpoint to convert from any unit to any
- [ ] Add switch to parameter to enable base unit saving
- [ ] Make the value of parameter optionally unit-safe
- [ ] Add dynamic input to the interface with unit-diverse input
- [ ] Opt: Detect unit from test
- [ ] Opt: Add babel to include translations of units
- [ ] Opt: Define custom units
- [ ] Opt: Add template tag to do unit transformation in reports

### Describe alternatives you've considered

N/A

### Examples of other systems

I know partskeepr could do something similar. The 'big-boy' PLMs have this integrated as well.

### Do you want to develop this?

- [X] I want to develop this.

matmair:
still intrested

EvilMustacheTwirl:
So there are a couple of django projects that do a lot of the base work but they seem to be gathering dust so I'm not sure if they would still work. 

https://github.com/coddingtonbear/django-measurement

https://github.com/bitlabstudio/django-units

It might be easier to just copy the currency conversion and make it a unit conversion then users can add the units and conversions themselves.

matmair:
@EvilMustacheTwirl thank you for your input. I would prefer pinet as it is used by a lot of EEs and MEs in academia so the conversions should be right. I do not really have time for this right now so open for an FR.

The currency conversion is provided by an external dependency.

SchrodingersGat:
@matmair I'd love to implement this for https://github.com/inventree/InvenTree/pull/4823

Parameters currently have "units" specified, but they don't really mean anything. It would be great to be able to set real physical units, say "metres" but values could be put in as `mm` or `cm` (and correctly converted for sorting / filtering)

matmair:
Nice!

Issue #3080, opened 2022-05-26
[FR] Parameter filtering 2.0


eeintech:
## 0.6.4 stable

![image](https://user-images.githubusercontent.com/4020546/170566376-3e4db1d7-1870-45b6-978c-fa1b252f454c.png)

## 0.7.0 stable

![image](https://user-images.githubusercontent.com/4020546/170566942-0c1d1b16-c35d-41cb-a097-ed5b0f652054.png)

## Bug description

See screenshots above, I would like to get back to the `0.6.4` table filter settings if possible, as I use this feature often :smiley:

matmair:
@eeintech seems to be a regression rather than a bug.

eeintech:
@matmair https://en.wikipedia.org/wiki/Software_regression

"A software regression is a type of [software bug](https://en.wikipedia.org/wiki/Software_bug) where a feature that has worked before stops working."

Anyway happy to label it otherwise in the title, lemme know

SchrodingersGat:
@eeintech this was probably broken in https://github.com/inventree/InvenTree/pull/2761 which moved the part parameter table to use the API

The 'table filter' function required all the data to be loaded into the table when the table is initialized. For large datasets this made the page take dozens of seconds to load (which is not tenable). So, we moved to using the API, and all was right with the world.

However, I did not realize that this "filter" feature relied on the old way of doing things.

If you want this feature back, we'll have to work out a way of filtering part parameters via the API.

This is something I've wanted to do for ages, to be able to say (for example) "give me all bits of wood which have a length between 20mm and 200mm"

Until we have that system in place, this table will be sadly lacking in any filters.

SchrodingersGat:
Parameter filtering will be a super powerful feature, will take some thinking to get it implemented though :)

matmair:
@SchrodingersGat I think we should try to get that working with the general filter setup

SchrodingersGat:
Absolutely!

eeintech:
Changed the title as while being a regression this is now oriented towards a better implementation!

SchrodingersGat:
@eeintech check out some of the work going on in https://github.com/inventree/InvenTree/pull/4823

eeintech:
@SchrodingersGat Really cool, thanks!

PR #4823, opened 2023-05-15
Parameter filtering

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/3080
- Closes https://github.com/inventree/InvenTree/issues/4209

This PR adds a number of key improvements to the concept of "part parameterrs"

### TODO

- [x] ~~Allows different "types" of parameter templates to be defined~~ (will implement later)
- [x] ~~Validates entered values against specified template type~~ (will implement later)
- [x] Add physical units to parameters - https://github.com/inventree/InvenTree/issues/4209
- [x] Automatically convert to base units based on SI units
- [x] Calculates and stores numerical value of parameter value (for efficient sorting)
- [x] Data migration to calculate "numerical value" for existing parameters
- [x] Unit test for above data migration
- [x] Background task to re-calculate parameter values if a template changes
- [x] Performs sorting of parts by parameter value in table
- [ ] ~~Performs querying of parts by parameter value in table (e.g. `eq` / `lt` / `gt` )~~ - moved to https://github.com/inventree/InvenTree/issues/4851
- [x] Documentation
- [x] Updates to demo dataset with new physical parameters - https://github.com/inventree/demo-dataset/pull/45


SchrodingersGat:
@matmair should we migrate the existing "units" field (which is just a string, no functional value) to a physical unit?

InvenTree/part/admin.py
```diff
@@ -122,9 +122,9 @@ class Meta(PartResource.Meta):
         ]
 
 
-class StocktakeInline(admin.TabularInline):
-    """Inline for part stocktake data"""
-    model = models.PartStocktake
```
matmair, line 127: Question: Is there no need for the old inline anymore?

InvenTree/part/models.py
```diff
@@ -3295,6 +3296,9 @@ class PartParameterTemplate(MetadataMixin, models.Model):
     Attributes:
         name: The name (key) of the Parameter [string]
         units: The units of the Parameter [string]
+        description: Description of the parameter [string]
+        type: The data type of the Parameter [integer]
+        choices: A comma-separated list of choices for the Parameter [string]
```
matmair, line 3301: Question: Where are type / choices defined?

matmair (commented):
Very cool development;

Some ideas/discussion points:
- I think it could be beneficial to be able to deactivate all calculations on a field
- It could be interesting to mark a part parameter as locked. Ie. the user can not change the parameter value - only plugins can
- An API endpoint or query string that returns a parameter in a specific unit would be nice
- This PR assumes we use SI; That should be documented somewhere

matmair:
> @matmair should we migrate the existing "units" field (which is just a string, no functional value) to a physical unit?
> 
> Edit: I've started to dive into using [pint](https://pypi.org/project/Pint/) and looks like I have a good path forward for dev

We can certainly try; not sure if this would lead to errors in non-English languages. Maybe we could add a button in the settings and prompt all superusers?

InvenTree/part/admin.py
SchrodingersGat: IMO no, it has been outdated since the recent stocktake updates

InvenTree/part/models.py
SchrodingersGat: @matmair I'm going to move this into https://github.com/inventree/InvenTree/issues/4835

InvenTree/build/migrations/0013_auto_20200425_0507.py
```diff
@@ -53,5 +49,5 @@ class Migration(migrations.Migration):
             field=models.PositiveIntegerField(db_index=True, default=0, editable=False),
             preserve_default=False,
         ),
-        migrations.RunPython(update_tree, reverse_code=nupdate_tree),
+        migrations.RunPython(update_tree, reverse_code=migrations.RunPython.noop),
```
matmair, line 52: Comment: Nice, I did not know that

matmair (approved):

InvenTree/build/migrations/0013_auto_20200425_0507.py
SchrodingersGat: New to me also :)

eeintech:
@SchrodingersGat Is it supposed to work on the demo version (engineer user)? I am seeing a few problems:

* Parameters tabs did not filter parts by the category selected
*  Only part name shows as there is no option to add parameters in new columns
* 
image

* Only common filters are available, no parameters in the list

image

I am not sure where to look for this new feature if not there?

SchrodingersGat:
@eeintech you might have tried the demo server before the new version was loaded:

image

> Parameters tabs did not filter parts by the category selected

This looks like a bug, I'll look into it

second-string:
I'm a bit confused, and maybe just misunderstanding, but I still cannot properly filter on parameter values. Both on my own server, and on the demo server, if I search for text that's in one of the parameter columns (but NOT in the part name/description), then those parts don't come up.

For example, on the demo server if you search for 'canvas' on the Parameters page, only one result comes up that has 'canvas' in the name. However there are multiple parts that have canvas as a parameter value. Clicking the sort arrow on a column properly sorts, but I can't figure out how to filter down by specific  parameter values.

There are also no params available in the filter dropdown, as eeintech showed in their screenshot above.

SchrodingersGat:
@second-string the parametric table can currently be "sorted" by parameter value, but not filtered.

With regards to searching by parameter value - this is a good addition but currently the search is only against the parts, not the parameter values. I'll raise a new issue to look into that.

Issue #4589, opened 2023-04-06
[FR] Make default stock status settable through settings page


standekker:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

I am testing with Inventree and noticed that the default status of a new stock item is always `OK`. For my use case, it would be great if we can set the default stock state through the settings page. For me, it would set it default `Attention Needed` so that every new stockpart can be checked at a later date.

This may be related to #2817, but as I read it #2817 request adding a new, expanded flow. I would be helped by just being able to set the default stock status through the settings page, which is why I am making the Feature Request.

### Suggested solution

Being able to set the default stock state through the settings page.

### Describe alternatives you've considered

Manually setting the stock status to something else than `OK` upon creation of stock.

### Do you want to develop this?

- [X] I want to develop this.

matmair:
@standekker there are several things regarding status - not only for stock. The big goal is #2817 but there are also #2903 #3501 and possibly #4289. There is some logic coupled with the stock status - so we would need to decouple that. An example could be how Snipe-It [handles it](https://snipe-it.readme.io/docs/overview#status-labels).

standekker:
Gotcha, I wasn't aware of #2903, #3501 and #4289, those also all sound promising! I have played with Snipe-It before and I do like how they handle it. Would this feature request still be something you would be willing to add (or approve if I can add it myself) or is this something that would be quickly superseded by #2817, #2903, #3501 or #4289 and thus not worth the trouble?

matmair:
It is probably simpler and faster to fulfil this FR now and add the other things later if/once we rewrite status in general.
There is no active dev on the mentioned issues except #3501, which I am testing in prod right now.

matmair:
Not stale

standekker:
Not stale

matmair:
Not stale

Issue #4289, opened 2023-02-01
Custom Stock Status


Fluxus00:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

It would be more user friendly if you were able to customize your stock status options. As an example I would like to use a status like "tested", "in production"  or similar ones. The current status options are not enough for me to use them properly.

### Suggested solution

Enable the option to create custom status options in the admin menu.

### Describe alternatives you've considered

Extending the amount of predefined status options.

### Do you want to develop this?

- [ ] I want to develop this.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/inventree) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/inventree/InvenTree/issues/4289">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/inventree/InvenTree/issues/4289/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/inventree/InvenTree/issues/4289/pledge.svg">
</picture>
</a>

gunstr:
When I migrated to InvenTree a year back I had similar need as @Fluxus00 primarily because I have different reasons than what is hard coded to why stock should be treated as available and more important what should be included in overall inventory valuation and not. My first solution was to anyway use the status codes but with a note on my desk to translate between what I saw on the screen and the actual status.

This was obviously error prone why I now in my local installation have changed the hard coded values in the [StockStatus ](https://github.com/inventree/InvenTree/blob/c6df0dbb2d44de013421299e1296769a36cb72d3/InvenTree/InvenTree/status_codes.py#L166) class. It works well for my use cases, only little drawback is that the status codes are obviously not propagated the app.

I have seen FR's for the same being raised before why there is probably more people being interested. But as @matmair mention a proper solution comes with so much more and unfortunately I have not yet managed to build enough competence in the architecture to feel comfortable to volunteer for a full implementation. Maybe later this year when I have retired from my full time work...

Just a thought - given the complexity of the current status code, would a "tag" field with no business logic impact and with configurable values be an option?

matmair:
@gunstr ping me or @SchrodingersGat here if you want some pointers how this could be done. There are many ways to achieve the requested functionality.
[...]

PR #4824, opened 2023-05-15
Machine integration

### LINKED ISSUE #4789: [FR] Support multiple Label printers within the same label plugin

wolflu05:
This PR adds the machines feature described in #4789 .

image
image

fixes #4789 

---

```[tasklist]
# 📝 Todos
- [x] 🚚 driver errors
- [x] 🚚 handle error function for machines and registry
- [x] 🚚 Init driver function
- [x] 🚚 Machine update hook function (only partially working)
- [x] 🚚 Machine has unapplied changes (restart required)
- [x] 🚚 Machine restart hook
- [x] 🚚 get_machines should by default only return initialized machines
- [x] 🚚 funtion to check if all required settings are defined
- [x] 🚚 Machine States (Ready, Connecting, …) (waiting on #4857)
- [x] 🚚 export everything required from the plugin app
- [x] 🚚 get all machines, can be maybe used in combination with event mixin
- [x] 🚚 get_machines function for driver to return all machines belonging to that driver
- [x] 🚚 separate machine type and base driver
- [x] 📫 API GET available machine types / drivers
- [x] 📫 API CRUD machines
- [x] 📫 API CRUD machine settings
- [x] 📫 API Permissions
- [x] 💄 PUI Settings Page
- [x] 💄 PUI machine list: table
- [x] 💄 PUI machine list: registry errors
- [x] 💄 PUI machine list: add machine
- [x] 💄 PUI machine detail: view
- [x] 💄 PUI machine type table with detail drawer: showing machine type info and list available drivers and Which plugin provides that type
- [x] 💄 PUI machine driver drawer with some info about it
- [x] 💄 PUI driver drawer error stack + show all machines
- [x] 🎰 LABEL PRINTING machine type
- [x] 🎰 LABEL PRINTING: print_label
- [x] 🎰 LABEL PRINTING: print_labels
- [x] 🎰 LABEL PRINTING: location setting (useful for basic location scoping)
- [x] 🎰 LABEL PRINTING: get_printers (implement at driver level, useful for filtering the printers based on certain hardware aspects)
- [x] 🎰 LABEL PRINTING: number of copies printing option
- [x] 🎰 LABEL PRINTING: include request in get_printers kwargs
- [x] 🎰 LABEL PRINTING: remember last used printer machine per user
- [x] ✨ Add Machine Settings and separat from driver settings (Machine settings could be used for e.g. printer location)
- [x] ✨ Machine active
- [x] ✨ Sync machine registry with creating/updating/deleting of machines
- [x] 🚒 dont initialize if machine is not fully configured (error handeling)
- [x] 🚒 Check if machine initialized without error (error handeling)
- [x] 🚒 Check if driver implements all required functions (error handeling)
- [x] ♻️ refactor to mixin for override detection (add/check required configuration to classes, required_{attributed,overrides})
- [x] ♻️ refactor machine pk should be UUID
- [x] ♻️ refactor rename base_driver to machine_type
- [x] ♻️ refactor move check into own check function isPlgRegistryLoaded (Prevent MachineConfig.ready to be executed twice)
```

```[tasklist]
# ⚠️ Before merge
- [x] check `TODO:` comments
- [x] fix style (precommit)
- [x] remove debug messages
- [x] Remove badly working django admin interface for machine integration
- [x] Cups plugin providing a fully working example driver (https://github.com/wolflu05/inventree-cups-plugin/pull/6)
- [x] Docs
- [x] Tests
- [x] bump api version
```

```[tasklist]
# ♻️ External PRs/Refactors
- [ ] https://github.com/inventree/InvenTree/pull/5005
- [ ] https://github.com/inventree/InvenTree/pull/5251
- [ ] https://github.com/inventree/InvenTree/pull/4834
- [ ] https://github.com/inventree/InvenTree/pull/4927
- [ ] https://github.com/inventree/InvenTree/pull/5224
- [ ] https://github.com/inventree/InvenTree/pull/5786
- [ ] https://github.com/inventree/InvenTree/pull/5837
```

```[tasklist]
### 💡Further ideas (Not this PR)
- [ ] label printing queue
- [ ] machine status polling function on driver (with option to disable status polling for all printers/for an individual printer)
- [ ] label printer dialog sort machines based on priorities provided as tuple with `get_printers` (not sure if that makes sense or if last used is just enough)
- [ ] machine permissions with custom permission levels per machine type (e.g. so that only specific roles/users can print labels with a specific printer - for label printers there is a permission level "print")
- [ ] Show the printer status badge in the machine selection dropdown (maybe by extending the ChoiceField with an optional status code property)
- [ ] Settings only required if conditions are met (e.g. for a usb/network printer select the host, port, user, password only makes sense for network printers, for usb printer a single select is enough.)
- [ ] Live Autocomplete Settings type key (Client posts current values to API and API responds with suggestion)
- [ ] Termination hook for machines, (already tried at 6bb0ba719bef6f61f4395823f2ff8ee8f7097a3e but not that easy when run in gunicorn), then the machine restart could be a simple terminate + re init
- [ ] add metadata to machine model
- [ ] How does scaling of the machine registry work if inventree runs on multiple nodes
- [ ] PnP machine type (#4164)
- [ ] 3d printer machine type (#5191)
- [ ] let driver auto discover physical machines in the network/computer and show them so that they can be added quickly (like homeassistant has it)
- [ ] Define icons for each machine type and show them along the UI
- [ ] machine actions (buttons on machine page that can trigger certain function on the driver)
- [ ] Issue plugin events on machine load/finish load (ref: https://github.com/inventree/InvenTree/pull/4824#issuecomment-1574104681) (Already tried, but: enqueues to much offloaded events that are unnecessary on simple setups on startup => the event system first needs a fix)
- [ ] send notification on error status change or machines capture errors 
- [x] Relation type settings key (Automatic drop down for selecting related field) (fixed with #4944)
```


InvenTree/common/models.py
wolflu05, line 1: @matmair I think this needs to be changed so that this model is more generic and doesn't require code for each different use case here.

matmair (approved):

InvenTree/machine/models.py
```diff
@@ -0,0 +1,146 @@
...
+
+from django.contrib import admin
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+import common.models
+from machine.registry import registry
+
+
+class Machine(models.Model):
```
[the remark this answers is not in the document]
wolflu05: Can you expand that idea a bit more?

InvenTree/machine/registry.py
```diff
@@ -0,0 +1,84 @@
...
+    def load_machines(self):
+        # Imports need to be in this level to prevent early db model imports
+        from machine.models import Machine
+
+        for machine in Machine.objects.all():
+            self.machines[machine.name] = machine
+
+        # initialize machines after all machine instances were created
+        for machine in self.machines.values():
+            machine.initialize()
```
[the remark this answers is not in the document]
wolflu05: Haven't thought a about failing the initialization yet, but I thought maybe one would need to get all machines for each initialization via  the driver. So I thought first adding all and then init them.

wolflu05:
@matmair Do you think the Base Drivers should be exported from the machine or plugin app? *(The notification adapters defined in common are exported through the plugin app)*

InvenTree/common/models.py
matmair: Not really sure how much we can do here regarding duplication, the settings systems have changed a lot in the last year. A clean rewrite could be interesting - especially when looking at adding typing and making it easier to extend.

InvenTree/machine/driver.py
```diff
@@ -0,0 +1,52 @@
+from typing import Dict
+
+
+class BaseDriver:
```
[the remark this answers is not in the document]
matmair: Then that needs to throw an error and not define. We will need testing around misdefined drivers, machine types, and machines anyhow.

InvenTree/machine/models.py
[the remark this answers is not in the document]
matmair: incrementing primary keys are a great setup for [IDOR](https://en.wikipedia.org/wiki/Insecure_direct_object_reference) problems. Therefore I prefer uuids for stuff we will probably expose to external systems. Also, the reference for objects is always the same length with uuid4.

matmair:
@wolflu05 if you are not against it I would suggest it. The idea is that there always is a simple and steady path for the things we expose for plugins while we are free to refactor the inner workings and structure without breaking plugins.

InvenTree/machine/__init__.py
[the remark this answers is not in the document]
wolflu05: @matmair With this suggested change, I get the following error:

<details>

```
(venv) vscode ➜ /workspaces/InvenTree (feature/machines) $  cd /workspaces/InvenTree ; /usr/bin/env /workspaces/InvenTree/dev/venv/bin/python /home/vscode/.vscode-server/extensions/ms-python.python-2023.8.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher 49417 -- /workspaces/InvenTree/InvenTree/manage.py runserver --noreload 
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/vscode/.vscode-server/extensions/ms-python.python-2023.8.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher/../../debugpy/__main__.py", line 39, in <module>
    cli.main()
  File "/home/vscode/.vscode-server/extensions/ms-python.python-2023.8.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 430, in main
    run()
  File "/home/vscode/.vscode-server/extensions/ms-python.python-2023.8.0/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 284, in run_file
    runpy.run_path(target, run_name="__main__")
  File "/home/vscode/.vscode-server/extensions/ms-python.python-2023.8.0/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 321, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/home/vscode/.vscode-server/extensions/ms-python.python-2023.8.0/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 135, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/home/vscode/.vscode-server/extensions/ms-python.python-2023.8.0/pythonFiles/lib/python/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 124, in _run_code
    exec(code, run_globals)
  File "/workspaces/InvenTree/InvenTree/manage.py", line 23, in <module>
    execute_from_command_line(sys.argv)
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/core/management/__init__.py", line 395, in execute
    django.setup()
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/apps/registry.py", line 91, in populate
    app_config = AppConfig.create(entry)
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/apps/config.py", line 212, in create
    mod = import_module(mod_path)
  File "/usr/local/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/workspaces/InvenTree/InvenTree/machine/__init__.py", line 1, in <module>
    from machine.registry import registry
  File "/workspaces/InvenTree/InvenTree/machine/registry.py", line 4, in <module>
    import InvenTree.helpers
  File "/workspaces/InvenTree/InvenTree/InvenTree/helpers.py", line 15, in <module>
    from django.contrib.auth.models import Permission
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/contrib/auth/models.py", line 3, in <module>
    from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/contrib/auth/base_user.py", line 48, in <module>
    class AbstractBaseUser(models.Model):
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/db/models/base.py", line 108, in __new__
    app_config = apps.get_containing_app_config(module)
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/apps/registry.py", line 253, in get_containing_app_config
    self.check_apps_ready()
  File "/workspaces/InvenTree/dev/venv/lib/python3.10/site-packages/django/apps/registry.py", line 136, in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
```

</details>

I guess this is because it tries to import that file while discovery where Django is not ready and it imports the registry which imports Inventree.helpers which then tries to import models which are not ready. Should I either move the inheritors function into an own helpers file in `machines` or should we import the models at function level in the inventree helpers?

wolflu05:
I'm planning on add the ability that machines can expose their current states. E.g. for a label printer: Ready, Printing, Disconnected, ... do you think this makes sense to define per Machine Type or per CustomMachineDriver? (Want to have a proper Interface with Enums).

Per MachineType has the ability that they are equal across all label printers. However something custom like "out of labels", "machine opened" ... is not possible. I guess there are more different states on other machine types like PnP. What about having both abilities? MachineType specific states and a machine status that's just a simple string?

InvenTree/machine/__init__.py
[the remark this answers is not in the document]
matmair: @wolflu05 The cleanest would be to refactor the registry code and move the import down into the class

matmair:
@wolflu05 status/state, in general, needs a bit of a refactor (see #4289, #4589). To keep this PR somewhat my suggestion would be to add enums to the machine type and enable drivers to define current state (the enum from machine) and state text (free text, by default from machine type state but can be set to "anything")

InvenTree/machine/models.py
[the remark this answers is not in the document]
wolflu05: Would you suggest using the uuid as a pk or as an additional field? And is the uuid replacing the integer `id` field?

InvenTree/machine/models.py
```diff
@@ -0,0 +1,150 @@
...
+    def __str__(self) -> str:
+        """String representation of a machine."""
+        return f"{self.name}"
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Override to set origignal state of the machine instance."""
+        super().__init__(*args, **kwargs)
+
+        self.errors = []
+        self.initialized = False
```
wolflu05, line 50: What is the best way to store such state as these model instances are recreated everytime I use them and therefor this state is not preserved.

matmair: I am not sure if the state should be saved to the database. What if the device changes while the instance is down? I think the state should be processed on initialization of the machine and the be dynamic.

wolflu05: No, this is no model field, doesn't this work with Django to have some fields that are stored only during runtime?

InvenTree/machine/models.py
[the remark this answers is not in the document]
matmair: I would use the uuid as the pk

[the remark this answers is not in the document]
wolflu05: So you would store this state in a dict in the registry and load it each time the Machine model gets instantiated?

InvenTree/machine/registry.py
```diff
@@ -0,0 +1,142 @@
...
+        Kwargs:
+            name: Machine name
+            machine_type: Machine type definition (class)
+            driver: Machine driver (class)
+            active: (bool, default: True)
+            base_driver: base driver (class | List[class])
+        """
+        allowed_fields = ["name", "machine_type", "driver", "active", "base_driver"]
+
+        kwargs = {**{'active': True}, **kwargs}
```
matmair, line 117: Question: is there a reason to not put this into the function definition with a default?

InvenTree/machine/machine_type.py
```diff
@@ -0,0 +1,91 @@
...
+
+        errors = []
+
+        if len(missing_attributes) > 0:
+            errors.append(f"did not provide the following attributes: {', '.join(missing_attributes)}")
+        if len(missing_overrides) > 0:
+            errors.append(f"did not override the required attributes: {', '.join(map(lambda attr: attr.__name__, missing_overrides))}")
+
+        if len(errors) > 0:
+            raise NotImplementedError(f"The driver '{cls}' " + " and ".join(errors))
```
matmair, line 49: Nice approach; Could we factor this out in a base that is used both by BaseDriver and BaseMachineType? We have similar things for most mixins too.

matmair (approved):

InvenTree/machine/registry.py
wolflu05: Because then the generic loop does not work that way?

InvenTree/machine/models.py
matmair: @wolflu05 the idea would be to do several steps on initialisation:
- load all database objects for machines
- issue 'machine_registry_init_start' event
- iterate over the objects
  - init machine from machine type with db object
  - here is the initial state set on the machine object
  - run a machine setup method to give it the option to discover "real" state
  - then save the machine object (not db ref) to the registry
- issue 'machine_registry_init_end' event

InvenTree/machine/machine_type.py
wolflu05: @matmair Where should that mixin be placed? Using `InvenTree/mixins.py` doesn't work, because in that class `Inventree.helpers` are imported that import models which throws an `AppRegistryNotReady` issue if I import my `ClassValidationMixin` in this file.

InvenTree/machine/models.py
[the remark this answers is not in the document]
wolflu05: @matmair Is the current implementation ok so? (595ccd6a2a9cb662db18bf6bbdd6aa83330bf83f) Do you have concerns about performance as the uuid is now also stored in the machine settings model as foreign key. Or should we have an additional Integer pk that is used for the foreign key and only expose the unique uuid field?

InvenTree/machine/registry.py
```diff
@@ -0,0 +1,155 @@
...
+
+        discovered_machine_types: List[BaseMachineType] = InvenTree.helpers.inheritors(BaseMachineType)
+        for machine_type in discovered_machine_types:
+            try:
+                machine_type.validate()
+            except NotImplementedError as error:
+                self.errors.append(error)
+                continue
+
+            machine_types[machine_type.SLUG] = machine_type
```
matmair, line 47: Should we check for slug duplication?

InvenTree/machine/registry.py
```diff
@@ -0,0 +1,155 @@
...
+            base_drivers.append(machine_type.base_driver)
```
matmair, line 48: Where is the driver validated in this case?

InvenTree/machine/registry.py
```diff
@@ -0,0 +1,155 @@
...
+        if initialize:
+            machine.initialize()
+
+    def get_machines(self, **kwargs):
+        """Get loaded machines from registry. (By default only active machines)
+
+        Kwargs:
+            name: Machine name
+            machine_type: Machine type definition (class)
+            driver: Machine driver (class)
```
matmair, line 119:
```suggestion
            driver: Maschine driver (class)
```

matmair (approved):

InvenTree/machine/machine_type.py
matmair: It is a class (but not model) mixin so probably in InvenTree/helpers - I am a bit surprised that models are imported on top level. Maybe move it into a separate file "registry_helpers" or something like that.

InvenTree/machine/registry.py
wolflu05: They don't need to, they are only the base drivers. Actually they won't validate without an error, because the only implement an abstract implementation. I only collected this list to subtract them from all inharitants.

InvenTree/machine/machine_type.py
wolflu05: Im not sure if this is long or term the best solution. Had this situation often that incant import helpers directly. Maybe we should refactor the helpers in a separate pr to import the models on function level? Also this ClassValidationMixin could be used for current plugins.

wolflu05:
@matmair I'm currently trying to get a termination hook running for machines analog to the initialize hook. I tried to do this in some djangonic way. I found [this](https://stackoverflow.com/a/20646978) example using Django signals. However this does not work really. If I add `--noreload` to the call args it works, but if multiple processes are spawned for reloading it directly exits. I expect the following log output:
```
Quit the server with CONTROL-C.
^C### After sleep
AFTER SIGNALS SEND
```

I found some ENV that is set if this is currently the main thread, but this is not set if `--noreload` is applied so I added a argv check, but this feels like a workaround. Do you have a better idea?

matmair:
@wolflu05 not really. I have never had to deal with the need for a shutdown signal. Your implementation looks fine to me.

InvenTree/machine/apps.py
```diff
@@ -12,11 +13,25 @@ class MachineConfig(AppConfig):
...
+
+    def on_shutdown(self, **kwargs):
+        # gracefully shutdown machines
+        from machine import registry
+
+        logger.debug("Shutting down InvenTree machines")
+        registry.terminate()
+
+        import time
+        time.sleep(3)
```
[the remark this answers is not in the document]
wolflu05: Just a debug statement for testing the termination code. Left this in so that I can ensure it didn't terminate before this code run through.

wolflu05:
I think the issue is not the signal part, but the Django threading part. Django starts two threads and I didn't get it to work with both (with and without --noreload).

matmair:
TBH my understanding of Django Threading is pretty limited, especially since they introduced async support.

wolflu05:
I have something working which works with runserver now, but I'm not sure if this is good, as this does not work with gunicorn. To get it to work there too I would need to override the gunicorn config with a function on_exit which wouldnt be a clean solution. What do you think?

matmair:
@wolflu05 as we ship the gunicorn config (I assume you are referring to the .py file) as part of core I do not see a problem in overriding/changing it.

wolflu05:
We would need one central config, I saw that there are multiple, for docker, beside the manage.py and in the docs folder.

InvenTree/machine/machine_type.py
wolflu05: @matmair Does importing the models at function level have any impact on performance? If so what about two helper files? One that contains model code and one that doesn't?

matmair: Not really, there is a very very small difference. Split sounds like a good idea.

matmair:
@wolflu05 there shouldn't be any difference between them - maybe we need the way we reference them so we can merge to one.

matmair:
@wolflu05 there shouldn't be any difference between them - maybe we need the way we reference them so we can merge into one.

InvenTree/machine/machine_type.py
wolflu05: Ref: #4927

wolflu05:
@matmair Which plugin events exactly do you imagine to be triggered during the machine registry lifecycle?

matmair:
@wolflu05 I thought about something like:
- event before loading anything into the registry
- an event before every machine
- an event after every machine
- an event after all machines are loaded
- an event if machines are activated
- an event if machines are deactivated
- (if termination is implemented) an event before and after machine termination

wolflu05:
Do you already have some use cases for these events in mind? I mean there is already a function that can be defined by the driver that gets called for the majority of events. But good idea to also provide a machine update hook function.

I cannot imagine where an machine specific event would make sense in the plugin.

matmair:
@wolflu05 the idea is to be able to intersect certain steps and change stuff around a specific foreign code section. A use case would be to change loaded versions for an init sequence so you can monkey patch plugins.

InvenTree/InvenTree/ready.py
```diff
@@ -60,3 +73,21 @@ def canAppAccessDatabase(allow_test: bool = False, allow_plugins: bool = False,
...
+    """
+    from django.conf import settings
+
+    # If plugins are not enabled, there wont be a second load
+    if not settings.PLUGINS_ENABLED:
+        return True
+
+    from plugin import registry
+
+    return not registry.is_loading
```
wolflu05, line 93: Do you think that has some unwanted behavior if we include that everywhere where we have that canAppAccessDatabase also the isInMainThread Check from above in this file?

matmair: The only area that could be problematic would be tasks (as they always run in the background). Would probably be an improvement overall

wolflu05: Should I add those two functions to all other app configs or should I move these two checks to the canAppAccessDatabase function?

(Sure I'll do this in a separate pr)

matmair: I would keep it as a separate function but add it to all configs.

wolflu05: But should I combine the isInMainThread and isPluginRegistryLoaded function?

matmair: I would keep them separate, they address different questions. Is `isMainThread` working now?

wolflu05: Oh right, have to check if isMainThread works first?

wolflu05:
@matmair currently trying to implement the states thing. Do you have an idea why I cannot import the StatusCode class like this? I didn't see why I get the `AppRegistryNotReady` error. 

<details>

```patch
diff --git a/InvenTree/machine/machine_type.py b/InvenTree/machine/machine_type.py
index e54af75fa..afbfb1a7d 100644
--- a/InvenTree/machine/machine_type.py
+++ b/InvenTree/machine/machine_type.py
@@ -1,6 +1,7 @@
 from typing import TYPE_CHECKING, Dict, Literal, Type
 
 from InvenTree.helpers_mixin import ClassProviderMixin, ClassValidationMixin
+from generic.states.states import StatusCode
 
 # Import only for typechecking, otherwise this throws cyclic import errors
 if TYPE_CHECKING:
@@ -14,6 +15,11 @@ else:  # pragma: no cover
         pass
 
 
+class MachineStatus(StatusCode):
+    """Base class for representing a set of machine status codes"""
+    pass
+
+
 class BaseDriver(ClassValidationMixin, ClassProviderMixin):
     """Base class for machine drivers
 
@@ -57,6 +63,7 @@ class BaseMachineType(ClassValidationMixin, ClassProviderMixin):
         DESCRIPTION: Description of what this machine type can do (default: "")
 
         MACHINE_SETTINGS: Machine type specific settings dict (optional)
+        MACHINE_STATUS: Set of status codes this machine type can have (optional)
 
         base_driver: Reference to the base driver for this machine type
     """
@@ -66,6 +73,7 @@ class BaseMachineType(ClassValidationMixin, ClassProviderMixin):
     DESCRIPTION: str
 
     MACHINE_SETTINGS: Dict[str, SettingsKeyType]
+    MACHINE_STATUS: MachineStatus
 
     base_driver: Type[BaseDriver]
```

</details>

wolflu05:
Seems like it's because in the `__init__.py` file the something is imported that imports something from the drf. I would suggest refactoring the generic package so that we can import StatusCode without expecting that apps are ready.

wolflu05:
The following patch would solve this problem, and I think it would be generally good to separate the pure stats abstraction from any drf things.

<details><summary>Patch to fix AppRegistryNotReady</summary>
<p>

```patch
diff --git a/InvenTree/build/api.py b/InvenTree/build/api.py
index 63ae43215..0feaf2463 100644
--- a/InvenTree/build/api.py
+++ b/InvenTree/build/api.py
@@ -11,7 +11,7 @@ from django_filters.rest_framework import DjangoFilterBackend
 from django_filters import rest_framework as rest_filters
 
 from InvenTree.api import AttachmentMixin, APIDownloadMixin, ListCreateDestroyAPIView, MetadataView
-from generic.states import StatusView
+from generic.states.api import StatusView
 from InvenTree.helpers import str2bool, isNull, DownloadFile
 from InvenTree.status_codes import BuildStatus, BuildStatusGroups
 from InvenTree.mixins import CreateAPI, RetrieveUpdateDestroyAPI, ListCreateAPI
diff --git a/InvenTree/generic/states/__init__.py b/InvenTree/generic/states/__init__.py
index 5137b7c3f..952c31557 100644
--- a/InvenTree/generic/states/__init__.py
+++ b/InvenTree/generic/states/__init__.py
@@ -6,10 +6,8 @@ There is a rendered state for each state value. The rendered state is used for d
 States can be extended with custom options for each InvenTree instance - those options are stored in the database and need to link back to state values.
 """
 
-from .api import StatusView
 from .states import StatusCode
 
 __all__ = [
-    StatusView,
     StatusCode,
 ]
diff --git a/InvenTree/order/api.py b/InvenTree/order/api.py
index b4a60a95a..99f04c5b2 100644
--- a/InvenTree/order/api.py
+++ b/InvenTree/order/api.py
@@ -16,7 +16,7 @@ from rest_framework.response import Response
 import common.models as common_models
 from common.settings import settings
 from company.models import SupplierPart
-from generic.states import StatusView
+from generic.states.api import StatusView
 from InvenTree.api import (APIDownloadMixin, AttachmentMixin,
                            ListCreateDestroyAPIView, MetadataView)
 from InvenTree.filters import SEARCH_ORDER_FILTER, SEARCH_ORDER_FILTER_ALIAS
diff --git a/InvenTree/stock/api.py b/InvenTree/stock/api.py
index 3011e2e9d..e1345113c 100644
--- a/InvenTree/stock/api.py
+++ b/InvenTree/stock/api.py
@@ -22,7 +22,7 @@ from build.models import Build
 from build.serializers import BuildSerializer
 from company.models import Company, SupplierPart
 from company.serializers import CompanySerializer
-from generic.states import StatusView
+from generic.states.api import StatusView
 from InvenTree.api import (APIDownloadMixin, AttachmentMixin,
                            ListCreateDestroyAPIView, MetadataView)
 from InvenTree.filters import (ORDER_FILTER, SEARCH_ORDER_FILTER,
```

</p>
</details>

wolflu05:
@matmair you can see my current implementation in 758550de56231e4f4a5d7de39bc53d65d8dd6483 , I cherry picked 88a8e0 to this branch, because without I get the AppRegistryNotReady error. Im also open for a different fix.

wolflu05:
Not stale, just waiting for:
1. the PUI framework to be ready
2. some kind of model framework for the printing options

wolflu05:
Still wip

wolflu05:
@SchrodingersGat @matmair regarding the printing dialog, do we want to:

1. first select between InvenTree pdf, InvenTree sheet and InvenTree machines and then use the printing options to show another select box to select the actual machine. The machine's printing options could then be lazy loaded using an dependent field (I already implemented this serializer type upstream in Python and cui, not in pui and not in the app)
2. or have the machines selectable in the exact same select as we can also decide between InvenTree pdf, ... and then forward the printing options 1:1 to the driver

One would now be easier to implement and is way more decoupled from the InvenTree code. However I feel like implementing option 1 sounds like bad UX. However I could see option too also implemented very decoupled if we add another `get_printers` method to the printing plugin Mixin, that also gets passed some context (like labels to print, location, ...) in. Then we could use that to return all available printers which then builds a list based on the return value of all plugins. But that would be a bit breaking.

Please let me know what you think about this.

SchrodingersGat:
I think 2) makes more sense to me. Select the printer "driver" first, (pdf, sheet, brother, zebra, etc). Then each driver (plugin) provides additional options (such as "select machine"). Most of this is already implemented via the custom printer options dialog API.

matmair:
Maybe it would make sense to also add a default printing setting so that you do not need multiple clicks to get a label printed

wolflu05:
> I think 2) makes more sense to me. Select the printer "driver" first, (pdf, sheet, brother, zebra, etc). Then each driver (plugin) provides additional options (such as "select machine"). Most of this is already implemented via the custom printer options dialog API.

That would be a third option, a combination between one and two. But each driver shouldn't have something todo with manually adding a machine select field, then we would be again like we are now. That should be handled by the machine framework. Do you think it makes sense to first select the brand with which we try to print and then the printer? I imagine a setup with only three printers, two zebra and one connected through cups. I then would have to first select cups and then the only existing printer. I would think it makes sense to do it like e.g. windows does. Just select the printer. When there is some printer filtering (based on location, label template, ...) inplace, the list shouldn't be to long, also for large setups. The question now is just if these printer list should be in the current select field where we can also choose InvenTree pdf or if it should be an extra select which gets shown if we selected machines?

matmair:
I think selecting the driver and then the machine is easier for now - the UI/UX of how to render different machines/driver combos and provide intelligent filtering is probably a bit of dev work and I would love to get this merged sooner than later to use it as a building block for MES and more.

wolflu05:
@SchrodingersGat I'm not really sure if you understand how I described the two options. Ill add some prototype recordings:

The way I always thought to implement (option 1) was to add another built in printing printing plugin "Machines". Then we can select between Inventree pdf/-sheet and machines by default. All other label printing plugins then should be implemented as a driver rather than a normal plugin with the printing mixin. The printing mixin then only can be used for plugins that doesn't connect to physical machines, but are some virtual things like putting multiple labels on a grid, ...

At the end it should look something like this (example):

https://github.com/inventree/InvenTree/assets/76838159/f834de0a-c82a-4b35-9631-f60c8fbca2dc

However option 2 would involve some more work because the required things are not in place currently, but would have much better UX in my opinion. For that we would need to add an function `get_printers(...)` to the LabelPrintingMixin in order to implement the machines registry with a built in plugin. That would look like this:

https://github.com/inventree/InvenTree/assets/76838159/0e43e7d9-b12c-41c6-a0e9-4de59fbbd06e

wolflu05:
> My expectation was that the "machines" interface allows a certain plugin to connect to multiple external machines. Let's take the brother-plugin as an example. With the machines interface, the plugin can support multiple connected brother printer machines.
> The brother-plugin knows what machines are available (within the subset of "brother" machines)

Until here everything is correct if you swap out plugin with driver. Basically the registry holds machines of specific types. Each machine type can have a different driver. And drivers are provided by plugins (having a driver defined in a separate class was the structure @matmair initially proposed to me a long time ago). Take a look at the started developer documentation over [HERE](https://inventree--4824.org.readthedocs.build/en/4824/extend/plugins/machines/), I already started there to explain the machine registry structure more in detail. For an example driver (my cups is not yet implemented), take a look at this [example](https://inventree--4824.org.readthedocs.build/en/4824/extend/plugins/machines/#example-driver). To the `MyXYZDriver` there will be added the methods `print_label` and `print_labels` like we already have them in the `LabelPrintingMixin`, but there is an extra machine parameter added with the reference to the machine where the driver can access all the settings. This driver then also extends from the `LabelPrintingBaseDriver` (which provides the pricing functions) instead of `BaseXYZDriver`. 

That would look like this:
```py
from plugin import InvenTreePlugin
from plugin.machine import BaseLabelPrintingDriver

class BrotherLabelPlugin(InvenTreePlugin):
    NAME = "BrotherLabels"
    SLUG = "brother"
    TITLE = "Brother Label Printer"
    # ...

    class BrotherDriver(BaseLabelPrintingDriver):
        SLUG = 'brother-driver'
        NAME = 'Brother label driver'
        DESCRIPTION = 'Print inventree labels with brother printers'

        MACHINE_SETTINGS = {
            'CONNECTION_URL': {
                'name': _('Connection url'),
                'description': _('Connection url to connect to the brother server'),
                'required': True,
            }
        }

        class PrintingOptionsSerializer(serializers.Serializer):
           copies = serializers.IntegerField() 
 
        def print_label(self, machine, **kwargs):
           conn_url = machine.get_setting("D", "CONNECTION_URL")
           conn = brother.get_connection(conn_url)
           conn.print(kwargs['pdf_file'], copies=kwargs['printing_options'].get('copies', 1))
```

So unless you want to do anything virtual (e.g. the inventree sheet printer where the concept of machines makes no sense) `LabelPrintingMixin` plugins have to be rewritten to provide a machine driver instead of implementing the `LabelPrintingMixin`

Having this extra class for a "Driver" structure enables us much more flexibility to implement other things like machine status (see [documentation](https://inventree--4824.org.readthedocs.build/en/4824/extend/plugins/machines/#machine-status)) or machine settings and don't directly clutter the plugin class. Afaik this structure is also used for notification methods. (And by how python handles nested classes, this would be the equivalent if we write the class at the root level which I prefer, so you find that way in the documentation. Because the drivers (and machine types) are discovered by looking for classes that inherit the base driver (like its done with notification methods)).

> and can present these to the "printing options" dialog (as it already exists), e.g. "select printer". This would just be another printer plugin option. IMO the printing dialog does not know about machines at all - each plugin decides whether or not to present "machines" as an option to the printer dialog.
> In this fashion I think that the first video you linked above looks like exactly this, no?

No, how I understand your thinking (what I named option 3 above) is that you can select between Inventree Pdf/-sheet/cups/brother/zebra and then in the printing options is an option to select between the machines of that particular type. But that's not what I proposed. In option 1 you select between inventree pdf/-sheet and **machines** and if you selected machines you get **all** printers, the cups, zebra and brother ones (later with filtering based on the current location and the printer location, label size, ...).

I hope the added documentation and this more detailed explanation help you to understand what machine types/drivers/the registry are and they work. If there is anything still unclear, don't hesitate to ask.

SchrodingersGat:
@wolflu05 thanks for taking the time to explain that in detail. I think that the "machines" side of things makes a lot of sense. However I do not see why the concepts of "machines" and "label printing" cannot be separated.

When you want to print a label, you select the "printing plugin" - this is an abstract plugin which recevies a PDF and "does something with it". This "something" may be download a file, or send to a printer, or even email to someone. The plugin is free to specify any additional options at run-time (via the very nice extra fields you have implemented).

The plugin is also free to hook into the "machines" interface (and even specify its own driver).

For example the brother plugin can define a machine driver for connecting to a network printer, and a driver for connecting to a USB printer, etc.

When the user wants to print a label using the "brother" label plugin they just select "brother labels" from the plugin dropdown. The plugin then can provide an extra drop-down where the user can select from a list of connected "machines".

Note that the printing pipeline never knows about the concept of "machines" - just "plugins"..

Does this make sense? I think that the separation here keeps things cleaner

wolflu05:
> I do not see why the concepts of "machines" and "label printing" cannot be separated. [...]
> Note that the printing pipeline never knows about the concept of "machines" - just "plugins".. [...]
> I think that the separation here keeps things cleaner

@SchrodingersGat Yes, that is exactly what I was planning, the printing pipeline doesn't know anything of machines, otherwise the code can get  pretty fast convoluted like it was where pdf printing was not implemented through a printing plugin.
 
I planned to implement the connection in a builtin plugin called something like `Label printing machines`. That plugin then uses the printing options to show a select with all available machines. If one machine is selected, it extends the printing options (using a dependent field) with the machines options. Then if a a user clicks "Print", the `print_labels` method in the builtin Label printing machines plugin is used to forward the call to the correct driver.   

> When you want to print a label, you select the "printing plugin" - this is an abstract plugin which recevies a PDF and "does something with it". This "something" may be download a file, or send to a printer, or even email to someone. The plugin is free to specify any additional options at run-time (via the very nice extra fields you have implemented).

The concept of the `LabelPrintingMixin` will be still there, but then to be only used for "virtual" things, as you named: emailing, exporting a pdf, ... (technically also for physical printer connections, but for that use case it's not that powerful as using a label printing driver.)

Machines and drivers should simplify things where the same code can be used with multiple settings. If each printing plugin would implement multiple printer selection (using printing options) and the config (e.g. server, port, usb) for each printer on their own it's much more complex to do so (also the UI may would look different from printing plugin to printing plugin), but would be possible, and is still possible after this is implemented.

I think we can still later change the implementation if we want to have it different if we keep the driver interface the same. That is also an advantage of having a separate driver class.

Does this make sense to you now? Or would you like to have something different?

SchrodingersGat:
I *think* that we are on the same page now... I am excited to see this progress :)

wolflu05:
Sounds good. Then I'll start implementing the label printer machine type like discussed.

wolflu05:
9 months passed now, this is my biggest PR and I learned a lot about InvenTree and Django during that time. Now this PR is finally ready for review 🚀 . @SchrodingersGat  @matmair 

I can recommend to start reviewing this by reading the [docs](https://inventree--4824.org.readthedocs.build/en/4824/extend/machines/overview/) to understand how the machine registry, custom drivers and types work before taking a deeper look into the code.

As an example, I adapted my cups label printer plugin to use the new driver interface. You can find it [here](https://github.com/wolflu05/inventree-cups-plugin/tree/driver-interface) if you want to see a real driver actually being implemented.

During this PR I collected a list of related things that can be further improvements (see initial comment) which I will post in a new issue once this is merged.

If you have any questions, feel free to comment, I'm really curious about your feedback.

matmair (commented):
LGTM; huge work effort, thanks @wolflu05

SchrodingersGat (approved):

SchrodingersGat:
@wolflu05 this is a *massive* effort to get this up and running, apologies for any road blocks along the way :)

This all looks good, but please appreciate that with such a large contribution, there are going to be some teething issue and bugs we have not spotted. It all appears to be in order and I am happy to merge now, so that we have some time before the next release to spot any obvious issues.

Without any machines to fill in myself, it is hard to create any test data to work with.

As we progress, it would be great to work on some more examples for the documentation of different types of machines, as we build those up.

No doubt you will be tagged in a number of issues / questions around all this :)

BTW regarding the style issue - once https://github.com/inventree/InvenTree/pull/6481 is merged, please merge master back in to your branch and we should be good to go.

SchrodingersGat:
Congrats and thanks @wolflu05 - let's hope it is a smooth integration from here!

wolflu05:
> This all looks good, but please appreciate that with such a large contribution, there are going to be some teething issue and bugs we have not spotted. [...] No doubt you will be tagged in a number of issues / questions around all this :)

This is actually the largest PR I've ever done. All good, feel free to ping me if there is anything I can help with. Hopefully we don't find to much issues 😆.

> Without any machines to fill in myself, it is hard to create any test data to work with.

If you want to test it, you can create a simple driver and place it anywhere in the code like this:

<details><summary>Testing driver</summary>

```py
from plugin.machine.machine_types import LabelPrinterBaseDriver

class TestLabelPrinterDriver(LabelPrinterBaseDriver):
    """Label printer driver for Dymo printers."""

    SLUG = "test-driver"
    NAME = "Test Driver"
    DESCRIPTION = "Test label printing driver for InvenTree"

    MACHINE_SETTINGS = {
        "CONNECTION_URL": {
            "name": "Connection url",
            "description": "Connection url for the machine in format user@server:port/route",
            "required": True,
        }
    }

    def print_labels(self, machine, label, items, request, **kwargs):
        print("==== Print label")
        print(machine, label, items, request, kwargs)
        print(machine.get_setting('CONNECTION_URL', 'D'))
        print("="*10)
```

</details> 

> As we progress, it would be great to work on some more examples for the documentation of different types of machines, as we build those up.

Definitely, I think the documentation around this has big potential, also the generated part from the docstrings.

> Congrats and thanks @wolflu05 - let's hope it is a smooth integration from here!

Thank you very much for merging, this is a great step forward, I actually really hadn't expected to get this merged that quick. And yes, let's hope it is a smooth integration 😆 .

PR #4834, opened 2023-05-16
Refactor: BaseInvenTreeSetting

wolflu05:
This PR tries to add typings for settings and make the `common.BaseInvenTreeSetting` model more generic. This simplifies the implementation for the #4824 , see https://github.com/inventree/InvenTree/pull/4824#discussion_r1193650747

Actually first, I planned to only add a `get_filters` method like the `get_kwargs` method until I now discovered that something like this was already there onetime and was removed. (See: 5ee9af7f0ebed1cb56ea488299824daf0db0afb8). Let me know if you have a better idea to make it more generic.

I tested this and discovered no bugs or issues that something doesnt work or break with this PR. Hopefully tests cover the other half 😄.

### DIFF
--- a/InvenTree/plugin/models.py
+++ b/InvenTree/plugin/models.py
@@ -172,12 +172,6 @@ class PluginSetting(common.models.BaseInvenTreeSetting):
 
         return super().get_setting_definition(key, **kwargs)
 
-    def get_kwargs(self):
-        """Explicit kwargs required to uniquely identify a particular setting object, in addition to the 'key' parameter."""
-        return {
-            'plugin': self.plugin,
-        }
-
 
 class NotificationUserSetting(common.models.BaseInvenTreeSetting):
     """This model represents notification settings for a user."""
@@ -199,13 +193,6 @@ class NotificationUserSetting(common.models.BaseInvenTreeSetting):
         kwargs['settings'] = storage.user_settings
 
         return super().get_setting_definition(key, **kwargs)
-
-    def get_kwargs(self):
-        """Explicit kwargs required to uniquely identify a particular setting object, in addition to the 'key' parameter."""
-        return {
-            'method': self.method,
-            'user': self.user,
-        }
 
     method = models.CharField(
         max_length=255,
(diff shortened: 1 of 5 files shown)


matmair:
@wolflu05 are you sure that the subclass checking is not required for plugin settings?

matmair (commented):
Looks good in general

InvenTree/plugin/models.py
matmair, line 1: Line 171 lose coverage with the change - https://coveralls.io/builds/60021910/source?filename=InvenTree%2Fplugin%2Fmodels.py#L171

matmair (changes requested):

wolflu05:
> are you sure that the subclass checking is not required for plugin settings?

I used project wide code search and checked all method calls individually where I changed something if there is a `plugin` kwarg. And then looked through the list for `plugin=` results, changed everything there what I found to call the `.plugin_config()` function there.

matmair:
So we can remove that section entirely?

wolflu05:
What section? The subclass checking in the BaseInventreeSettingClass? Yes. Ideally everything is moved out now and should be implemented in new code like this.

Do you want me to add general tests in this PR? But how do we test an abstract model?

InvenTree/plugin/models.py
wolflu05: Removed that line, as it is not needed anymore as it only gets a `PluginConfig` now.

matmair:
Something seems off with the coverage - I will let it run again

matmair (commented):
Looks good but not sure if the uniqueness will work this way

wolflu05:
How do you mean that?

SchrodingersGat (approved):

matmair:
> How do you mean that?

I was worried about removing the kwargs passed to the uniqueness check for user settings, in local testing it seems to work this way though.

matmair (approved):

matmair:
@SchrodingersGat 0.12.0 is looking like a big change under the hood - I think especially this one needs a bit of testing in the real world before we release a stable

SchrodingersGat:
@wolflu05 thanks for the contribution

@matmair agreed, we won't be pushing out this release immediately!

PR #4857, opened 2023-05-19
Refactor states/status

matmair:
Refactor for generic StatusCode definition. The goal is to make the code self-sufficient, 100% covered by unit tests and easy to use in any place needed (including plugins).

Specific changes:
- [x] factors state/status out to generic method
- [x] makes API endpoint safer to use
- [x] switches StatusCode to enums - making the definition less repetitive and shorter


InvenTree/InvenTree/context.py
```diff
@@ -57,16 +55,8 @@ def status_codes(request):
...
-        # Expose the StatusCode classes to the templates
-        'ReturnOrderStatus': ReturnOrderStatus,
-        'ReturnOrderLineStatus': ReturnOrderLineStatus,
-        'SalesOrderStatus': SalesOrderStatus,
-        'PurchaseOrderStatus': PurchaseOrderStatus,
-        'BuildStatus': BuildStatus,
-        'StockStatus': StockStatus,
-        'StockHistoryCode': StockHistoryCode,
-    }
+    classes = inheritors(StatusCode)
```
SchrodingersGat, line 58: Nice simplification!

wolflu05:
1. Is this from any help for my machine integration (#4824)? Maybe you remember, I want to add some in RAM stored machine states that could be set by the driver. Does this help so I need to take a closer look?
2. Do you plan to refactor the abstract settings class in this generic app too?

matmair:
1. Yes, that is the goal (the enum can be expanded on demand
2. No, I see states as a simple enum (a value that can be represented in multiple ways) - that can be saved into a field - and settings as typed values, that can have complex validation, set and get methods and are always related to a specific database value

matmair:
I need a bit of feedback @inventree/triage. Not really looking for a full review but input on the following:

This PR changes the signature for status codes from:
``` python
class SalesOrderStatus(StatusCode):
    """Defines a set of status codes for a SalesOrder."""

    PENDING = 10  # Order is pending
    IN_PROGRESS = 15  # Order has been issued, and is in progress
    SHIPPED = 20  # Order has been shipped to customer

    options = {
        PENDING: _("Pending"),
        IN_PROGRESS: _("In Progress"),
        SHIPPED: _("Shipped"),
    }

    colors = {
        PENDING: 'secondary',
        IN_PROGRESS: 'primary',
        SHIPPED: 'success',
    }

    # Open orders
    OPEN = [
        PENDING,
        IN_PROGRESS,
    ]
```
to 
```python
class SalesOrderStatus(StatusCode):
    """Defines a set of status codes for a SalesOrder."""

    PENDING = 10, _("Pending"),   # Order is pending
    IN_PROGRESS = 15, _("In Progress"), 'primary'   # Order has been issued, and is in progress
    SHIPPED = 20, _("Shipped"), 'success'           # Order has been shipped to customer


class SalesOrderStatusGroups:
    """Groups for SalesOrderStatus codes."""

    # Open orders
    OPEN = [
        SalesOrderStatus.PENDING.value,
        SalesOrderStatus.IN_PROGRESS.value,
    ]
```
A lot of functions are refactored in the background to be more efficient/more strongly checked. 

But there are caveats. Assigning a value changes:
```
- self.status = BuildStatus.COMPLETE
+ self.status = BuildStatus.COMPLETE.value
```
Status Groups are in a separate class
Logical operations do not change.

Thoughts?

wolflu05:
Looks much cleaner now. Need to do some further review at some point to get how to use this for machine integration.

Regarding the .value thing. Can't you override some python __getter__ function to return what you like instead of the need for the .value?

matmair:
@wolflu05 I will also provide usage docs in the new module, it is pretty simple.
I couldn't get overriding the getter to work we also lose all the 'special' features Enums provide.

wolflu05:
Have you had a look how [models.TextChoices](https://docs.djangoproject.com/en/3.2/ref/models/fields/#enumeration-types) handled this?

matmair:
@wolfu05 no, haven't thought about that - checking now; thanks!

matmair:
@SchrodingersGat this would be ready for a review. I moved the database additions to a new PR as this one is becoming big and hard to review with the database components intertwined.

SchrodingersGat:
@matmair looking now, sorry I didn't respond to your previous ping.

InvenTree/order/migrations/0096_alter_returnorderlineitem_outcome.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('order', '0095_salesordershipment_delivery_date'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='returnorderlineitem',
+            name='outcome',
+            field=models.PositiveIntegerField(choices=[(10, 'Pending'), (20, 'Return'), (30, 'Repair'), (40, 'Replace'), (50, 'Refund'), (60, 'Reject')], default=10, help_text='Outcome for this line item', verbose_name='Outcome'),
```
SchrodingersGat, line 16: Is it possible to change the "choices" field to not have to enumerate all the valid choices into the migration?

It would be handy potentially to be able to change the choices without generating a new migration?

A similar issue exists when adding support for a new currency

SchrodingersGat (approved):
LGTM in general, but there's a lot of files touched :)

InvenTree/order/migrations/0096_alter_returnorderlineitem_outcome.py
matmair: @SchrodingersGat seems possible, I will look into it.

matmair: @SchrodingersGat I have looked into this and though about it. I have created a PR that changes the behaviour for currencies #4975.
As the choices for states have an impact on business, I think at least the default value for states should be in migrations.

wolflu05:
Does this trick also work with now autogenerated new migrations of is there some manual adjustment needed afterwards before committing the migration files?

matmair:
If you change the choices you will have to adjust it again.

matmair:
Hopefully the upstream fixes make the CI go through now

matmair:

SchrodingersGat:
@matmair ok something screwy is going on:

- https://github.com/inventree/InvenTree/actions/runs/5216970511
- https://github.com/inventree/InvenTree/actions/runs/5216970511/jobs/9416333186
- https://github.com/inventree/InvenTree/actions/runs/5216970511/jobs/9416333108

image
image

Looks like we need to give this one some serious attention!

Issue #4769, opened 2023-05-05
Email sending fails with 'Invalid address "" '


crvio:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [X] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

A fresh docker deployment of Inventree fails when trying to send an email to verify an email address.
When trying to resend the message for email verification from the admin profile settings page, I get a message of "Correo electrónico enviado a xxx@yyy.zz" (which roughly translates to "Email sent to xxx@yyy.zz", being that the email address it should be sent to.), but immediately a new error notification shows in the upper right corner.
When clicking on it, the following stacktrace gets shown: (pasted it in the relevant log output section)
By reading a bit into it I see that it is probably trying to email to an empty string address, yet the methods used belong to django modules, so tracing back the issue gets a little tricky for me.
Any help would be much appreciated.

### Steps to Reproduce

I followed the docs in [here](https://docs.inventree.org/en/latest/start/docker_prod/).
Logged in and started to set it up.
Email server, connection settings and credentials were input through the .env file pointed to from the docker-compose.yaml file.
When trying to add a new user (through the login page) no email was sent and the error was shown.
It might as well be the case that I'm missing some important step, but I couldn't find the error somewhere else with the solution as it usually is the case when I do.

### Relevant log output

```bash
Traceback (most recent call last):

File "/root/.local/lib/python3.9/site-packages/django/core/mail/message.py", line 82, in sanitize_address
token, rest = parser.get_mailbox(addr)

File "/usr/local/lib/python3.9/email/_header_value_parser.py", line 1791, in get_mailbox
token, value = get_name_addr(value)

File "/usr/local/lib/python3.9/email/_header_value_parser.py", line 1760, in get_name_addr
if value[0] in CFWS_LEADER:

IndexError: string index out of range


During handling of the above exception, another exception occurred:


Traceback (most recent call last):

File "/home/inventree/InvenTree/InvenTree/forms.py", line 277, in send_mail
result = super().send_mail(template_prefix, email, context)

File "/root/.local/lib/python3.9/site-packages/allauth/account/adapter.py", line 140, in send_mail
msg.send()

File "/root/.local/lib/python3.9/site-packages/django/core/mail/message.py", line 284, in send
return self.get_connection(fail_silently).send_messages([self])

[...]

PR #4492, opened 2023-03-14
Email settings not configured - remove need for user and password

matmair:
Fixes #4473

This PR removes the need to set password and user for email.

PR #4862, opened 2023-05-21
Improved error handling for email support

SchrodingersGat:
- Prevent email sending if email not configured
- Check for tx email address before sending

(cherry picked from commit de541f811ede030ea5eb3136132731e1dafccc31)

Ref: https://github.com/inventree/InvenTree/issues/4769


InvenTree/email.py
```diff
@@ -0,0 +1,79 @@
...
+        # Display warning unless in test mode
+        if not settings.TESTING:  # pragma: no cover
+            logger.debug("EMAIL_HOST is not configured")
+
+    # Display warning unless in test mode
+    if not settings.TESTING:  # pragma: no cover
+        logger.debug("EMAIL_HOST_USER is not configured")
+
+    # Display warning unless in test mode
+    if not settings.TESTING:  # pragma: no cover
```
matmair, line 39: Question: Where is the check for user or password?

matmair (approved):

InvenTree/email.py
SchrodingersGat: https://github.com/inventree/InvenTree/pull/4492

matmair: Seems like an error in the initial PR by me. We should check before logging - not error out though

SchrodingersGat: @matmair I'm not 100% across what you're suggesting here - want to make a suggested change?

InvenTree/email.py
```diff
@@ -0,0 +1,79 @@
...
+    if not settings.EMAIL_HOST:
+        configured = False
+
+        # Display warning unless in test mode
+        if not settings.TESTING:  # pragma: no cover
+            logger.debug("EMAIL_HOST is not configured")
+
+    # Display warning unless in test mode
+    if not settings.TESTING:  # pragma: no cover
+        logger.debug("EMAIL_HOST_USER is not configured")
```
matmair, line 36:
```suggestion
    if not settings.EMAIL_HOST_USER and not settings.TESTING:  # pragma: no cover
        logger.debug("EMAIL_HOST_USER is not configured")
```

InvenTree/email.py
matmair, line 39:
```suggestion
    if not settings.EMAIL_HOST_PASSWORD and not settings.TESTING:  # pragma: no cover
```

matmair (changes requested):

matmair (approved):

PR #4880, opened 2023-05-23
Make nginx more clear that changing the port is 'dangerous'

matmair:
See https://github.com/inventree/InvenTree/issues/4859#issuecomment-1560025566


matmair:
@SchrodingersGat is this wording understandable?

SchrodingersGat:
@matmair thanks I think this makes it a lot clearer

PR #4881, opened 2023-05-23
Improve cli experience

matmair:
This PR:
- Makes sure inventree run cli works without activation
- Adds a version task to the invoke tasks (set as a default)

Printout of the version task
```

InvenTree - inventree.org
The Open-Source Inventory Management System

========================================
Installation paths:
Base        /opt/inventree
Config      /etc/inventree/config.yaml
Media       /opt/inventree/data/media
Static      /opt/inventree/data/static

Versions:
Python      3.8.10
Django      3.2.18
InvenTree   0.11.0
API         107

Commit hash:None
Commit date:None

You are probably running the package installer / single-line installer. Please mentioned that in any bug reports!

Use '--list' for a list of available commands
Use '--help' for help on a specific command
```

### DIFF
--- a/InvenTree/InvenTree/version.py
+++ b/InvenTree/InvenTree/version.py
@@ -9,7 +9,7 @@ import subprocess
 
 import django
 
-from InvenTree.api_version import INVENTREE_API_VERSION
+from .api_version import INVENTREE_API_VERSION
 
 # InvenTree software version
 INVENTREE_SW_VERSION = "0.12.0 dev"
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 web: env/bin/gunicorn --chdir $APP_HOME/InvenTree -c InvenTree/gunicorn.conf.py InvenTree.wsgi -b 0.0.0.0:$PORT
 worker: env/bin/python InvenTree/manage.py qcluster
-cli: env/bin/python -m invoke
+cli: . env/bin/activate && exec env/bin/python -m invoke
--- a/tasks.py
+++ b/tasks.py
@@ -11,8 +11,6 @@ from platform import python_version
 
 from invoke import task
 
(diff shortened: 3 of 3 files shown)


SchrodingersGat (approved):

SchrodingersGat:
Nice addition :) ready to merge?

InvenTree/InvenTree/version.py
```diff
@@ -9,7 +9,7 @@
 
 import django
 
-from InvenTree.api_version import INVENTREE_API_VERSION
+from .api_version import INVENTREE_API_VERSION
```
wolflu05, line 12: Why do you decided to import like this? What is the preferred way for imports? Sometimes I see you using relative imports. Sometimes you only  import the direct function, sometimes you import the complete module.

matmair:
@SchrodingersGat yes!

InvenTree/InvenTree/version.py
matmair: Imports in Python are a whole thing. In this specific case, the file will be called both by invoke from the context of / and from other functions in the Django app from the context of /InvenTree. So the absolute import would need to be both InvenTree.InvenTree.api_version and InvenTree.api_version.

We could set the working directory when starting invoke but that will lead to problems once external automation is used (ansible eg.). Therefore I decided that a relative import is the best solution here. In general, I prefer absolute imports for code that should be called from other modules and relative imports when a module is self-reliant and has a defined import point.

wolflu05: And what about the second half of my question?

> Sometimes you only import the direct function, sometimes you import the complete module.

matmair: It is easier to not import 20 different functions and there are some interesting corner cases with Django where it helps to import the module and not a specific function. Over 90% of the time it is the first reason

Issue #4428, opened 2023-02-28
Multiple lines of "fatal: not a git repository (or any of the parent directories): .git" during update


simonkuehling:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

Probably not a bug per se, since everything is working alright - but maybe something to take a look into:

During `docker-compose run inventree-server invoke update` I always get multiple lines of the error message
`fatal: not a git repository (or any of the parent directories): .git`

Can I debug this further to understand where to look for fixing this?

### Steps to Reproduce

Run  `docker-compose run inventree-server invoke update` in a docker production installation.

### Expected behaviour

No fatal errors unless something is actually wrong ;-)

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.11.0 dev
Django Version: 3.2.18
Commit Hash: e9a1b57
Commit Date: 2023-02-28
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}, {'name': 'IPNAutoGenerator', 'slug': 'ipnautogenerator', 'version': None}, {'name': 'Brother Labels', 'slug': 'brother', 'version': '0.6.0'}]

### Relevant log output

```shell
docker-compose run inventree-server invoke update
Creating network "inventree-production_default" with the default driver
Creating inventree-db ... done
Creating inventree-production_inventree-server_run ... done
Loading config file : /home/inventree/data/config.yaml
Installing required python packages from 'requirements.txt'
Requirement already satisfied: pip in /usr/local/lib/python3.9/site-packages (23.0.1)
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
Requirement already satisfied: arrow==1.2.3 in /root/.local/lib/python3.9/site-packages (from -r requirements.txt (line 7)) (1.2.3)
Requirement already satisfied: asgiref==3.6.0 in /root/.local/lib/python3.9/site-packages (from -r requirements.txt (line 9)) (3.6.0)
[...]

PR #4882, opened 2023-05-23
Possible fix for git messages

matmair:
This PR might fix https://github.com/inventree/InvenTree/issues/4428


wolflu05:
Do you think this line is also responsible for these messages? 

https://github.com/inventree/InvenTree/blob/86b72549a2cc68a9262e7710e3dfd3e726906ac8/InvenTree/plugin/helpers.py#L119

matmair:
@wolflu05 to a part - yes. Another big source was the lookup for commit number and date.

matmair:
Strangely enough, the tests no longer throw the error - bare metal still does. @SchrodingersGat this def makes things better. Let's merge this in and I will submit another PR once I have determined the rest comes from.

wolflu05:
> @wolflu05 to a part - yes. Another big source was the lookup for commit number and date.

So you won't add your check also to this usage of git?

matmair:
> > @wolflu05 to a part - yes. Another big source was the lookup for commit number and date.
> 
> So you won't add your check also to this usage of git in this PR?

No, it broke in some tests with the check added. I need a better one for the packages.

SchrodingersGat:
Very clever fix @matmair :)

matmair:
Not really TBH but thanks

PR #4898, opened 2023-05-25
Add Metadata to more models

miggland:
Closes https://github.com/inventree/InvenTree/issues/4202

Adds metadata to:
- [x] Part test templates
- [x] Part parameters
- [x] Part category parameter templates
- [x] BOM item substitute
- [x] Part relateds
- [x] Stock item test result
- [x]  Return order
- [x] Return order lines
- [x] Return order extra lines
- [x] Company contacts

Contains one bug fix

Todo:
- [x] Tests


wolflu05:
I'm not sure if the 0.11.3 milestone is correct here as this is getting merged in master? Also it probably makes sense to not backport this change as this is no bugfix.

miggland:
There is a small bugfix in there - and the rest is not really new, just an extension of existing functionality to more models

matmair (approved):

matmair:
@miggland we aim to keep the bugfix branches stable and add new stuff only to stable - that needs to be documented better for sure.

Could you cherry-pick the bugfix to a separate branch and add a PR against 0.11.x? I think it might be a while till the next release based on the current milestone set.

miggland:
I'll finish the complete package first, then I can think about a separate PR for the bugs..

SchrodingersGat:
@miggland LMK when this is ready for review

miggland:
I'm confused by the test results.

It seems like the contact edit API endpoint is not working, because it is looking for a URL without the final slash:
```
Not Found: /api/company/contact/6
.Not Found: /api/company/contact/1
Unexpected GET response at '/api/company/contact/1': status_code = 404
```

Any ideas why?

matmair:
The interesting part is that it works in one scenario but not another. I will try my hand at a fix @miggland

matmair:
Maybe this was a fluke - can not reproduce locally on sqlite or pgsql. I have restarted the CI, maybe it runs through now 🤞

miggland:
Not a fluke, and it is reproducible on my side.

I changed `company/api.py`, and the URL has changed (includes /). However, the error remains the same.
```
--- a/InvenTree/company/api.py
+++ b/InvenTree/company/api.py
@@ -561,7 +561,7 @@ company_api_urls = [
     ])),
 
     re_path(r'^contact/', include([
-        re_path(r'^(?P<pk>\d+)/?', include([
+        re_path(r'^(?P<pk>\d+)/', include([
             re_path('^metadata/', MetadataView.as_view(), {'model': Contact}, name='api-contact-metadata'),
             re_path('^.*$', ContactDetail.as_view(), name='api-contact-detail'),
         ])),
```

```
.Not Found: /api/company/contact/1/
Unexpected GET response at '/api/company/contact/1/': status_code = 404
```

When I try the same URL manually, it works..

miggland:
An explanation:
The failing test is in the class `ContactTest`, and used a fixed PK of 1 in `test_edit` and 6 in `test_delete` for testing the edit/delete functions.
However, these PKs were not defined through fixtures but from creating the contacts through the API, _without a defined PK_.
To be able to test metadata, each model needs an object in the database. Thus, I added a fixture used in class `CompanyMetadataAPITest`, in which the contact has PK=1.
Since `ContactTest` was run after `CompanyMetadataAPITest`, but without fixtures, there was no contact with PK=1, but with auto-incremented, higher values. Thus a 404 error.

I have changed the failing tests to get the first/last contact item from the database, then use that PK in the API tests. Now it works (at least on my dev machine).
I don't quite understand why the PK is incremented even when the data is not present, but maybe the database is not truncated properly after each test class, just all rows dropped?

miggland:
> Could you cherry-pick the bugfix to a separate branch and add a PR against 0.11.x? I think it might be a while till the next release based on the current milestone set.

OK, see: https://github.com/inventree/InvenTree/pull/4947

Assuming the test passes, this is ready for review/merge.

matmair:
@miggland there was a new migration upstream

SchrodingersGat (approved):

SchrodingersGat:
@miggland thanks for the additions :)

SchrodingersGat:
@miggland @matmair I think this PR has broken database migrations for live data (demo instance is down currently). Will have to look into this ASAP

SchrodingersGat:
image

SchrodingersGat:
@miggland OK I can see what happened here:

- You renamed part migration 0012 to 0013
- You inserted your new migration as "0012"

In systems where *old* 0012 had already been run,  the new fields had been applied to the partparameter model.

Basically, if you have to mess around with the order of migrations, you should never change one which has already been run! I should have checked this more carefully before merging in. The CI checks passed because they operate on a clean database, and the tables here had not already been created.

I'll make a PR to address this.

SchrodingersGat:
Bottom line, we should *never* adjust the ordering of previously applied migrations - any new migrations should be applied *after* old ones.

If there are ever conflicts of migrations this needs to be carefully checked (I dropped the ball on this one!)

matmair:
I also didn't notice. Is this something we could check for in CI?

miggland:
Sorry about that. I have no problem with your explanation, and that wasn't what I intended to do either, but I must have messed up and moved the wrong one. Classical case of wanting to get things finished and not checking properly.

Being more aware of the problem helps though - I'll be more careful in future.

SchrodingersGat:
@matmair unsure. The issue is subtle:

1. Previous migration file `0012` added new fields to the `PartParameter` model
2. Database migrations were updated, such that `0012` was applied
3. Previous migration file `0012` was renamed to `0013`
4. *New* migration file inserted as `0012` which added new metadata fields
5. Database migrations run again (i.e. `python manage.py migrate`)
6. `0012` is recorded as already having been run (*and is not run again*) - thus metadata fields are not added
7. `0013` is a *new* migration, and so DB tries to apply it. However it is actually the *old* `0012` so causes a "field already exists" error
8. At this point we cannot roll back either, because we can't apply `0013` to then un-apply it

To test this in CI, maybe we maintain an active database, separate from the repo (or in a separate branch), and the data migrations have to successfully apply to *that* database, in addition to an *empty* database (which is the current check).

1. New PR is created
2. As part of CI, *old* database is fetched which is running on top of current master
3. If database migrations apply, all is *OK*
4. Once PR is merged, triggers a workflow to update the database with the new migrations

This is probably the closest we can replicate an "active" database in CI

PR #4899, opened 2023-05-25
Add API Calendar for Return Orders

miggland:
Activate the calendar for return orders


SchrodingersGat (approved):

matmair (approved):
Clean change; would it make sense to make the event body customizable?

matmair:
Regarding milestones. We normally push new features only to the new release, not the bugfix channel so that would be 0.12.0 right now

miggland:
> Clean change; would it make sense to make the event body customizable?

I thought about that, but didn't figure out a simple enough way to do this.

> Regarding milestones. We normally push new features only to the new release, not the bugfix channel so that would be 0.12.0 right now

Fair enough. In this case, one could argue that it is a bug-fix, since the link was missing for the Return Order.. However, 0.12.0 is fine with me.

SchrodingersGat (approved):

SchrodingersGat:
@miggland LGTM - ready for merge?

miggland:
Yep, ready.

SchrodingersGat:
Great :) thanks for the continuing contritbutions!

PR #4908, opened 2023-05-26
Fix completeSalesOrderShipment typo in sales_order.js

### LINKED ISSUE #4906: Shipping items from sales order fails due to typo
https://github.com/inventree/InvenTree/blob/master/InvenTree/templates/js/translated/sales_order.js#L399

Should be completeSalesOrderShipment instead of completeSalseOrderShipment, now gives the error "cant find variable completeSalseOrderShipment"

I know creating a PR would be the way to go, but im on mobile atm.

martonmiklos:
Fixes #4906

### DIFF
--- a/InvenTree/templates/js/translated/sales_order.js
+++ b/InvenTree/templates/js/translated/sales_order.js
@@ -396,7 +396,7 @@ function completePendingShipments(order_id, options={}) {
  */
 function completePendingShipmentsHelper(shipments, shipment_idx, options={}) {
     if (shipment_idx < shipments.length) {
-        completeSalseOrderShipment(shipments[shipment_idx].pk,
+        completeSalesOrderShipment(shipments[shipment_idx].pk,
             {
                 buttons: [
                     {


matmair (approved):

SchrodingersGat:
Nice catch @martonmiklos - we really need to get some CI testing on the front end code ASAP

martonmiklos:
> Nice catch @martonmiklos -

Credits goes to @jelgersma

> we really need to get some CI testing on the front end code ASAP

What is the plan for that (what technology do you prefer)?

SchrodingersGat:
> What is the plan for that (what technology do you prefer)?

I don't really have enough knowledge about the unit testing frameworks available to make a call on that! Have to read some more. Can you make any recommendations?

wolflu05:
I would say some linting would be enough for the first, as there is anyway the plan to switch to react with typescript?

Linting, eg with eslint would cover the undefined function references.

SchrodingersGat:
We already do run eslint - https://github.com/inventree/InvenTree/blob/master/.eslintrc.yml - so we should look into why it did not catch this incorrect function

PR #4911, opened 2023-05-26
Bump deps

matmair:
Bump deps - no code changes needed.


SchrodingersGat:
Interesting that updating the pint library has broken here. Do we want to:

1. Revert pint library
2. Investigate deeper what has broken it here

SchrodingersGat (approved):

PR #4915, opened 2023-05-28
Test for JS linting

SchrodingersGat:
- add bad .js code
- Call function which has not been defined
- Should throw JS lint error

### DIFF
--- a/InvenTree/templates/js/translated/part.js
+++ b/InvenTree/templates/js/translated/part.js
@@ -57,6 +57,7 @@ function partGroups() {
         attributes: {
             title: '{% trans "Part Attributes" %}',
             collapsible: true,
+            test: callFunctionWhichDoesNotExist(),
         },
         create: {
             title: '{% trans "Part Creation Options" %}',


SchrodingersGat:
@wolflu05 looks like the eslint config has been out of date for a while - should have been catching these errors!

SchrodingersGat:
So there are a *lot* of fails here, so how should we proceed?

1. Manually fix them "as is" (by using the es-lint "globals" and "exported" comments at the top of the file)
2. Move to a "modules" approach (not familiar with this, but have done some brief reading on it).
3. Prioritize the migration to react
4. Some combination of the above?

wolflu05:
I think a first good approach would be to ignore the translated files first and see how many errors pop up then.

matmair:
2. Might help a lot with other things like bundling, dependency updates etc.

Do we know why it did not work?

SchrodingersGat:
> Do we know why it did not work?

We were using the "google" config for eslint, which previously had thrown errors on these types of faults. I changed to "eslint:recommended" and it is now throwing again. So maybe the "google" config was changed at some point?

wolflu05:
Why do the localized files get checked for every language? Wouldn't it make sense to only check one language? Having the same error dozens of times doesn't really help.

SchrodingersGat:
> Why do the localized files get checked for every language? Wouldn't it make sense to only check one language? Having the same error dozens of times doesn't really help.

Because of a very subtle bug which crept in a while ago - the actual translated strings were not escaped properly, and were causing issues for certain languages.

For example:

```javascript
runFunction('{% trans "Such is life" %}');
```

would get translated into

```javascript
runFunction('c'est la vie');
```

(which obviously is invalid and throws an error.

So, while we also implemented some pre-processing of the translations to escape such characters, we also run the translated files through to ensure that the same thing doesn't happen again without us at least knowing about it.

wolflu05:
No, you missed one 😢 - but good work through.

.eslintrc.yml
```diff
@@ -19,6 +19,7 @@ rules:
   valid-jsdoc: off
   no-multiple-empty-lines: off
   comma-dangle: off
+  no-unused-vars: off
```
wolflu05, line 22: why do you added this ignore?

SchrodingersGat: There are *many* functions where we do things like:

```javascript
renderer: function(value, row) {
    // stuff here
    // we use "row" but not "value"
}
```

SchrodingersGat:
> Why do the localized files get checked for every language? Wouldn't it make sense to only check one language? Having the same error dozens of times doesn't really help.

[here's a good example](https://github.com/inventree/InvenTree/actions/runs/5121136141/jobs/9208524820?pr=4915)

.eslintrc.yml
wolflu05: Try prefixing the variable name with `_` this is an convention for intentionally unused variables like in this case.
```js
renderer: function(_value, row) {
    // stuff here
    // we use "row" but not "value"
}
```

SchrodingersGat: That's fair, but as we are moving away from raw javascript and refactoring the entire javascript codebase, the "reward to effort" ratio for fixing this is too low :)

PR #4938, opened 2023-05-31
Add user settings to remember the last/favourite label template

### LINKED ISSUE #4932: [FR]Add option to remember the lsat used label type

### Problem statement

When printing labels the label type is always set to the first label type available. 
When printing labels with the label type other than the first the label type always need to be selected.
This is could be quite inconvenient.

### Suggested solution

Set autocomplete flag on the label type comboboxes.

If approved I can fix it quite easily.

Adding field/table for the a "favourite" label type per user per label type (part/stock/stocklocation).

### Examples of other systems

N/A

- [X] I want to develop this.

martonmiklos:
Fixes #4932

### DIFF
--- a/InvenTree/templates/js/translated/label.js
+++ b/InvenTree/templates/js/translated/label.js
@@ -78,9 +78,6 @@ function selectLabel(labels, items, options={}) {
     }
 
     var modal = options.modal || '#modal-form';
-    {% settings_value 'DEFAULT_LOCATION_LABEL_TEMPLATE' user=request.user as default_location_template_pk %}
-    {% settings_value 'DEFAULT_PART_LABEL_TEMPLATE' user=request.user as default_part_template_pk %}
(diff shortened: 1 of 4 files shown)


InvenTree/templates/js/translated/label.js
```diff
@@ -92,6 +94,16 @@ function selectLabel(labels, items, options={}) {
         },
         function(item) {
             return item.pk;
+        },
+        null,
+        function(item) {
+            if (options.key == 'part')
+                return item.pk == "{{ default_part_template_pk }}";
```
SchrodingersGat, line 101: You can access user settings directly, this might be more readable as:

```javascript
return item.pk == user_settings.DEFAULT_LOCATION_LABEL_TEMPLATE;
```

SchrodingersGat (approved):

SchrodingersGat:
JS linting is failing because you can't use `settings_value` in a translated file - the translated file will no longer have access to any templated values :)

SchrodingersGat:
LGTM, thanks for the contribution @martonmiklos

PR #4944, opened 2023-06-01
Fix/settings bugs and added model SettingKeyType typing

wolflu05:
This PR fixes various issues related to settings:

- setting get cached even there are not saved yet (have no pk) which leads to trying to recreate them which fails
   This was pretty hard to find, I located that issue with `git bisect` using 61481b4eb as bad commit, 0ffe7861a as good commit and got 16ff1936 as first bad commit. 
- add `model` key to `SettingsKeyType` (was actually the initiator of this PR)
- plugin slugs that contain `-` can't use settings because the API `re_path` doesn't include `-` chars
- related fields cannot be reseted, instead there is an js error thrown that `data.value.toString()` cannot be called on `null`
- spelling mistakes discovered while debugging in the code (I have this [extension](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) installed to prevent them while I'm coding)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,7 +97,7 @@ node_modules/
 maintenance_mode_state.txt
 
 # plugin dev directory
-./plugins/
+InvenTree/plugins/
 
 # Compiled translation files
 *.mo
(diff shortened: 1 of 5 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@wolflu05 some subtle fixes here, looks good :)

PR #4956, opened 2023-06-02
Fix common spelling mistakes

matmair:
We have a lot of spelling mistakes in the codebase, as recent PRs showed. This PR:
- adds spell checking to the hook pipeline
- fixes existing mistakes so we have a clean running CI


wolflu05:
Nice addition, was already using a vscode extension, but doing this as part of linting is the better solution.

SchrodingersGat:
Wow quite a large number of errors 😕

InvenTree/stock/api.py
```diff
@@ -364,7 +364,7 @@ class Meta:
         ]
 
     # Relationship filters
-    manufactuer = rest_filters.ModelChoiceFilter(label='Manufacturer', queryset=Company.objects.filter(is_manufacturer=True), field_name='manufacturer_part__manufacturer')
+    manufacturer = rest_filters.ModelChoiceFilter(label='Manufacturer', queryset=Company.objects.filter(is_manufacturer=True), field_name='manufacturer_part__manufacturer')
```
matmair, line 367: @SchrodingersGat I do not think that the change in this filter will break anything - am I overseeing anything?

SchrodingersGat (approved):
Some *actual* bugs in here which the spellcheck has fixed, too...

SchrodingersGat:
Thanks @matmair for looking into this :)

PR #4964, opened 2023-06-03
Enable and disable plugins via the API

SchrodingersGat:
This PR does the following:

- Plugin activation is now handled directly from the settings page
- Plugin activation uses the API
- Plugins can now also be *deactivated* via the API
- Renders table of plugins via API access, rather than templated on the server side (because who wants templates?)


InvenTree/plugin/serializers.py
```diff
@@ -171,6 +171,26 @@ def save(self):
 
 class PluginConfigEmptySerializer(serializers.Serializer):
     """Serializer for a PluginConfig."""
+    ...
+
+
+class PluginActivateSerializer(serializers.Serializer):
+    """Serializer for activating a plugin"""
```
matmair, line 178: Question: Should this maybe be renamed to reflect that it can activate/deactivate a plugin?

matmair (approved):
looks good in general

InvenTree/plugin/serializers.py
SchrodingersGat: Done :)

matmair (approved):

SchrodingersGat:
For reference, here's a screenshot of the new interface:

image

PR #4966, opened 2023-06-04
[FR] Switch from git to dulwich

### LINKED ISSUE #4942: [FR] Switch from pure git to dulwich

### Problem statement

The calling of external processes via subprocess is creating problems with logs.

### Suggested solution

Switch to using https://github.com/jelmer/dulwich - a pure Python version

There is also GitPython - which still requires git on the system.

### Examples of other systems

netbox uses this lib

- [X] I want to develop this.

matmair:
This PR switches from using git to dulwich. This means we do not need git on the OS to discover git repos and do not rely on git's interface. This should remove all warnings and also make the code behave more predictable.
Fixes #4942

### DIFF
--- a/InvenTree/InvenTree/version.py
+++ b/InvenTree/InvenTree/version.py
@@ -3,16 +3,26 @@
 Provides information on the current InvenTree version
 """
 
+import datetime
 import os
+import pathlib
 import re
-import subprocess
 
 import django
+
+from dulwich.repo import NotGitRepository, Repo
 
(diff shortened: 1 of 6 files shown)


SchrodingersGat:
@matmair LGTM. On the "production" docker build we don't ship with the git logs, and use environment variables to report hash and revision date. Is this still supported?

matmair:
@SchrodingersGat the lookups for that are kept the same as before.

PR #4973, opened 2023-06-05
Add migration test workflow

SchrodingersGat:
Adds a new CI workflow step.

Test that the database migrations apply cleanly to an existing "production" database which has all previous migrations applied.

Keep track of "production" databases at each stable release, to ensure that updates from previous stable installations won't immediately fail.

-------------------

Future work: Keep track of "master" production database, to ensure latest commits merge and migrate correctly. Push any changes to the migrations back upstream.

-------------------

Ref: https://github.com/inventree/InvenTree/pull/4984


wolflu05:
I can recommend using [tmate](https://github.com/mxschmitt/action-tmate), that way you don't need to push and wait for new ideas to fix this.

matmair:
@SchrodingersGat is there a reason to keep this separate from QC or is this just for testing?

PR #4977, opened 2023-06-06
Migration bug fix 2

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/4898
Ref: https://github.com/inventree/InvenTree/pull/4961

This is a "new and improved" bug fix for migration shenanigans due to issues with #4989 and #4961

There is a small window for corrupted migrations to have *already* crept in, so this fix ensures that affected database columns are not created "twice".


miggland:
Tested this by starting a new instance based on commit `c0dafe155fbaa872110bc53c5ce190b3fde53ada`, stopping it, switching to `a3150d9cb35bf24d4d27e1285180e1a9e75d0323`, and running ìnvoke update`.

The migration fails, because of:
```
  File "/home/inventree/data/env/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.DuplicateColumn: column "checkbox" of relation "part_partparametertemplate" already exists
```

This occurs even if the check shows that the field does not exist:
```
PartParameterTemplate._meta.fields = (<django.db.models.fields.AutoField: id>, <django.db.models.fields.CharField: name>, <django.db.models.fields.CharField: units>, <django.db.models.fields.CharField: description>, <django.db.models.fields.json.JSONField: metadata>)
Column 'checkbox' does not exist (skipping)
```

The same goes for `choices`.

Suggested fix in https://github.com/inventree/InvenTree/pull/4979

PR #4979, opened 2023-06-06
Migration fix - Remove check for field, just try to get rid of it

miggland:
See https://github.com/inventree/InvenTree/pull/4977

### DIFF
--- a/InvenTree/part/migrations/0112_auto_20230525_1606.py
+++ b/InvenTree/part/migrations/0112_auto_20230525_1606.py
@@ -22,14 +22,12 @@ def delete_columns(apps, schema_editor):
     # Check if the 'checkbox' column exists
     try:
         print("Checking for column 'checkbox' in table 'part_partparametertemplate'")
-        PartParameterTemplate._meta.get_field('checkbox')
         schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN checkbox;")
     except (AttributeError, FieldDoesNotExist):
         print("Column 'checkbox' does not exist (skipping)")
 
     try:
         print("Checking for column 'choices' in table 'part_partparametertemplate'")
-        PartParameterTemplate._meta.get_field('choices')
         schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN choices;")
     except (AttributeError, FieldDoesNotExist):
         print("Column 'choices' does not exist (skipping)")


SchrodingersGat:
@miggland the current approach seems to be working ok - what's the rationale here?

SchrodingersGat:
Looks like it doesn't work for all db types

miggland:
Yes, that syntax unfortunately only works for PG. I have some more ideas, more is coming

miggland:
Both checks together, then it works - at least with Postgres and SQLite.

InvenTree/part/migrations/0112_auto_20230525_1606.py
matmair, line 1: This breaks on mysql https://github.com/matmair/InvenTree/actions/runs/5193906520/jobs/9365058518

matmair (changes requested):

InvenTree/part/migrations/0112_auto_20230525_1606.py
matmair: @miggland is there reason for using raw SQL or can we use at least the [schema_editor](https://docs.djangoproject.com/en/4.2/ref/schema-editor/#remove-field)? Pure Mysql can not to "IF EXISTS"

SchrodingersGat: schema_editor looks appropriate, I was not aware of this, should have done that to start with

InvenTree/part/migrations/0112_auto_20230525_1606.py
```diff
@@ -2,7 +2,7 @@
 
 from django.core.exceptions import FieldDoesNotExist
 from django.db import migrations, models
-
+from django.db.utils import OperationalError
```
matmair, line 5:
```suggestion
from django.db.utils import OperationalError
from django.db.transaction import TransactionManagementError
```

InvenTree/part/migrations/0112_auto_20230525_1606.py
```diff
@@ -22,17 +22,22 @@ def delete_columns(apps, schema_editor):
...
-        schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN checkbox;")
-    except (AttributeError, FieldDoesNotExist):
-        print("Column 'checkbox' does not exist (skipping)")
+        schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN IF EXISTS checkbox;")
+    except (OperationalError):
+        # OperationalError caused by syntax not allowed in sqlite - try again without
+        try:
+            schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN checkbox;")
+        except (OperationalError):
+            print("Column 'checkbox' does not exist (skipping)")
```
matmair, line 31:
```suggestion
        schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN IF EXISTS checkbox;")
    except (OperationalError):
        # OperationalError caused by syntax not allowed in sqlite - try again without
        try:
            schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN checkbox;")
        except (OperationalError):
            print("Column 'checkbox' does not exist (skipping)")
    except (TransactionManagementError):
        # TransactionManagementError caused by trying to run a transaction inside a transaction - mysql
        in_atomic_block = schema_editor.connection.in_atomic_block
        schema_editor.connection.in_atomic_block = False
        try:
            migrations.RemoveField(
                model_name='part_partparametertemplate',
                name='checkbox',
            )
        except (FieldDoesNotExist):
            print("Column 'checkbox' does not exist (skipping) - mysql")
        finally:
            schema_editor.connection.in_atomic_block = in_atomic_block
```

InvenTree/part/migrations/0112_auto_20230525_1606.py
```diff
@@ -22,17 +22,22 @@ def delete_columns(apps, schema_editor):
...
-        PartParameterTemplate._meta.get_field('choices')
-        schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN choices;")
-    except (AttributeError, FieldDoesNotExist):
-        print("Column 'choices' does not exist (skipping)")
+        schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN IF EXISTS choices;")
+    except (OperationalError):
+        try:
+            schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN choices;")
+        except (OperationalError):
+            print("Column 'choices' does not exist (skipping)")
```
matmair, line 40:
```suggestion
        schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN IF EXISTS choices;")
    except (OperationalError):
        try:
            schema_editor.execute("ALTER TABLE part_partparametertemplate DROP COLUMN choices;")
        except (OperationalError):
            print("Column 'choices' does not exist (skipping)")
    except (TransactionManagementError):
        # TransactionManagementError caused by trying to run a transaction inside a transaction - mysql
        in_atomic_block = schema_editor.connection.in_atomic_block
        schema_editor.connection.in_atomic_block = False
        try:
            migrations.RemoveField(
                model_name='part_partparametertemplate',
                name='choices',
            )
        except (FieldDoesNotExist):
            print("Column 'choices' does not exist (skipping) - mysql")
        finally:
            schema_editor.connection.in_atomic_block = in_atomic_block
```

matmair:
The suggested changes fix the myqsl migrations on mysql and mariadb; on-prem and on github (docker running on ubuntu 20.04)

SchrodingersGat:
Another option:

The original PR which broke this renamed `part.migrations.0112_auto_20230531_1205.py` to `part.migrations.0113_auto_20230531_1205.py`

So, for anyone who ran `0112_auto_20230531_1205`, there should be a log in their migration history.

Anyone who did *not* run this (now deleted) migration is fine.

So, can we check if there is an entry in the migration log for this file?

matmair:
@SchrodingersGat that should also be possible

InvenTree/part/migrations/0112_auto_20230525_1606.py
miggland: I did not try it, and kept it from @SchrodingersGat original implementation.

However, since Django seems to not realise that the field is there (the reason #4977 doesn't quite work), I suspected using `RemoveField` wouldn't work either.

miggland: It seems I suspected correctly - since the fields do not show up in the list, `RemoveField` does not do anything - just like checking for the field in the list.

miggland:
> The suggested changes fix the myqsl migrations on mysql and mariadb; on-prem and on github (docker running on ubuntu 20.04)
```

Did you actually test this out, by going from a state before the botched migration to this? When I do that (see #4977 for details) I get this:
Checking for column 'choices' in table 'part_partparametertemplate'
  File "/home/inventree/data/env/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/home/inventree/data/env/lib/python3.9/site-packages/django/db/backends/mysql/base.py", line 73, in execute
    return self.cursor.execute(query, args)
  File "/usr/local/lib/python3.9/site-packages/MySQLdb/cursors.py", line 206, in execute
    res = self._query(query)
  File "/usr/local/lib/python3.9/site-packages/MySQLdb/cursors.py", line 319, in _query
    db.query(q)
  File "/usr/local/lib/python3.9/site-packages/MySQLdb/connections.py", line 254, in query
    _mysql.connection.query(self, query)
MySQLdb.OperationalError: (1060, "Duplicate column name 'checkbox'")
```

SchrodingersGat:
Interestingly the duplicate migration doesn't even error out in sqlite:

image

miggland:
I've added another suggestion, which should work for MySQL even on an existing DB.

@matmair Could you double-check that this doesn't break the test in mysql?

SchrodingersGat:
@wolflu05 I agree it is far from ideal but we have to make it seamless for the user.

SchrodingersGat:
@miggland your latest patch works for me on postgresql. I can successfully:

- Start with a fresh db from *before* your original commit (with the original migration applied)
- Migrate forwards (fields get deleted first)
- Migrate backwards
- Migrate forwards again (fields get skipped)

miggland:
> @miggland your latest patch works for me on postgresql. I can successfully:
> 
>     * Start with a fresh db from _before_ your original commit (with the original migration applied)
> 
>     * Migrate forwards (fields get deleted first)
> 
>     * Migrate backwards
> 
>     * Migrate forwards again (fields get skipped)

Awesome. Had not tried migrating backwards, but I did try it on postgres, sqlite and mysql in docker, going from before to after the broken migration.

SchrodingersGat:
@miggland @matmair  FYI I have an [alternative approach](https://github.com/inventree/InvenTree/pull/4987) which does not require any custom SQL

SchrodingersGat:
@miggland other PR is merged, feel free to close this out now

PR #4984, opened 2023-06-07
Fix reference to pack_size in stock migration

mattbnz:
5dd6f18 renamed the pack_size field of SupplierPart to pack_quantity, but this migration was missed in the change. Trying to migrate from a DB older than two weeks ago therefore fails.

### DIFF
--- a/InvenTree/stock/migrations/0094_auto_20230220_0025.py
+++ b/InvenTree/stock/migrations/0094_auto_20230220_0025.py
@@ -16,12 +16,12 @@ def fix_purchase_price(apps, schema_editor):
     Due to an existing bug, if a PurchaseOrderLineItem was received,
     which had:
 
-    a) A SupplierPart with a non-unity pack size
+    a) A SupplierPart with a non-unity pack quantity
     b) A defined purchase_price
 
     then the StockItem.purchase_price was not calculated correctly!
 
-    Specifically, the purchase_price was not divided through by the pack_size attribute.
+    Specifically, the purchase_price was not divided through by the pack_quantity attribute.
 
     This migration fixes this by looking through all stock items which:
 
@@ -57,7 +57,7 @@ def fix_purchase_price(apps, schema_editor):
             if line.part == item.supplier_part:
                 # Unit price matches original PurchaseOrder (and is thus incorrect)
                 if item.purchase_price == line.purchase_price:
-                    item.purchase_price /= item.supplier_part.pack_size
+                    item.purchase_price /= item.supplier_part.pack_quantity
                     item.save()
 
                     n_updated += 1


SchrodingersGat:
@mattbnz great catch, thanks for providing this :)

SchrodingersGat:
@mattbnz I'd like to understand this a little bit better; changes to data migrations can cause lots of headaches!

The migration you are changing, `stock.migrations.0094_` was written on `2023-02-20`, and references "supplier_part__pack_size".

However the migration which added the `supplier_part__pack_quantity` field is `company.migrations.0059_` which was added on `2023-05-19` and should not have been applied yet! Maybe they are being applied out of order here.

In your setup, can you please run `python ./manage.py showmigrations company` and `python ./manage.py showmigrations stock` to see where the migration state is?

I think that the *better* approach here is to ensure that the migrations are run in the correct order

mattbnz:
Trying to run `python ./manage.py showmigrations` before running `invoke update` fails with the following error: `sqlite3.OperationalError: no such table: part_partpricing`

Running the showmirgations command in the old environment (running on a very old June 2022 git checkout from  9b4e443289e45ea0e8100b42938145bfc3b18b9d) that I'm trying to upgrade from shows that company is up to `0045_alter_company_notes` and stock is up to `0078_add_stockitem_buildorder_detail` - complete showmigratinos output from this environment is at: https://pastebin.pl/view/4cd56dcf

Then I'm copying that database across to a new environment against the current git head, which gives the above error when trying to run showmigrations. Running update shows that the company migrations are indeed being completely applied before the stock migrations are applied - from my niave interpretation of the output it looks like it's going through the models in alphabetical order and applying all outstanding migrations for each model before proceeding to the next one, it's not applying migrations in date order as you might be expecting... 

Full output at https://pastebin.pl/view/406538b9

SchrodingersGat:
Given the current *excitement* we are having over in https://github.com/inventree/InvenTree/pull/4987 due to a bad patch on a migration file, I want to ensure this is done *perfectly*

Can you please try the following steps:

- BACKUP YOUR DATABASE
- Ensure your database is in the "old" state (before any updates
- Open the file [company.migrations.0059_supplier_part_pack_units.py](https://github.com/inventree/InvenTree/blob/master/InvenTree/company/migrations/0059_supplierpart_pack_units.py)
- Add the following dependency to the `dependencies` list on line 10: `('stock', '0094_auto_20230220_0025'),`
- Run your migrations
- Hopefully, should all pass now

These steps should ensure that `stock.0094` (which references the "pack_size" field) is run *before* `company.0059` (which then removes this field)

mattbnz:
No, I still get the original error. 

Thinking about this a bit more after a nights sleep, I think that makes sense - the issue is not so much the order the migrations are being run in, but the fact that a migration (regardless of what order it is running in) is trying to access an attribute on a model that no longer exists in the codebase that is being run... 

Put another way, the dependency is not on the data in the database (which migration ordering can solve), but on the code itself. Any attempt to execute the stock.0094_auto_20230220_0025 migration run in a codebase synced past 5dd6f18495e9899bb6646e7acff5c0c1d15c4762 is going to fail, because the codebase no longer contains SupplierPart.pack_size - so the migration cannot try and access that field regardless of database state/migration execution order. 

Basically by removing pack_size from the model, 5dd6f18495e9899bb6646e7acff5c0c1d15c4762 creates a situation where migrations that reference it *must* have been executed prior to that commit, or they will now fail. Presumably this is an inherent limitation on migrations and removing fields...  A quick search leads to https://code.djangoproject.com/ticket/23861 where Django itself seemed to hit this case, and whcih resulted in the following being added: https://docs.djangoproject.com/en/4.2/topics/migrations/#considerations-when-removing-model-fields

So I think you either need to accept the change to update the code of the migration (I agree risky, I haven't fully tested it in all scenarios), or ideally not remove the field and instead use the system_check_removed_details method to remove it instead of actually removing it, but even that seems like a lot of effort to revise at this point. 

I don't envy the situation sorry! FWIW, I'm past this now and not blocked, was just sending the PR to help others who might also be caught, but I guess up to you if the trade-offs are worth it or not, how common is it for InvenTree users to be a whole year behind like I was...

SchrodingersGat:
I think this has been addressed in #5027 so closing this one out.

Issue #4992, opened 2023-06-08
Uninstalling a tracked/serialized item does not return it to available stock


mattbnz:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When a tracked/serialized stockitem is installed in another item, there is an option to then `uninstall` it, however taking this option does not make it available to install into another item (the expected behaviour) because while the belongs_to field on the stockitem is cleared, the consumed_by field remains set to the value of the build order that constructed the item that it was originally installed into.

This behaviour was introduced by https://github.com/inventree/InvenTree/pull/4817, prior to this I believe the behaviour was as I expect - uninstalling a tracked/serialized item returns it to available stock ready for installation in another item. 

I think this is simpler to fix than the more general case in https://github.com/inventree/InvenTree/issues/4487 given the "uninstall" action is already explicitly acting on a single stockitem which the user is specifying they want back in stock (I think?, at least I do!)

If you agree the behaviour is a regression, I think the fix is as simple as clearing the consumed_by field at https://github.com/inventree/InvenTree/blob/master/InvenTree/stock/models.py#L1263

Shall I send a PR? 

### Steps to Reproduce

1. Create a serialized/tracked stockitem (A)
2. Install it into another item (B)
3. Uninstall (A) from (B)

### Expected behaviour

(A) should be available to be assigned to another item. I think this is both the previous behaviour and what is implied by the "uninstall into location" action - if the intent was to scrap/keep the item consumed, it's not going back into a location... 

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

Running against HEAD at a3150d9cb35bf24d4d27e1285180e1a9e75d0323

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

SchrodingersGat:
@mattbnz I agree with your approach that clearing the `consumed_by` field would be required here. Please submit a PR and I will review :)

PR #4994, opened 2023-06-08
Place uninstalled items back in stock

mattbnz:
reset the consumed_by field as well as belongs_to, so an uninstalled item can be reused.

Fixes: #4992


SchrodingersGat (approved):

SchrodingersGat:
Cheers for the quick fix @mattbnz  :)

PR #4814, opened 2023-05-14
[WIP] Add "project_code" field to Build model

SchrodingersGat:
(cherry picked from commit 0a1147b71d99c515e4b721e5a81708f7c70625a6)

Minimum working example to see what errors are thrown

PR #4997, opened 2023-06-08
Respect timezone when handeling git commit times

matmair:
This is a small fix for #4966 as local machine timezone offsets where not respected by the original PR.

### DIFF
--- a/InvenTree/InvenTree/version.py
+++ b/InvenTree/InvenTree/version.py
@@ -3,10 +3,11 @@
 Provides information on the current InvenTree version
 """
 
-import datetime
 import os
 import pathlib
 import re
+from datetime import datetime as dt
+from datetime import timedelta as td
 
 import django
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@matmair for reference this PR has broken docker builds:

https://github.com/inventree/InvenTree/actions/runs/5210919131/jobs/9402593907

This is the same issue we are seeing in:

- https://github.com/inventree/InvenTree/pull/4732
- https://github.com/inventree/InvenTree/pull/4808
- https://github.com/inventree/InvenTree/pull/4814

Looks like we really need to give this some attention!

matmair:
@SchrodingersGat I do not really see how this change could cause the errors - are you sure this is the cause?

SchrodingersGat:
> are you sure this is the cause?

Not completely - looking at the docker builds, this one ran and stopped some previous runs - so perhaps those previous runs hold the culprit. Will have to investigate tomorrow

Issue #4999, opened 2023-06-08
Fix broken docker build


SchrodingersGat:
Docker build is borked, needs fixing before next release.

https://github.com/inventree/InvenTree/pull/4997#issuecomment-1582513974

```[tasklist]
### Tasks
- [x] Check previous versions
- [x] https://github.com/inventree/InvenTree/actions/runs/5210537105/jobs/9405692470
```

matmair:
@SchrodingersGat I will watch the runs and restart them in the line of PRs

matmair:
Actually looking at the things you linked: The runs did not fail, they were stopped due to concurrency rules

matmair:
See:
* https://github.com/inventree/InvenTree/actions/runs/5210537105/jobs/9405692470
* https://github.com/inventree/InvenTree/actions/runs/5210674654/jobs/9402043420
* https://github.com/inventree/InvenTree/actions/runs/5210919131/jobs/9403505957

matmair:
This seems to be a known issue https://github.com/orgs/community/discussions/8336
@SchrodingersGat should we shut off concurrency-checks for now? I can create a PR for it

matmair:
@SchrodingersGat it seems like the failed docker builds were just cancelled runs - I have submitted a PR that would remove concurrency:
- pro: no failed runs because of cancellation
- con: unneeded builds (ie burning energy for nothing)

PR #5002, opened 2023-06-08
Remove concurrency checks

matmair:
Fixes https://github.com/inventree/InvenTree/issues/4999


SchrodingersGat:
@matmair I'm not convinced this is the right approach - I'd like to dig into the failed builds a bit further before we throw out the concurrency checks :)

PR #5005, opened 2023-06-08
Skip ready functions if not in main thread or plugins are not loaded yet

wolflu05:
⚠️ This PR touched the `gunicorn.conf.py` and added the `preload_app` option there. Maybe it's a good idea to mark this as a breaking change and list that change in the release notes. For docker this is no issue, but I'm not sure for installer and bare metal.

---

This improves the ready function of all apps. I started implementing this in the machines app (Ref: #4824)

Basically there are 4 different options to start the server. This PR (the `isInMainThread` check function) ensures that the `AppConfig.ready` function only gets called once for each start in all different starting scenarios:

1. `python manage.py runserver` - this starts a Django dev server with reload function. Django internally starts the app in two threads, the ready function only runs in the main thread
2. `python manage.py runserver --noreload` - like 1, but it only starts one thread and there is no reload capability, that way we can ignore this case, the ready function already only runs once
3. `invoke server` - basically like 1, but via invoke tasks with a `python manage.py check_for_db` before which is already filtered by `canAppAccessDatabase`
4. `gunicorn -c gunicorn.conf.py InvenTree.wsgi` - this starts inventree in a lot of different threads. The `preload_app = True` setting in the `gunicorn.conf.py` ensures that the ready functions are executed in the main thread before the application continues to execute in the couple of workers.

And secondly this PR should also execute the ready function's only once, because the plugin app reloads all Django apps if there is a AppMixin plugin. This is done by setting a flag after the reload has been initiated.


matmair (approved):
looks good - will increase performance at least a bit for sure

wolflu05:
Interesting why tests failed. Do you have an idea?

matmair:
Judging from the number of errors/failures something significant broke

SchrodingersGat:
It looks like the plugins aren't loading (at least in the test environment).

Do plugins still load locally for you?

wolflu05:
I'll need to test that locally if they load for tests. Running normally worked.

wolflu05:
I think the tests are somehow broken that is not related to my code. I now added debug statements and always return true. So there shouldn't be any change but psql tests are failing.

wolflu05:
I'm continuing my work here. (I adjusted the PR description to explain the changes a bit more). The isInMainThread thing is already working. Currently I'm stuck with the plugin registry apps check.

Is it correct, the AppMixins _activate_mixin function is the only function in the complete source code that triggers a full reload of the django INSTALLED_APPS?
And only if there is a plugin that uses the AppMixin, then its basically not possible with the current architecture to detect that and dont run the AppConfigs ready functions twice, because at the place where we have to decide if there is a second reload we dont have finished discovering plugins, so we cant tell.

- Solution 1: Always reload the apps again so we can always expect a second call to the ready functions
- Solution 2: Always run the ready functions only the first time and ignore the second time.

After implementing solution 2 I found out that this actually doesn't work by design. My intention behind this check was that e.g. machines are not initialized two times in the short starting period, but if I ignore the second time, there is no way plugins can provide custom drivers, because the machine registry is already initialized with missing drivers. (Same applies for notification methods - *that's I guess why the tests fail*)

So I have two options now:
1) use second ready call (*1. from above*)
2.1) call the machine init, notification methods discovery from the registry after all plugins are initialized

What do you think @matmair ?

matmair:
Touching the 2-load approach plugins currently take is playing with fire. A lot of the more advanced dog functions rely on a cleanly loaded app structure - which is why the reload exists. I do not see how AppMixins are supposed to work correctly without the reload.
You could implement solution 1 but I do not see how that is an improvement.
The problem is that you do not know if there is going to be a reload until Django itself is ready and the database can be accessed to check if AppMixins are even allowed and present.

What you could do in regards to #4824: move that django app behind plugins in the plugin stack and set a flag in the plugins ready function if the current load is the final one (it knows at the end of the plugin activations sequence) - only act trigger machine init functions then.

wolflu05:
You're right, this works now. But one thing is still not optimal. The plugin reload mechanism. I want that this also only triggers the ready methods once, but there is again an issue by design:

https://github.com/inventree/InvenTree/blob/804af73c2e4dc55f000856f45b552950d130eb14/InvenTree/plugin/registry.py#L212-L214

These lines unload all plugins which triggers the ready functions and then loads them which triggers the ready functions again. What we could do is:
```py
    with maintenance_mode_on():
           self.plugins_loaded = False
           self.unload_plugins(force_reload=force_reload)
           self.plugins_loaded = True
           self.load_plugins(full_reload=full_reload)
```

This will ensure that the unload_plugins function only reloads the apps and don't execute the init stuff in the ready functions, BUT if there is no (active) AppMixin plugin found by load_plugins, there won't be a second call what we expected at the beginning. 

@matmair Do you see a way we can tell if there is a second reload here?

wolflu05:
The easiest fix would be in `AppMixin._activate_mixin` to remove the `has_changed` check and then always reload there like we also do in `AppMixin._deactivate_mixin` (there is no `has_changed` check). Implemented in 2ff4141dd38ac2bd340e37cd783aee14c3d0f2e4

wolflu05:
This would be ready for review from my POV if there are no objections from @matmair

InvenTree/plugin/base/integration/AppMixin.py
```diff
@@ -36,7 +36,6 @@ def _activate_mixin(cls, registry, plugins, force_reload=False, full_reload: boo
 
         if settings.PLUGIN_TESTING or InvenTreeSetting.get_setting('ENABLE_PLUGINS_APP'):
             logger.info('Registering IntegrationPlugin apps')
-            apps_changed = False
```
matmair, line 39: This adds to the testing time significantly - when we added the check for testing it remove around 5 mins of runtime. With the reduced overall time the impact might be lower but needlessly reloading every test init seems like a waist.
If you want to remove the change check better replace it with a flag for testing

InvenTree/gunicorn.conf.py
```diff
@@ -8,3 +8,6 @@
 
 max_requests = 1000
 max_requests_jitter = 50
+
+# preload app so that the ready functions are only executed once
+preload_app = True
```
matmair, line 13: This will stop runtime reloads - while upstream InvenTree is not using that due to its simple plugins I know several plugins that do. This will be breaking any plugin that does so this needs to be mentioned prominently in the changelog. @wolflu05 have you measured how much this actually improves? This will force hard restarting the application when loading custom settings.py during runtime or importing global modules instead of allowing for a soft rollover of workers.

matmair (changes requested):
Looks good in general, a few minor questions remain

matmair:
> like we also do in `AppMixin._deactivate_mixin`

Deactiating uses nearly no ressources as it just cleans out a few variables. In my tests checking took more ressources then just deactivating (there was no settings caching at that point - that was probably the resaon).


> Do you see a way we can tell if there is a second reload here

A flag in global scope (like the registry) in a django app that is always included would probably do the trick

matmair:
I am still contemplating if it wouldn't be a better solution to defer machine initialisation to a point after all apps are ready, I do not see a scenarion in which a machine would need to be loaded at the point of the first request to the app.

InvenTree/gunicorn.conf.py
wolflu05: > have you measured how much this actually improves?

This is the whole key point. Without that change, the machine init function would run in all workers which is definitely not the expected bahavior.

wolflu05:
> Deactiating uses nearly no ressources as it just cleans out a few variables.

No, it also reloads the apps.

> A flag in global scope (like the registry) in a django app that is always included would probably do the trick

I do not see how this is possible. If the apps change, we need the load_plugins function to see if there are now apps that require a reload. So we can't tell if the unload or also the load reloads plugins 


> I am still contemplating if it wouldn't be a better solution to defer machine initialisation to a point after all apps are ready, I do not see a scenarion in which a machine would need to be loaded at the point of the first request to the app.

I think that was what I suggested with 2.1. but I do not see now how I can solve this after a reload was triggered.


The only thing I could do is add this check if we are in testing like you suggested, but I think you don't like this solution. Maybe could you please take a look how you would solve this issue?

matmair (approved):
It is foreseeable that a lot of users will not change their guncorn.py file so whatever system you build upon this PR should deal with that case in a non-catastrophic way.

Not really fond of the added overhead in testing but I currently do not have the required capacity to work on this.

InvenTree/plugin/base/integration/AppMixin.py
```diff
@@ -46,8 +46,12 @@ def _activate_mixin(cls, registry, plugins, force_reload=False, full_reload: boo
                         settings.INSTALLED_APPS += [plugin_path]
                         registry.installed_apps += [plugin_path]
                         apps_changed = True
+
             # if apps were changed or force loading base apps -> reload
-            if apps_changed or force_reload:
+            # Ignore reloading if we are in testing mode and apps are unchanged
```
wolflu05, line 51:
```suggestion
            # Ignore reloading if we are in testing mode and apps are unchanged so that tests run faster
```

wolflu05:
What are we gonna do with this PR? From my POV this would be ready finally for merge. @matmair @SchrodingersGat

matmair:
LGTM

PR #5007, opened 2023-06-08
Refactor: Dockerimage

### LINKED ISSUE #4924: Reduce size of docker image
InvenTree docker image is quite large, and could *probably* be reduced significantly in size.

But, requires someone to look at it closely...

wolflu05:
This PR refactors the docker image to reduce its size dramatically. It now weights `524MB`.

```[tasklist]
# TODO
- [x] remove gnupg
- [x] rewrite init script with ash and remove bash
- [x] test with existing installation
- [x] test with all dbms
- [x] figure out how to get the other db drivers to work, there are some native dependencies missing
```

image

fixes #4924


wolflu05:
Is git actually required in the image now? I would say yes, as some plugins are installed via git.

SchrodingersGat:
> Is git actually required in the image now? I would say yes, as some plugins are installed via git.

To install via git we would need the git executable, yeah

wolflu05:
Now the image weights about 629MB. I'd love if we could reduce the needed dependencies for the database drivers.

Dockerfile
```diff
@@ -52,26 +52,34 @@ LABEL org.label-schema.schema-version="1.0" \
...
 # RUN apt-get upgrade && apt-get update
-RUN apt-get update
 
 # Install required system packages
-RUN apt-get install -y  --no-install-recommends \
-    git gcc g++ gettext gnupg libffi-dev libssl-dev \
+RUN apt-get update && apt-get install -y  --no-install-recommends \
+    git gnupg gettext \
+    # Image format support
+    # libjpeg-dev webp libwebp-dev \
```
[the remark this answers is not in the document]
SchrodingersGat: @wolflu05 you mean the jpeg and webp packages? They are needed so we can handle a wider range of modern web image formats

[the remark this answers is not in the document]
wolflu05: Ok. So they need to be in.

wolflu05:
What's really interesting and what could be used to consider using an alpine image would be this [Dockerfile](https://github.com/TandoorRecipes/recipes/blob/develop/Dockerfile). It's an other open source project that's also based on Django and also has to deal with lots of images (from your cooked meals) and also has ldap support already built in and weights out 350mb (uncompressed).

SchrodingersGat:
@wolflu05 two things immediately stick out in the linked image

- Running purge step after install
- Running all install steps on the one line

Might be worth applying that to our image

wolflu05:
> Running purge step after install

I already do this as much as I can.

> Running all install steps on the one line

I tried squashing all image layers, but this also doesn't reduce the size of the image.

wolflu05:
I just tried how alpine turns out. Here is a little comparison with the current <TODO add commit ref> state of pushed files:

| Name | Size | Build time \* |
|--------|--------|--------|
| Old | `1.03GB` | `119.0s` |
| New debian slim based | `778MB` | `135.8s` |
| New alpine based | `556MB` | `141.8s` |

\* all builds on MacBook Pro 2019 (docker cache disabled)

SchrodingersGat:
The alpine image is looking pretty appealing then!

wolflu05:
FYI, adding LDAP needed packages adds about 5MB to the alpine image, so it weights: `561MB`. And about 5MB to the Debian image so it weights: `781MB`.

wolflu05:
Interested to hear what @matmair has to say to this alpine approach.

SchrodingersGat:
It certainly saves a *lot* of bandwidth. And the LDAP addition seems minor, but I have no horse in that race :)

matmair:
If it works why not. Have you verified that ldap actually resolved all objectTypes and both ldap and ldaps with the installed dependencies?

The size reduction is certainly impressive. Is there something significantly different to the linked project? They seem to be even smaller so I wonder how much we need to change to achieve that size. With the amount the image gets pulled this PR certainly has positive carbon-footprint implications.

The change from Debian to Alpin should to be marked breaking and be mentioned in the release notes, plugins will need to adapt their (native) install instructions to work on the docker as apt will not work.

wolflu05:
> If it works why not. Have you verified that ldap actually resolved all objectTypes and both ldap and ldaps with the installed dependencies?

Not fully, need some further testing.

> The size reduction is very impressive. Is there something significantly different to the linked project? They seem to be even smaller so I wonder how much we need to change to achieve that size. With the amount the image gets pulled this PR certainly has positive carbon-footprint implications.

They don't use lots of things we need:
- weasyprint
- mariadb
- git
- gnupg
- bash (maybe we rewrite the init script to use ash)

> The change from Debian to Alpin should to be marked breaking and mentioned in the release notes, plugins will need to adapt their (native) install instructions to work on the docker as apt will not work.

Yes definitely it will break some things.

**Idea:** provide both alpine and Debian for the next release for a smooth transition period until plugins are adapted.

**Further analysis needed:** Does switching out the images work easily with an already installed InvenTree. Maybe the venv that's mounted via volume contains some files that are not compatible with alpine.

SchrodingersGat:
> Further analysis needed: Does switching out the images work easily with an already installed InvenTree. Maybe the venv that's mounted via volume contains some files that are not compatible with alpine.

We definitely need to investigate this before we push this :)

wolflu05:
Oh and I forgot the biggest point why our alpine image is bigger than tandoors. Our codebase weights about 90MB with all the translated templates and po files.

matmair:
> Oh and I forgot the biggest point why our alpine image is bigger than tandoors. Our codebase weights about 90MB with all the translated templates and po files.

That should be solved by #2789 as we will only ship messages

matmair:
@wolflu05 I think we can remove gnupg - we do not ship the code that used it anymore.

Would be interested in your thoughts on https://github.com/inventree/InvenTree/pull/5017 - [background](https://www.docker.com/blog/generate-sboms-with-buildkit/)

SchrodingersGat:
@wolflu05 just a friendly bump to see where you are up to with this :)

wolflu05:
Need to remove gnupg and do some more testing with other db drivers and also with existing installs.

matmair:
@wolflu05 do you want/need any support on this one? Would like to use this for #5011

wolflu05:
I'll need to update a few things and then would be good if you could test it too so we don't oversee something

wolflu05:
Should I replace the old dockerfile and put the dev thing into a Dockerfile.dev so we still use Debian for development?

SchrodingersGat:
> * Should I replace the old Dockerfile with the alpine one and put the dev thing into a Dockerfile.dev so we still use Debian for development?
> * should I remove bash from the alpine image and rewrite the init script to use ash so we even save some more mb?

1. No, I don't think we should have a different base image for dev or prod
2. I think so, I don't see any real reason to keep bash

matmair:
So we are clear: We are fully switching to Alpine? I am implementing the build step for react rn and do not want to use the wrong image ;-)

wolflu05:
I would think so from my side. What do you mean @SchrodingersGat .


But nevertheless, it's totally irrelevant for you with what base image we went later. You can use the `FROM node:18 as frontend_builder` image for building and then a `COPY --from frontend_builder <...> <...>` to copy in whatever our prod image is.

matmair:
The question is in which file as there is `Dockerfile` and `Dockerfile.alpine`. I will do it in the current file for now.

wolflu05:
I would plan to remove the .alpine extension (replace the old file), was first just for trying it out to keep both side by side for comparison.

wolflu05:
Are we still want to support the dev target from our current image? I don't think it's necessary as we have devcontainer and native install.

SchrodingersGat:
> I would think so from my side. What do you mean @SchrodingersGat .

I support moving entirely to alpine for base image

> Are we still want to support the dev target from our current image? I don't think it's necessary as we have devcontainer and native install.

I suppose this is a fair point? But is there a downside to leaving it in, at least for a while, for anyone who would still be using this?

wolflu05:
> I suppose this is a fair point? But is there a downside to leaving it in, at least for a while, for anyone who would still be using this?

I plan to replace the current Dockerfile completely with the .alpine one. And in the alpine one is no dev target. There is just the production target. So it would be more effort to recreate a dev target with alpine smoothly.

wolflu05:
@matmair @SchrodingersGat I'm done now with the changes I wanted to do. The image now weights `524MB`. The biggest things are (estimated):
- weasyprint, fonts, ... (`~100MB`)
- mysql related stuff (`~70MB`)
- translated js templates (`~50MB`)
- git (`~20MB`)

### 🚧 Help needed
I already test a lot of scenarios with existing installs, other dbms, already installed plugins, ... and everything worked so far. Before this is ready to merge, I'd like to ask if you could test this too so we do not overlook something.

SchrodingersGat:
@wolflu05 awesome :) I'll take a look at this tonight

SchrodingersGat:
The development image works as expected on my end. @matmair any final thoughts before we merge this in?

matmair:
LGTM; I take it that we are fully removing support for the Raspberry Pi architecture with this as the package resources are removed?

Dockerfile
```diff
@@ -51,44 +53,48 @@ LABEL org.label-schema.schema-version="1.0" \
...
-# RUN apt-get upgrade && apt-get update
-RUN apt-get update
-
-# Install required system packages
-RUN apt-get install -y  --no-install-recommends \
-    git gcc g++ gettext gnupg libffi-dev libssl-dev \
-    # Weasyprint requirements : https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#debian-11
-    poppler-utils libpango-1.0-0 libpangoft2-1.0-0 \
+RUN apk add --no-cache \
+    git gettext py-cryptography \
```
matmair, line 57: Not sure we need git any more

Dockerfile
```diff
@@ -98,33 +104,14 @@ ENV INVENTREE_COMMIT_HASH="${commit_hash}"
...
-#RUN adduser inventree
-#RUN chown -R inventree:inventree ${INVENTREE_HOME}
-#USER inventree
-
-# Install InvenTree packages
-RUN pip3 install --user --disable-pip-version-check -r ${INVENTREE_HOME}/requirements.txt
-
-# Server init entrypoint
-ENTRYPOINT ["/bin/bash", "./init.sh"]
+COPY InvenTree ./InvenTree
```
matmair, line 107: Should we clean up what we ship in a next step? Ie. remove contrib things, assets and anything else that is not needed to run the app.

matmair (approved):

Dockerfile
SchrodingersGat: What about to install a plugin from a remote git repo?

matmair: I do not think it is a good idea to push that way to install plugins; discovery is slower and the support for nested plugins is limited; pip as a distribution way is way cleaner

wolflu05: I would argue to keep git and add a big note somewhere to the plugin dev
section. Sometimes you just want to test something and then git is easier.

Dockerfile
wolflu05: I don't think there is much we can remove, we already only copy the InvenTree folder.

Dockerfile
SchrodingersGat: > pip as a distribution way is way cleaner 

Absolutely agreed, but there may be users who have their own plugins not available via pip?

SchrodingersGat:
> LGTM; I take it that we are fully removing support for the Raspberry Pi architecture with this as the package resources are removed?

@wolflu05 does removing RPi support gain us any space? I would imagine not? Perhaps it is beneficial to leave it in for anyone who wants to target raspbian

wolflu05:
Not sure, haven't tested, because I have no access to one, should I add it and Someone of you test it?

SchrodingersGat:
> Not sure, haven't tested, because I have no access to one, should I add it and Someone of you test it?

I don't have a way to test it really, but I think maybe we should just leave those lines in. Unless you are building to target raspbian, they don't *do* anything

SchrodingersGat:
@wolflu05 thanks for the fine work here :)

SchrodingersGat:
@wolflu05 this PR truly was a *breaking* change: https://github.com/inventree/InvenTree/actions/runs/5506646753/jobs/10035587151

Looks like we did not run through the complete set of unit tests

PR #5011, opened 2023-06-09
Platform UI - React integration

### LINKED ISSUE #2789: [FR] Switch UI to React

### LINKED ISSUE #3901: Epic: React

matmair:
This PR introduces the base for the new UI based on React. The _current_ UI is referred 'classic' while the new React one is named 'platform' - as it is based on a new UI platform.
The UI itself is de facto empty - it is just a shell to show how the mechanism works. There will be a parallel PR with a first basic interface.

Detailed mechanisms:
- there are env/config.yaml mechanisms to select which UI urls should be served
- there are invoke tasks to build the code for the frontend and collect it into static
- there are yarn tasks to extract/compile language files (based on gettext)

Missing:
- [ ] dev docs
- [x] min version of node. technically 16 but I would like to set a higher one (19/20) from the start
- [x] build step for docker (multistep - to not include node on the image) maybe after https://github.com/inventree/InvenTree/pull/5007
- [ ] build step for releases
- [x] merging of translation files (or mapping - still not sure about that)

Fixes #3901 Fixes #2789


InvenTree/web/urls.py
```diff
@@ -0,0 +1,26 @@
...
+        )
+
+
+spa_view = ensure_csrf_cookie(TemplateView.as_view(template_name="web/index.html"))
+
+
+urlpatterns = [
+    path("assets/<path:path>", RedirectAssetView.as_view()),
+    re_path(r"^(?P<path>.*)/$", spa_view),
+    path("", spa_view),
```
wolflu05, line 25: What is the advantage about passing this as a template instead of using only static files for the spa? Doesn't this also have an impact about performance?

src/frontend/public/inventree.svg
wolflu05, line 1: I suggest to use a plain (maybe also compressed export- so everything is in one line) export here, as this file is much smaller than. (Also everywhere else that logo is used. For an Inkscape editable we could commit it somewhere else)

InvenTree/web/urls.py
matmair: CSRF protection and session management mostly. The data is also only loaded once/redirected so the proxy will only load full loads from the backend, everything else is cached/served from static.

src/frontend/public/inventree.svg
matmair: will be interesting later on, I would prefer to have a functioning UI before optimizing a potential dead-end

SchrodingersGat:
Exciting times!

Regarding translation - we will need to be careful that during this transition process we don't delete any existing translated strings (as much as we possibly can). Otherwise we lose all the great work that has been done on crowdin.

SchrodingersGat:
@matmair when this is ready to look at, can you provide a set of instructions to build and run a MVP?

SchrodingersGat:
@matmair is this ready to look at? How should I go about getting this running locally - anything I need to build?

matmair:
Ok, @inventree/triage this should be ready for a first review now. It should just *(tm)*  work with dev containers.
The `inv update` command was modified to include everything necessary to set up and build the frontend. 

Missing:
- [ ] dev docs
- [ ] min version of node. technically 16 but I would like to set a higher one (19/20) from the start
- [ ] build step for docker (multistep - to not include node on the image) maybe after #5007 
- [ ] build step for releases
- [ ] merging of translation files (or mapping - still not sure about that)

.devcontainer/postCreateCommand.sh
```diff
@@ -16,6 +16,7 @@ python3 -m venv dev/venv
 pip install invoke
 invoke update
 invoke setup-dev
+invoke frontend-install
```
wolflu05, line 19: Won't we add this task to the setup dev?

wolflu05:
We already sort python imports, maybe it's worth sorting ts imports to. There is [this](https://github.com/trivago/prettier-plugin-sort-imports) prettier plugin.

wolflu05:
Is there already a plan what to use for api requests? I can recommend [TanStack/query](https://github.com/TanStack/query). This offers a lot of caching, state management, refetching already out of the box. One thing we should also consider is building something to auto generate ts types from the open api definition. This would help us identify errors if we change some fields on the backend that are referenced in the fronted, because ts then would throw an error.

.devcontainer/postCreateCommand.sh
matmair: You don't really need the frontend deps to do backend dev so my idea was to only install when it is needed.

matmair:
The frontend stack is already kinda done, I am using axios and react-query for now. The branch is [here](https://github.com/matmair/InvenTree/tree/platform-ui).
I thought about generating bindings but it created huge diffs in testing. My plan is to just test as much as possible of the frontend to catch this kind of issue. Still searching for a good solution to track frontend coverage.

wolflu05:
> My plan is to just test as much as possible of the frontend to catch this kind of issue.

I'm not sure if tests are the best way for eliminating that issue. They take a long time to run and also set up, especially e2e tests. I think if ts is set up correctly there are not that much bugs that could be catched with e2e tests that couldn't be also catched with typescript and im not sure if maintaining e2e tests are worth that low improvement.

What do you mean with huge diffs? You mean because generating generates them in different orders each time? We could sort them by name so the order is the same every time.

.devcontainer/postCreateCommand.sh
wolflu05: Yeah, maybe it's worth spearing them  like setup-dev-frontend and setup-dev-backend and then let the setup-dev do both. I guess there are a lot of beginners which are not sure what they need and later ask why they get errors changing something on the frontend.

matmair:
@wolflu05 a thousand lines in the git diff/commit for a simple endpoint addition/move. I do not see how ordering will resolve the problem - the openAPI changes themselves easily get sizeable. I tested with [this](https://openapi-generator.tech/) generator.

wolflu05:
But do you think it's better to define all types manually each time there are some changes? Maybe someone only changes backend logic and the ts types are forgotten. I would love if we would have something auto generated.

wolflu05:
ok, so would you be ok with creating a FR for auto generated types and do this later if we already have something running, its not ideal, but I think definitely the right step.

matmair:
Sure, we have a ton of `horizon` milestones. I like the idea but recognise the work needed to get something stable. A clean schema could also be used for the Python API (and maybe more languages).

SchrodingersGat:
@matmair ok getting setup to test this locally, have installed rpm / node / yarn.

```bash
root@36e94a55b060:/home/inventree# invoke frontend-compile
Installing frontend dependencies
00h00m00s 0/0: : ERROR: [Errno 2] No such file or directory: 'install'
```

wolflu05:
@SchrodingersGat This works fine for me, tested by just opening the branch in codespaces. Google shows that there is something wrong with your yarn setup. [See](https://stackoverflow.com/questions/46013544/yarn-install-command-error-no-such-file-or-directory-install)

wolflu05:
@SchrodingersGat you need to be at the root of the repo. The task steps into the directory with the package.json file where the `build` command is defined at `scripts.build`

.pre-commit-config.yaml
```diff
@@ -54,3 +54,23 @@ repos:
...
           docs/docs/webfonts/.*
       )$
+- repo: https://github.com/pre-commit/mirrors-prettier
+  rev: "v3.0.0-alpha.9-for-vscode"
+  hooks:
+    - id: prettier
+      files: ^src/frontend/.*\.(js|jsx|ts|tsx)$
+      additional_dependencies:
+      - "prettier@^2.4.1"
+      - "@trivago/prettier-plugin-sort-imports"
```
wolflu05, line 64: Are you sure there is no additional configuration needed about the groups, I always set sort groups manually where I used this plugin.

SchrodingersGat:
@matmair we'll need to add the required node / yarn dependencies to the development Dockerfiles

.pre-commit-config.yaml
matmair: In my tests, it sorted without groups set but I will add the default

matmair:
@SchrodingersGat I already added it to the dev container - shipping it with the distributed container will enlarge them, not sure if that is required.

SchrodingersGat:
This is not the distributed (production) images though. Just the development image for those who use it for dev

.pre-commit-config.yaml
matmair: Add ed

wolflu05:
If I just open this branch with codespaces, I end up with a non clean working tree.

```bash
On branch plattform
Your branch is up to date with 'origin/plattform'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   .pre-commit-config.yaml
        modified:   src/frontend/src/locales/de/messages.d.ts
        modified:   src/frontend/src/locales/de/messages.ts
        modified:   src/frontend/src/locales/en/messages.d.ts
        modified:   src/frontend/src/locales/en/messages.ts
        modified:   src/frontend/src/locales/hu/messages.d.ts
        modified:   src/frontend/src/locales/hu/messages.ts
        modified:   src/frontend/src/locales/pseudo-LOCALE/messages.d.ts
        modified:   src/frontend/src/locales/pseudo-LOCALE/messages.ts

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        InvenTree/web/static/

no changes added to commit (use "git add" and/or "git commit -a") 
```

SchrodingersGat (approved):

wolflu05:
Why are you removing eslint and prettier from the dev dependencies?

matmair:
We do not need them; already running with pre-commit

wolflu05:
Are you sure the vscode prettier and eslint extensions work? They need to have prettier and eslint installed in node_modules as far as is remember.

matmair:
I have neither tested that nor plan to. formatting and lining is done by pre-commit both on the Django side and the new UI.

wolflu05:
I highly recommend trying it out. It saves so much time, it directly highlights if something is wrong. Also you can type without caring about indentation in tsx and in combination with Format ok save, you just hit cmd+s and everything looks pretty. That doesn't mean I don't use the precommit, that's the last check for me if I overlooked something.

Dockerfile
```diff
@@ -84,6 +84,15 @@ RUN \
 COPY ./docker/requirements.txt base_requirements.txt
 RUN pip install --disable-pip-version-check -U -r base_requirements.txt
 
+# Frontend builder
+
+FROM inventree_base as frontend
```
wolflu05, line 89: Why don't you use the official node image as base?

matmair:
Most extensions do not work well with web IDE which means I can not use them at work so my interest is limited

Dockerfile
matmair: We need a full InvenTree environment for tags to work

wolflu05: Sorry, which tags? vite build should run fine?

matmair: Django template tags - we need those for plugins rendering panels

wolflu05: Sorry, do you have an example of such a usage already?

matmair: Nothing I have time to polish rn - plugins are not the focus of this PR.

matmair:
@SchrodingersGat can we push docs/bare metal releases back and merge this in early?

Dockerfile
```diff
@@ -98,6 +98,17 @@ RUN chmod +x init.sh
 
 ENTRYPOINT ["/bin/sh", "./init.sh"]
 
+# Frontend builder
+
+FROM inventree_base as frontend
+
+RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash - && apt-get install -y nodejs npm && apt-get autoclean && apt-get autoremove && \
```
wolflu05, line 105: Installing nodejs and npm is as easy as just doing `apk add nodejs npm` (apt is not available, as the base image now is alpine). See official nodejs [install docs](https://nodejs.org/en/download/package-manager#alpine-linux) for alpine. Same issue below at the dev target.

wolflu05 (commented):
I still have the issue with a non clean working tree if I just open your branch freshly in a new devcontainer.

Dockerfile
matmair: replace them

matmair:
We have to choose between guaranteed up-to-date pseudo language files or a 'clean ' working branch on the first start. I would rather have the first.

wolflu05:
1. What are these pseudo languages?
2. the translations are not all, also the build output is not in the gitignore

SchrodingersGat:
@matmair I'd like to wait for response to @wolflu05 questions above :)

matmair:
@wolflu05 
1. they are required to check if all parts of the UI are translateable
2. if does not make sense to split out one language, will just lead to inconsistencies
3. yes because the compiled static need to be included to run without node

Those files should be committed - I just removed them for this PR because the UI is empty rn and it makes reading the diff harder without adding any benefit. It is the same as with the language files - we automated that part but at some point they need to be updated and included.

matmair:
I would appreciate it if you would not extend your content meaningful after commenting by editing - I mainly read on mails to decide if I need to answer and do not get updates for edits @wolflu05

wolflu05:
> Those files should be committed - I just removed them for this PR because the UI is empty rn and it makes reading the diff harder without adding any benefit. It is the same as with the language files - we automated that part but at some point they need to be updated and included.

I don't like commiting the build output. As you said, the diff gets really huge, that means also the repo size rapidly grows => it takes an eternity to clone the repo, currently it already needs 40s to clone. I would say that node is required if you develop frontend parts and if you don't like to install it, use the devcontainer. It's a really bad idea if this is the only argument to commit these huge diffs later.


And regarding the languages and pseudo languages, you plan to commit them before merge?

SchrodingersGat:
With regards to translation files:

- We need to be very careful that we don't remove the current string translations which have been done on crowdin
- Any *new* translations (i.e. in react) should match the strings used in existing translations
- We need to very carefully ensure that adding in new translation files happens in a way which is compatible with our current translation setup
- At some point, we'll remove translation for html templates and js templates (way down the track)

@matmair with this in mind, can you please briefly outline how the translation files are calculated / stored in the new framework? So we can determine how we plan for integration with crowdin.

matmair:
@wolflu05 if you see another way to distribute the front end I am open to ideas. The alternative way still has to work for bare metal installs - we would brake the upgrade path for most users otherwise judging from the issues.

matmair:
@SchrodingersGat translations work basically the same as they do with the current frontend. Strings are discovered when the frontend-trans task is run, placed in .po files, those are picked up by crowdin, translated, pushed from crowdin to the translation branch and then merged into master before release.



> * We need to be very careful that we don't remove the current string translations which have been done on crowdin

the new translations would be separate, like the app translations




> * Any _new_ translations (i.e. in react) should match the strings used in existing translations

Not sure this will be possible - a lot of the interface is underdocumented so there will probably be a lot of new tooltips, menu entries etc.



> * We need to very carefully ensure that adding in new translation files happens in a way which is compatible with our current translation setup

Uses the same mechanism as the current setup



> * At some point, we'll remove translation for html templates and js templates (way down the track)

A lot of strings are also used by the backend so that will not remove a lot of translatable files - a bunch of *glue* pieces for the current way of translations will be removed tough

SchrodingersGat:
@matmair ok thanks, I'm happy with that. Sounds like we have a way forward at least

SchrodingersGat (approved):

SchrodingersGat:
Anything holding us back from merging at this point?

wolflu05:
> the front end I am open to ideas. The alternative way still has to work for bare metal installs - we would brake the upgrade path for most users otherwise judging from the issues.

We can build them and upload a zip file with the point releases. If you want to install from a none point release, require node. I know a lot of softwares that require node for their frontend build (e.g. gitea). And for apt packages, I guess there is already some build step where this could be build.

SchrodingersGat:
My feeling is that the compiled web files should *not* be included in the repo. If we run through the different ways to use InvenTree:

| Setup Method | How to get src files |
| Docker | Compiled as part of docker image build |
| Installer | Compiled as part of install process | 
| Bare Metal | Compiled as part of install procedure |
| Devcontainer | Probably not needed, run using yarn? |
| Other ... | ??? |

matmair:
Compiling the frontend on the server with bare metal and installer means that every user needs to install thousands of external packages on their machine. If only one of those is vulnerable (pretty likely) web proxies or on-device scanners will trigger - effectively making inventree uninstallable in the enterprise/goverment or for responsible sysadmins.
If this is the plan I need to hand-break my involvement in this whole thing hard as all usability in my deployments is gone. Requiring nodejs on the server means no installation in my environments - the whole frontend integration is designed around this requirement. If the user installs nodejs you could just use the build output of that but at that point, I can not deploy anymore. Also, the installation footprint shoots up significantly if you need all the modules on the server - that is above 1gb of unneeded junk for a 5-10 mb frontend.

wolflu05:
Why not upload a build output to each point release as a zip file?

SchrodingersGat:
> Compiling the frontend on the server with bare metal and installer means that every user needs to install thousands of external packages on their machine. If only one of those is vulnerable (pretty likely) web proxies or on-device scanners will trigger - effectively making inventree uninstallable in the enterprise/goverment or for responsible sysadmins.
> If this is the plan I need to hand-break my involvement in this whole thing hard as all usability in my deployments is gone. Requiring nodejs on the server means no installation in my environments - the whole frontend integration is designed around this requirement. If the user installs nodejs you could just use the build output of that but at that point, I can not deploy anymore. Also, the installation footprint shoots up significantly if you need all the modules on the server - that is above 1gb of unneeded junk for a 5-10 mb frontend.

You make some great points here - the security implications in particular I had not considered.

I had been considering this front-end code in a similar way to how we currently render static files with `invoke static`. But if we are not going to be shipping node at all, as it is not required to actually *serve* the compiled frontend code, then this makes sense.

I'm happy to move forward with compiled front-end files "baked in". Let's keep this going, this is a super important path forward for the project.

wolflu05:
I see the points with the vulnerabilities, but really, we shouldn't commit the build output. There are a lot of downsides:
1. The repo contains 14.5k commits and the folder is huge, `209MB` currently. Cloning takes about 40s on my machine. If we commit the build files we dramatically increase that time and also the size.
2. PRs contain huge diffs of build changes
3. The build output is not very consistent, if you change only a few components it can result in a changed build output from ground up, because e.g. you have different versions of node packages. The build output files are normally minified, so they are all in one line
4. Merging get a PITA, especially with multiple concurrent merges as every pr changes these files

### How to properly solve this problem:
As stated by @SchrodingersGat we have a few targets. My suggesting would be the following:

We have to distinguish between development and production. 

#### Development
I think for development on the frontend it's not a big deal to install nodejs, you need it anyway. If you only want to do backend, go ahead and use some API client like postman or the integrated one. For devcontainer, its not a big deal the devcontainer contains node anyway.

#### Production
| Setup method | How to deliver the static build output |
| --- | --- |
| Docker ✅ | Compiled as part of docker image multistage build |
| Installer ❓ | Compiled as part of the deb package build process |
| Bare Metal ❓ | Compiled as part of a release pipeline and uploaded as a zip folder to the GitHub releases. The setup instructions could then just download the zip folder, extract it and place it inside static |
| Bare Metal (no point release) ❓ | Can't help you, you need node, or at least docker and build it yourself, very small user base and also not the recommended setup for production |

Please, evaluate the downsides, they are big for me. I really don't like to have them in here. I really know we want to move this forward, but once we decide to commit the repo will definitely grow fast and we can't reverse it.

SchrodingersGat:
Maybe due to me being a complete react noob, but is it possible you guys are talking about different things?

@wolflu05 you are talking about 200mb worth of files - which directory, specifically, and what purpose do these files serve?

@matmair you are talking about 10mb worth of files - which directory, specifically, and what purpose do these files serve?

wolflu05:
I'm talking about the files in `InvenTree/web/static/` that are generated by `invoke frontend-build`. They are only 5-20MB even the project grows, but committing a diff in those files only a couple of times increases the repo size, commit diff, checkout time by a lot.

matmair:
I would rather have devs download a bigger repo from time to time then most users (especially the ones that have been using this for years when bare metal was the only real deployment option) dealing with node. Manual changes to the update methodology are about the most brutal thing you can do to a sysadmin.

@SchrodingersGat I am talking about the real files that are acutally present

@wolflu05 with that argument we should stop translations

matmair:
I will step back from discussions for a few days as I do not feel that it will be productive in the end. Favouring one deployment method because of the lack of the others will only lead to more support load and if that is the result I'd rather ship the frontend as a plugin or reconsider the rewrite.

SchrodingersGat:
@matmair @wolflu05 are we happy to move on from here, and merge this (without the compiled front-end files), with the understanding that this code will not go into production without a complete, robust method for distribution that does not require the user to compile from node?

matmair:
fine with me @SchrodingersGat

wolflu05:
Yeah, also fine with me, merging this now definitely simplifies things e.g. the huge diff in the other PRs.

SchrodingersGat:
Fantastic :) @matmair are there any other changes you want to make here, or happy to merge as-is?

matmair:
Yes @SchrodingersGat let's merge

PR #5019, opened 2023-06-11
Add metadata to plugin configs

matmair:
This PR adds metadata for plugin configs. Thought we already have it, now we actually do.


matmair:
@SchrodingersGat this is ready for review and merge; I have removed the metadata from the settings page, with a large enough body of JSON it started to crash randomly.

wolflu05:
Crash only for the plugins? Or also on every other model that has metadata in the admin settings page?

matmair:
@wolflu05 i have only noticed on plugins after loading 2mb of json 😄

Issue #4833, opened 2023-05-16
[FR] display installed sub-items in stock views


T0jan:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

It would be nice if one could check the installed items into another item from the stock view or installed items view. Consider and example where you have `device A` where `device 1` and `device 2` are installed into and `device 1` then has two items `device X` and `device Y` installed. Tree view:

                      device A
                     /        \
             device 1           device 2
            /        \
    device X          device Y

### Suggested solution

it would be very nice to have a `load installed items` button similar to the `load bom for subassambly` which would would load and display the items for any level the user wants to see.

### Describe alternatives you've considered

an alternative solution could be to permanently add the info of installed subitems to the item/stock view as additional column. But I think this would limit the usefulness as you only could view one level lower with this approach without making it a cluttered table.

### Do you want to develop this?

- [ ] I want to develop this.

matmair:
Interesting idea @T0jan; I am not sure how this will work from a performance view - especially with larger BOMs. Interested to see what others have to say about this.

SchrodingersGat:
A "load on demand" button (like in the BOM view) would work well here. We'd need to annotate a "installed_items" count to the stock item serializer to determine which lines get buttons. But should not be much of a hit on performance. 

@T0jan do you want to implement this? I'm happy to run you through the basics and point to likely areas of attack in the code.

matmair:
I think this makes sense as an extra endpoint for the stockitem to keep the main endpoint small with large items (I have instances with 300+ subitems, that could impact the performance)

SchrodingersGat:
@matmair I am not suggesting adding the sub-items themselves into the main StockItemSerializer. Instead add a field `installed_item_count` which is an annotation of the total number of sub items. Does not add any significant DB overhead.

To *request* the subitems, you simply query the stockitem endpoint again, with the `?installed_in=<parent_id>`  query

matmair:
@SchrodingersGat so how would the API that returns the items for frontend rendering look like?

SchrodingersGat:
All API calls would be to the same API endpoint:

1. Call `/api/stock/?<original_query>`
2. For any items that have `installed_items > 0` we add a little "load more" button
3. User presses "load more" button for a particular stock item
4. Call `/api/stock/?installed_in=<stock_id>` which returns another set of stock items
5. Items are displayed in a "tree" style just like the current BOM table

[...]

PR #5030, opened 2023-06-13
Stock installed items

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/4833


T0jan:
@SchrodingersGat works great, thank you. how difficult would it be to add this load options also to the installed items view of a specific stock item. I recon this would be the first point were users would actually look which sub-items are part of an assembly

PR #5042, opened 2023-06-14
Improve StockItem API speed

SchrodingersGat:
- Removes child detail fields which cannot be effectively annotated
- Prefetch required fields
- Add unit test method for checking query count

In a test setup, reduces a query of 200 items from ~1s to ~200ms (200 DB hits down to 18)


matmair:
@SchrodingersGat does this need a raised API version?

Issue #4886, opened 2023-05-24
[bug] displayed numbers on front page aren't independent of row settings


T0jan:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

hi again,

there is a small bug on the front page which in my opinion limits the usefulness of the page all together. If you have more then 25 items in a specific category, lets say low stock for example, then instead of the total number only a `25` will be displayed if the rows per page is set to 25:

![image](https://github.com/inventree/InvenTree/assets/22519396/fbf7ad02-d54e-4b65-9acc-63d8e1f55eba)
![image](https://github.com/inventree/InvenTree/assets/22519396/fd6fb0c1-3436-48db-a3c7-03de8ac07e92)

![image](https://github.com/inventree/InvenTree/assets/22519396/b4ea7faf-a203-471d-b1c7-9a0b9109824f)
![image](https://github.com/inventree/InvenTree/assets/22519396/e8f1d9c6-bf94-4cb4-85fe-d09fa1a154bf)

### Steps to Reproduce

1. have more then 25 items with low stock flag
2. go to `front page`
3. got to `low stock` entry
4. set `rows per page` to 25

### Expected behaviour

Independent of the `rows per page` settings the total number of fitting entries should be shown.

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.12.0 dev
Django Version: 3.2.18

Database: postgresql
Debug-Mode: True
Deployed using Docker: False
Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}]

matmair:
Can reproduce, thanks for the report @T0jan

matmair:
I will leave this sitting for a bit for any beginner to pick if they want as it is a localised issue.

Basically, a filter is set by default on API calls by bootstrap-table, our table library. One fix is to set the limit to a higher, fixed value (maybe 101) and add a plus symbol to the badge if the number of records is higher than the set value (so display 100+ if the number of results is 101). You can set the parameter via the options. The specific one is documented [here](https://bootstrap-table.com/docs/api/table-options/#queryparams).

PR #5057, opened 2023-06-16
Homepage Improvements

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/4886

Some basic improvements to the homepage, which has not been touched in a while.

Minimal effort here, as this will all be replaced with react anyway


matmair (approved):
LGTM; Should we mark this breaking as it changes current behaviour?

SchrodingersGat:
I don't think this really qualifies as a breaking change, more of a bug fix

Issue #5059, opened 2023-06-16
Part responsible is limited to a User, groups not allowed


miggland:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

Part responsible can only be set to a User.

In Orders (and others?) groups and users can be chosen.

### Steps to Reproduce

1. Create a group
2. Create a part
3. Try to set responsible of part to group: find nothing

### Expected behaviour

Fields with the same name should behave in a similar manner - allowing groups is the expected behaviour.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.12.0 dev
Django Version: 3.2.19
Commit Hash: 31ff359
Commit Date: 2023-06-16

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-5.15.0-71-generic-x86_64-with-glibc2.36
Installer: DOC

Active plugins: False

### Please verify if you can reproduce this bug on the demo site.

- [X] I can reproduce this bug on the demo site.

matmair:
I would not classify it as a bug, it is intended behaviour. As long as the API stays consistent it should be fine to change.

miggland:
Still relevant.

PR #5060, opened 2023-06-16
Part responsible

miggland:
Fixes https://github.com/inventree/InvenTree/issues/5059


miggland:
Good question, I didn't think about that.

I'll check and get back to you.

SchrodingersGat:
I think this will need to be split into three separate migrations:

1. Add new field
2. Perform data migration
3. Delete old field

Additionally a unit test for the new data migration :)

miggland:
Will be done (probably..)

SchrodingersGat:
@miggland have you had a chance to look at the migration changes and unit tests for this?

miggland:
I started, but got stuck with the migration and then other things came up..

If you know of migrations which do something similar that would be helpful, I'd have somewhere to start..

Issue #4948, opened 2023-06-01
after update to .11.x can no longer set stock status


gablesm:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

selecting Stock, and then any stock item, previous instances allowed us to select the stock item, choose stock options and then set stock status.  

the interesting detail is that the list appears to show but no text is visible.  you can scroll down the list and blindly select a value and it will update the DB.  

on the stock status page it actually shows the proper stock status.

thanks for the help!

### Steps to Reproduce

1 - Go to Stock, stock items, (select any item)
2 - select stock options, then scroll down to 'Change stock status'
3 - new window appears but the only option that is visible is '---------'.  if you scroll up or down, you will see other the cursor highlight change and you can accept it.  it will update the DB and the stock status value but no text available within window so you must guess as to the selection.
![image](https://github.com/inventree/InvenTree/assets/1297480/81d9bd6a-62c0-4dee-b794-4765276a569f)

### Expected behaviour

we would expect to see the standard stock status items like OK or DAMAGED

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.11.2
Django Version: 3.2.18

Database: postgresql
Debug-Mode: False
Deployed using Docker: False
Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}]

matmair:
Have you checked https://docs.inventree.org/en/latest/faq/#feature-x-does-not-work-after-update and restarted the server after the update?

gablesm:
Yes.  We’ve been running inventree for about a year with no issues on
upgrades.

matmair:
@gablesm can you provide screenshots of every step or try to reproduce on demo.inventree.org? I have problems reproducing with the provided instructions (demo is on master, not 0.11.2).

walkermc20:
Just checked the demo server and the "Change Stock Status" is not included on the 0.12.0 dev branch?  Does not appear to be an option.  
![Screenshot 2023-06-16 115344](https://github.com/inventree/InvenTree/assets/34684351/44149e4f-51d5-4ffa-9375-da9a9bfc9dcb)

SchrodingersGat:
[...]

PR #5064, opened 2023-06-17
Stock status change API

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/4948

Reintroduced the ability to change the "status" field for multiple stock items simultaneously.

Previously this was achieved with multiple (sequential) API calls, which is super inefficient and also error prone. 

This PR adds a new API endpoint where this can all be achieved with two database hits.

### DIFF
--- a/InvenTree/stock/serializers.py
+++ b/InvenTree/stock/serializers.py
@@ -722,7 +722,7 @@ class StockChangeStatusSerializer(serializers.Serializer):
             transaction_notes.append(
                 StockItemTracking(
                     item=item,
-                    tracking_type=InvenTree.status_codes.StockHistoryCode.EDITED,
+                    tracking_type=InvenTree.status_codes.StockHistoryCode.EDITED.value,
                     date=now,
                     deltas=deltas,
                     user=user,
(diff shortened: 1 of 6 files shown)


InvenTree/stock/serializers.py
```diff
@@ -641,6 +643,100 @@ def save(self):
...
+
+            item.updated = now
+            item.status = status
+            items_to_update.append(item)
+
+            # Create a new transaction note for each item
+            transaction_notes.append(
+                StockItemTracking(
+                    item=item,
+                    tracking_type=InvenTree.status_codes.StockHistoryCode.EDITED,
```
matmair, line 725: Have you tested this? I think this needs to be

```suggestion
                    tracking_type=InvenTree.status_codes.StockHistoryCode.EDITED.value,
```

matmair (approved):
efficiency is always great!

mabroens:
Yes please reintroduce this feature. i was just looking for it and found it not working indeed.

Br Martijn

mabroens:
Additionally how can i add a status? In Test for example

SchrodingersGat:
> Yes please reintroduce this feature. i was just looking for it and found it not working indeed.

As this PR has been merged - the feature is indeed now available on the development version.

> Additionally how can i add a status? In Test for example

Custom status codes are not supported, if that is what you are looking for

mabroens:
Thank Oliver,

yes custom status was what i was looking for. understood.
perhaps add InTest. this status can be used for RMA's or for production.

PR #5065, opened 2023-06-17
Add Contacts to admin panel

LavissaWoW:
* Contact model added to admin panel


LavissaWoW:
Just a smallie I stumbled across when adding Address to the admin panel :>

SchrodingersGat:
Thanks @LavissaWoW :)

Issue #5070, opened 2023-06-19
[CI] Migration issues with MySQL


SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/discussions/5066

> Another thing I have noticed is a warning when I run 'invoke update':
> Applying part.0109_auto_20230517_1048...System check identified some issues:
> WARNINGS:
> company.Address: (models.W036) MySQL does not support unique constraints with conditions.
> HINT: A constraint won't be created. Silence this warning if you don't care about it.

And as also seen in [this failed run](https://github.com/inventree/InvenTree/actions/runs/5308767825/jobs/9608654049): 

![image](https://github.com/inventree/InvenTree/assets/10080325/154d1b32-14a7-4c84-abc8-d4e1dd9966f0)

This was introduced in PR [https://github.com/inventree/InvenTree/pull/4732](#4372) by @LavissaWoW .

We need to address this before next stable release.

- https://docs.djangoproject.com/en/4.2/ref/models/constraints/
- https://stackoverflow.com/questions/57631395/why-django-checkconstraint-doesnt-works-on-mysql

SchrodingersGat:
Best approach might be to add a custom "validate_unique" method, and handle it there, rather than at the database level

PR #5076, opened 2023-06-19
Fix constraint for address model

SchrodingersGat:
- Do not handle at database level
- Add a "validate_unique" method to the address model
- Fixes https://github.com/inventree/InvenTree/issues/5070


matmair (approved):
LGTM; At some point, I would like to discuss depreciating MySQL/MariaDB support. Going all in on Postgres for production would allow usage of a bunch of cool vendor-specific features.

SchrodingersGat:
> LGTM; At some point, I would like to discuss depreciating MySQL/MariaDB support. Going all in on Postgres for production would allow usage of a bunch of cool vendor-specific features.

Makes sense to consolidate in terms of reducing support effort, too.

PR #5100, opened 2023-06-23
Synergise label filter help texts

LavissaWoW:
* Help text for label models euqalised between models
* This also fixes a couple of typos in these strings


SchrodingersGat:
Thanks for the fixes ☺️

PR #5106, opened 2023-06-26
Adapt discovery methods

### LINKED ISSUE #5081: Issue loading plugins from external directory
When loading a plugin from the "plugins" directory (either "internal" or "external"), it can take upwards of 60 seconds to actually load the plugin. This can cause the worker to timeout (and thus not actually start the server at all).

Ref: https://github.com/inventree/InvenTree/discussions/5066

I have narrowed down the problem to the `get_git_log` function:

https://github.com/inventree/InvenTree/blob/66ec82d4d123fc75cb55b62a89a5829b0d637604/InvenTree/plugin/helpers.py#LL112C1-L112C1

When loading from an external directory, this code spends an *eternity* walking directories:

image

Either this needs to be optimized, or removed.

matmair:
Closes #5081


matmair:
@gunstr can you try out this branch an see if it solves the issues?

SchrodingersGat:
@matmair have you profiled this for external plugins, and found any difference?

gunstr:
@matmair Sorry but I see no difference, the system is still not responding.

When I run `invoke update` I still see in the terminal output that the process stops for ~20 seconds each time it reaches `Loading plugin ValidateUniqueIPN`

In the server.err.log you can also see the same timeout and a restart every 30 seconds. I have the backround worker timeout set to 90 seconds so why it happens after 30seconds I do not know (if at all related?)- but I guess it's anyway too long...

image


[server.err.log](https://github.com/inventree/InvenTree/files/11879248/server.err.log)

SchrodingersGat:
background worker timeout is separate to the foreground worker timeout. The foreground worker is the process which serves web data

PR #5113, opened 2023-06-28
Get git log fix

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/5081
- Replaces https://github.com/inventree/InvenTree/pull/5106

This PR simplfies the process for extracting git commit information for externally loaded plugins.

In a simple test environment, reduces plugin load times from 90 seconds to 0.2 seconds

@matmair would you be able to test on your end? I have verified that it provides correct data for plugins dropped into the "plugin" directory, if they have commit information supplied. However this approach does not have "depth of discovery" - can you expand on why this is required?

### DIFF
--- a/InvenTree/plugin/helpers.py
+++ b/InvenTree/plugin/helpers.py
@@ -2,6 +2,7 @@
 
 import inspect
 import logging
+import os
 import pathlib
 import pkgutil
 import sysconfig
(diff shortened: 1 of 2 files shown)


InvenTree/plugin/helpers.py
```diff
@@ -116,25 +117,26 @@ def get_git_log(path):
...
+    if os.path.exists(path) and os.path.isfile(path):
+        path = os.path.dirname(path)
 
     # only do this if we are not in test mode
     if not isInTestMode():  # pragma: no cover
 
         try:
-            walker = Repo.discover(path).get_walker(paths=[path.encode()], max_entries=1)
-            try:
-                commit = next(iter(walker)).commit
```
matmair, line 127: I think this is a regression.
The current implementation highlights the last change to the particular plugin, not the repo as a whole. This is why we use a `walker` that seeks a change in the particular file. Your implementation just shows the latest commit to the repo. You can see this with the build modules - they now all show the latest change to the whole repo, not the particular plugin.

SchrodingersGat: Surely we want to see the commit hash for the `HEAD` of the git repo, right? The exact state of the particular file can be directly inferred from that.

Additionally each individual plugin reports its own version.

To be clear, what you are saying is that given the "current implementation", a git repo containing *multiple* plugins would potentially report a *different* git hash for each plugin, even though they are all running from the same `HEAD` commit?

matmair: The current implementation is to look at each plugin. You can nest plugins (like we do in the samples folder) and you would want the individual change dates and authors to show up, not the head. At that point, the whole git integration becomes a lot less useful.

SchrodingersGat: So what you would want to see is the latest commit which touched any file associated with a particular plugin? I can see how this could introduce difficulties.

SchrodingersGat:
@matmair this is the last blocker for 0.12.0 and provides a significant performance improvement. Would you be happy for me to merge this in, and work on a version which provides the functionality you want, but without the performance hit, for the next release?

PR #5117, opened 2023-06-28
Transfer stock location

SchrodingersGat:
Pre-fill location field when transferring stock items

### DIFF
--- a/InvenTree/templates/js/translated/stock.js
+++ b/InvenTree/templates/js/translated/stock.js
@@ -1152,8 +1152,8 @@ function adjustStock(action, items, options={}) {
         // If a common location is specified, use that as the default
         let commonLocation = null;
 
-        for (var idx = 0; idx < items.length; idx++) {
-            let item = items[idx];
+        for (var idx = 0; ii < items.length; ii++) {
+            let item = items[ii];
 
             if (item.location == commonLocation) {
                 continue;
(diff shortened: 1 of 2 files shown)


InvenTree/templates/js/translated/stock.js
```diff
@@ -1148,7 +1148,25 @@ function adjustStock(action, items, options={}) {
...
+
+            if (item.location == commonLocation) {
+                continue;
+            } else if (commonLocation == null) {
+                commonLocation = item.location;
+            } else {
+                commonLocation = null;
+                break;
+            }
+        }
```
wolflu05, line 1166: - Why do you declare a variable idx and increment ii?
- I strongly suggest also using let here
- but a for(const item of items) look should do the trick also.
- Maybe you can also remove the Else if because you have continue already above, this will make it a lot more readable.

SchrodingersGat: Thanks for the feedback @wolflu05

wolflu05: No problem, but I personally ask myself many times if I should comment some more modern js improvements or not as we anyway switch to react which should enforce this modern style anyway. Do you want me to comment such things or should I stop? Mostly such var usages trigger me to think if I should comment.

SchrodingersGat: It's cool, keep us all to a higher standard :)

matmair: Keep the tips coming @wolflu05, the language for sure can be used better than I do. Regarding style I would love some good linting for Platform UI, if you have suggestions feel free to voice them.

PR #5136, opened 2023-07-03
change invoke setup-dev to invoke setup-test --dev

xeno27:
Chnage postCreateCommand in .devcontainer to invoke setup-test --dev.

This will setup this dev container with all demo data (which is from my perspective the right way).

Setup-dev will "only" insall all required things for development, setup-test --dev will first setup the demo data and then run setup-dev. 

We can also do setup-dev --tests, but they way i choos is documented in the docs.

### DIFF
--- a/.devcontainer/postCreateCommand.sh
+++ b/.devcontainer/postCreateCommand.sh
@@ -15,7 +15,7 @@ python3 -m venv dev/venv
 # setup InvenTree server
 pip install invoke
 invoke update
-invoke setup-dev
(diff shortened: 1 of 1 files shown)


wolflu05:
This is not a good idea, because this script runs on every container recreate. So it possibly Setup the env again and again with the demo dataset which is actually already done on the first creation. That's why it's recommended to use vscode tasks to import the demo data.

xeno27:
OK from this perspective it make sense.

PR #5144, opened 2023-07-04
Database config spelling error

RitchieP:
Possible spelling error for `INVENTREE_TIMEZONE` env variable.

If this is intended, maybe we should add a clarification.


SchrodingersGat:
@RitchieP thanks for the fix :)

PR #5168, opened 2023-07-04
Bump django - fix CVE-2023-36053

matmair:
This PR fixes a CVSS 7.5 vul in django that affects us.


matmair:
@SchrodingersGat we need to release after this is merged ASAP - we are affected

PR #5192, opened 2023-07-06
fix: generic status import

wolflu05:
This fixes an issue where you cannot import `StatusCode` at some point where the App are not initialized yet so you get an `AppRegistryNotReady` error raised. (needed for #4824)

### DIFF
--- a/InvenTree/generic/states/__init__.py
+++ b/InvenTree/generic/states/__init__.py
@@ -6,10 +6,8 @@ There is a rendered state for each state value. The rendered state is used for d
 States can be extended with custom options for each InvenTree instance - those options are stored in the database and need to link back to state values.
 """
 
-from .api import StatusView
 from .states import StatusCode
 
 __all__ = [
-    StatusView,
     StatusCode,
 ]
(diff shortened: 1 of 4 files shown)


InvenTree/generic/states/__init__.py
matmair, line 1: @wolflu05 I do not see how this is necesarry. Can you be a bit more clear why you are removing the simpler import route?

matmair: At what point do you define an API view before apps are ready?

wolflu05: You can find a more detailed description here: https://github.com/inventree/InvenTree/pull/4824#issuecomment-1622496888

wolflu05: > At what point do you define an API view before apps are ready?

I don't define an api, the thing is that I need to import statuscode somewhere where the apps are not ready.

wolflu05: @matmair Do you have another idea?

matmair: I do not have time to work on this rn; let's just merge this, it is good enough

PR #5193, opened 2023-07-06
Fix 404 link in docker_prod.md

wolflu05:
Redis broke their links so the old link now returns 404. Comparing the new page to the one I can serve via the web archive it seems like the new link is the same page.


matmair (approved):
LGTM

PR #5204, opened 2023-07-08
Fix cli on 22.04

matmair:
This PR fixes cli runs on ubuntu 22.04 and higher with packed installs.
Needs a release to be pushed to install base.


SchrodingersGat:
@matmair I believe that this change may be why the `inventree run cli` command is not working - as per [this discussion](https://github.com/inventree/InvenTree/discussions/5791)

If the installer is running on Ubuntu 20.04 (like my WSL setup) then the extra `echo "" &&` at the start is seeming to cause the rest of the line not to be run. This is why @imager2019 is not getting any output.

PR #5206, opened 2023-07-08
Packages - force python 3.9

matmair:
On stock 20.04 python3 refers to python 3.8 - we need 3.9.
This PR explicitly lists 3.9 as a reference.


matmair:

PR #5213, opened 2023-07-10
Docker: add missing reporting libs

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/5007

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,7 +58,7 @@ RUN apk add --no-cache \
     # Image format support
     libjpeg libwebp zlib \
     # Weasyprint requirements : https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#alpine-3-12
-    py3-pip py3-pillow py3-cffi py3-brotli pango poppler-utils pango-dev cairo-dev \
+    py3-pip py3-pillow py3-cffi py3-brotli pango poppler-utils \
     # SQLite support
     sqlite \
     # PostgreSQL support
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,7 +66,7 @@ RUN apk add --no-cache \
     # MySQL / MariaDB support
     mariadb-connector-c-dev mariadb-client && \
     # fonts
-    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-noto terminus-font && fc-cache -f \
+    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-noto terminus-font && fc-cache -f
 
 EXPOSE 8000
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,7 +58,7 @@ RUN apk add --no-cache \
     # Image format support
     libjpeg libwebp zlib \
     # Weasyprint requirements : https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#alpine-3-12
-    py3-pip py3-pillow py3-cffi py3-brotli pango \
+    py3-pip py3-pillow py3-cffi py3-brotli pango poppler-utils pango-dev cairo-dev \
     # SQLite support
     sqlite \
     # PostgreSQL support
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -66,7 +66,9 @@ RUN apk add --no-cache \
     # MySQL / MariaDB support
     mariadb-connector-c-dev mariadb-client && \
     # fonts
-    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-noto terminus-font && fc-cache -f
+    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-noto terminus-font && fc-cache -f \
+    # report generation
+    poppler-units libpango-1.0-0 libpangoft2-1.0-0
```
wolflu05, line 71:
```suggestion
    poppler-units
```

The later two packages do not exist on alpine, [see](https://pkgs.alpinelinux.org/packages?name=libpangoft*&branch=edge&repo=&arch=&maintainer=), hopefully it works even without them.

wolflu05:
> Was this intentional, or an oversight? We'll need to add these back in.

Missed the poppler-untis one, but the rest was intentional as they do not exist on alpine, [see](https://pkgs.alpinelinux.org/packages?name=libpangoft*&branch=edge&repo=&arch=&maintainer=), hopefully it works even without them. Unfortunately, I already testet label generation and it worked.

But your changes do not work, I suggest adding only `poppler-units` to this line https://github.com/inventree/InvenTree/blob/61840bc2f815e35d533f77a024e7758edd8d4964/Dockerfile#L61

SchrodingersGat:
> But your changes do not work

Do not work how? The docker image builds OK for me?

wolflu05:
> Do not work how? The docker image builds OK for me?

Yes, it's basically the same as without your changes. Look closely at the last line where you added the `\` the `fc-cache` command just ignores the extra args you provided. You need to add them to a `apk add` command. I suggest adding it to L60, because there is the other report stuff.

SchrodingersGat:
> Yes, it's basically the same as without your changes. Look closely at the last line where you added the `\` the `fc-cache` command just ignores the extra args you provided. You need to add them to a `apk add` command. I suggest adding it to L60, because there is the other report stuff.

Ah, yes ok that is obvious now :)

Dockerfile
```diff
@@ -58,15 +58,15 @@ RUN apk add --no-cache \
...
+    py3-pip py3-pillow py3-cffi py3-brotli pango poppler-utils pango-dev cairo-dev \
     # SQLite support
     sqlite \
     # PostgreSQL support
     postgresql-libs postgresql-client \
     # MySQL / MariaDB support
     mariadb-connector-c-dev mariadb-client && \
     # fonts
-    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-noto terminus-font && fc-cache -f
+    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-noto terminus-font && fc-cache -f \
```
wolflu05:
```suggestion
    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-noto terminus-font && fc-cache -f
```

wolflu05:
Keep an eye on the size, not sure if we really need the added packages, I looked at the [official docs](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#alpine-3-12) and that's why I didn't add them.

SchrodingersGat:
> Keep an eye on the size, not sure if we really need the added packages, I looked at the [official docs](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#alpine-3-12) and that's why I didn't add them.

Yep, looks like unit tests pass correctly without them, so have removed again :)

SchrodingersGat:
@wolflu05 very nice work :) The size difference is impressive

Screenshot 2023-07-10 at 10 44 03 pm

PR #5218, opened 2023-07-10
[WIP] Mantine datatables

SchrodingersGat:
This PR is a major refactor of the existing table infrastructure, moving from [bootstrap-table](https://bootstrap-table.com/) to [mantine datatable](https://icflorescu.github.io/mantine-datatable/)

Example:

image

```[tasklist]
### Functionality
- [x] API integration
- [x] Search Bar
- [x] Column Sorting
- [x] Server-side pagination
- [x] Toggle column visibility
- [x] Custom filters
- [x] Refresh button
- [x] Custom actions
- [ ] https://github.com/inventree/InvenTree/issues/5320
- [x] Download data action
- [x] Adhere to global theme properties
```

### Notes

- Requires https://github.com/inventree/InvenTree/issues/2789


SchrodingersGat:
Screenshot 2023-07-13 at 10 02 25 pm

**New Features**

- Search box
- Column visibility toggling
- Refresh buttons
- Custom field rendering

src/frontend/src/components/tables/InvenTreeTable.tsx
```diff
@@ -0,0 +1,347 @@
...
+    enableRefresh?: boolean;
+    pageSize?: number;
+    printingActions?: any[];
+    barcodeActions?: any[];
+    customActionGroups?: any[];
+    customFilters?: TableFilter[];
+}) {
+
+    // Check if any columns are switchable (can be hidden)
+    const hasSwitchableColumns = columns.some((col: any) => col.switchable);
```
wolflu05: For performance reasons you could wrap this in a useMemo here and all the functions in a useCallback react hooks, arrow functions also look much clear and you can save the {} here.

src/frontend/src/components/tables/InvenTreeTable.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+        // Find matching column:
+        // If column provides custom ordering term, use that
+        let column = dataColumns.find((col) => col.accessor == key);
+
+        if (column && column.ordering) {
+            return column.ordering;
+        } else {
+            return key;
+        }
```
wolflu05: Just a tip, to avoid these long && chains you could use optional chaining (`column?.ordering`) and you could also combine that with the  or operator or even the nullish coalescing operator like this:
```suggestion
        return column?.ordering ||key;
```

src/frontend/src/components/tables/InvenTreeTable.tsx
```diff
@@ -0,0 +1,347 @@
...
+        for (let idx = 0; idx < newColumns.length; idx++) {
+
+            if (newColumns[idx].accessor == columnName) {
+                newColumns[idx].hidden = !newColumns[idx].hidden;
+            }
+
+            if (newColumns[idx].hidden) {
+                hiddenColumns.push(newColumns[idx].accessor);
+            }
+        }
```
wolflu05: You could utilize the es6 array map, filter, reduce functions here. They become pretty handy and look much more modern than such cpp style for loops.

src/frontend/src/components/tables/InvenTreeTable.tsx
```diff
@@ -0,0 +1,347 @@
...
+        // Handle sorting
+        let ordering = getOrderingTerm();
+        if (sortStatus.direction == 'asc') {
+            queryParams.ordering = ordering;
+        } else {
+            queryParams.ordering = `-${ordering}`;
+        }
+            
+        return api
+            .get(`http://localhost:8000/api/${url}`, {  // TODO: Don't hardcode the base URL here!
```
wolflu05: Url should not be hardcoded

src/frontend/src/components/tables/InvenTreeTable.tsx
```diff
@@ -0,0 +1,347 @@
...
+                return [];
+            }).catch(function(error) {
+                setMissingRecordsText(t`Error` + ": " + error.message);
+                return [];
+            });
+    }
+
+    const { data, isError, isFetching, isLoading, refetch } = useQuery(
+        [`table-${tableKey}`, sortStatus.columnAccessor, sortStatus.direction, page],
+        async() => fetchTableData(),
```
wolflu05: You don't need this arrow function wrapper, just pass the function in.

src/frontend/src/components/tables/build/BuildOrderTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+export function BuildOrderTable({
+    params={}
+} : {
+    params?: any;
+}) {
+
+    // Add required query parameters
+    let tableParams = Object.assign({}, params);
+
+    tableParams.part_detail = true;
```
wolflu05, line 107: I would wrap this in a useEffect/useState or useMemo and instead of Object.assign I would use the modern spread operator ...

wolflu05 (commented):
Just scrolled through the code and added a few comments.

SchrodingersGat:
@matmair  having trouble working out how to use your theme code to influence the styling of the tables here - any pointers?

matmair:
@SchrodingersGat what are you struggling with? The main theme is automatically used if you do not set colours/styles.

SchrodingersGat:
> @SchrodingersGat what are you struggling with? The main theme is automatically used if you do not set colours/styles.

After pulling down your latest changes, it is working with dark theme now. Not sure what was blocking it? But, it is working well now:

image

matmair:
Coming together nicely, I see. Regarding custom filters: You could but the whole section for that on a HoverCard, that way you do not consume a lot of space when it is not in use but lots of for the inputs etc.

yarn.lock
```diff
@@ -960,10 +960,10 @@ loose-envify@^1.0.0, loose-envify@^1.4.0:
   dependencies:
     js-tokens "^3.0.0 || ^4.0.0"
 
-mantine-datatable@^2.8.5:
```
wolflu05, line 963: Actually, why do we need this second lock file?

src/frontend/src/pages/Index/Stock.tsx
```diff
@@ -0,0 +1,20 @@
+import { Trans } from '@lingui/macro';
+import { Group } from '@mantine/core';
+
+import { PlaceholderPill } from '../../components/items/Placeholder';
+import { StylishText } from '../../components/items/StylishText';
+import { StockItemTable } from '../../components/tables/stock/StockItemTable';
+
+export default function Part() {
```
wolflu05, line 8: I think you overlooked this after pasting, same for the build.tsx

src/frontend/src/components/tables/Search.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      if (value != searchValue) {
+        setSearchValue(value);
+        searchCallback(value);
+      }
+    }, 500);
+    return () => clearTimeout(timer);
+  });
```
wolflu05, line 21: This denounce logic could be replaced by a useDebounce hook from [mantine](https://mantine.dev/hooks/use-debounced-value/).

package.json
```diff
@@ -7,6 +7,6 @@
     "@mantine/notifications": "^6.0.17",
     "eslint": "^8.41.0",
     "eslint-config-google": "^0.14.0",
-    "mantine-datatable": "^2.8.5"
+    "mantine-datatable": "^2.9.0"
```
wolflu05, line 10: Same Wort the package.json as with the lockfile, why do we need this in the root folder?

SchrodingersGat:
> Coming together nicely, I see. Regarding custom filters: You could but the whole section for that on a HoverCard, that way you do not consume a lot of space when it is not in use but lots of for the inputs etc.

Current draft is to have a button which shows number of active filters, click to expand:

image

image

Still pretty rough, but the mechanics are getting there

src/frontend/src/components/tables/InvenTreeTable.tsx
```diff
@@ -68,7 +68,7 @@ function loadActiveFilters(tableKey: string, filterList: TableFilter[]) {
  * @param filters : any - map of active filters, {name: value}
  */
 function saveActiveFilters(tableKey: string, filters: TableFilter[]) {
-  let active = {};
+  let active: any = {};
 
   filters.forEach((f) => (active[f.name] = f.value));
```
wolflu05: You could simplify this to the following. Also, `any` types are not that great, in my project I decided to also disallow them strictly.
```suggestion
  const active = Object.fromEntries(filters.map((f) => [f.name, f.value]));
```

SchrodingersGat:
Custom filters are now working quite well. The UX might need a bit of polish, but the functionality is all there:

image
image
image

SchrodingersGat:
@wolflu05 @matmair given that the image auth question might be somewhat involved, I'm happy to proceed with this, and handle images in a separate PR.

LMK if there are any outstanding issues with this before merge

wolflu05:
There are some unresolved conversations. I do not have permissions to resolve them so maybe you should take a look at that and see what has already been addressed.

SchrodingersGat:
Ok, merging this one in - any issues pop up we can address them. But it will be good to have this framework in place to work frmo

src/frontend/vite.config.ts
```diff
@@ -14,5 +15,10 @@ export default defineConfig({
   build: {
     manifest: true,
     outDir: '../../InvenTree/web/static/web'
+  },
+  server: {
+    watch: {
+      usePolling: true
+    }
```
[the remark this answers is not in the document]
wolflu05: @SchrodingersGat why do you added this here? I also have a few other comments here.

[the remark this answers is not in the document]
SchrodingersGat: Did you see my previous comment? On my setup (windows / devcontainer) it won't auto-reload unless I have this configured.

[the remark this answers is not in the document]
SchrodingersGat: Guilty :)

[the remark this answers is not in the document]
wolflu05: Maybe we can apply this setting conditionally like checking if we run on windows.

src/frontend/src/components/items/Thumbnail.tsx
```diff
@@ -0,0 +1,58 @@
...
+    alt = t`Thumbnail`,
+    size=20,
+} : {
+    src: string;
+    alt?: string;
+    size?: number;
+}) {
+
+    // TODO: Use api to determine the correct URL
+    let url = "http://localhost:8000" + src;
```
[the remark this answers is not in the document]
wolflu05: You would need to do something like this: `const [host] = useLocalState((state) => [state.host])`, but I think this gets replaced if we want to get this to work with auth.

src/frontend/vite.config.ts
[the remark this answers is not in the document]
wolflu05: Something like:
```suggestion
  },
  server: {
    watch: {
      usePolling: process.platform === "win32"
    }
```

src/frontend/src/components/tables/InvenTreeTable.tsx
wolflu05:
```ts
import {useMemo} from "react";
const hasSwitchableColumns = useMemo(() => columns.some((col: any) => col.switchable), [columns]);
```

And for an useCallback example:
```ts
const updateSearchTerm = useCallback((term: string) => {
        term = term.trim();
        // Ignore identical search terms
        if (term == searchTerm) return;

        setSearchTerm(term);
        latestSearchTerm = term;
        refetch();
}, [searchTerm, refresh]
```
You can read more on the [react docs](https://react.dev/reference/react/useCallback) page.

src/frontend/src/components/items/Thumbnail.tsx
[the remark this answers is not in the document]
SchrodingersGat: This bit of code will be refactored entirely when we implement a function for retrieving and caching images, so I'll leave it "as is" until then.

package.json
matmair: All code and support files for the frontend are in src/frontend, no need to modify this one

yarn.lock
matmair: just `./src/frontend/yarn.lock`, not sure where you are getting package-lock.json files with yarn

PR #5226, opened 2023-07-11
Native LDAP support

### LINKED ISSUE #3746: [FR] Add native LDAP

### Problem statement

We currently do not provide a native LDAP integration - which is often needed together with SAML for enterprise users.

### Suggested solution

Native inclusion of ldap - maybe via django-auth-ldap. A problem I see is the native dependency tree for python-ldap; Maybe it could be modified to use python-ldap3 to make the footprint slimmer.

Current way with cusotm middleware and auth backends works too - but only for advanced users.

### Examples of other systems

netbox

- [ ] I want to develop this.

wolflu05:
This PR adds native ldap support. I already played a bit while my refactor (#5007) and found out that this doesn't add that much megabytes more to the image. The docker image only grows about `6MB` (from `528MB` to `534MB`) which is not that much.

```[tasklist]
- [x] don't require ldap system packages on bare metal install/development/gh pipeline
```

fixes #3746 
fixes #5098 
fixes #3745
fixes #4464


wolflu05:
1. Do you have any suggestions how to handle installs where no native ldap dependencies are available, e.g. some bare metal, development (especially on windows), gh actions, ...
2. What is wrong with the dependency files? Somehow it removed a few things and ci now fails.

matmair:
1. Disable the function ;.)
2. You are on 3.10, the target of the CI is 3.9 so additional dependencies are needed
3. Not to my knowledge, we will have to polish the experience around this a lot as the defaults are lacking compared to what SSO/built-in support rn (for example group assignments, restrictions, onboarding flow)

wolflu05:
> 1. Disable the function ;.)

That's already done, but the problem  exists on the pip install process. If the packages are added like this, the ldap system headers are needed already for the pip install step. How can I get around this and only install these ldap packages if needed, e.g. docker file, ... 

How should I handle bare metal installs and the docker pipeline?

matmair:
that is difficult - maybe split out this requirement in a separate file?

wolflu05:
And how would you trigger the install e.g. on a bare metal install?

matmair:
That is a bit more difficult. We currently only provide packages for ubuntu/debian so we could add a check for those operating systems and then install the required packages. Maybe this should be tackled in the packaging stage / with an additional package that specifies the required deps and ensures lockstep-updating. I have to think about this a bit - maybe try getting it running in docker and I while report back once I have a conclusion.

wolflu05:
Ok. It already works in docker, the only problem now is where we install the python-ldap and Django-auth-ldap package so that you could decide if you want to have those packages.

matmair:
If they are this small I would just ship them with docker. On a CIS level 1 hardened ubuntu 22.04 the required apt packages add 255mb of packages so bare metal might need another way

wolflu05:
One option would be to install them via `docker/requirements.txt`. That way normal installs still work and also all ci stuff. For all bare metal stuff we could add a few commands to install the required packages manually to the docs like this: 
```bash
apt install build-essential python3-dev libldap2-dev libsasl2-dev
# activate venv
pip install python-ldap django-auth-ldap
``` 

1. Is there a simple command that can be used to activate the venv on all bare metal installs, (never used the bare metal install)?
2. Should I document this feature somewhere in the docs, and if so at what place?

I think we shouldn't care much about dozens of different install methods, we should focus on functionality. So I think docker and these 3 lines setup instructions should be enough. What do you mean?

wolflu05:
This would be ready for review from my POV. Test failure is unrelated (already addressed by #5244).

InvenTree/InvenTree/settings.py
matmair, line 1: The TLS/STARTLS/CA options are missing.
This setup also always binds on the provided user which will raise issues with audits. There should be an option to bind with the authenticating user.

InvenTree/config_template.yaml
```diff
@@ -223,6 +223,29 @@ remote_login_header: HTTP_REMOTE_USER
 #     KEYCLOAK_URL: 'https://keycloak.custom/auth'
 #     KEYCLOAK_REALM: 'master'
 
+# Add LDAP support
+# ldap:
+#   enabled: false
+#   server_uri: ldap://example.org
```
matmair, line 229: ldap is unencrypted ie unsecure and should not be suggested.
```suggestion
#   server_uri: ldaps://example.org
```

InvenTree/config_template.yaml
```diff
@@ -223,6 +223,29 @@ remote_login_header: HTTP_REMOTE_USER
...
+#   bind_dn: cn=admin,dc=example,dc=org
+#   bind_password: admin_password
+#   search_base_dn: cn=Users,dc=example,dc=org
+
+#   # uncomment for advanced filter search, default: uid=%(user)s
+#   # search_filter_str:
+
+#   # uncomment for advanced user attribute mapping
```
matmair, line 237:
```suggestion
#   # uncomment for advanced user attribute mapping (in the format <InvenTree attribute>: <LDAP attribute>)
```

matmair:
> I think we shouldn't care much about dozens of different install methods, we should focus on functionality. So I think docker and these 3 lines setup instructions should be enough. What do you mean?

I do not support this attitude. We support 3 different install methods in 2 variants. If we do not want to support install methods they should be removed and marked as obsolete with enough headroom for users to switch.
Docker is not a serious release method with the current security posture. Nearly none of our images pass an automated security checker. So they can not be used by people with security requirements or needs.
Moving fast and breaking things works for a social media site but not for the auth layer of a system that holds business-critical information like inventory or orders.

I will look into support for this with the installer as soon as I am done with the currently pending major refactor. Support will either require changes to the installer script or the package mechanism, should be minor though.

wolflu05:
@matmair ok, good attitude actually, you're right, this is not just any feature, its security relevant. Thank you for taking a look sometime on the bare metal implementation. I added the requested options now.

wolflu05:
This is ready from my POV in docker, but waiting for the package installer integration.

matmair:
I do not know when I will have time to figure this out for packager install. I think adding a note about that and emerging is the prudent option.

matmair (approved):

SchrodingersGat:
@wolflu05 can you please add a note to the docs specifying that this is not currently available in the 'installer'

wolflu05:
@SchrodingersGat I highlighted that this is not available by default for non docker installs now. Please let me know if that is ok, or if I should remove the docs about how to configure this with a bare metal install. I didn't checked this myself with the inventree apt package installer, but I provided the commands that were necessary to set this up in the old Debian based docker image.

matmair:
This will stop working with every point release of inventree as envs are reset at that point. I would just add a line that itdoes not work with the package installer.

wolflu05:
Ok, didn't knew that, that's why I was asking. Will do later.

wolflu05:
Is that ok so?

matmair:
LGTM

SchrodingersGat:
Thanks @wolflu05

PR #5228, opened 2023-07-11
add sonar active check

matmair:


matmair:
wrong repo 😱

matmair:
I should really stop multitasking and coding at the same time; if this works I will propose it - but with cleaner commits and messages 😆

Issue #5227, opened 2023-07-11
Protected settings not working


wolflu05:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

While working on #5224 I discovered, that the initial implementation of protected settings (#2339) doesn't work correctly (anymore). You can still see the protected settings values in the plugin settings page and they aren't redacted. 

### Steps to Reproduce

1. create a plugin with protected settings
2. goto the plugins settings page and set a value
3. reload settings page and see the NOT redacted values

### Expected behaviour

protected settings should be redacted with `***`.

### Deployment Method

- [X] Docker
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.13.0 dev
Django Version: 3.2.20
Commit Hash: c30190e
Commit Date: 2023-07-11
Commit Branch: feature/required-settings
Database: sqlite3
Debug-Mode: True
Deployed using Docker: False
Platform: Linux-5.15.49-linuxkit-pr-x86_64-with-glibc2.31
Installer: GIT

Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}]

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

SchrodingersGat:
Thanks @wolflu05 - fix incoming in https://github.com/inventree/InvenTree/pull/5229

PR #5229, opened 2023-07-11
Protected settings fix

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/5227

### DIFF
--- a/InvenTree/common/serializers.py
+++ b/InvenTree/common/serializers.py
@@ -27,6 +27,10 @@ class SettingsValueField(serializers.Field):
         """
         return '***' if instance.protected else instance.value
 
+    def to_internal_value(self, data):
+        """Return the internal value of the setting"""
+        return data
+
 
 class SettingsSerializer(InvenTreeModelSerializer):
     """Base serializer for a settings object."""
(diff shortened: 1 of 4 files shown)


wolflu05:
Are you sure this covers every place where you can access settings? What is the expected behavior of get_setting?

SchrodingersGat:
AFAIK the behaviour is to protect the setting from display in the interface / API, not prevent it from being accessed at all, right? Otherwise, how would it actually be used?

SchrodingersGat:
@matmair any thoughts here? I have not ever used any protected settings in practice

matmair:
One of the refactors in settings seems to have broken this. Maybe there should be a test for this.
The idea is to not let the values go out of the system - python code still needs to be able to access it though so this should work as intended.

wolflu05:
> AFAIK the behaviour is to protect the setting from display in the interface / API, not prevent it from being accessed at all, right? Otherwise, how would it actually be used?

Oh yeah 🤦‍♂️ didn't thought about that. Makes totally sense, otherwise this feature isn't useful anyhow.

SchrodingersGat:
> One of the refactors in settings seems to have broken this. Maybe there should be a test for this.

I'll add in a unit test

PR #5245, opened 2023-07-13
fix: make protected setting again changeable through admin ui

wolflu05:
#5229 hides the settings completely from admin ui and also disables the button to change it. This PR adds the change button back, but still hides the settings value.


matmair (approved):
LGTM

SchrodingersGat:
Nice work thanks

PR #5248, opened 2023-07-13
Remove unneeded gunicorn

matmair:
Not sure we need this file anywhere anymore - seems like there is no obvious referecne in the docs.


SchrodingersGat:
Seems reasonable

SchrodingersGat:
Thanks @matmair

PR #5259, opened 2023-07-16
Fix settings function callback

matmair:
Fixes the currently broken settings function lookup - seems to be broken at least since 0.12.0


SchrodingersGat:
How is is currently broken?

matmair:
It does not work for settings in plugins as the filter needed for filtering them out (that is supplied in get_filters_for_instance) is never called.

SchrodingersGat:
Interesting, it had been working for me in master... I have tested your patch locally and looks all good, but let's keep an eye on it.

PR #5260, opened 2023-07-17
Adds trigger when plugins are loaded

matmair:
This PR adds an even trigger when all plugins were activated. This can be listened for in plugins.


SchrodingersGat:
Looks like this might cause some issues during database loading, maybe a race condition? CI is failing, in any case

matmair:
Yeah there seems to be a fundamental problem - will look into it

matmair:
Works now @SchrodingersGat !

PR #5275, opened 2023-07-18
P UI: Added frontend publishing for bare-metal (git)

matmair:
This PR adds the required tasks and CI for deploying the frontend to bare-metal git installs. Code by @wolflu05

PR #5277, opened 2023-07-18
P UI: Added frontend publishing for bare-metal (git) 

matmair:
This PR adds the required tasks and CI for deploying the front end to bare-metal git installs. Code by @wolflu05

Relaces #5275


wolflu05:
Yeah, these few changes you reverted where just in for testing, I don't wanted to always cancel the other Workflows.

matmair:
@wolflu05 is there anything else we should change? I am testing the integration into the 'update' task right now - seems to work

wolflu05:
I'll try to review all prs tomorrow.

tasks.py
```diff
@@ -761,3 +798,132 @@ def frontend_build(c):
...
+        print("[ERROR] Either ref or tag needs to be set.")
+
+    if tag:
+        tag = tag.lstrip("v")
+        try:
+            handle_download(f"https://github.com/{repo}/releases/download/{tag}/frontend-build.zip")
+        except Exception as e:
+            if not isinstance(e, requests.HTTPError):
+                raise e
+            print(f"""[ERROR] An Error occurred. Unable to download frontend build, release or build does not exist,
```
wolflu05, line 902: just FYI, I used three concatenated strings as I wanted the indentation to be aligned, but its also fine like that if you prefer that way, Ill use it the next time.

tasks.py
```diff
@@ -102,6 +103,26 @@ def yarn(c, cmd, pty: bool = False):
...
+        try:
+            return str(subprocess.check_output([cmd], stderr=subprocess.STDOUT, shell=True), encoding='utf-8').strip()
+        except subprocess.CalledProcessError:
+            return None
+        except FileNotFoundError:
+            return None
+
+    yarn_version = check('yarn --version')
+    node_version = check('node --version')
+    return ret((not yarn_version or not node_version), node_version, yarn_version)
```
wolflu05, line 123: I would not enforce yarn, I would rather show a message that you should install it via `npm i -g yarn`, that's not a big deal from my side if you already have node installed, and even if you don't want yarn, you can also use npm. You can also install it only project wide and run via npx. Maybe add a flag to download no matter if you have node.

tasks.py
```diff
@@ -761,3 +798,132 @@ def frontend_build(c):
...
+            print("[ERROR] Cannot find frontend-build.zip attachment for current sha")
+            return
+        print(f"Found artifact {frontend_artifact['name']} with id {frontend_artifact['id']} ({frontend_artifact['size_in_bytes']/1e6:.2f}MB).")
+
+        print(f"""
+GitHub doesn't allow artifact downloads from anonymous users. Either download the following file
+via your signed in browser, or consider using a point release download via invoke frontend-download --tag <git-tag>
+
+    Download: https://github.com/{repo}/suites/{qc_run['check_suite_id']}/artifacts/{frontend_artifact['id']} manually and
+    continue by running: invoke frontend-download --extract --file <path-to-downloaded-zip-file>""")
```
wolflu05, line 929: Good catch, I would remove the `--extract` option here, as it defaults to True and its also not used in all other suggestions. The reason why it was here is, that I first planned to use the extract option to also supply the path, but then quickly noticed that its not possible with a boolean setting, but I already had written this print statement.

tasks.py
```diff
@@ -304,8 +327,11 @@ def update(c, skip_backup=False):
     # Perform database migrations
     migrate(c)
 
-    # Compile frontend
-    frontend_compile(c)
+    # Decide if we should compile the frontend or try to download it
+    if node_available():
+        frontend_compile(c)
+    else:
+        frontend_download(c)
```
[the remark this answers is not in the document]
matmair: And i amended the docs

tasks.py
matmair: I find this way better understandable but no hard feelings either way

tasks.py
matmair: All the frontend commands expect yarn and our locking file is shipped for yarn so I thin enforcing it is the safer choice.

wolflu05: Ok, so there will be no warning if you already have node but not yarn, it'll default to download the frontend. Maybe just add a print here?

wolflu05:
Perfect, just only one little thing, the `docker-compose run inventree-server invoke update` may already be something users documented somewhere on their own updating guides. I would default to `--no-frontend` if the `INVENTREE_DOCKER` env is `true`.

wolflu05 (approved):
LGTM

SchrodingersGat (approved):

PR #5281, opened 2023-07-18
Update InvenTree demo url

hasecilu:
When customizing InvenTree on the `config.yaml` file I found that the url to the demo page is outdated.
Although there is a clickable url redirection I think it's better to update the url unless it's planned to change again.


SchrodingersGat:
Nice catch :)

PR #5304, opened 2023-07-21
Show built in label printing plugin even if the external plugins are …

martonmiklos:
…disabled


SchrodingersGat:
Good find, I must not have had the right setup for my testing.

PR #5314, opened 2023-07-22
P UI: Show server version on auth

matmair:
This PR adds the API and server versions to the login screen.


SchrodingersGat:
Most users are not going to need to know this, and potentially might be confusing? Is this for debug purpose?

matmair:
@SchrodingersGat I think this could be helpful for error reports and is a small footprint
grafik
grafik
The more detailed info is only shown on the instance selector that will go away once the backend stuff for server settings is ready
grafik

SchrodingersGat:
It is quite unobtrusive, that's fair enough :)

PR #5318, opened 2023-07-23
Fix selecting the default stock item label template

martonmiklos:
I have discovered an another mistake in the #4938: the auto selection of the stock item label type did not functioned due to using bad key/default value settings name. I apologize for the issue, I will take more care next time.


SchrodingersGat:
Thanks for chasing this up @martonmiklos

PR #5319, opened 2023-07-23
Fix admin url to point to right model

Nerade:


SchrodingersGat:
@Nerade thanks for the fix :)

PR #5327, opened 2023-07-23
Fix translation generation

matmair:
This PR fixes the translation extraction step for P UI.


matmair:
@SchrodingersGat this should fix the errors in the translation pipeline like [this one](https://github.com/inventree/InvenTree/actions/runs/5636446129/job/15268531738#step:6:82).

PR #5335, opened 2023-07-24
Fix "your" word duplicated

hasecilu:


hasecilu:
@SchrodingersGat, on some files is used "Send mail", should it be renamed to "Send email"? email was used on previous strings.
```
src/frontend/src/pages/Auth/Reset.tsx
41:              <Trans>Send mail</Trans>

src/frontend/src/pages/Auth/Set-Password.tsx
111:              <Trans>Send mail</Trans>

src/frontend/src/components/forms/AuthenticationForm.tsx
133:              <Trans>Send mail</Trans>
```

matmair:
The interface is very unstable. I would recommend letting us build something that is working and then starting to perfect it. There is a 50/50 chance any of those strings fly out.

hasecilu:
Ok, so the new section on Crowdin is still WIP? I'll check the other strings.

PR #5339, opened 2023-07-25
Remove `container-name:`s from Compose file

ssterling:
`container-name:` overrides any use of the [command line option `-p`](https://docs.docker.com/engine/reference/commandline/compose/#use--p-to-specify-a-project-name).  By not overriding container names, users may use `-p` to run multiple instances of InvenTree, for example: `docker compose -p personal up` creates containers `personal-inventree-worker-1`, `personal-inventree-proxy-1`, etc. while `docker compose -p business up` creates containers `business-inventree-worker-1` and so forth.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,7 +20,6 @@ services:
     # Use PostgreSQL as the database backend
     # Note: This can be changed to a different backend if required
     inventree-dev-db:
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@ssterling seems like a good improvement, thanks.

Issue #5330, opened 2023-07-24
OAuth login broken with Microsoft Graph due to failing sso_check_provider


hvraven:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

Since the upgrade from 0.10 first to 0.11, now 0.12.1 the previously working OAuth Login via Microsoft Graph is broken on our installation. My primary problem is the lack of any output from inventree on whats not working. When acccessing the login URL I get the error message, stating "The selected SSO provider is invalid, or has not been correctly configured", but no information in the error log or on the log in the container. I increased the log level to DEBUG, but that added primarily access logs.

Is there any way to get more log information out of Inventree and figure out why it doesn't like the configuration. I verified that all the required entries are (still) there.

### Steps to Reproduce
* Install inventree, configuring basic setup incl. e-mail
* Configure backends & middlewares. Section of my `config.yaml`:
```yaml
authentication_backends:
  - 'django.contrib.auth.backends.ModelBackend'
  - 'allauth.account.auth_backends.AuthenticationBackend'

#  Custom middleware, sometimes needed alongside an authentication backend change.
middleware:
  - 'django.middleware.security.SecurityMiddleware'
  - 'x_forwarded_for.middleware.XForwardedForMiddleware'
  - 'django.contrib.sessions.middleware.SessionMiddleware'
  - 'django.middleware.locale.LocaleMiddleware'
  - 'django.middleware.common.CommonMiddleware'
  - 'django.middleware.csrf.CsrfViewMiddleware'
  - 'corsheaders.middleware.CorsMiddleware'
  - 'django.contrib.auth.middleware.AuthenticationMiddleware'
  - 'InvenTree.middleware.InvenTreeRemoteUserMiddleware'
  - 'django_otp.middleware.OTPMiddleware'
  - 'django.contrib.messages.middleware.MessageMiddleware'
  - 'django.middleware.clickjacking.XFrameOptionsMiddleware'
  - 'InvenTree.middleware.AuthRequiredMiddleware'
  - 'maintenance_mode.middleware.MaintenanceModeMiddleware'
  - 'InvenTree.middleware.InvenTreeExceptionProcessor'

# Add SSO login-backends (see examples below)
social_backends:
  - 'allauth.socialaccount.providers.microsoft'

# Add specific settings for social account providers (if required)
social_providers:
  microsoft:
    TENANT: 'uuid-of-my-tenant'
```
[...]

PR #5340, opened 2023-07-25
Fix SSO check comparing id against name and extend log output

hvraven:
While trying to debug my problems described in #5330 I found that the failing check was the problem. This adds some error log to inform the user about the failure, otherwise there is no indication that the check has not passed.

### DIFF
--- a/InvenTree/part/templatetags/sso.py
+++ b/InvenTree/part/templatetags/sso.py
@@ -38,12 +38,12 @@ def sso_check_provider(provider):
     from allauth.socialaccount.models import SocialApp
 
     # First, check that the provider is enabled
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@hvraven while this will appear in the logs, it won't be visible to the front-end user. This is a good starting point, thought.

hvraven:
Yes, I am aware. This is only targeted at the admin. Right now the admin doesn't receive any information either. There are for sure better solution than this, but it's simple :-)

hvraven:
I included a fix for the problem described in #5330. I did however not check it on any other providers than the microsoft graph provider.

PR #5343, opened 2023-07-25
[P_UI] Added django settings for p_ui

wolflu05:
This PR improves a few things:
- adds the ability to define normal Django settings that can be used in P_UI
- adds the proxy option to the vite config, that way we the api and vite dev server run on the same url
- added current url als default server
- make host list configurable via Django settings
- added option to don't render the host list edit button
- use polling only for wsl


InvenTree/config_template.yaml
```diff
@@ -240,6 +240,15 @@ remote_login_header: HTTP_REMOTE_USER
 #   hide_admin_link: true
 #   hide_password_reset: true
 
+# Platform UI options
+# pui_settings:
+#   server_list:
+#     my_server1:
+#       host: https://demo.inventree.org/api/
+#       name: InvenTree Demo
```
matmair, line 248: We need a key for the server items too.

InvenTree/web/templatetags/spa_helper.py
```diff
@@ -34,3 +34,9 @@ def spa_bundle():
     return mark_safe(
         f"""<script type="module" src="{settings.STATIC_URL}web/{index['file']}"></script>{imports_files}"""
     )
+
+
+@register.simple_tag
+def spa_settings():
+    """Render settings for spa."""
+    return mark_safe(f"""<script>window.INVENTREE_SETTINGS={json.dumps(settings.PUI_SETTINGS)}</script>""")
```
matmair, line 42: Question: Can we move the serialisation step into global as the settings will not change during runtime?

src/frontend/src/components/forms/AuthFormOptions.tsx
```diff
@@ -19,12 +19,16 @@ export function AuthFormOptions({
...
-          {server.version} | {server.apiVersion}
-        </Text>
+        {window.INVENTREE_SETTINGS.show_server_selector && (
+          <>
+            <Tooltip label={hostname}>
+              <IconServer onClick={toggleHostEdit} />
+            </Tooltip>
+            <Text c={'dimmed'}>
+              {server.version} | {server.apiVersion}
+            </Text>
```
matmair, line 29: Suggestion: I think that versions should still be shown even if the sector is disabled

src/frontend/vite.config.ts
```diff
@@ -14,5 +14,14 @@ export default defineConfig({
   build: {
     manifest: true,
     outDir: '../../InvenTree/web/static/web'
+  },
+  server: {
+    proxy: {
+      '/api': {
+        target: 'http://localhost:8000',
+        changeOrigin: true,
+        secure: true
```
matmair, line 23: Question: How does this handle certificates? Ie if the Django server is only available with https

matmair (approved):
LGTM

src/frontend/vite.config.ts
wolflu05: Not sure, but this is only for development (I haven't setup https there) and secondly you don't need to use that feature and develop like you already do for now.

InvenTree/config_template.yaml
wolflu05: A key for what, can you please clarify what you mean with server items

InvenTree/web/templatetags/spa_helper.py
wolflu05: Into what global? As I discovered, to my suprise, these template tags only get called once after each server start for the first request.

src/frontend/src/components/forms/AuthFormOptions.tsx
wolflu05: We definitely can, but though I thought the current ui also doesn't show them. And there is one little bug which I was not able to fix. If you open the page with a clean local storage (e.g. private tab) the server version is not shown, you need to reload the page so that the hostList is loaded from local storage. Do you know how to fix that?

src/frontend/vite.config.ts
matmair: Sounds good

src/frontend/src/components/forms/AuthFormOptions.tsx
matmair: Probably because the state is not fetched. I will submit a path to your branch.

InvenTree/web/templatetags/spa_helper.py
matmair: File globals so it gets loaded when Django start up. The tags probably get cached if you use the same client and inputs. The page also only gets loaded by the browser if you do a hard refresh or new tab - I tried to reduce that as much as possible to get away from django template rendering.

InvenTree/config_template.yaml
matmair: A key per listed server. The `HostList` type (frontend) consists of a string as key and a dict with host and name. A key that is reliable between reloads is needed so that the selected host does not change between reloads (when the server list gets appended for example).

InvenTree/web/templatetags/spa_helper.py
wolflu05: where are these file globals?

InvenTree/config_template.yaml
wolflu05: And what is you proposed change here now? Sorry don't get it.

src/frontend/src/components/forms/AuthFormOptions.tsx
wolflu05: Found how to fix it in 5bd3782bad22d1adb8928f5a35b788fa252078ef

src/frontend/src/main.tsx
```diff
@@ -2,6 +2,26 @@ import React from 'react';
...
+      server_list: HostList;
+      default_server?: string;
+      show_server_selector: boolean;
+    };
+  }
+}
+
+window.INVENTREE_SETTINGS = {
+  server_list: {},
+  show_server_selector: true,
```
wolflu05, line 20: Do you think this should be enabled by default? My suggestion would be also to remove the demo server from the host list, if you need it you can configure it. But if you want to leave it, feel free, then Ill do that only for my deployments via configuration.

SchrodingersGat: I think for development, it's fine. But *most* installs will want to keep the simplistic single-server approach by default. I think we should probably disable it for production release.

wolflu05:
@matmair Can you help me here with the remaining issues?

src/frontend/src/main.tsx
wolflu05: This should be resolved now. If we are in dev mode, the demo server and the current url are the default. (That's also why I added the proxy option). And if we are not in dev, only the current path is in the host list which is also the default and we don't show the server selector.

InvenTree/web/templatetags/spa_helper.py
matmair: Everything in module level is global in Python ([longer description](https://realpython.com/python-scope-legb-rule/#modules-the-global-scope)). So if you move the json.dumps out of the function onto the same level as spa_settings it will only need be run once

src/frontend/src/main.tsx
```diff
@@ -1,7 +1,44 @@
...
+  interface Window {
+    INVENTREE_SETTINGS: {
+      server_list: HostList;
+      default_server: string;
+      show_server_selector: boolean;
+    };
+  }
+}
+
+const IS_DEV = import.meta.env.DEV;
```
matmair, line 19: @wolflu05 how do we enable this on Netlify while also building with production settings?

InvenTree/web/templatetags/spa_helper.py
wolflu05: Oh yeah, now got what you want. I thought you wanted a global template tag or something what I didn't have knowledge on. Refactored now.

src/frontend/src/main.tsx
wolflu05: Added a check via env variables. Let me know if that's ok for you.

wolflu05:
Would be ready for merge from my POV, I added the translating commands to the netlify build process so that the ui previews right.

matmair:

wolflu05:
@SchrodingersGat just a gently ping on this pr.

SchrodingersGat:
@wolflu05 thanks for the reminder :)

Can you run me through how the settings get from `.py` to `.ts`? Are they simply rendered into the `index.html` page?

If so, does this get rendered "live" by the server? What happens if the backend settings change?

wolflu05:
> Are they simply rendered into the index.html page?

Yes, they get rendered into the index.html page accessible on the global window object.

> If so, does this get rendered "live" by the server? What happens if the backend settings change?

Settings change in the config file/env variable requires a server restart until the new settings are available. Due to the implementation from @matmair the server list settings only get changed if they are not already saved in the local storage. So you need to clean your local storage until the new server list takes affect or use a new private tab for testing each time.

SchrodingersGat:
Ok, happy with that. Thanks for getting this up and running :)

Issue #5348, opened 2023-07-26
Get error "metadata: Metadata must be a python dict object" when importing Companies/Parts/etc through csv/json file


SchrodingersGat:
### Discussed in https://github.com/inventree/InvenTree/discussions/5322

<div type='discussions-op-text'>

<sup>Originally posted by **2old4it** July 24, 2023</sup>
Way to reproduce:
- Add 1 company in web interface
- do export -> tried with csv and json
- Add 1 company in the export file
- do import in web interface of same export file

After export I get a file with:

`[{"id": 1, "metadata": "{}", "notes": "", "name": "DigiKey", "description": "", "website": "https://www.digikey.com/", "phone": "", "email": "", "contact": "", "link": "", "image": "", "is_customer": "0", "is_supplier": "1", "is_manufacturer": "0", "currency": "USD"}]`

The metadate is filled with two brackets. However if I add another company using the exact same template:

`[{"id": 1, "metadata": "{}", "notes": "", "name": "DigiKey", "description": "", "website": "https://www.digikey.com/", "phone": "", "email": "", "contact": "", "link": "", "image": "", "is_customer": "0", "is_supplier": "1", "is_manufacturer": "0", "currency": "USD"},{"id": 2, "metadata": "{}", "notes": "", "name": "WhatEver", "description": "", "website": "https://www.wahtever.somewhere/", "phone": "", "email": "", "contact": "", "link": "", "image": "", "is_customer": "0", "is_supplier": "1", "is_manufacturer": "0", "currency": "USD"}]`

I get the error: **"metadata: Metadata must be a python dict"** when I try to import the file using the web interface.

My installation is a baremetal one within a python (3.11) virtual environment. It all seams to work.
`git log => commit 93b44ad8e6abdc970969360708aed3ff8189ec93 (HEAD -> stable, tag: 0.12.1, origin/stable)`

Any suggestions?
</div>

PR #5349, opened 2023-07-26
Exclude some common fields from django-import-export

SchrodingersGat:
- Add "get_fields()" method to InvenTreeResource
- Override default behaviour and exclude some common fields
- Will flow down to any inheriting classes
- Closes https://github.com/inventree/InvenTree/issues/5348

### DIFF
--- a/InvenTree/InvenTree/admin.py
+++ b/InvenTree/InvenTree/admin.py
@@ -31,3 +31,15 @@ class InvenTreeResource(ModelResource):
                 row[idx] = val
 
         return row
+
+    def get_fields(self, **kwargs):
+        """Return fields, with some common exclusions"""
+
+        fields = super().get_fields(**kwargs)
+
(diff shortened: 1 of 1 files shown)


InvenTree/InvenTree/admin.py
```diff
@@ -31,3 +31,15 @@ def export_resource(self, obj):
...
 
         return row
+
+    def get_fields(self, **kwargs):
+        """Return fields, with some common exclusions"""
+
+        fields = super().get_fields(**kwargs)
+
+        fields_to_exclude = [
+            'metadata',
```
wolflu05, line 41: But doesn't that mean one could not import metadata anymore if one wants?

SchrodingersGat: Yes, for now at least. The existing implementation was inconsistent - metadata was excluded from import on a bunch of models already.

PR #5359, opened 2023-07-27
P-UI: Add playground for new components

matmair:
The playground can be used to show new functions and possibilities


wolflu05:
Maybe we can hide this for prod by using the `IS_DEV_OR_DEMO` variable exported by `main.tsx` in #5343

matmair:

Issue #5333, opened 2023-07-24
Internal Error when Attempting to Verify Email Address


antbarney:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

An Internal Error is thrown when following the link to "Verify" an email address

### Steps to Reproduce

1. Send Email Verification email.
2. Follow the link to verify email address

### Expected behaviour

Address is verified 

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.13.0 dev
Django Version: 3.2.20
Commit Hash: 9b377cc
Commit Date: 2023-07-23

Database: postgresql
Debug-Mode: True
Deployed using Docker: True
Platform: Linux-6.1.34-Unraid-x86_64-with
Installer: DOC

Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}, {'name': 'InvenTreeLabel', 'slug': 'inventreelabel', 'version': '1.0.0'}]

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

### Relevant log output

```shell
ImproperlyConfigured at /api/auth/registration/account-confirm-email/Mg:1qO0Cd:hlqLndq87ETqUeWWTHXfzPaYql4lxX2IuQLSH3tMe_0/
TemplateResponseMixin requires either a definition of 'template_name' or an implementation of 'get_template_names()'
Request Method:	GET
Request URL:	http://10.10.20.10:8000/api/auth/registration/account-confirm-email/Mg:1qO0Cd:hlqLndq87ETqUeWWTHXfzPaYql4lxX2IuQLSH3tMe_0/
Django Version:	3.2.20
Exception Type:	ImproperlyConfigured
Exception Value:	
TemplateResponseMixin requires either a definition of 'template_name' or an implementation of 'get_template_names()'
Exception Location:	/usr/local/lib/python3.10/site-packages/django/views/generic/base.py, line 147, in get_template_names
Python Executable:	/usr/local/bin/python
Python Version:	3.10.12
Python Path:	
['/home/inventree/InvenTree',
 '/home/inventree',
 '/usr/local/bin',
 '/usr/local/lib/python310.zip',
[...]

Issue #3531, opened 2022-08-13
[FR] Add email link based logins


matmair:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

We currently have no password-less forms of logging in

### Suggested solution

Implement magic links like slack provides.
Ref: https://github.com/aaugustin/django-sesame

### Describe alternatives you've considered

Allauth options are also good for now

### Examples of other systems

Slack

### Do you want to develop this?

- [X] I want to develop this.

rkalman:
Using WebAuthn can be a good alternative for passwordless logins.
https://github.com/inventree/InvenTree/issues/2326#issuecomment-1299080622

matmair:
@rkalman WebAuthn has nothing to do with link based logins (often called magic links) and should be its separate FR if you want it.

WebAuthn is an additional factor for login while link-based logins replace the login entirely with a mail with a short lived token for single-use login. The first is used to add security while the latter is mostly used for users that seldomly use a platform.

PR #5396, opened 2023-08-03
Email fix

SchrodingersGat:
Fixes issues with confirming email addresses

Closes https://github.com/inventree/InvenTree/issues/5333
Reference https://django-rest-auth.readthedocs.io/en/latest/faq.html

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -608,7 +608,7 @@ REMOTE_LOGIN_HEADER = get_setting('INVENTREE_REMOTE_LOGIN_HEADER', 'remote_login
 
 # Magic login django-sesame
 SESAME_MAX_AGE = 300
-LOGIN_REDIRECT_URL = "/platform/logged-in/"
+LOGIN_REDIRECT_URL = "/logged-in/"
 
 # sentry.io integration for error reporting
 SENTRY_ENABLED = get_boolean_setting('INVENTREE_SENTRY_ENABLED', 'sentry_enabled', False)
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -608,7 +608,8 @@ REMOTE_LOGIN_HEADER = get_setting('INVENTREE_REMOTE_LOGIN_HEADER', 'remote_login
 
 # Magic login django-sesame
 SESAME_MAX_AGE = 300
-LOGIN_REDIRECT_URL = "/logged-in/"
+# LOGIN_REDIRECT_URL = "/platform/logged-in/"
+LOGIN_REDIRECT_URL = "/index/"
 
 # sentry.io integration for error reporting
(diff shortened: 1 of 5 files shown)


SchrodingersGat:
@matmair I've worked out why the endpoint did not have a template - but, the LOGIN_REDIRECT_URL is now hardcoded for the "platform" UI - this seems like an oversight, as this is not anywhere near being the default UI yet?

InvenTree/InvenTree/settings.py
```diff
@@ -608,7 +608,7 @@
 
 # Magic login django-sesame
 SESAME_MAX_AGE = 300
-LOGIN_REDIRECT_URL = "/platform/logged-in/"
+LOGIN_REDIRECT_URL = "/logged-in/"
```
matmair, line 611: @SchrodingersGat pretty sure this endpoint does not exist - `/plattform/logged-in/` is a jsx endpoint on P UI, not a Django endpoint

InvenTree/InvenTree/urls.py
```diff
@@ -79,13 +80,16 @@
...
-    path('auth/social/<int:pk>/disconnect/', SocialAccountDisconnectView.as_view(), name='social_account_disconnect'),
+    # Auth API endpoints
+    path('auth/', include([
+        re_path(r'^registration/account-confirm-email/(?P<key>[-:\w]+)/$', ConfirmEmailView.as_view(), name='account_confirm_email'),
+        path('registration/', include('dj_rest_auth.registration.urls')),
+        path('providers/', SocialProvierListView.as_view(), name='social_providers'),
+        path('social/', include(social_auth_urlpatterns)),
+        path('social/', SocialAccountListView.as_view(), name='social_account_list'),
+        path('social/<int:pk>/disconnect/', SocialAccountDisconnectView.as_view(), name='social_account_disconnect'),
+        path('', include('dj_rest_auth.urls')),
```
matmair, line 91: This should be the first item so we can overwrite several defaults from dj-rest-auth

matmair:
@SchrodingersGat LOGIN_REDIRECT_URL is needed for magic login so we either need to cut that feature (https://github.com/inventree/InvenTree/issues/3531) from 0.13 or write some view that redirects the user and selects somehow which UI the users currently uses.

InvenTree/InvenTree/urls.py
SchrodingersGat: @matmair sorry which line in particular should be first?

SchrodingersGat:
> @SchrodingersGat LOGIN_REDIRECT_URL is needed for magic login so we either need to cut that feature (#3531) from 0.13 or write some view that redirects the user and selects somehow which UI the users currently uses.

For now, the solution needs to ensure that the current (legacy) UI is hard-coded. A future PR can be added to support platform too, but current master assumes the login redirect will go to "platform" which is not correct

PR #5403, opened 2023-08-05
Increased settings value length from 200 to 2000

### LINKED ISSUE #5400: Plugin configuration values are limited to 200 chars, which is not enough for apprise

### Describe the bug*

I ran into a problem while configuring inventree-apprise, where the required configuration field is not long enough (reported at the plugin here: https://github.com/matmair/inventree-apprise/issues/1). In short the MS Teams integration requires a long URI with several tokens which does not fit into the 200 chars limit of the configuration field.

### Steps to Reproduce

Add `inventree-apprise` and attempt to configure a MS Teams plugin by setting `NOTIFICATION_APPRISE_URL` using a URI longer than 200 chars. The value will be cut at 200 chars, rendering the config invalid.

### Expected behaviour

the string will fit into the config value

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

0.12.3 with inventree-apprise 1.0

### Relevant log output

_No response_

matmair:
This PR enables longer settings - see discussion in https://github.com/matmair/inventree-apprise/issues/1 why this might be needed.
Closes #5400


SchrodingersGat:
Thanks @matmair

PR #5405, opened 2023-08-06
Add mermaid to docs

matmair:
This PR adds mermaid support to the docs - this is needed for DAGs and other graphs - useful for dev docs.


matmair:
Not stale

wolflu05:
@matmair What is the state of this pr? Would be cool to use mermaid for the machine registry lifecycle diagram.

Do you plan to mark this PR as ready, or do you wait if we need a different plugin when the docs are possibly handled with docusaurus?

matmair:
I stopped thinking about it because of the shift to docusaurus - which I think will be bad for docs in the long run because they will be less portable.
If you need it I will finish it right now, should be pretty simple.

matmair:
grafik
working now; have to check where the CI error comes from

wolflu05:
Thank you very much for finishing it. Really appreciate it.

Would be interested in what you mean with 
> because they will be less portable.

At the end docusaurus still uses markdown files like mkdocs does so too. For mermaid we would just need to add a [docusaurus mermaid plugin](https://docusaurus.io/docs/markdown-features/diagrams) like in mkdocs if we switch.

matmair:
> markdown files like mkdocs does so too

From what docusaurus documentations states and what the PR proposes this is not true. It uses primarily MDX - which is markdown + JSX in one file - A pattern which is cursed for portable and long living docs. Building a PDF or something similar with JSX involved is not fun. You can rebuild from markdown + yaml pretty easily on any static generator - JSX/React is very much one stack and a fast moving one too.

Looking at docusaurus [latest major release](https://docusaurus.io/blog/releases/3.0) alone gives me multiple headaches about maintainabiltity of any system build on it but I have already stated in the PR that I will not work on anything using it so not really a me problem.

matmair:
@SchrodingersGat this is ready for review

wolflu05:
First of all, thank you very much for spending the time on adding mermaid in this PR. I now tried it out in my branch and it seems like it's not really working correctly with *larger diagrams*. The labels just stick out of the boxes, markdown inside is not supported and there is no text wrap. 

image


Compared to on mermaid.live:
image


Do you have any idea how I can get the max box size configured so that the text wraps?

wolflu05:
Mh, tested a lot of things including setting the mermaid config by adding this to `docs/javascripts/extra.js`
```js
window.mermaidConfig = {
	flowchart: {
		useMaxWidth: true,
		htmlLabels: false,
		wrappingWidth: 100,
	}
};
```

but none of this seems to work. (I only got it to work so that the labels do not overflow, but no wrapping) I think that's not worth the effort, I'll then just try to document that using normal text.

For reference this also doesn't look great with the gh mermaid renderer:

```mermaid
flowchart LR
    A["`**Server start**`"] --> B
    B["`**Stage 1: Discover machine types**
    by looking for classes that inherit the BaseMachineType class`"] --> C
    C["`**Stage 2: Discover drivers**
    by looking for classes that inherit the BaseDriver class (and are not referenced as base driver for any discovered machine type)`"] --> MA
    subgraph MA["Stage 3: Machine loading"]
        direction TB
        D["`For each MachineConfig in database instanciate the MachineType class (drivers get instaciated here as needed and passed to the machine class. But only one instance of the driver class is maintained along the registry)`"] --> E
        E["`The driver.init_driver function is called for each used driver`"] --> F
        F["`The machine.initialize function is called for each machine, which calls the driver.init_machine function for each machine`"]
    end
    MA --> X["`**Done**`"]
```

matmair:
@wolflu05 It also looks weired when I render it locally - maybe mermaid live is using a newer/better version of mermaid?

wolflu05:
I also tried to override the version and checked that v10.8.0 is used, but it seems like that's not the issue. I also tried to use the mermaid2 plugin for mkdocs but with no success. I think I'll just let it out then. But thank you for your time.

matmair:
I think this would be helpful nonetheless for architecture graphs etc @SchrodingersGat

PR #5410, opened 2023-08-08
Bring stable up to speed with 0.12.x

matmair:
This PR is to bring the stable branch (which is the default package branch) up to speed with the 0.12.x improvements up to 0.12.5.
The 0.12.x branch should not be deleted.


SchrodingersGat:
@matmair this should occur automatically when all the checks for `0.12.4` pass

Issue #5420, opened 2023-08-10
Export records example creates the output in ephemeral, unaccessible location inside container


Petrox:
### Body of the issue

https://docs.inventree.org/en/latest/start/migrate/#export-data

Command: 

````
invoke export-records -f data.json

````

Suggestion to mention docker scenario 

````
invoke export-records -f data/backup/data.json
````

or

````
docker compose run inventree-server invoke export-records -f data/backup/data.json
````

SchrodingersGat:
@Petrox good point, would cause issues for docker users. Would you be willing to submit a PR to the documentation to address this case?

Petrox:
Let me know if this works for you. 

I intentionally didn't alter any other part of the document.

SchrodingersGat:
Closed in https://github.com/inventree/InvenTree/pull/5428

PR #5428, opened 2023-08-11
Update migrate.md to handle exporting json via docker volume #5420

Petrox:
Extra paragraph to handle exporting JSON from docker container to docker volume.


SchrodingersGat:
Very nice, thanks @Petrox

Issue #5417, opened 2023-08-09
invtree.org should link to STABLE docs instead of LATEST (which itself has commands that lead to error messages due to version differences)


Petrox:
### Body of the issue

Thank you for your time and effort to create this app, looks like a good initiative. 
I'm a new user and my early contributions will be trivial onboarding experience issues

My issue is that "assuming LATEST docs works for STABLE users" is mostly true, but is not technically correct. The main webpage should link to STABLE, since that is ready for production use anything else is not.

This might be fixed with a single HREF change on the homepage.

User experience:

 - Start: https://inventree.org/
 - click docs https://docs.inventree.org/en/latest/ 
 - click INSTALL,  https://docs.inventree.org/en/latest/start/intro/ 
 - follow the manual to docker https://docs.inventree.org/en/latest/start/docker_prod/
 - follow the guide to install the STABLE version of the application (while reading the LATEST docs).
 - https://docs.inventree.org/en/latest/start/docker_prod/#initial-database-setup

Command: 
````
docker compose run inventree-server invoke update --no-frontend
````

Output containing error message:

````
[+] Building 0.0s (0/0)
[+] Creating 2/0
 ✔ Volume "invtreepg_inventree_data"   Created                             0.0s
 ✔ Container invtreepg-inventree-db-1  Created                             0.0s
[+] Running 1/1
 ✔ Container invtreepg-inventree-db-1  Started                             0.3s
[+] Building 0.0s (0/0)
Creating directory /home/inventree/data/static
Creating directory /home/inventree/data/media
Creating directory /home/inventree/data/backup
Copying config file to /home/inventree/data/config.yaml
No idea what '--no-frontend' is!
````
ERROR

Docker images:

````
$ docker image ls
REPOSITORY               TAG       IMAGE ID       CREATED         SIZE
inventree/inventree      stable    cc009d342f7f   9 days ago      1.15GB
postgres                 13        ebc0be6af283   12 days ago     407MB
nginx                    stable    55d65c7ff24d   12 days ago     142MB
````

I have double checked and 

https://docs.inventree.org/en/stable/start/docker_prod/#initial-database-setup
has the correct command.

This means:
 - docs for any specific version of the app is probably OK
 - but the LATEST version of the docs could link to the STABLE docs when it comes to INSTALL PRODUCTION DOCKER
 - also the homepage should link to the STABLE docs since most newbie users don't want to eat the hot potato esp not if they want production.
[...]

PR #5431, opened 2023-08-11
Add comments and warnings to documentation so readers would follow the correct version 

Petrox:
proposed solution for #5417


SchrodingersGat:
@Petrox thanks :)

PR #5440, opened 2023-08-14
Fix for potential NoReverseMatch error

SchrodingersGat:
- Check that the database model really does exist in the template code

### DIFF
--- a/InvenTree/stock/templates/stock/item_base.html
+++ b/InvenTree/stock/templates/stock/item_base.html
@@ -368,7 +368,11 @@
     <tr>
         <td><span class='fas fa-th-list'></span></td>
         <td>{% trans "Sales Order" %}</td>
-        <td><a href="{% url 'so-detail' item.sales_order.id %}">{{ item.sales_order.reference }}</a> - <a href="{% url 'company-detail' item.sales_order.customer.id %}">{{ item.sales_order.customer.name }}</a></td>
+        <td>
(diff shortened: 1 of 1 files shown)


InvenTree/stock/templates/stock/item_base.html
```diff
@@ -368,7 +368,11 @@ <h5>{% if item.quantity != available %}{% decimal available %} / {% endif %}{% d
     <tr>
         <td><span class='fas fa-th-list'></span></td>
         <td>{% trans "Sales Order" %}</td>
-        <td><a href="{% url 'so-detail' item.sales_order.id %}">{{ item.sales_order.reference }}</a> - <a href="{% url 'company-detail' item.sales_order.customer.id %}">{{ item.sales_order.customer.name }}</a></td>
+        <td>
+            <a href="{% url 'so-detail' item.sales_order.id %}">{{ item.sales_order.reference }}</a>
+            {% if item.sales_order.customer %}
+            - <a href="{% url 'company-detail' item.sales_order.customer.id %}">{{ item.sales_order.customer.name }}</a></td>
+            {% endif %}
```
wolflu05, line 375: The td closing Tag only gets rendered of the condition is true, this might be an issue.

SchrodingersGat: @wolflu05 ah, yeah good catch. I'll fix that.

SchrodingersGat: https://github.com/inventree/InvenTree/pull/5445

PR #5480, opened 2023-08-25
Fix missing filters for get settings validator

wolflu05:
This PR fixes the missing filters to get the settings validators for plugins to work again.


matmair:
@wolflu05 you might need to merge/wait for https://github.com/inventree/InvenTree/pull/5486 to fix the doc's errors

SchrodingersGat:
@wolflu05 please check out the CI errors

wolflu05:
I didn't manage to identify why I sometimes not get the kwargs but in the tests I get them. This is the easiest fix that works. Do you have a better idea of fixing the validators for plugins?

SchrodingersGat:
@wolflu05 happy with this fix, but given this might change some pretty low-level stuff, can you please provide some updated unit tests? In particular related to the edge cases you were seeing

SchrodingersGat:
@wolflu05 would you be able to throw in some extra unit tests here?

wolflu05:
@SchrodingersGat I now found the place where the clean function is called with additional kwargs which is against Django's implementation, so I removed the kwargs completely and replaced them with the `self.get_filters_for_instance()` function. I added two tests, one directly for the Settings and one more integrated into the plugin registry to make sure validators actually work with settings that have extra relations too. Now I only need to find out, why the test passes locally but not in CI.

SchrodingersGat:
These types of errors are always very frustrating to track down! Thanks for adding the tests

wolflu05:
@SchrodingersGat I found the culprit. I came across this after connecting via tmate and adding several prints to the code while running the tests over and over and see how far it runs. I then discovered that the validator doesn't run because it's not callable, but he is a static method. I then googled why static methods are sometimes not callable (I used a static method for the validator) and found [this](https://stackoverflow.com/questions/45375944/python-static-method-is-not-always-callable). I then asked myself why it works on my machine and on ci not and then found out, I run python 3.10 in the devcontainer and ci 3.9. Python must have changed that behavior. I can validate that by running the provided snippet in the StackOverflow post and on my machine with 3.10 it returns True for both and on ci only for the direct check. 

So **TL;DR**, make sure we don't use static methods for validators anywhere, as they are not marked as callable in some circumstances. The tests should pass now after moving the validator into the global scope as a normal function.

SchrodingersGat:
@wolflu05 that's some fine detective work, well done!

wolflu05:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|0.12.x|[](https://github.com/inventree/InvenTree/pull/5522)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

PR #5500, opened 2023-09-03
PUI: Scan Page

matmair:
This PR adds an experimental new page: the Scan Page.
grafik
This page allows users to scan multiple barcodes at once and execute actions against them. Scans are saved locally.

This is currently WIP as it depends on actions (for parts, stockitems, ...) to be implemented in PUI so they can be called from this page. Nevertheless I would be happy to get feedback.


matmair:
Updated screenshot
grafik

matmair:
@inventree/triage this would be ready for a feedback round. You can reach the page by selecting "Scanning" in the navigation pane.
You either need to run on localhost or serve via https to test this as access to camera APIs is restricted.

I would be interested if:
- the structure makes sense
- scanning of barcodes works on your end
- the texts make sense

and what actions should be exposed for the available barcodes (see screenshot above)

SchrodingersGat:
@matmair new screenshot looks amazing! I'll try to find some time to look at it this weekend :)

SchrodingersGat:
@matmair to clarify the intent here, this dedicated barcode page is intended to be separate from the "quick scan" barcode action?

### Quick Scan Action

Scans a barcode, and if identified as an "inventree" barcode (e.g part / location / etc) it will navigate the UI to the relevant page. No other actions, just scan

### New Barcode Screen

Dedicated for more complex barcode data entry. Scan multiple barcodes, and then perform specific actions with those barcodes:

- check into location
- receive against purchase order
- assign to build order

matmair:
@SchrodingersGat yes, you described it well.

SchrodingersGat:
I'm happy with the general idea, do you want to merge in now and update as we go?

matmair:
Sounds good @SchrodingersGat

Issue #5119, opened 2023-06-28
Add support to scan digiKey partlabels


davrex:
Add support to scan digiKey partlabels QR for easier adding of stock items.
Scan QR 
find matching part,
populate the quantity
![20230628_223507](https://github.com/inventree/inventree-app/assets/2648476/7ba75c43-2fba-4c0f-a2c2-afbe81581ee1)

SchrodingersGat:
not stale, still a good idea!

30350n:
Hey! I'd be interested in helping to implement this. I've already managed to get barcode scanning working by adding a barcode plugin, but in the current state this isn't really useful for checking in parts after receiving an order.

For me personally, the ideal workflow would be to:
1. Add the Part/ManufacturerPart/SupplierPart definitions to Inventree automatically after ordering, via something similar to [Ki-nTree](https://github.com/sparkmicro/Ki-nTree) (I'm currently working on my own simplified version of this. Ideally this would also have to be accessible from within the InvenTree interface directly).
2. Add the Order by providing the BOM from the manufacturer. (This already works pretty well, although I've had some issues with part value detection).
3. Once the order is received: Scan in each items barcode, directly followed by a Stock Location barcode, to automatically mark the item as received and create the new Stock item.

For me personally, I'd like to be able to do 3. quick and easily via the App. I'm thinking of adding a Scanning Action in the Purchase Orders section, like this:
<img src="https://github.com/inventree/InvenTree/assets/34186858/6e997012-0a54-44f4-a7bf-3021d311ae46" width="300">

The information present in the 2D barcode could then be used to automatically figure out the Order etc.

I'm not sure though how an implementation for the InvenTree web interface would like.
Would be very happy to get some feedback/more ideas on this topic! :)

matmair:
I am currently actively working on 3) here #5500 and have ideas for 1) in #3791

matmair:
Implementing this in the App shouldn't be too difficult, the source code is here: https://github.com/inventree/InvenTree-app

30350n:
> Implementing this in the App shouldn't be too difficult, the source code is here: https://github.com/inventree/InvenTree-app

Yep, I'm already on it! Main issue is I've never messed with app development/dart before really, but it looks pretty straight forward, so I'll see what I can come up with.

matmair:
> > I am currently actively working on 3) here #5500 and have ideas for 1) in #3791
> 
[...]

Issue #421, opened 2019-07-07
Add /api/order/ endpoint to apidoc


SchrodingersGat:
Order api is not currently appearing in the autodoc for the API.

PR #5509, opened 2023-09-05
Add basic support for scanning vendor barcodes

30350n:
This adds 2 simple (and very similar) builting plugins for scanning digikey and mouser 2D barcodes.

The code parses the barcode data fields and return a matching `SupplierPart`, based on SKU (for digikey) and MPN + supplier name (for mouser, because there sadly is no SKU present in the Mouser barcodes).

Related to #5119 (though not providing a full implementation) and probably also #3261.


matmair:
looks great @30350n! Do you have a mouser packaging label you could provide for testing? I only have Digikey shipping labels available.

30350n:
> looks great @30350n! Do you have a mouser packaging label you could provide for testing? I only have Digikey shipping labels available.

Sure, here's two!

### Mouser Barcode 1

Barcode Data: `[)>06KINC-2023-03-28-0000114K0111PMC34063ADRQ311K0731242494LMX1VTI`
(`INC-2023-03-28-00001` is a custom order number I provided during checkout).



### Mouser Barcode 2

Barcode Data: `>[)>06K2142055214K0331PLDK320ADU33RQ3211K0609387694LCN1VSTMicro`



And finally here's a stone old (2017 or 2018 iirc) LCSC QR code (haven't implemented a handler for that one yet):

Barcode Data: `{pbn:PICK2009290002,on:SO2009290140,pc:C312270,pm:ST-1-102-A01-T000-RS,qty:2,mc:,cc:1,pdi:34421807}`



**If anyone has a more recent one from LCSC, I'd be happy to take a look at implementing a handler!**

wolflu05:
This is really cool, and I'm sure many InvenTree users use mouser and digikey, but I'm not sure if this should be part of InvenTree core as not everybody is using InvenTree for electronics with those two vendors. I would suggest publish this as a separate plugin and maybe combine it in some time with an import function if this plugin feature is available, so that there are one plugin for each vendor handling, barcodes, imports and possibly also other things.

matmair:
I disagree, including some batteries to get to a functioning system faster seems better then having to explain how to install plugins for the n th time.
There are no external dependencies so I do not see any downsides if we are including it.

30350n:
I also think this is minimal and useful enough to include as a builtin plugin. You can also always disable it or disable plugins all together if you don't need the functionality. (Though admittedly the main reason for implementing this as a **builtin** plugin for me was because it seemed to be the easiest ^^)

I could see this getting kind of bloaty with one addon per vendor though, if I were to add even more vendors (LCSC, Farnell, RS-Components etc. come to mind).
@matmair do you think it'd be a good idea to leave this as a single plugin which could handle all kinds of vendor barcodes, or do you prefer to leave it as is?

matmair:
@30350n I prefer separate plugins.

Aslo Mouser Barcode 1 seems to match a Digikey barcode?

wolflu05:
I would say this is an argument to improve the plugin installing system then. I agree, currently it is to complicated: searching via google or the plugin page on the website for a specific plugin, copy the package name and paste it to InvenTree and install it. This should all be integrated int InvenTree like octoprint has it. In the end I imagine installing a plugin is a one click task after searching for the plugin.

We need some guidelines where we make a cut and don't install it by default. I'm sure many don't want to clutter their plugin list with unnecessary builtins that cannot be removed neither disabled. And in my opinion mouser and digikey are some specific vendors. We need to make a cut somewhere, because otherwise some other people come and implement reichelt, rs, farnell, ... and then we end up with a long list of plugins not everyone is using which makes the whole plugin system useless in my opinion.

1. I would suggest creating an inventree/inventree-plugins repository then and make a folder for each of these "official" plugins then and publish them individually as pip packages.
2. I would suggest improving the plugin list a bit so that it can be fetched via some kind of api and directly integrated into InvenTree like octoprint, nodered and iobroker have it.
3. I would suggest that if the import Mixin is there, these plugins also provide official support for importing parts and maybe also syncing orders and pricing.

matmair:
> this is an argument to improve the plugin installing system

You are very welcome to do that. Octoprint is written in flask which is very different from Django implementation-wise.

All the suggestions sound good but I would rather ship something that just works now than dream about something like #3261 for a year.

30350n:
> @30350n I prefer separate plugins.
> 
> Aslo Mouser Barcode 1 seems to match a Digikey barcode?

Huh, yeah, right, I actually didn't test with that one, but the second one. 1. is newer though, so I guess they decided to use the same "header" or "magic" or whatever as digikey ... which kind of makes them more complicated to distinguish.

matmair:
@30350n might that be an industry standard?

InvenTree/plugin/builtin/barcodes/mouser_barcode.py
```diff
@@ -0,0 +1,71 @@
...
+        if not manufacturer_parts or len(manufacturer_parts) > 1:
+            logger.warning(f"Found {len(manufacturer_parts)} manufacturer parts for MPN {barcode_fields.get('mpn')} with MouserBarcodePlugin plugin")
+            return
+        manufacturer_part = manufacturer_parts[0]
+
+        supplier_parts = SupplierPart.objects.filter(
+            manufacturer_part=manufacturer_part.pk)
+        for supplier_part in supplier_parts:
+            if "mouser" in supplier_part.supplier.name.lower():
+                break
```
matmair, line 59: I think this should be replaced with a plugin-level setting to select the "mouser" supplier.

InvenTree/plugin/builtin/barcodes/mouser_barcode.py
```diff
@@ -0,0 +1,71 @@
...
+logger = logging.getLogger('inventree')
+
+
+class MouserBarcodePlugin(BarcodeMixin, InvenTreePlugin):
+    """BarcodePlugin for matching Mouser barcodes."""
+
+    NAME = "MouserBarcode"
+    TITLE = _("Mouser Barcodes")
+    DESCRIPTION = _("Provides support for scanning Mouser barcodes")
+    VERSION = "0.1.0"
```
matmair, line 20: Suggestion: If it works it should be 1.0

InvenTree/plugin/builtin/barcodes/digikey_barcode.py
```diff
@@ -0,0 +1,66 @@
...
+logger = logging.getLogger('inventree')
+
+
+class DigiKeyBarcodePlugin(BarcodeMixin, InvenTreePlugin):
+    """BarcodePlugin for matching DigiKey barcodes."""
+
+    NAME = "DigiKeyBarcode"
+    TITLE = _("DigiKey Barcodes")
+    DESCRIPTION = _("Provides support for scanning DigiKey barcodes")
+    VERSION = "0.1.0"
```
matmair, line 20: Suggestion: If it works it should be 1.0

matmair (commented):
Works well overall; before this can be merged we should address 2 things:
- adding a few sentences to the docs about this detection feature
- adding test coverage to ensure this does not break in refactorings etc.

30350n:
>We need some guidelines where we make a cut and don't install it by default. I'm sure many don't want to clutter their plugin list with unnecessary builtins that cannot be removed neither disabled.

But you can disable builtin plugins though, right? I kind of get your point, but i feel like the solution would be to simply disable all builtin plugins by default (which kind of seems to be the case already?).

> This should all be integrated into InvenTree like octoprint has it. In the end I imagine installing a plugin is a one click task after searching for the plugin directly in the admin ui.

Yes, but that kind of system doesn't exist yet and once it **does exist**, it'll be trivial to integrate these plugins into it.

>I would suggest that if the import Mixin is there, these plugins also provide official support for importing parts and maybe also syncing orders and pricing.

I agree that this is probably a sensible way to move forward, but imo this is another argument **for** having these plugins as builtins. Because that way if other people would like to contribute new vendors or part importing or similar things, there'd be a central place for that code to exist instead of it being hidden away and scattered across multiple small repos all over the place.

wolflu05:
> All the suggestions sound good but I would rather ship something that just works now than dream about something like #3261 for a year.

So you're not against moving this out into a own package that is not installed by default if such a one-Click install flow I proposed above is available? Then I would be fine with merging this but move this out at a later point.

matmair:
> > All the suggestions sound good but I would rather ship something that just works now than dream about something like #3261 for a year.
> 
> So you're not against moving this out into a own package that is not installed by default if such a one-Click install flow I proposed above is available? Then I would be fine with merging this but move this out at a later point.

No. But I do not anticipate any such one-click solution any time soon. I tried building this multiple times over my 5+ years with django and it is a very complex effort that IMO is not worth the effort.

30350n:
> @30350n might that be an industry standard?

Yeah, probably. Though more like industry """standard""". I think this is some kind of quasi-standard for 2D barcodes, but the way digikey and mouser setup the data is just different enough to make it impractical to handle them together.
If you take a look at the implementation, the "per entry headers" are sometimes the same (`1P` and `Q` for MPN and quantity) and sometimes similar, but not quite the same (`10K`/`11K` for invoice number, etc.).

wolflu05:
> But you can disable builtin plugins though, right? 

If I remember correctly, they are forced to be active, but can't test currently.

> across multiple small repos all over the place

That's why I proposed the third point (using an official inventree/inventree-plugins monorepo)

matmair:
@wolflu05  If you do not mind I would appreciate moving the general plugin discussion to a "discussion", threading is hard in issues.

@30350n there seems to be a standard: https://www.ecianow.org/assets/docs/ECIA_Specifications.pdf - would you be up to implement/look into that?
grafik

30350n:
> @30350n there seems to be a standard: https://www.ecianow.org/assets/docs/ECIA_Specifications.pdf - would you be up to implement/look into that? grafik

Sure! Haven't seen that one yet, but that seems to be relatively straight forward.
Vendors using this barcode format would still possibly have to be handled differently though, so the two plugins should stay right?
I think I'd personally just implement this as a single function which parses a "ECIA 2D Barcode", but where would this function live? Should I add something like a `ecia_2d_barcode.py` to `plugin/base/barcodes`?

matmair:
@30350n I think a single plugin/function should be sufficient. Might be valuable to add the parsed data to the response for usage in custom frontends

InvenTree/plugin/builtin/barcodes/mouser_barcode.py
30350n: Done.

InvenTree/plugin/builtin/barcodes/mouser_barcode.py
30350n: Done.

InvenTree/plugin/builtin/barcodes/digikey_barcode.py
30350n: Done.

30350n:
I've separated out the ECIA barcode parsing code now into an extra `barcode2d.py` file (could probably move this elsewhere if desired. I've decided to keep the two separate plugins because:
1. Mouser barcodes still need to be handled a little differently, because they (for whatever reason) don't contain a SKU. Thus I've added a setting to the plugin now, to select the "Mouser Supplier" by name (seems like that's the way to go).
and
2. I'd think they'd act as a good base for part importing / #3261

30350n:
I also include the `quantity` and `order_number` in the response like this now:
```py
response = {
    "supplierpart": {
        "pk": 42,
        ...,
        "quantity": 1337,
        "order_number": "<the order number>", # this can be a custom order number too
    },
}
```
Do you think it would be cleaner to put them into a separate field?
Something like this comes to mind:
```py
response = {
    "supplierpart": {
        "pk": 42,
        ...,
    },
    "vendorbarcode"
        "quantity": 1337,
        "order_number": "<the order number>",
    },
}
```

InvenTree/plugin/builtin/barcodes/mouser_barcode.py
```diff
@@ -0,0 +1,92 @@
...
+    DESCRIPTION = _("Provides support for scanning Mouser barcodes")
+    VERSION = "1.0.0"
+    AUTHOR = _("InvenTree contributors")
+
+    SETTINGS = {
+        "MOUSER_SUPPLIER_NAME": {
+            "name": _("Mouser Supplier Name"),
+            "description": _("Name of the Supplier which acts as 'Mouser'"),
+            "default": "Mouser",
+            "validator": lambda name: bool(name),
```
wolflu05, line 29: Use a model field like this:
```suggestion
            "validator": lambda name: bool(name),
            "model": "company.company"
```

SchrodingersGat:
@30350n a broad implementation question. Is the intent here that when you scan a barcode, you either:

a) Provide information back to the scanner, but perform no action
b) Create new stock item, receive items against purchase order, and return stock item

I would suggest that implementing `b)` means that the caller (could be app, UI, or external API) just needs to scan the barcode, and all the magic happens on the server.

InvenTree/plugin/builtin/barcodes/mouser_barcode.py
SchrodingersGat: Yep, if you use a "model" attribute you can allow the user to select the supplier directly. So, you don't need to have them input a name and use name lookup (and then potentially change the config if the company name ever changes). They can just select "Mouser" (e.g.) from a drop-down list of companies in the system.

InvenTree/plugin/builtin/barcodes/digikey_barcode.py
```diff
@@ -0,0 +1,60 @@
...
+    def scan(self, barcode_data):
+        """Process a barcode to determine if it is a DigiKey barcode."""
+
+        if not (barcode_fields := parse_ecia_barcode2d(barcode_data)):
+            return
+
+        if not (sku := barcode_fields.get("supplier_part_number")):
+            return
+
+        supplier_parts = SupplierPart.objects.filter(SKU__iexact=sku)
```
SchrodingersGat, line 33: There's a fair bit of shared code here between the Digikey and Mouser plugins. I think it would be worth creating a top-level class which implements the generic ecia barcode functionality with helper methods such as `get_supplier_part()`?

InvenTree/plugin/builtin/barcodes/mouser_barcode.py
30350n: Thanks! I've been looking for something like that for a solid 15 minutes yesterday, but didn't find anything ^^

30350n: I've also added back some code from an ealier revision which tries to automatically guess the Mouser supplier if it hasn't been set yet (I feel like that's a good QOL addition).

InvenTree/plugin/builtin/barcodes/digikey_barcode.py
30350n: To me this is one of thoses cases where applying DRY would actually lead to less maintainable, more convoluted code. The `get_supplier_part()` function would only be like 3 lines and would have to live in yet another file probably. Not worth it imo.

The one thing I would like to abstract out is "adding the additional vendor barcode data to the response", this part:
```py
if quantity := barcode_fields.get("quantity"):
    try:
        data["quantity"] = int(quantity)
    except ValueError:
        logger.warning(
            f"Failed to parse quantity '{quantity}' with "
            f"MouserBarcodePlugin plugin"
        )

if order_number := barcode_fields.get("purchase_order_number"):
    data["order_number"] = order_number
```
I'm just still not set on the exact format of this, like I mentioned [here](https://github.com/inventree/InvenTree/pull/5509#issuecomment-1707411820) (I think I'd prefer the second example there).

30350n:
> @30350n a broad implementation question. Is the intent here that when you scan a barcode, you either:
> 
> a) Provide information back to the scanner, but perform no action b) Create new stock item, receive items against purchase order, and return stock item
> 
> I would suggest that implementing `b)` means that the caller (could be app, UI, or external API) just needs to scan the barcode, and all the magic happens on the server.

a) Because b) isn't really possible I think, the main problem being that to receive a purchase order item and return a stock item, you have to scan **2** barcodes. First the items vendor barcode, then the barcode for the stock location it'll be assigned too (alternatively this could also be set by hand, but as I see it, it has to be set *somewhere*).

My current approach for the App would be to have an option in the PO section to "Scan Received Items" which works like this:
1. Scan the items barcode and find the relevant PO + Line (if this fails, we just throw an error and do nothing)
2. Show a widget containing the Quantity (if possible, from the barcode result), Status and Stock Location (with the option to choose the Stock Location by scanning a barcode). This would be pretty much exactly the same as the existing widget for receiving items.
3. With that data, do another API request to receive the item and create the new stock item + also assign the vendor barcode to the stock item, **but only if it contains unique information like the order number** (DigiKey and Mouser barcodes always contain this, but other vendors might not), like you mentioned [here](https://github.com/inventree/InvenTree/issues/5119#issuecomment-1707504034).

30350n:
Refactored pretty much everything to make it easier to integrate #3791 to this in the future. I've also added LCSC barcode support after confirming with someone that their barcode format hasn't really changed from the 3 year old codes I've got.

SchrodingersGat:
There may be a slightly different way to approach this:

1. Scan the barcode, which receives stock against purchase order, and creates the new stock item.
2. Return the stock item API data (much like scanning a stock item now)
3. This pulls up the "stock item detail" screen (in either the web interface or app)
4. You can then scan into a location, or set the location manually

Hopefully I have explained this sufficiently. Let me know what you think

30350n:
Ah okay, I assumed that setting the StockLocation was required to create a StockItem in the API, because that's how it works in the app with the current "Receive Item" action.

Though tbh, I think I'd still prefer a separate "Scan Received Items" action. Your version would be easier to implement of course, but I think it'd be harder to understand for an end user and it'd also be kind of hidden away. I'd also prefer it if the App were to force you to assign a StockLocation, just because it makes it less likely that you forget to assign one ^^

SchrodingersGat:
From a workflow point of view, you can already assign an "location" to an incoming purchase order, and also to individual lines in the order. So, if that data is provided, it can be assigned to the stock item by the plugin.

Additionally, you could specify an "incoming goods" location which is set by default. The user can always "override" this by scanning into a different location later on.

So, location in order of preference is:

1. Location specified in purchase order
2. Default incoming goods location
3. No location

After receipt of goods, users can still transfer stock wherever.

If 1) or 2) is used, then the extra step can be skipped entirely

matmair:
@SchrodingersGat but how do you determine to which order the stock belongs? You can order the same part multiple times or am I missing something?

30350n:
> @SchrodingersGat but how do you determine to which order the stock belongs? You can order the same part multiple times or am I missing something?

Via the order number which is present in Digikey, Mouser and LCSC barcodes.

I've also already thought about how this could work for other suppliers which don't specify the order number. If there's the same part in multiple active orders, there's no way to find out which one it belongs to, as you said. But that's an edge case really. In most cases there'll probably only be a single matching item, in which case finding out the order number is trivial.

30350n:
> From a workflow point of view, you can already assign an "location" to an incoming purchase order, and also to individual lines in the order. So, if that data is provided, it can be assigned to the stock item by the plugin.
> 
> Additionally, you could specify an "incoming goods" location which is set by default. The user can always "override" this by scanning into a different location later on.

Perfect, that's very helpful actually! I think I'll take a look at implementing the required actions in the app tomorrow for that and create a PR there ^^

SchrodingersGat:
@30350n congratulations on being the projects' 3000th star :)

matmair:
@30350n that will require that the order number on the barcode and in inventree match - this can not be assumed.
I think it would make sense to create a new endpoint for any action that actively changes stuff in InvenTree when a barcode is scanned. The current function of the barcode endpoint is to lookup information, not change things. Modifying that might lead to unintended consequences for API users. If we do not do that we should mark this as breaking.

wolflu05:
Why not implement the changing operation in a second request made by the frontend after the barcode data is queried using the existing api?

30350n:
> @30350n that will require that the order number on the barcode and in inventree match - this can not be assumed.

Why not? That's similar to assuming the SKU from the barcodes matches a supplierpart, no?

>I think it would make sense to create a new endpoint for any action that actively changes stuff in InvenTree when a barcode is scanned. The current function of the barcode endpoint is to lookup information, not change things. Modifying that might lead to unintended consequences for API users. If we do not do that we should mark this as `breaking`.

That's actually not what I want to implement though. That more or less was @SchrodingersGat s idea of how this could be implemented. I'd still prefer to create a separate action which handles calling the correct API endpoints in the App (and Web Interface) as discussed before.

30350n:
> @30350n congratulations on being the projects' 3000th star :)

Haha, that's a funny coincidence ^^.

SchrodingersGat:
So, what is the outcome here in terms of workflow? As far as I can determine from the discussions above there are two separate ideas:

### Information Only

Scanning the digikey (or other) barcode returns a JSON object with all the information which can be gleaned from the barcode. Part / supplier part / purchase order / line item / etc etc etc.

No *action* is taken, and so the implementing scanner (could be the web / app / external integration) then has to perform an action based on the returned data. Actions could be:

- Navigate to purchase order
- Receive item into stock
- Lookup digikey page
- ???

### Automated Receive

In addition to the above, the first time the barcode is scanned, it receives the items into stock, and marks them as received against the purchase order. From a workflow perspective this provides the following advantages:

- Single operation for user who receives incoming goods
- Works directly with existing web interface and app (no code changes needed there)
- "Simpler" implementation

Have I got this right, or missing a potential option? How should we move forward here

30350n:
Well, what I currently am thinking of implementing is kind of an inbetween.

### *kinda* Information Only (current implementation)
Scanning a supplier barcode returns a `"supplierpart"` json response, **if** the barcode handler managed to match the barcode to a supplier part. Additionally, if present, the `"order_number"` and `"quantity"` from the barcode get added to said response as extra, optional fields.

### Automated Receive (what I'm currently looking into implementing)

If a supplier barcode gets scanned by a **specific action**/scanner which job it is to receive parts, that action can use the extra response fields from above, to automatically populate the order number and quantity for creating a new stock part. Additionally, if a stock part gets successfully added like this, the supplier barcode gets assigned to the new stock item, thus it's going to be handled by the internal barcode handler in the future.

The thing I find important here is, that this **only happens with the end user intending to receive an item**. If we'd automatically do this when a user scans a barcode with the default barcode scanner, this would only result in confusion imo.

30350n:
I agree. But instead of having a plugin-specific setting, I still think it'd be best to just trigger it intentionally.
If I understand your implementation idea correctly, you'd like to be able to just scan a barcode with the default scanner, resulting in the item being automatically added to stock.
My implementation would do the exact same thing, I'd just not do it with the default scanner, but with a specific action, like this in the App for example:

This would still result in the single scan functionality, just without abusing the default barcode scanner for it.

SchrodingersGat:
I like that idea, a specific endpoint for barcode scanning which is distinct from the generic "scan barcode" endpoint to avoid confusion

matmair:
> Why not? That's similar to assuming the SKU from the barcodes matches a supplierpart, no?

No, the SKU must be entered so it can be assumed. The order reference is generated by InvenTree when the order is created and would need to be transferred to the supplier - that does not always work in my experience. Not all suppliers support replacing their order number with a custom one and I had Digikey ignore it multiple times and just print theirs.

> This would still result in the single scan functionality, just without abusing the default barcode scanner for it.

That is what I had in mind. We could maybe add a checkbox in the WebUI to enable actions. The important thing from my POC is to not accidentally change what API endpoints are meant to do.

30350n:
> No, the SKU must be entered so it can be assumed. The order reference is generated by InvenTree when the order is created and would need to be transferred to the supplier - that does not always work in my experience. Not all suppliers support replacing their order number with a custom one and I had Digikey ignore it multiple times and just print theirs.

There's a field for a "Supplier Reference" though. Which I guess you are supposed to set to the order number, no?
I'm currently implementing the action in the app and my approach is to check both, the reference and supplier references for an exact match to the order number from the barcode.

> The important thing from my POV is to not accidentally change what API endpoints are meant to do.

Exactly! Great, so I guess everyones roughly on the same page now :)

SchrodingersGat:
So the required "back-end" change is to implement a new barcode API endpoint, which is purpose-built for receiving items against purchase orders.

Barcode mixin plugins can then choose to implement a "handle_purchase_order_scan" method (or similar) if they support this method.

From the "purchase order" detail in the user interface (and app) this custom barcode endpoint can be used. Extra context (API fields) can be provided if available, such as "which purchase order" (if coming from the context of a particular purchase order).

30350n:
Hm okay, well that would probably be better yeah. I've almost finished implementing the required logic .... in the app .. in Dart, so uhh guess I'll have to translate that into python again ^^

So there's 2 things I'm still not sure about:
1. Like you described [here](https://github.com/inventree/InvenTree/pull/5509#issuecomment-1709234915) there's a certain order of preference to automatically selecting a stock location (location specified in line item -> default location specified in part -> no location). If the first two aren't set, should we really always set no location? Wouldn't it be better to launch an extra form again to specify the location in those cases? (I think it'd be fine to set no location if there aren't any stock locations defined in general).
2. What happens if the barcode doesn't specify a quantity? One option would be to assume the "full quantity ordered" another one would be to simply reject the whole request, but once again, wouldn't it be nice to let the user specify the quantity explicitly in those cases?

So I guess what I'm proposing more or less is having two possible outcomes from the API endpoint:
1. "receive_item_success" => which indicates the item has been successfully received
2. "receive_item_additional_information_required" => which return the partial data gathered and could be followed up by another normal "receive_item" API request later.

30350n:
I've refactored everything yet again and added the new `api-barcode-po-receive` endpoint. The endpoint can return 3 different results:
1. `"success"` if the supplier part in the barcode got matched to a purchase order line item and quantity and location information could be resolved
2. `"action_required"` if the supplier part in the barcode got matched to a purchase order line item **but either quantity or location information is missing**
3. `"error"` if an error happened during this operation

Additionally, I've also implemented a BarcodeHandler which handles this in [inventree-app #421](https://github.com/inventree/inventree-app/pull/421).

@matmair Would this be compatible with #5500 in its current state?

If so, it be great if someone could rereview this again. From my POV the only thing that's still missing is mentioning this in the barcode docs. Otherwise this would be ready for merging. :)

SchrodingersGat (commented):
Very clean implementation here! I hope to have some time to actually test it soon :)

InvenTree/plugin/base/barcodes/api.py
```diff
@@ -232,7 +234,7 @@ def post(self, request, *args, **kwargs):
                 instance.unassign_barcode()
 
                 return Response({
-                    'success': 'Barcode unassigned from {label} instance',
+                    'success': f'Barcode unassigned from {label} instance',
```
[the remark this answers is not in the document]
30350n: Don't thank me, thank the VSCode code highlighting ^^

matmair:
@30350n #5500 should be easy to adapt

matmair (approved):
LGTM, good work @30350n

InvenTree/plugin/base/barcodes/api.py
```diff
@@ -241,13 +243,94 @@ def post(self, request, *args, **kwargs):
...
+    def post(self, request, *args, **kwargs):
+        """Respond to a barcode POST request."""
+
+        data = request.data
+        if not (barcode_data := data.get("barcode")):
+            raise ValidationError({"barcode": _("Missing barcode data")})
+
+        purchase_order = None
+        if (purchase_order_pk := data.get("purchase_order")):
+            purchase_order = PurchaseOrder.objects.get(pk=purchase_order_pk)
```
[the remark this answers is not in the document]
30350n: Fixed using the second one, I don't like try except blocks ^^

InvenTree/plugin/base/barcodes/api.py
```diff
@@ -241,13 +243,94 @@ def post(self, request, *args, **kwargs):
...
+
+    def post(self, request, *args, **kwargs):
+        """Respond to a barcode POST request."""
+
+        data = request.data
+        if not (barcode_data := data.get("barcode")):
+            raise ValidationError({"barcode": _("Missing barcode data")})
+
+        purchase_order = None
+        if (purchase_order_pk := data.get("purchase_order")):
```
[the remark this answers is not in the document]
30350n: Done! Probably had a not there before I guess.

InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -37,3 +51,203 @@ def scan(self, barcode_data):
         """
 
         return None
+
+    def scan_receive_item(self, barcode_data, user, purchase_order=None, location=None):
+        """Scan a barcode to receive a purchase order item.
+
+        Default return value is None
+        """
```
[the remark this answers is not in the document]
30350n: Added a bunch more information to this. Kinda forgot about it because when I wrote that I wasn't sure on the exact implementation myself ^^

InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -37,3 +51,203 @@ def scan(self, barcode_data):
...
+                return {"error": _(f"Found multiple placed purchase orders for '{order_number}'")}
+
+            purchase_order = purchase_orders[0]
+
+        #  find the first incomplete line_item that matches the supplier_part
+        line_items = purchase_order.lines.filter(
+            part=supplier_part.pk, quantity__gt=F("received"))
+        if not line_items:
+            return {"error": _("Failed to find pending line item for supplier part")}
+
```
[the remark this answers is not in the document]
30350n: Well as the comment above says, my idea was pretty much that it'd just be fine to return the **first** one then. I think duplicated SKUs in the same purchase order would already be an edge case, right? Not sure how that would happen really.
But anyways, so in those cases, once you confirm enough "quantity" of the first line item, this would automatically switch to the next one.

InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -37,3 +51,203 @@ def scan(self, barcode_data):
...
+            # 2. check if it's defined on the part
+            # 3. check if there's 1 or 0 stock locations defined in InvenTree
+            #    -> assume all stock is going into that location (or no location)
+
+            if location := line_item.destination:
+                pass
+            elif location := supplier_part.part.get_default_location():
+                pass
+            elif StockLocation.objects.count() <= 1:
+                if stock_locations := StockLocation.objects.all():
```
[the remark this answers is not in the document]
30350n: Done! Didn't know about `first()` really useful, refactored a bunch of code using that now ^^

30350n:
> @30350n #5500 should be easy to adapt

Great! I'll keep an eye on that one ^^

InvenTree/plugin/base/barcodes/mixins.py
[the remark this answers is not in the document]
matmair: Having the same SKU multiple times is easy if you order for multiple BOs/multiple projects in one PO.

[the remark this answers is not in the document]
30350n: Ah okay interesting, well one possible improvement to this I could see would be matching the quantity, if supplied aswell.
I.e. if we have multiple line items, check each of them and see if one of them matches the quantity provided in the barcode. If not, just match the first one (as it's now).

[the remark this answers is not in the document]
SchrodingersGat: @30350n that is almost sufficient, I think. You will also need to ensure that the received quantity makes sense for the "first" item, if there is no exact match. What if you receive a partial quantity shipment?

SchrodingersGat:
@30350n as this PR provides an update to the API, please bump the api version number in `InvenTree/api_version.py` (with a reference back to this PR)

InvenTree/plugin/base/barcodes/mixins.py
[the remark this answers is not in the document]
30350n: That shouldn't be a problem. In those cases a stock item for the partial quantity gets created, but the line item is still "unfinished". So once you scan the item containing "the rest", it returns the same line item again.

> You will also need to ensure that the received quantity makes sense for the "first" item, if there is no exact match.

That would mean the line items quantity should ideally be greater than the barcode quantity in those cases then, right?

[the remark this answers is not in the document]
SchrodingersGat: Yes, that's right.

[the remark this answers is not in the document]
30350n: Should be good now (I added a bunch of for else logic :P)

30350n:
> @30350n as this PR provides an update to the API, please bump the api version number in `InvenTree/api_version.py` (with a reference back to this PR)

Also done!

SchrodingersGat:
Thanks @30350n :)

Looks like there are still some issues with your tests though

30350n:
Ah yeah, I missed one. Not sure what that last one is about that's still failing ...
It passes for me locally.

SchrodingersGat:
Odd. Have you made any changes at all to the test that is still failing?

30350n:
> Odd. Have you made any changes at all to the test that is still failing?

Nope, absolutely no clue why that one's failing, not my test btw if that wasn't obvious. From what I can tell, it's meant to test disabling a plugin. So it just takes the first one from the plugin configurations and disables it.
So I thought: I've added 3 new plugins, maybe it's taking one of those and then somehow failing ... but now, as you can see [here](https://github.com/inventree/InvenTree/pull/5509#issuecomment-1724758596) it's picking the default barcode plugin. Which I didn't really touch afaik.

The error itself is also really weird btw: the `original_display_for_field` function from the admin module of `django_money` *somehow* hits maximum recursion depth.

30350n:
Also it's passing for me locally so, even less of a clue what's going on.

wolflu05:
One tip: change the workflow to not depend on each other, then add the tmate action and run the specific tests via ssh by modifying the code. This is much quicker than waiting 30mins each time. I did the same thing in [this pr](https://github.com/inventree/InvenTree/commit/a23c89fbca5db19b188c359fc8063d023bd98d3f).

30350n:
> One tip: change the workflow to not depend on each other, then add the tmate action and run the specific tests via ssh by modifying the code. This is much quicker than waiting 30mins each time. I did the same thing in [this pr](https://github.com/inventree/InvenTree/commit/a23c89fbca5db19b188c359fc8063d023bd98d3f) (the referenced commit removes those debugging things)

Thanks, will do next time I have to do something like this, I almost got it tracked down now. Tbh ideally I could just run this locally, but for whatever reason I don't get an error there. Also the logs are really not helpful (or I'm blind).

InvenTree/plugin/base/barcodes/mixins.py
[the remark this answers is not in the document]
matmair: @30350n do you know about docstrings? We are using [google docstyle](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings), that way you get nice formatting in vscode or similar IDEs.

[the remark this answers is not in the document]
30350n: Uhhhh not really sure what you mean with that? Is there some kind of vscode extension for that? (I mainly just tried to copy the docstring style from existing functions.)

30350n:
Hmmm. So after looking further into this, it kind of seems to me like the database state is shared between test cases. Is this expected behavior? Doesn't seem like a good idea to me and probably what's causing the issue here...

30350n:
It's [this](https://github.com/30350n/InvenTree/blob/91bad761bb601448bfd907843ccd35fcc1027436/InvenTree/plugin/builtin/suppliers/test_supplier_barcodes.py#L193) test case that's introducing the error btw (although from what I can tell, it is not directly causing it ....).

The only ""special"" thing I do there is create 2 new stock locations. Could this be the issue and if so: why!?

matmair:
@30350n it could be if database errors are thrown. Please merge in master, the error logs should be a lot clearer now.

30350n:
Thanks for looking into this! Let's see...

30350n:
Hah, everything passing now. Knew that one wasn't on me. Thanks for the fix!

30350n:
Just added a small note regarding this functionality to the docs, if the tests pass now, I think this is ready to be reviewed/merged. :)

InvenTree/plugin/builtin/suppliers/tme.py
```diff
@@ -0,0 +1,60 @@
...
+    NAME = "TMEPlugin"
+    TITLE = _("Supplier Integration - TME")
+    DESCRIPTION = _("Provides support for scanning TME barcodes")
+    VERSION = "1.0.0"
+    AUTHOR = _("InvenTree contributors")
+
+    def parse_supplier_barcode_data(self, barcode_data):
+        """Get supplier_part and barcode_fields from TME QR-Code or DataMatrix-Code."""
+
+        if not isinstance(barcode_data, str):
```
SchrodingersGat, line 29: There is a fair amount of repeated code in this `parse_supplier_barcode_data` method which could be pushed to the `SupplierBarcodeMixin` class

InvenTree/plugin/base/barcodes/mixins.py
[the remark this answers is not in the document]
SchrodingersGat: Have a look at our general commenting elsewhere. Methods (especially complex ones) should enumerate arguments, provide a return type and any potential errors raised:

image

InvenTree/plugin/builtin/suppliers/digikey.py
```diff
@@ -0,0 +1,38 @@
...
+        """Get supplier_part and barcode_fields from DigiKey DataMatrix-Code."""
+
+        if not isinstance(barcode_data, str):
+            return None
+
+        if not (barcode_fields := self.parse_ecia_barcode2d(barcode_data)):
+            return None
+
+        sku = barcode_fields.get("supplier_part_number")
+        if not (supplier_part := self.get_supplier_part(sku)):
```
SchrodingersGat, line 35: Here, as we *know* this is digikey, we should be able to pass in the supplier pk to `get_supplier_part`

InvenTree/plugin/builtin/suppliers/lcsc.py
```diff
@@ -0,0 +1,54 @@
...
+        if not (match := LCSC_BARCODE_REGEX.fullmatch(barcode_data)):
+            return None
+
+        barcode_pairs = (pair.split(":") for pair in match.group(1).split(","))
+        barcode_fields = {
+            BARCODE_FIELD_NAME_MAP.get(field_name, field_name): value
+            for field_name, value in barcode_pairs
+        }
+
+        sku = barcode_fields.get("supplier_part_number")
```
SchrodingersGat, line 41: Why not extract and use MPN here (if it is available in the barcode)?

InvenTree/plugin/builtin/suppliers/mouser.py
```diff
@@ -0,0 +1,73 @@
...
+            return None
+
+        if not (mouser := self.get_mouser_supplier()):
+            return None
+
+        if not (barcode_fields := self.parse_ecia_barcode2d(barcode_data)):
+            return None
+
+        sku = barcode_fields.get("supplier_part_number")
+        mpn = barcode_fields.get("manufacturer_part_number")
```
SchrodingersGat, line 69: Here you extract MPN, and make use of it - but not for the other suppliers?

InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -37,3 +51,277 @@ def scan(self, barcode_data):
...
+            response["lineitem"]["location"] = location.pk
+
+        # if either the quantity is missing or no location is defined/found
+        # -> return the line_item found, so the client can gather the missing
+        #    information and complete the action with an 'api-po-receive' call
+        if not quantity or (not location and not no_stock_locations):
+            response["action_required"] = _("Further information required to receive line item")
+            return response
+
+        purchase_order.receive_line_item(
```
SchrodingersGat, line 244: I would wrap this in a `try/except` block - it may fail due to some reason, and you want to be able to throw the right error message back to the user

SchrodingersGat (changes requested):
@30350n I have spent some time this evening playing around with this - please see the review comments. Ping me if you have questions on any of them :)

SchrodingersGat:
@30350n this is very promising, I did manage to get it working with some digikey barcodes I had laying around. Please address the outstanding issues and I'm keen to review again. This is a very exciting feature!

SchrodingersGat:
@30350n let me know when you're ready for me to look at this again!

30350n:
> @30350n let me know when you're ready for me to look at this again!

Just finished going through everything, let me know what you think!

matmair:
@30350n it seems that some APIs are now returning a 500 error code - see [here](https://github.com/inventree/InvenTree/actions/runs/6486132371/job/17613716433?pr=5509).
This is only on this branch and not master so probably related to the changes proposed.

30350n:
> @30350n it seems that some APIs are now returning a 500 error code - see [here](https://github.com/inventree/InvenTree/actions/runs/6486132371/job/17613716433?pr=5509). This is only on this branch and not master so probably related to the changes proposed.

Yeah I'm pretty sure I know what the issue is actually, I think I encountered this earlier but kinda forgot about it 😅 .
Will fix when I'm @home again.

matmair:
Looking good

InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -37,3 +54,334 @@ def scan(self, barcode_data):
...
+            return
+
+        actual_data = barcode_data.split(HEADER, 1)[1].rsplit(TRAILER, 1)[0]
+
+        return actual_data.split("\x1D")
+
+    @staticmethod
+    def get_supplier_parts(sku: str, supplier: Company = None, mpn: str = None):
+        """Get a supplier part from SKU or by supplier and MPN."""
+        supplier_parts = SupplierPart.objects
```
SchrodingersGat, line 118: I would suggest you start with `supplier_parts = SupplierPart.objects.all()` (and then see comment below)

InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -37,3 +54,334 @@ def scan(self, barcode_data):
...
+            supplier_parts = supplier_parts.filter(supplier=supplier.pk)
+            if len(supplier_parts) == 1:
+                return supplier_parts
+
+        if mpn:
+            supplier_parts = SupplierPart.objects.filter(manufacturer_part__MPN__iexact=mpn)
+            if len(supplier_parts) == 1:
+                return supplier_parts
+
+        if supplier_parts is SupplierPart.objects:
```
SchrodingersGat, line 135: If you use `SupplierPart.objects.all()` above, you don't need this check

InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -37,3 +54,334 @@ def scan(self, barcode_data):
...
+            supplier_parts = []
+
+        logger.warning(
+            "Found %d supplier parts for SKU '%s', supplier '%s', MPN '%s'",
+            len(supplier_parts),
```
SchrodingersGat, line 140: Instead you can use `supplier_parts.count()`

InvenTree/plugin/base/barcodes/mixins.py
SchrodingersGat: Please wrap `receive_line_item` method here in a try/except

InvenTree/plugin/builtin/suppliers/mouser.py
```diff
@@ -0,0 +1,73 @@
...
+class MouserPlugin(SupplierBarcodeMixin, SettingsMixin, InvenTreePlugin):
+    """Plugin to integrate the Mouser API into Inventree."""
+
+    NAME = "MouserPlugin"
+    TITLE = _("Supplier Integration - Mouser")
+    DESCRIPTION = _("Provides support for scanning Mouser barcodes")
+    VERSION = "1.0.0"
+    AUTHOR = _("InvenTree contributors")
+
+    SETTINGS = {
```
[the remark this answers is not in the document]
SchrodingersGat: @30350n the other supplier plugins (digikey / lcsc) still need the `SUPPLIER_ID` setting attribute

SchrodingersGat (changes requested):

SchrodingersGat:
@30350n nice progress - a few more things I'd like to see addressed :) I have not had time to test your updated code with a real barcode and scanner, but if you can ping me when you've made the changes as above I'll check it out :)

30350n:
Thanks for reviewing again! :)

InvenTree/plugin/builtin/suppliers/mouser.py
```diff
@@ -0,0 +1,49 @@
...
+from django.utils.translation import gettext_lazy as _
+
+from plugin import InvenTreePlugin
+from plugin.base.barcodes.mixins import SupplierBarcodeData
+from plugin.mixins import SettingsMixin, SupplierBarcodeMixin
+
+logger = logging.getLogger('inventree')
+
+
+class MouserPlugin(SupplierBarcodeMixin, SettingsMixin, InvenTreePlugin):
```
SchrodingersGat, line 17: The `MouserPlugin` is the only one which inherits from the `SettingsMixin` - so the settings don't yet work for the other supplier plugins. Please add `SettingsMixin` to the other ones (e..g digikey / tme / etc)

SchrodingersGat (changes requested):
Looking good, still a couple of issues. In particular, the classes must all inherit from SettingsMixin

InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -37,3 +54,334 @@ def scan(self, barcode_data):
...
+                )
+                return None
+
+        if not (supplier_name := getattr(self, "DEFAULT_SUPPLIER_NAME", None)):
+            return None
+
+        suppliers = Company.objects.filter(name__icontains=supplier_name, is_supplier=True)
+        if len(suppliers) != 1:
+            return None
+        self.set_setting("SUPPLIER_ID", suppliers.first().pk)
```
SchrodingersGat, line 365: @30350n nice idea, to update the supplier if not set

InvenTree/plugin/base/barcodes/mixins.py
SchrodingersGat: @30350n can you address this? We should handle any errors which occur during `receive_line_item`

SchrodingersGat (changes requested):

SchrodingersGat:
@30350n getting very close! One more change, see notes

30350n:
Yeah, but the `receive_line_item` should not throw an error unless there's an actual bug in the code, right?
In which case you **want** to get an Exception.

As I wrote before:

The way I understand it, if some InvenTree code throws an Exception you get one of those error notifications and an entry in the admin panel. Imo, this is exactly what should happen here. If receive_line_item throws an error, there is a bug in the code which has to be fixed. Thus, if we were to wrap it and throw a "soft" error, it would seem as if it's an expected error (you'd also get no tracebacks and it'd just generally make everything much harder to debug).

Sorry, but this still seems like an anti feature to me.

SchrodingersGat:
@30350n sorry, I must have missed your previous comment, and you make a good argument. On balance I think your approach is correct and thus I think we are good to merge!

SchrodingersGat (approved):

SchrodingersGat:
@30350n thanks for the *huge* effort here, this is a great new feature! Please keep a lookout for any bugs, and if you can spare some time soon for the associated documentation, that would be great :)

SchrodingersGat:
@30350n it has been a pleasure working with you too - please keep contributing!!

PR #5515, opened 2023-09-07
Issued by filter

SchrodingersGat:
Allow build order table data to be filtered by which user issued the order

### DIFF
--- a/InvenTree/InvenTree/api_version.py
+++ b/InvenTree/InvenTree/api_version.py
@@ -2,7 +2,7 @@
 
 
 # InvenTree API version
-INVENTREE_API_VERSION = 133
+INVENTREE_API_VERSION = 132
 
 """
 Increment this API version number whenever there is a significant change to the API that any clients need to know about
--- a/InvenTree/InvenTree/api_version.py
+++ b/InvenTree/InvenTree/api_version.py
@@ -2,10 +2,13 @@
 
 
 # InvenTree API version
-INVENTREE_API_VERSION = 131
+INVENTREE_API_VERSION = 133
 
 """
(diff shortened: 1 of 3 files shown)


InvenTree/InvenTree/api_version.py
```diff
@@ -2,11 +2,14 @@
 
 
 # InvenTree API version
-INVENTREE_API_VERSION = 131
+INVENTREE_API_VERSION = 133
```
wolflu05, line 5: Is this a typo?

InvenTree/templates/js/translated/table_filters.js
```diff
@@ -18,6 +18,30 @@
...
+            let users = {};
+
+            inventreeGet('{% url "api-user-list" %}', {}, {
+                async: false,
+                success: function(response) {
+                    for (let user of response) {
+                        users[user.pk] = {
+                            key: user.pk,
+                            value: user.username
+                        };
```
wolflu05, line 35: Just if you want to learn something more modern, you could do it with reduce like:
```js
response.reduce((users, user) => ({...users, [user.pk]: { key: user.pk, value: user.username }}), {})
```
Or using the map function and Object.fromEntries in combination.

InvenTree/InvenTree/api_version.py
SchrodingersGat: Nice catch, thanks

InvenTree/templates/js/translated/table_filters.js
SchrodingersGat: Thanks @wolflu05 for the tip. This will all soon be replaced with React - and I am certainly focusing on writing more up-to-date code over there :)

PR #5521, opened 2023-09-10
[PUI] Part detail page

SchrodingersGat:
Adding in a barebones framework for a "part detail" page. Will need the actual content added at a later date. The components developed here can be used across the app in multiple views.


SchrodingersGat:
Screenshot 2023-09-10 at 8 53 42 pm

PR #5522, opened 2023-09-10
[0.12.x] Fix missing filters for get settings validator (#5480)

wolflu05:
# Backport

This will backport the following commits from `master` to `0.12.x`:
 - [Fix missing filters for get settings validator (#5480)](https://github.com/inventree/InvenTree/pull/5480)

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

### DIFF
--- a/InvenTree/common/models.py
+++ b/InvenTree/common/models.py
@@ -166,7 +166,7 @@ class BaseInvenTreeSetting(models.Model):
 
         do_cache = kwargs.pop('cache', True)
 
-        self.clean(**kwargs)
+        self.clean()
         self.validate_unique()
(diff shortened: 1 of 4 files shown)


wolflu05:
Don't wonder, I tried out this back port tool locally by adding a PAT because I thought it will be easier, but no, actually it was some kind of weird first. Not sure if it wouldn't be easier to do it with git directly.

wolflu05:
Should work now, so it can be merged. I think the rest of the failures are not related to my changes, because they are also present in the previous runs on the 0.12.x branch.

PR #5525, opened 2023-09-11
[PUI] Attachment table

SchrodingersGat:
Implements generic "attachments" table for react interface

### DIFF
--- a/InvenTree/InvenTree/api.py
+++ b/InvenTree/InvenTree/api.py
@@ -228,6 +228,12 @@ class AttachmentMixin:
 
     filter_backends = SEARCH_ORDER_FILTER
 
+    search_fields = [
+        'attachment',
+        'comment',
+        'link',
+    ]
+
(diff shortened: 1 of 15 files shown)


SchrodingersGat:
Screenshot 2023-09-11 at 9 29 12 pm

Example screenshot of the "part attachment" table

wolflu05:
This looks really cool. On a general note, is there a previewing feature planned, which maybe be also integrated into the plugin system. This could then be used for e.g. directly previewing pdf, pictures. And with plugin support also for wirewiz files. I imagine this like an api endpoint which returns html content that previews that file. This could then be easily integrated into every platform, also the app via a web view. And in addition, one could also implement custom editors, so that one could edit attachments in place, useful for tables, wirewiz files with live preview.

SchrodingersGat:
@wolflu05 that's not in the scope for this PR but should be workable with the new interface. I suggest you make a new issue / FR for that :)

PR #5527, opened 2023-09-11
Fix for `%` (percent) unit (missing unary operator "%")

30350n:
If using the newest version from [`master`](https://github.com/inventree/InvenTree/tree/master), adding a `PartParameter` with a `PartParameterTemplate` that uses `%` as its unit results in a `pint.errors.DefinitionSyntaxError: missing unary operator "%"` error.

This seems to be a pint issue (see [pint #1758](https://github.com/hgrecco/pint/issues/1758)) which is sadly still present on pint 0.22 and 0.23. The user who created the mentioned issue provided a workaround though, which this patch applies. Imo this is the way to go for now, because otherwise the `%` unit is just unusable.

### DIFF
--- a/InvenTree/InvenTree/conversion.py
+++ b/InvenTree/InvenTree/conversion.py
@@ -106,7 +106,7 @@ def convert_physical_value(value: str, unit: str = None, strip_units=True):
                 val = ureg.Quantity(value, unit)
             else:
                 # Convert to the provided unit (may raise an exception)
-                val = val.to(unit)
+                val = val.to(ureg.Unit(unit))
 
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@30350n thanks for the fix, and nice addition of unit tests :)

30350n:
No problem! :) Just fixed the fomatting, Style checks should pass now hopefully.

SchrodingersGat:

lafrech:
I believe the workaround I proposed in https://github.com/hgrecco/pint/issues/1758 is not needed anymore since Pint 0.23.

Issue #5539, opened 2023-09-14
[Docker] Automated builds broken


SchrodingersGat:
Automated builds have been broken for around a week now:

https://github.com/inventree/InvenTree/actions/workflows/docker.yaml

Specific error details:

```
======================================================================
ERROR: test_plugin_install (plugin.test_api.PluginDetailAPITest)
Test the plugin install command.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/***/InvenTree/plugin/test_api.py", line 55, in test_plugin_install
    self.assertEqual(data['success'], True)
KeyError: 'success'
```

SchrodingersGat:
@matmair any changes to the plugin installation API recently?

matmair:
@SchrodingersGat nothing I submitted in the last month at least

SchrodingersGat:
Interesting, the results over on https://github.com/inventree/InvenTree/pull/5544 show that it's related to pip installation of the plugin:

Trying to install 'minimal' via the API:

```
2023-09-15T00:42:00.0547323Z Finished reloading plugins
[...]

PR #5544, opened 2023-09-15
Docker CI test

SchrodingersGat:
Trying to work out why docker builds are failing.

Ref: https://github.com/inventree/InvenTree/issues/5539

### DIFF
--- a/InvenTree/plugin/test_api.py
+++ b/InvenTree/plugin/test_api.py
@@ -41,6 +41,8 @@ class PluginDetailAPITest(PluginMixin, InvenTreeAPITestCase):
             expected_code=201,
         ).data
 
+        print("A:", data)
+
         self.assertEqual(data['success'], True)
 
         # valid - github url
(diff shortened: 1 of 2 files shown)


wolflu05:
If you don't want to wait so long each time, and you cannot reproduce it locally, I would suggest using the tmate action and comment out the regular test, and the needs other jobs. Then you can docker exec into the container and run only the failing test, change things and rerun it again, that's much quicker than running all tests, waiting, changing, running again.

InvenTree/plugin/test_api.py
```diff
@@ -41,6 +41,8 @@ def test_plugin_install(self):
             expected_code=201,
         ).data
 
+        print("A:", data)
```
wolflu05, line 44: @SchrodingersGat I think you forgot the debug comment here?

SchrodingersGat: Ah, nice spot, thanks

PR #5568, opened 2023-09-19
Remove --no-frontend option

SchrodingersGat:
- Until the new frontend is fully ready to be shipped, adding this option only serves to confuse users (and break the install process potentially)
- Remove this option entirely from the invoke script
- Update documentation accordingly

@matmair we have a lot of changes ready to push for [0.13.0](https://github.com/inventree/InvenTree/milestone/44) but the frontend is still far from ready to ship.

I would like to tweak the current invoke script (and docs) so that the process is unchanged from 0.13.0 (i.e. users do not have to include the "--no-frontend" option).

This will aid us to shipping 0.13.0 asap. The frontend options are still there for anyone who wants to test drive, but won't get in the way of everyone else doing an upgrade.

### DIFF
--- a/docs/docs/start/docker_dev.md
+++ b/docs/docs/start/docker_dev.md
@@ -68,7 +68,7 @@ If desired, the user may edit the environment variables, located in the `.env` f
 Perform the initial database setup by running the following command:
 
 ```bash
-docker compose run inventree-dev-server invoke update --no-frontend
+docker compose run inventree-dev-server invoke update
 ```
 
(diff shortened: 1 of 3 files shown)


matmair:
I am not sure releasing a half-implemented new UI hidden in the background is a good idea, will probably confuse a lot of users.

SchrodingersGat:
> I am not sure releasing a half-implemented new UI hidden in the background is a good idea, will probably confuse a lot of users.

It won't be "hidden" - they can still build it. And I will ensure that we document it clearly in the release blog

PR #5569, opened 2023-09-19
Path detail API

SchrodingersGat:
Adds "path" data to API endpoints for:

- StockLocation
- StockItem
- PartCategory
- Part

### Justification

This data needs to be exposed to the API for the upcoming react framework as it is not currently accessible via the API.

### Notes

- This additional data is optional, if the `?path_detail=1` query parameter is provided
- Only for the "detail" endpoints (not the list endpoints, too expensive)

### DIFF
--- a/InvenTree/InvenTree/email.py
+++ b/InvenTree/InvenTree/email.py
@@ -45,7 +45,7 @@ def is_email_configured():
         configured = False
 
         if not testing:  # pragma: no cover
-            logger.warning("DEFAULT_FROM_EMAIL is not configured")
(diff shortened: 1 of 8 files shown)


InvenTree/InvenTree/email.py
```diff
@@ -45,7 +45,7 @@ def is_email_configured():
         configured = False
 
         if not testing:  # pragma: no cover
-            logger.warning("DEFAULT_FROM_EMAIL is not configured")
+            logger.debug("DEFAULT_FROM_EMAIL is not configured")
```
wolflu05, line 48: Thank you :D - this bothered me since a long time.

Issue #5523, opened 2023-09-10
Plugin - admin.site.register() throws an error now


afkiwers:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

In 0.12.0 dev everything was working, However suddenly I can't access the admin backend anymore without getting an error. I have to remove admin.site.register(SelectedCategory, SelectedCategoryAdmin) from my admin.py to not run into an error.

![image](https://github.com/inventree/InvenTree/assets/42826581/9ac1a888-2bf2-4071-b49f-f9fa6ea1d7c7)

### It was working here
```
# Version Information:
InvenTree-Version: 0.12.0 dev
Django Version: 3.2.19
Commit Hash: 6e64c27
Commit Date: 2023-06-28
Commit Branch: run-locally
Database: sqlite3
Debug-Mode: True
Deployed using Docker: False
Platform: Windows-10-10.0.22621-SP0
Installer: GIT

Active plugins: [{'name': 'KiCadLibraryPlugin', 'slug': 'kicad-library-plugin', 'version': '1.0.0'}]

```

### Error message
```
Environment:


Request Method: GET
Request URL: http://127.0.0.1:8000/admin/

Django Version: 3.2.20
Python Version: 3.11.4
Installed Applications:
['django.contrib.admin',
 'build.apps.BuildConfig',
 'common.apps.CommonConfig',
 'company.apps.CompanyConfig',
 'plugin.apps.PluginAppConfig',
 'label.apps.LabelConfig',
 'order.apps.OrderConfig',
 'part.apps.PartConfig',
 'report.apps.ReportConfig',
 'stock.apps.StockConfig',
 'users.apps.UsersConfig',
 'web',
 'generic',
 'InvenTree.apps.InvenTreeConfig',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'user_sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.sites',
 'maintenance_mode',
 'django_filters',
 'rest_framework',
 'rest_framework.authtoken',
 'corsheaders',
 'crispy_forms',
 'import_export',
 'django_cleanup.apps.CleanupConfig',
 'mptt',
 'markdownify',
 'djmoney',
 'djmoney.contrib.exchange',
 'error_report',
 'django_q',
 'formtools',
 'dbbackup',
 'taggit',
 'flags',
 'allauth',
 'allauth.account',
[...]

PR #5572, opened 2023-09-20
Plugin loading fixes

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/5523

This patch fixes two bugs which seem specific to AppMixin plugins:

### External Plugins Dir

Loading AppMixin plugins from external (custom) plugins dir did not work correctly - there was no mechanism available to add these plugins to `INSTALLED_APPS`. Some extra path checks fix this

### Admin URLs

Admin integration breaks (as reported in #5523) due to changes in 3e37469350535ddae7db76571ba4913104b9534e

Bring the admin URLs up one level to fix this.

### DIFF
--- a/InvenTree/InvenTree/config.py
+++ b/InvenTree/InvenTree/config.py
@@ -281,6 +281,12 @@ def get_plugin_file():
     return plugin_file
 
 
+def get_plugin_dir():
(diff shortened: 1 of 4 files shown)


matmair (approved):
Lgtm

PR #5575, opened 2023-09-20
[P UI] Router syntax switch

matmair:
This PR suggests a change of the used react-router syntax to make it more concise and easier to read - especially for nested routes.


SchrodingersGat:
Yeah, this is heaps better

PR #5577, opened 2023-09-20
[P UI] Make base url configurable

matmair:
This PR makes the base URL (by default platform) configurable. This way the admin can select under which subpath P UI should be server (or if it should not use a subpath at all).

### DIFF
--- a/InvenTree/InvenTree/middleware.py
+++ b/InvenTree/InvenTree/middleware.py
@@ -64,7 +64,7 @@ class AuthRequiredMiddleware(object):
             elif request.path_info.startswith('/accounts/'):
                 authorized = True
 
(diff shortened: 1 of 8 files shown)


lippoliv:
Does it also support Ports ("http://some-nas:1337")? Because 0.12.7 doesn't :/

matmair:
Ports already work, our dev environments work that way.

PRs are not the right place to report problems if you want help though. That would be either discussions or issues.

lippoliv:
Agreed. I'll collect information and report a bug. I just thought it's a full makeover so a bug would be obsolete. Thanks 🙏

SchrodingersGat:
Seems like a straightforward change, thanks @matmair

matmair:
@SchrodingersGat I am just splitting out all the side noice from #5559 so you can review that ;-)

PR #5588, opened 2023-09-21
Feature/location types

### LINKED ISSUE #5561: [FR] Storage location types

### Problem statement

I have different types of boxes and 1. want to keep track of what containers I have and 2. display the same icon for the same box even if I update it at one place without messing with the db directly.

### Suggested solution

Add a storage location type table (like Parameter templates) where bins, bags, boxes, drawers, shelves, cases, toolboxes can be created with their properties like name, description, icon (and in the future maybe also an image or size can be added?). Then each location is optionally related to a storage location type. These types then can be edited in settings. And if the type contains an icon, the storage location uses that.

Actually in the end I also like to deprecate the location icon field and make it readonly, but I'm not sure how migration could look like. Maybe a script which groups by icon and creates location types according to that which get assigned to the locations. 

Use sql queries to update the icons on all parts, but this is not a really easy solution for non programmers.

### Examples of other systems

/

- [X] I want to develop this.

wolflu05:
This PR implements the basic functionality of StockLocation types like proposed in #5561 . 

I'm not sure how far PUI is, but as far as I see, there is currently nothing where I could implement this, so it can be done when implementing the stock location forms and settings.

To keep backwards compatibility I renamed the model field (but kept the db field named `icon`) to `custom_icon` and added a setter / getter approach to the icon field. But there is one drawback with this setter, if the `icon` field gets loaded as default value by any form which is later saved, the location no longer inherits its icon from the location type, because the icon is now saved to the location. Let me know if I should remove the setter or if you have a better idea.

## Todo

- [x] tests
- [x] bump api version
 

fixes #5561


wolflu05:
This would be ready from my POV. The CI issues are as far as I see related to the exchange host api which currently is not working, see https://github.com/Formicka/exchangerate.host/issues/236

mlucchelli:
`RecursionError: maximum recursion depth exceeded`

wolflu05:
> `RecursionError: maximum recursion depth exceeded`

If you look at other test runs from [other branches](https://github.com/inventree/InvenTree/actions/runs/6269211315/job/17025327893#step:5:1), you can see that the recursion error is present there too in combination with the two exchange rate tests that are failing.

SchrodingersGat:
@wolflu05 this looks great, thanks for taking the time to implement this, I think it is a very clean implementation :)

PR #5595, opened 2023-09-22
Feature/Tree picker

wolflu05:
This bothered me since a long time. It is very hard if one has locations with the same repeating name to search for them by a tree path. To improve the experience I added a tree picker to the forms which shows the tree in a new modal to select the correct path. Additionally, this improves the backdrop for secondary, ... modals, so that only the top modal is infront of the backdrop.

https://github.com/inventree/InvenTree/assets/76838159/e89aea4e-3ba7-4a96-848b-ffc1313fc08e

This could have some performance issues like the top navigation has them too, but this is related to the jQuery tree view library. I hope we can implement something like this in PUI sometime too. But for now this PR is ready to review.

### DIFF
--- a/InvenTree/part/serializers.py
+++ b/InvenTree/part/serializers.py
@@ -113,6 +113,7 @@ class CategoryTree(InvenTree.serializers.InvenTreeModelSerializer):
             'name',
             'parent',
             'icon',
+            'structural',
         ]
 
 
--- a/InvenTree/InvenTree/api_version.py
+++ b/InvenTree/InvenTree/api_version.py
@@ -7,7 +7,7 @@ INVENTREE_API_VERSION = 136
 """
 Increment this API version number whenever there is a significant change to the API that any clients need to know about
 
-v136 -> 2023-0-23 : https://github.com/inventree/InvenTree/pull/5595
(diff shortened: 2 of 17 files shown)


InvenTree/part/serializers.py
```diff
@@ -113,6 +113,7 @@ class Meta:
             'name',
             'parent',
             'icon',
+            'structural',
```
matmair, line 116: API change: we will need to bump the api version.

matmair (approved):
Lgtm; the inclusion of the structural field warrants an API version bump

wolflu05:
You're right, missed that, added it now.

SchrodingersGat:
@wolflu05 sorry for the delayed response on this one - it looks fantastic! No notes :)

wolflu05:
Thanks that you like this. To build this in PUI, I think it would be better to build the tree component first. From my research, we could use [this](https://github.com/diogofcunha/react-virtualized-tree) library which is unsing virtualization what drastically improves performance on large data sets. They show an [example](https://diogofcunha.github.io/react-virtualized-tree/#/examples/large-collection) with 1848897 nodes and the performance is awesome.

PR #5604, opened 2023-09-24
Fix controlling log level in devcontainer

lippoliv:
I followed https://docs.inventree.org/en/latest/develop/devcontainer/ (which is very nice by the way) to have a proper dev setup on my mac.

I was wondering why I do not see info logs, as I do on my production setup. Found this small thing, hope it helps.


SchrodingersGat:
Nice find, pretty obviously a bug :)

SchrodingersGat:
Thanks for the contribution @lippoliv

PR #5605, opened 2023-09-24
Add worker task to VS Code

lippoliv:
After following https://docs.inventree.org/en/stable/develop/devcontainer/ I was wondering why the queue isn't processed. Maybe I used wrong search keywords, but I haven't found another solution than to invoke a worker manually.

So I extended `tasks.json` to have vscode support.

I created this PR as a draft, because it's probably the wrong way to deal with workers. Maybe the devcontainer should have a worker already running? So I'm open for feedback.


SchrodingersGat:
I think it makes sense to implement it like this - although we should also include in this PR an update to the documentation (in the devcontainer section) describing the new task for the background worker.

lippoliv:
Ok nice. But documentation is another repository, right? So should I create a PR there as well?

wolflu05:
No, documentation was moved into the main repo a while a go. You find it in the `docs/` folder.

lippoliv:
OK then this should be part of this PR as well, I'll try to push additional changes

matmair (approved):

lippoliv:
@matmair @SchrodingersGat I merged in master and added documentation, happy to see some feedback :)

docs/docs/develop/devcontainer.md
```diff
@@ -45,6 +45,10 @@ If you need some demo test-data, run the `setup-test` task. This will import an
 
 If you only need a superuser, run the `superuser` task. It should prompt you for credentials.
 
+#### Run background workers
+
+If you want need to process your queue with background workers, run the `worker` task.
```
wolflu05, line 50: Im not a native speaker, but in my opinion this sentence would sound better without the `want`. @SchrodingersGat what do you think?

lippoliv: Good catch. I missed that after rephrasing it :) Fix incoming.

(not an native speaker either, but I didn't want to write it this way)

SchrodingersGat:
@lippoliv looks good to me, thanks for the contribution :)

Issue #5594, opened 2023-09-22
Endless background jobs


lippoliv:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

I don't know how to describe it, I'll try my best, please ask if you'r missing something.

### Background
I'm onboarding a small business to InvenTree. It has a 10+ years history in eCommerce (but very small) and moved over to handmade products, so InvenTree will come into play to deal with stock. I decided to import "the past" (re-create stock movements so to say) for some resons. It's partially automated, so it's a fun deep dive into InvenTree.

### Setup
I host it on DS918+ (INTEL Celeron J3455, 1.5 GHz) with 8 GiB RAM. It runs on Docker via Protainer.

### Dimension
Currently I have 10 Parts, 10 Purchase Orders, 137 Sale Orders.  
I would say that's not much.

### Problem
I can create new customers, I can create Sale Orders. I can issue Sale Orders. All this runns quiet smooth. BUT: When I ship the order items, things start to go south.

Per shipment, roughly 200 or 300 background jobs are spawned (while being processed), which take probably 5 minutes to go back down to 0. For sure it decreases every now and then, but also it increases again.

If I ship 3 or 4 orders, the queue is averaging about 700 - 900 queued jobs, that lasts for 20 minutes or so.

### Steps to Reproduce

1. Ship a ton of orders

### Expected behaviour

Maybe it's to blue-eyed, but I would say shipping some orders will just do some database updates. That should be processed within no time.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.12.7
Django Version: 3.2.19
Commit Hash: 6563b4c
Commit Date: 2023-09-16

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-4.4.180+-x86_64-with-glibc2.36
Installer: DOC

[...]

PR #5606, opened 2023-09-24
Stop cascade update_pricing if prices didn't change

lippoliv:
I analysed #5594 further and came across a thought which I think is worth sharing: Why cascade price changes if nothing changed? So I implemented a POC and improved number of events significantly (counter 25 down to 2).

# Steps to reproduce

1. Create P1 as variant
2. Create P2 as component with P1 in BOM
3. Wait for workers to finisih
4. In P2 BOM change quantity of P1 needed for Build
5. Debug (or review logs) how often `update_pricing` is executed

# Idea this change

If `update_pricing` whould cascade, but prices haven't changed, what's the matter? So the idea is to even if `cascade` is wanted, don't do it if no prices changed.

# POC implementation

I don't fully like the code change tbh. It looks a little heavy to maintain. But I have no idea what other implementation is better, maybe some more skilled dev can help me out? I'm open for feedback.

### DIFF
--- a/InvenTree/part/models.py
+++ b/InvenTree/part/models.py
@@ -2350,6 +2350,8 @@ class PartPricing(common.models.MetaMixin):
     - Detailed pricing information is very context specific in any case
     """
 
+    price_modified = False
+
     @property
     def is_valid(self):
         """Return True if the cached pricing is valid"""
@@ -2476,7 +2478,7 @@ class PartPricing(common.models.MetaMixin):
             pass
 
         # Update parent assemblies and templates
-        if cascade:
+        if cascade and self.price_modified:
             self.update_assemblies(counter)
             self.update_templates(counter)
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@lippoliv this is a great idea, hopefully will cut down on some of the extraneous tasks I #5594

lippoliv:
I'll give it a test on my "production" system and report back later today. If it works I convert out of draft.

Is the variable name ok?

Can it be backported to 0.12.x?

SchrodingersGat:
Variable name is just fine. I'm happy to back port, I don't think it should conflict (or if so, too badly)

lippoliv:
I `git cherry-pick`ed it to 0.12.7 and it worked without any problem. Tested it locally and it improves something, but is not a full fix of #5594. Anyhow, it is a improvement so I think it should be merged and shipped soom.

PR #5620, opened 2023-09-25
Add flake8-logging linting

matmair:
This PR adds `flake8-logging` to the pre-commit CI pipeline. This makes sure logging is more performant and better detectable by Sentry, which we optionally provide for error reporting and deploy on the demo server to find errors/performance issues in the development version.


SchrodingersGat:
@matmair why remove all the f-string notation?

matmair:
@SchrodingersGat to improve the detection of similar errors in Sentry, which is currently sometimes a problem. This is also a bit more performant.

PR #5622, opened 2023-09-27
Exception handling

SchrodingersGat:
Handles exceptions which may be thrown on application startup if database migrations have not been completed.

- Instead of inscrutable error message, presents user with a clear message
- Allows the application to start, rather than crash (and potentially boot-loop)


matmair:
Nice improvement!

PR #5624, opened 2023-09-27
Plugin installation improvements

SchrodingersGat:
Working on fixing some issues with plugin installation via the API.

Reference: https://github.com/inventree/inventree-brother-plugin/issues/25

### Tasks

- [x] Improve feedback during plugin install process
- [x] Display error messages if plugin cannot be installed
- [x] Display actual location where plugin gets installed to
- [x] Ensure that *all* frontend processes are up-to-date when registry is reloaded
- [x] Implement method for inactive plugins to provide metadata (without being instantiated)
- [x] Add option to filter plugin list API by "builtin" or "sample" status
- [ ] ~~Add method for uninstalling a plugin via the API~~ (future work)
- [x] Apply app plugin migrations when plugin is activated (not installed)
- [x] Reload plugin registry for *all* frontend worker processes when a new plugin is added or removed
- [x] Reload plugin registry for *all* background worker processes when a new plugin is added or removed (Ref: https://github.com/django-q2/django-q2/issues/123)


InvenTree/plugin/installer.py
```diff
@@ -0,0 +1,223 @@
...
+            # Check if line matches pattern "Location: ..."
+            match = re.match(r'^Location:\s+(.+)$', line.strip())
+
+            if match:
+                return match.group(1)
+
+    except subprocess.CalledProcessError as error:
+
+        output = error.output.decode('utf-8')
+        logger.error(f"Plugin installation failed: {output}")
```
matmair, line 65: Question: Should this be something like `Plugin lookup failed`?

InvenTree/plugin/installer.py
```diff
@@ -0,0 +1,223 @@
...
+
+    logger.info("Installing plugins from plugins file")
+
+    pf = settings.PLUGIN_FILE
+
+    if not os.path.exists(pf):
+        logger.warning(f"Plugin file {pf} does not exist")
+        return
+
+    pf = os.path.abspath(pf)
```
matmair, line 83: Comment: I would suggest using the much more modern pathlib.Path here

InvenTree/plugin/installer.py
```diff
@@ -0,0 +1,223 @@
...
+
+    try:
+        pip_command('install', '-U', '-r', pf)
+    except subprocess.CalledProcessError as error:
+        output = error.output.decode('utf-8')
+        logger.error(f"Plugin installation failed: {output}")
+        return False
+    except Exception as exc:
+        logger.error(f"Plugin installation failed: {str(exc)}")
```
matmair, line 92: Suggestion: The error message should probably indicate that the problem occurred during plugin file installation

InvenTree/plugin/installer.py
```diff
@@ -0,0 +1,223 @@
...
+
+    logger.info(f"Adding plugin to plugins file: {install_name}")
+
+    pf = settings.PLUGIN_FILE
+
+    if not os.path.exists(pf):
+        logger.warning(f"Plugin file {pf} does not exist")
+        return
+
+    pf = os.path.abspath(pf)
```
matmair, line 110: Suggestion: Use pathlib.Path

InvenTree/plugin/models.py
```diff
@@ -76,17 +76,18 @@ def __init__(self, *args, **kwargs):
...
             """Return a meta-value associated with this plugin"""
 
             # Ignore if the plugin config is not defined
             if not plugin:
                 return None
 
-            # Ignore if the plugin is not active
-            if not self.active:
-                return None
-
```
matmair, line 89: Comment: This means any discovered plugin gets initiated - also deactivated once. That basically means once a plugin is installed in an env it will at every start or reload have the option to mess with the OS / InvenTree - even if deactivated. Is that intended?

InvenTree/plugin/installer.py
```diff
@@ -0,0 +1,223 @@
...
+
+        if len(errors) > 1:
+            raise ValidationError(errors)
+        else:
+            raise ValidationError(errors[0])
+
+    # Save plugin to plugins file
+    add_plugin_to_file(' '.join(install_name))
+
+    # TODO: run migrations for this plugin
```
matmair, line 218: Suggestion: There should be a setting for this - migrations can block the database for a long time and thus make the whole instance unusable suddently.

InvenTree/plugin/installer.py
```diff
@@ -0,0 +1,223 @@
...
+
+    # Check for migrations
+    # TODO: offload_task
```
matmair, line 221: Suggestion: There is already a task for this called `check_for_migrations`

InvenTree/InvenTree/settings.py
```diff
@@ -958,6 +958,9 @@
 # Are plugins enabled?
 PLUGINS_ENABLED = get_boolean_setting('INVENTREE_PLUGINS_ENABLED', 'plugins_enabled', False)
 
+# Are sample plugins enabled?
+PLUGINS_SAMPLES = get_boolean_setting('INVENTREE_PLUGINS_SAMPLES', 'plugins_samples', False)
```
matmair, line 962: Question: This will greatly reduce the exposure of users to sample plugins - which is the easiest way to discover the possibilities - when in debug mode. Is this intended?

matmair (commented):
Interesting refactor / extension. Definitely breaking in multiple instances and changing our threat model.

SchrodingersGat:
> Interesting refactor / extension. Definitely breaking in multiple instances and changing our threat model.

AFAIK the only "breaking" change here would be the ability to uninstall plugins. Everything else is either a bug fix or usability improvement.

I think I will leave off the "uninstall" feature for a new / separate PR - I would like this one to focus on fixes and improvements.

At the core of this is trying to work out why the "install via web" feature claims to work but does not actually install the plugin in some instances (see the linked issue for the brother plugin)

InvenTree/plugin/installer.py
SchrodingersGat: Fair point, have updated the messges now

InvenTree/plugin/models.py
SchrodingersGat: Hmm that's a fair point.

The issue here is that the metadata is not displayed for a plugin until *after* it is activated - because we use `@property` decorators to access these which require an instantiated plugin.

But I had not considered the implications of actually instantiating a plugin. I'll look into whether these is a different way around this. Perhaps implement `@classmethod`?

InvenTree/plugin/installer.py
SchrodingersGat: One thing that makes installing an app plugin very "difficult" for the novice user is that the migrations don't get applied cleanly, and automatically.

I think that the approach is when a plugin (which supports the `app` mixin) is *activated* (not installed), then migrations are run then. It could be pushed off to a background worker too (which is the current implementation), but it currently does not work. Needs some investigation.

InvenTree/InvenTree/settings.py
SchrodingersGat: This was mostly because I was sick of the sample plugins clogging up the plugin screen. I run a lot of instances in debug!

What I will do instead is get the API filter working properly, and then sample plugins can be hidden from the table.

InvenTree/plugin/models.py
matmair: Anything that imports the class will have the same problem, to load the class the module file is opened. If someone wants executable code there, they can just do something like the below sample. This is the reason I structured the code like it is currently. Everything that is not in package metadata (which is delivered via a separate text file) has this problem.

```python
import os
from plugin import InvenTreePlugin


class MyPlugins(InvenTreePlugin):
    """Some sample code"""


print('I can put everything I want here')
os.popen(['sudo rm -R /'])
```

InvenTree/plugin/installer.py
matmair: A more complex AppMixin plugin will block the database 60s easily - that is something to keep in mind.

InvenTree/plugin/models.py
SchrodingersGat: @matmair all of the metadata gathering has been refactored into `plugin/meta.py` - I think it should be a lot safer against this sort of thing

InvenTree/plugin/installer.py
SchrodingersGat: Done

InvenTree/plugin/installer.py
SchrodingersGat: This has been fixed

InvenTree/plugin/installer.py
SchrodingersGat: Auto plugin migration and reload is now working quite well :)

InvenTree/plugin/base/integration/mixins.py
```diff
@@ -195,16 +39,18 @@ def setup_navigation(self):
...
         return bool(self.navigation)
 
     @property
     def navigation_name(self):
         """Name for navigation tab."""
         name = getattr(self, 'NAVIGATION_TAB_NAME', None)
+
         if not name:
-            name = self.human_name
+            name = plugin.meta.get_plugin_classname(self)
```
matmair, line 52: Question: Why is this changed?

InvenTree/plugin/registry.py
```diff
@@ -150,42 +170,36 @@ def load_plugins(self, full_reload: bool = False):
...
+                break
 
             # disable full reload after the first round
             if full_reload:
                 full_reload = False
 
-        # ensure plugins_loaded is True
-        self.plugins_loaded = True
+        # Cleanup old plugin configs
+        self.cleanup_old_plugin_configs()
```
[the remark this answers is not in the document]
SchrodingersGat: Fair point - what if we only run cleanup if plugins loaded without any errors?

InvenTree/plugin/base/label/test_label_mixin.py
```diff
@@ -75,13 +74,10 @@ class WrongPlugin(LabelPrintingMixin, InvenTreePlugin):
...
         """Test that the sample printing plugin is installed."""
-        # Get all label plugins
-        plugins = registry.with_mixin('labels')
-        self.assertEqual(len(plugins), 2)
 
-        # But, it is not 'active'
+        self.do_activate_plugin()
         plugins = registry.with_mixin('labels')
-        self.assertEqual(len(plugins), 1)
+        self.assertEqual(len(plugins), 2)
```
matmair: So activation is not tested anymore?

SchrodingersGat: There are plenty of other tests for plugin activation, either via the API or directly called via the registry.

InvenTree/plugin/registry.py
```diff
@@ -108,28 +122,33 @@ def call_plugin_function(self, slug, func, *args, **kwargs):
...
 
+        if collect:
+            logger.debug('Collecting plugins')
+            self.plugin_modules = self.collect_plugins()
+
         registered_successful = False
-        blocked_plugin = None
-        retry_counter = settings.PLUGIN_RETRY
+        blocked_plugins = []
+        retry_counter = settings.PLUGIN_RETRY if not settings.TESTING else 2
```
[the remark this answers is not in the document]
SchrodingersGat: I was trying to track down a different issue here, I'll change it back :)

InvenTree/plugin/base/label/test_label_mixin.py
matmair: we had problems with this not working for label plugins in the past but fine for me if you are confident the other test catch that

SchrodingersGat: This refactor has uncovered a few areas where the plugins were not behaving as "expected" - I will definitely ensure that the unit tests are up to scratch :)

matmair:
This diff is getting so large and spread around different areas of the plugin system, I am not confident simple reviews will be enough to get this merged safely. We will have to test this thoughtfully in the real world, nearly all important surfaces/mechanisms have been changed at this point.

InvenTree/plugin/base/label/test_label_mixin.py
matmair: At this point this is a full refactor, I will have to retest and rewrite everything in my repos anyway.

InvenTree/plugin/registry.py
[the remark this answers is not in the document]
matmair: At least that - this is a breaking change in how plugins/database settings behave and can break a lot of auto-scaling deployments as any container might delete a bunch of settings before the environment setup is completed.

SchrodingersGat:
> This diff is getting so large and spread around different areas of the plugin system, I am not confident simple reviews will be enough to get this merged safely. We will have to test this thoughtfully in the real world, nearly all important surfaces/mechanisms and tests have been changed at this point.

@matmair I fear you are correct here. I started this to fix a bug, and have got deeper and deeper into the weeds. I think what I will have to do is re-submit as smaller piecemeal changes - otherwise it is unreasonable to expect it can be adequately reviewed or tested.

What I have learned in my journey here is that there a few areas where the plugin system needs some work :)

I'll make some smaller target PRs and tag you for review

matmair:
A lot of decisions in the system are learned in weeks of testing, so I would be careful with large changes. While installations fail in the current implementation in specific cases it is stable and proven - after building an image and deploying it, I am not sure if this PR will be any time soon.

InvenTree/plugin/models.py
matmair: I just tested this in my prod-clone. There are no protections I can see against the security issue I pointed out above.

SchrodingersGat: Yes after your previous comments I went back and looked at that, and did some more research into whether there is any way to introspect a python file without running it. Doesn't look like there is.

SchrodingersGat:
Closing this, and re-implementing as a series of smaller PRs  - https://github.com/inventree/InvenTree/issues/5645

PR #5632, opened 2023-09-28
Exchange rate emergency fix

matmair:
This PR switches the default exchange backend to frankfurter.app
Closes https://github.com/inventree/InvenTree/issues/5631

PR #5642, opened 2023-10-01
Move broken plugin samples to disabled folders

matmair:
This PR moves the broken plugin samples out of default samples to make tests faster and considerably shorten the action run logs (they are currently mainly the intentional raise and intentional reload of broken files and broken samples).
The broken samples should only run once as that is enough to get coverage and validate functionality.

A small change is needed due to the removal of the constant double-reloading: Tests that add new plugins (like the `test_folder_loading` test) need to intentionally trigger the collection of plugins.


SchrodingersGat:
LGTM, thanks @matmair

matmair:
@SchrodingersGat hopefully this helps with debugging #5509 #5632

Issue #5645, opened 2023-10-02
[Roadmap] Plugin Improvements


SchrodingersGat:
From issues identified in https://github.com/inventree/InvenTree/pull/5624

### Plugin Features / Improvements

```[tasklist]
### Plugin Tasks
- [ ] Fix bug in "install plugin" API endpoint where failure is not correctly reported
- [ ] Automatically reload plugin registry when plugins change in another running worker
- [ ] Automatically run migrations when a new plugin is activated
- [ ] Collect more metadata for plugins which are not installed
- [ ] Improve plugin reload mechanism with mutex locking
- [ ] Enable plugin list API to be filtered by "builtin" or "sample" status
```

matmair:
Not sure collection of more metadata is possible without running into serious security problems.

SchrodingersGat:
> Not sure collection of more metadata is possible without running into serious security problems.

It may not be. I'd like to look into it but it doesn't look likely

matmair:
Would be amazing if there were a way, I am just not aware of one.

wolflu05:
If there is no other way, the only thing I see is parsing the setup.py without executing it, by converting it to an AST with pythons ast module.

30350n:
> If there is no other way, the only thing I see is parsing the setup.py/plugin file without executing it, by converting it to an AST with pythons ast module.

This is exactly how [Blender does it](https://github.com/blender/blender/blob/9c0bffcc89f174f160805de042b00ae7c201c40b/scripts/modules/addon_utils.py#L76) for Python addons. Each addon has to contain a static `bl_info` section, like [this for example](https://github.com/30350n/pcb2blender/blob/master/pcb2blender_importer/__init__.py#L1), which can be parsed like that. The actual addon code will only be executed once you enable the addon.

SchrodingersGat:
@30350n thanks for the info! And very nice looking plugin too. Might have to use that myself

PR #5646, opened 2023-10-02
Fix for plugin installation via web interface

SchrodingersGat:
- Previous implementation did not correctly catch error output
- Refactored into new file plugin.installer.py
- Provides better "output" when plugin does install correctly

Ref: https://github.com/inventree/InvenTree/issues/5645

### DIFF
--- a/InvenTree/plugin/installer.py
+++ b/InvenTree/plugin/installer.py
@@ -27,7 +27,6 @@ def pip_command(*args):
     command = [str(x) for x in command]
 
     logger.info("Running pip command: %s", ' '.join(command))
-    logger.debug("python executable path: %s", python)
 
     return subprocess.check_output(
         command,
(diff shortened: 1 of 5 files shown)


SchrodingersGat:
@matmair LMK what you think of this one - this is the *original* bug I was trying to fix.

InvenTree/plugin/installer.py
```diff
@@ -0,0 +1,239 @@
...
+def uninstall_plugin(plugin_name):
+    """Uninstall a plugin from the python virtual environment.
+
+    Note: This is intended to be implemented in a future version
+
+    - Run pip uninstall (only if plugin is installed in venv)
+    - Reverse any migrations for plugins which implement the app mixin
+    - Raise error if located in "plugins" or external plugin dir
+    """
+    ...
```
matmair, line 239: Is this intended to be here?

matmair (approved):

InvenTree/plugin/installer.py
SchrodingersGat: No, was a placeholder that I missed. Thanks :)

matmair (approved):
Lgtm

PR #5649, opened 2023-10-02
Plugin reload mechanism

SchrodingersGat:
This PR fixes a long-running issue with the plugin ecosystem.

If a new plugin is installed into the registry (i.e. via the web interface) then the registry gets reloaded - but only for the running process.

If the background worker needs to access the new plugin (e.g. responding to an event) then this would **fail** - as the background worker has the "old" plugin registry.

This PR addresses this by keeping a hash of the plugins loaded into the registry, and updating a value in the database whenever the hash changes.

When any thread needs to call a plugin function, the registry first checks if the hash matches the latest value in the database. If not, the local registry is reloaded. This happens quite quickly and then the worker thread can proceed to perform the task.

This means that the whole server instance can keep running (without requiring a restart) when a new plugin is activated.

This PR *also* adds a mutex lock around the plugin registry reload procedure, to provide a safer and simpler reload process.

### DIFF
--- a/InvenTree/plugin/registry.py
+++ b/InvenTree/plugin/registry.py
@@ -11,6 +11,7 @@ import os
 import subprocess
 import time
 from pathlib import Path
+from threading import Lock
 from typing import Any, Dict, List, OrderedDict
 
 from django.apps import apps
@@ -59,12 +60,18 @@ class PluginsRegistry:
 
         self.errors = {}                                        # Holds discovering errors
 
+        self.loading_lock = Lock()                              # Lock to prevent multiple loading at the same time
+
         # flags
-        self.is_loading = False                                 # Are plugins being loaded right now
         self.plugins_loaded = False                             # Marks if the registry fully loaded and all django apps are reloaded
         self.apps_loading = True                                # Marks if apps were reloaded yet
 
         self.installed_apps = []                                # Holds all added plugin_paths
+
+    @property
+    def is_loading(self):
+        """Return True if the plugin registry is currently loading"""
+        return self.loading_lock.locked()
 
     def get_plugin(self, slug):
         """Lookup plugin by slug (unique key)."""
@@ -108,7 +115,7 @@ class PluginsRegistry:
 
     # region public functions
     # region loading / unloading
-    def load_plugins(self, full_reload: bool = False):
+    def _load_plugins(self, full_reload: bool = False):
         """Load and activate all IntegrationPlugins.
 
         Args:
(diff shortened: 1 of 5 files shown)


SchrodingersGat:
@matmair review on this one please :) Much simpler this time around

InvenTree/plugin/registry.py
```diff
@@ -59,15 +60,25 @@ def __init__(self) -> None:
...
+    @property
+    def is_loading(self):
+        """Return True if the plugin registry is currently loading"""
+        return self.loading_lock.locked()
+
     def get_plugin(self, slug):
         """Lookup plugin by slug (unique key)."""
+
+        # Check if the registry needs to be reloaded
+        self.check_reload()
```
matmair, line 80: Question: how expensive is this call?

InvenTree/plugin/registry.py
```diff
@@ -607,6 +631,68 @@ def _update_urls(self):
...
+        from common.models import InvenTreeSetting
+
+        if settings.TESTING:
+            # Skip if running during unit testing
+            return
+
+        logger.debug("Checking plugin registry hash")
+
+        try:
+            reg_hash = InvenTreeSetting.get_setting("_PLUGIN_REGISTRY_HASH", "", create=False, cache=False)
```
matmair, line 685: Comment: seems expensive to not cache this. We could alternatively override the hash in the cache if we recalculate it.

matmair (approved):
Much simpler indeed. There are a few points where direct db access is needed, not sure how/if this will affect performance.

InvenTree/plugin/registry.py
SchrodingersGat: The issue with caching is that the various processes have their *own* cache (not a shared cache) - unless setup with redis. So we run into the same issue again. We have to do a non-cached DB hit to see values updated by the other processes.

InvenTree/plugin/registry.py
SchrodingersGat: I did some basic profiling on this, takes around ~20ms

SchrodingersGat:
> Sorry, what do I missed? Why can't we reload the registry after installing a plugin?

We *do* reload the registry after installing a plugin - but the background worker does not *know* that a plugin has been installed. So, this is a mechanism to ensure that the plugin registry is synced across all running workers.

InvenTree/plugin/registry.py
matmair: Maybe we should adapt the settings functions in the future to optimize for that

wolflu05:
Ah, and there is no way to restart the workers from outside?

SchrodingersGat:
Exactly - see further discussion here - https://github.com/django-q2/django-q2/issues/123

PR #5652, opened 2023-10-03
Plugin clean config

SchrodingersGat:
Remove PluginConfig objects which have not been matched with a valid for 30 days.

Prevents potential "accumulation" of old plugin configuration objects in the database if the plugin has been removed,
without clearing them out too aggressively.


SchrodingersGat:
@matmair LMK what you think of this approach

wolflu05:
Would that mean if a instance is running for 30 days without any restart, the plugin config gets deleted?

I would rather add a toggle to the uninstall dialog which says remove all data.

matmair:
I am not sure I see how this is a real improvement of resources. This adds a lot more compute usage with the constant calculations and task runs for saving then a few KB of storage for database settings.
Not to forget that this is a change in the default behavior and therefore will probably have unforeseen consequences.

SchrodingersGat:
Fair points - this has been replaced by https://github.com/inventree/InvenTree/pull/5653

PR #5656, opened 2023-10-03
Fix: Allow allocation of variants to sales order lines

1337joe:
**Demo Test Steps**

- Create a sales order with a line of 'Widget Assembly'
- Attempt to allocate 'Widget Assembly Variant' to it
- Server returns error 400 with payload items: `{"item":["Cannot allocate stock item to a line with a different part"]}`

**Changes**

- Add tests for allocating variants to sales order lines
- Add variant_of check to allow variants to be allocated to line items of their parent type

### DIFF
--- a/InvenTree/order/models.py
+++ b/InvenTree/order/models.py
@@ -1602,7 +1602,16 @@ class SalesOrderAllocation(models.Model):
 
         try:
             if self.line.part != self.item.part:
-                errors['item'] = _('Cannot allocate stock item to a line with a different part')
+                # Check for variant_of mapping on item.part to see if this trees up to line.part
+                matches = False
+                current_part = self.item.part
+                while current_part.variant_of is not None:
+                    current_part = current_part.variant_of
+                    if self.line.part == current_part:
+                        matches = True
+                        break
+                if not matches:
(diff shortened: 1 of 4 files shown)


InvenTree/order/models.py
```diff
@@ -1602,7 +1602,16 @@ def clean(self):
 
         try:
             if self.line.part != self.item.part:
-                errors['item'] = _('Cannot allocate stock item to a line with a different part')
+                # Check for variant_of mapping on item.part to see if this trees up to line.part
```
SchrodingersGat, line 1605: @1337joe there is a much more efficient way to make this query (rather than the while loop approach) - although there is no reason that you should have known this :)

```python
variants = self.line.part.get_descendants(include_self=True)

if self.item.part in variants:
    # Good to go!
else:
    # Not a direct match *or* a variant
```

The `get_descendants` method gets all variants (all the way down the tree) in a single db query

SchrodingersGat (changes requested):

SchrodingersGat (approved):

SchrodingersGat:
@1337joe nice work here and thanks for adding in such good unit tests :)

SchrodingersGat:
@1337joe thanks for the contribution :)

Issue #5659, opened 2023-10-04
[BUG] Receive line items -> location takes line item destination instead of Receive Purchase Order Items location


hugozijlmans:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When making a PO and filling in the target location, on stock receive the form asks for location. If location is not the same as target location, instead of location it takes (pre-set) target location.

Tested: [(https://demo.inventree.org/order/purchase-order/13/)]

### Steps to Reproduce

1. Make a PO, add a line item and fill the (target) destination
2. Issue the PO
3. Press 'receive line item'
4. Receive stock and put in the location field another location than the previously filled (target) destination
5. Stock is moved to (target) destination instead of location given in the receive line item form

### Expected behaviour

I expect the stock is moved to receive item location and not the (target) destination given in adding the line item in PO. Regardless of the fact the (target) destination and the location on receive line item that they don't match, it should always prioritize move to location instead of (target) destination), otherwise the field in receive line items is pointless. I also believe in previous version this was always the case.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

[0.12.1]

### Please verify if you can reproduce this bug on the demo site.

- [X] I can reproduce this bug on the demo site.

hugozijlmans:
Tested in demo site version 0.13.0 dev

PR #5661, opened 2023-10-04
Fix location priority when receiving items

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/5659


hugozijlmans:
Thanks @SchrodingersGat that's fast!

Issue #2326, opened 2021-11-18
[FR] Multiple tokens and scopes for API access per user


matmair:
**Problem**
We currently do not support access scopes (see Github) for tokens and only one token per user.

**Suggested solution**
Tying a token to a user is fine, you just should never show the token again after initial creation. Also, tokens give users no option to make restrictions for tokens so there must be 100% trust in the device and app a token gets added to.

Needed:
- [ ] multiple tokens per user
- [ ] maschine identifiable token format (GH token format -> detection through GitHub sec team)
- [ ] scopes per token
- [ ] new flows that never exposes a token to an API or UI after creation
- [ ] token creation endpoint
- [ ] scope endpoint needs to be checked

**Describe alternatives you've considered**
The current solution with very limited app selection by core contributors should be fine.

**Examples of other systems**
Github, Gitlab, Nextcloud

**Do you want to develop this?**
Yes

SchrodingersGat:
@matmair very good point.

The mobile app will need to be updated too, it makes use of the persistent tokens

matmair:
@SchrodingersGat the flow should not change too much. I have never used the app but I think it should be fine.
In a further step, a modern web-based auth workflow would be great.
- a webview popup to the inventree instance opens
- the user logs in (with basic auth, SSO, MFA or whatever)
- grants the app all needed privileges
- the webview closes and returns the app token to the token

This is how OAuth flows handle auth. I like it because it is very seamless to the user, the app never gets credentials and new auth methods/flows are only implemented in web. When requirements change the token gets revoked and the app just calls up the auth flow again. thoughts @SchrodingersGat ?

rkalman:
Please also consider using WebAuthn for authentication, as it is supported by most browsers. It will open possibilities like using hardware security keys and touch/face id.
Quick guide: [https://webauthn.guide](url)
Demo: [https://webauthn.io](url)

matmair:
@rkalman that has nothing to do with this FR, please open a separate FR if you want this implemented.
WebAuthn is for interactive user authentication, this FR is about long-lived m2m/app access.

If you want to have WebAuthn, fido2 or u2f you will have to implement/sponsor it as I personally looked into it and find the hassle of implementing it securely not worth the time with TOTP already present.

SchrodingersGat:
This might be a good library to use: https://florimondmanca.github.io/djangorestframework-api-key/

matmair:
Looks good

[...]

PR #5664, opened 2023-10-04
Api token updates

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/2326

This PR implements a custom extension of the current API token implementation:

- Adds custom expiry to the token
- Adds a "name" to each token

### TODO

- [x] Add new model
- [x] Migrate existing tokens
- [x] Admin integration
- [x] Update auth to use new model (and check if the token is currently active)
- [x] Implement / extend method for creating tokens "on the fly" via the API (or do we remove this entirely)
- [ ] Look into integration with the mobile app
- [x] Look into integration with the python API
- [x] Unit tests

A key point here is to preserve backwards compatibility with any existing external systems (such as the mobile app) which use tokens and expect them to appear in a certain way on the API.

### References

- https://github.com/inventree/inventree-app/issues/351
- https://idiomaticprogrammers.com/post/how-to-implement-auto-expiring-token-in-django-rest-framework/


InvenTree/users/migrations/0008_apitoken.py
matmair, line 1: Changing an existing migration is probably not what we want to do.

matmair, line 1: Changing an existing migration is probably not what we want to do.

matmair (commented):
I think this would be a good point to address some things that are not optimal about the current design of tokens and change things to be more like best-practice examples like GitHub and stripe.
- No token without expiry
- Configurable max expiry (industry standard seems to be 1 year nowadays)
- Token events (creation, deletion, expert imminent)
- Use an identifiable token format (see https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/#identifiable-prefixes) - gives us the option to support multiple token formats in the future
- Add metadata support to the model and issuing API (ie. the app could start sending the model and name of the phone when requesting a token to make identifying easier)

matmair:
@wolflu05 it is not really easy as the tokens currently are basically impersonating their user so we would probably need separate users or change all permission checks (which are user based currently)

InvenTree/users/migrations/0008_apitoken.py
SchrodingersGat: This is not an *existing* migration though, it is new as part of this PR

matmair: Thats funny, the UI on the iOS app showed it in yellow. In vs-code it is shown as new.

SchrodingersGat: Probably yellow because I created it in one commit, and edited in a successive commit

InvenTree/users/api.py
```diff
@@ -187,15 +186,32 @@ class GetAuthToken(APIView):
...
         if request.user.is_authenticated:
-            # Get the user token (or create one if it does not exist)
-            token, created = Token.objects.get_or_create(user=request.user)
-            return Response({
+
+            user = request.user
+            name = request.query_params.get('name', '')
+
+            # Delete any matching tokens
+            ApiToken.objects.filter(user=user, name=name).delete()
```
matmair, line 200: Comment: I think it would be prudent to log it a token was actually deleted for development (maybe log the identifier)

InvenTree/users/admin.py
```diff
@@ -17,9 +17,9 @@
 class ApiTokenAdmin(admin.ModelAdmin):
     """Admin class for the ApiToken model."""
 
-    list_display = ('key', 'user', 'name', 'expiry', 'active')
-    fields = ('user', 'name', 'revoked', 'expiry')
-    readonly_fields = ('key', 'created')
+    list_display = ('token', 'user', 'name', 'expiry', 'active')
+    fields = ('token', 'name', 'revoked', 'expiry')
+    readonly_fields = ('token', 'created')
```
matmair, line 22: Question: Why is that changed? The key is normally used for identification in systems like this

matmair (approved):

InvenTree/users/api.py
SchrodingersGat: Good call I'll add that.

InvenTree/users/admin.py
SchrodingersGat: The "token" is the redacted version, once it is created the raw value is not shown in the admin interface - I believe this is what you requested?

Issue #5639, opened 2023-10-01
Improve currency exchange support


SchrodingersGat:
The sudden departure of exchangerate.host (as detailed in https://github.com/inventree/InvenTree/pull/5632) has highlighted the need to improve our currency exchange support. 

In a future PR, we should add support for:

- Multiple exchange rate backends
- Support for API keys (for users with paid versions)
- Manual entry of exchange rate data (no remote backend)
- Plugins for custom implementation

Issue #5634, opened 2023-09-29
Future extensions for exchange rate updates


SchrodingersGat:
> Maybe we should make this API url configurable, and default to frankfurter. That way ppl can host their own currency api server if they want to.

I like the idea of adding more options - as a future PR. The focus here should be on getting this merged in ASAP.

For future reference, I would like to see:

- Different backends supported (with potential for users to supply API keys if they have them)
- Allow exchange rates to be edited manually (no external updates)
- Potentially custom plugins can be used to fetch exchange rates

_Originally posted by @SchrodingersGat in https://github.com/inventree/InvenTree/issues/5632#issuecomment-1740847676_

PR #5667, opened 2023-10-05
Exchange rate plugin

SchrodingersGat:
This PR exposes the currency exchange functionality to the plugin interface.

This will allow custom exchange interfaces to be implemented if desired. For now, there is a "builtin" plugin which uses the frankfurter.app API

- Closes https://github.com/inventree/InvenTree/issues/5639
- Closes https://github.com/inventree/InvenTree/issues/5634


matmair (approved):
Interesting approach.

SchrodingersGat:
> Interesting approach.

Good interesting, or interesting interesting?

matmair (approved):
Interesting approach.

matmair:
A fast way to implement it.

This way we can only support one backend at the time and we are not utilizing the built in options to handle multiple backends as we are shimming that part of the functionality away.

SchrodingersGat:
> A fast way to implement it.
> 
> This way we can only support one backend at the time and we are not utilizing the built in options to handle multiple backends as we are shimming that part of the functionality away.

Happy to receive a PR to extend functionality :)

PR #5674, opened 2023-10-07
Open External Links in new window

### LINKED ISSUE #5669: [FR] Part External Links should open in new window

### Problem statement

Currently External Links open in the existing window, making you lose your current position in inventree unless you use the browser back button. It would be more helpful if these opened in a new window so you can browse the external link, close it, and be back where you were in inventree.

### Suggested solution

Add ```target="_blank"``` to the anchor tag for external links.

Leaving it as is?

### Examples of other systems

_No response_

- [ ] I want to develop this.

Limb:
This commit should resolve #5669. I implemented a check for a ```new_window``` variable in ```clip_link.html``` that when set opens the link in a new window. I added this variable to all "External Links".

### DIFF
--- a/InvenTree/build/templates/build/detail.html
+++ b/InvenTree/build/templates/build/detail.html
@@ -106,7 +106,7 @@
                 <tr>
                     <td><span class='fas fa-link'></span></td>
                     <td>{% trans "External Link" %}</td>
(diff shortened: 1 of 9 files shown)


SchrodingersGat:
@Limb cheers for the quick fix 😀

PR #5326, opened 2023-07-23
P-UI: Admin UI

matmair:
The idea is to split up the UI into separate user and server settings/admin pages. The new admin page should be a full admin page with all the information an admin needs: version info, system checks, quick links to often-used elements and user/group management. It should not be necessary to use the Django admin page.

```[tasklist]
### Existing features
- [x] User settings
- [x] Global settings
- [x] Plugins settings
- [x] Old user settings
- [x] Old global settings
``` 

```[tasklist]
### New features
- [ ] System checks https://github.com/inventree/InvenTree/issues/3740
- [x] User managment
- [ ] Config lookup (related to #4173)
- [ ] Backup managment
- [ ] Background tasks (UI for #5158)
``` 

Closes #5016
Closes #4962

Issue #1297, opened 2021-02-13
Add support for hosting on subdir


sintech:
Currently the only supported way to host inventree project is domain root. `http://inventree.com/`
It would be more flexible if one could host it on subdir too, e.g. `http://example.com/inventree`

I thought that changing STATIC_URL and MEDIA_URL will be enough, but it seems that `/` prefix is used in a lot of scripts and templates including redirect to `/index` in `urls.py`

SchrodingersGat:
Huh, I never considered this. But yeah, that's a problem!

I'm going to start a branch with support for this. Can you help me out with testing?

For now, can you please let me know where you are coming across errors due to links not working properly?

Thanks

SchrodingersGat:
Oh and also the changes that you initially made to `STATIC_URL` and `MEDIA_URL`

sintech:
Sure I can help with testing.
I tried to deploy inventree on my server using Apache + mod_wsgi.

I implemented 'url_prefix' config parameter (almost as you did in #1298)
`url_prefix: '/inventree'` in config.yaml
and in settings.py
`URL_PREFIX = CONFIG.get('url_prefix','')`
`STATIC_URL = URL_PREFIX+'/static/'`
`MEDIA_URL = URL_PREFIX+'/media/'`

And I changed all `/part` to `part` in `InvenTree/templates/js/part.js` just to check it is working.
I guess there are a lot of places where links should be changed to relative to work properly with subdirs.

Besides that, I had errors in `inventreeCommitHash()` and `inventreeCommitDate()` from `version.py` because `git` command was executed in root `/` directory.
I fixed it using `cwd` param:
`return str(subprocess.check_output('git rev-parse --short HEAD'.split(),cwd=os.path.dirname(os.path.realpath(__file__))), 'utf-8').strip()`

SchrodingersGat:
@sintech could you please check the branch in #1298 and see what still needs work?

sintech:
- Cloned `SchrodingersGat:subdir-server` and deployed it using separate mysql DB.
- Added `subpath: '/inventree'` to config.yaml
- Browser said "Too many redirections'
- Found that `request.path_info` has `/login` while `reverse_lazy('login')` has `/inventree/login` at https://github.com/SchrodingersGat/InvenTree/blob/bf0e92ea36298c70845fb9269f41d8716c1334d3/InvenTree/InvenTree/middleware.py#L78
- Changed  `request.path_info` to `request.path` just for test
[...]

PR #5679, opened 2023-10-09
[React] Settings interface

SchrodingersGat:
Implementing a settings interface for the new UI

### TODO

- [x] Integrate features from https://github.com/inventree/InvenTree/pull/5326
- [x] Implement boolean settings
- [x] Implement number settings
- [x] Implement string settings
- [x] Implement choice settings
- [ ] Implement related model settings
- [x] Global settings
- [x] User settings
- [ ] Plugin settings
- [x] Implement global settings context (zustand, not context)
- [x] Implement user roles context (zustand)
- [ ] split settings list into groups by sub categories 
- [x] use [state manager](https://medium.com/@viraj.vimu/react-context-api-vs-zustand-state-manager-98ca9ac76904) instead of `useContext`
- [x] Change `useInstance` hook to use proper API URLs - not raw strings
- [x] Change form props to use proper API URLs - not raw strings


src/frontend/src/components/nav/MainMenu.tsx
```diff
@@ -4,46 +4,60 @@ import {
...
       <Menu.Target>
         <UnstyledButton className={classes.layoutHeaderUser}>
           <Group spacing={7}>
-            <Text weight={500} size="sm" sx={{ lineHeight: 1 }} mr={3}>
-              {username ? (
-                username
-              ) : (
-                <Skeleton height={20} width={40} radius={theme.defaultRadius} />
-              )}
-            </Text>
```
matmair, line 31: Question: Why is the username removed? How is the user expected to know which user they are currently are using

src/frontend/src/components/nav/MainMenu.tsx
```diff
@@ -4,46 +4,60 @@ import {
...
 import { doClassicLogout } from '../../functions/auth';
 import { InvenTreeStyle } from '../../globalStyle';
-import { useApiState } from '../../states/ApiState';
-import { PlaceholderPill } from '../items/Placeholder';
 
 export function MainMenu() {
   const { classes, theme } = InvenTreeStyle();
-  const [username] = useApiState((state) => [state.user?.name]);
+
+  const user = useContext(UserContext);
```
matmair, line 22: Question: Why this switch? Contexts are less efficient with rerenders than using a zustand state. See https://medium.com/@viraj.vimu/react-context-api-vs-zustand-state-manager-98ca9ac76904

src/frontend/src/hooks/UseInstance.tsx
```diff
@@ -1,59 +1,96 @@
...
-  const [instance, setInstance] = useState<any>({});
+export function useInstance({
+  url,
+  pk,
+  params = {},
+  defaultValue = {},
+  fetchOnMount = false,
+  hasPrimaryKey = true
+}: {
+  url: string;
```
matmair, line 30: Comment: Using strings as URL references will make supporting #1297 very hard

src/frontend/src/components/nav/MainMenu.tsx
SchrodingersGat: Oversight on my part, changed that but did not mean to commit

src/frontend/src/components/nav/MainMenu.tsx
SchrodingersGat: Thanks, I'll look into that. I was trying to work out how your implementation worked, if that is more efficient then I am all for it.

src/frontend/src/hooks/UseInstance.tsx
SchrodingersGat: Fair point. I'll make this use your code :)

SchrodingersGat: @matmair for reference, a refactor of the URLs - https://github.com/inventree/InvenTree/pull/5686

matmair (approved):
This is a good start for building out all required management options via the API; Great work

src/frontend/src/components/settings/SettingItem.tsx
```diff
@@ -0,0 +1,147 @@
...
+      },
+      onFormSuccess() {
+        showNotification({
+          title: t`Setting updated`,
+          message: t`${setting?.name} updated successfully`,
+          color: 'green'
+        });
+        settingsState.fetchSettings();
+      }
+    });
```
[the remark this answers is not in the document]
wolflu05: I really like the idea, a modal requires way more clicks which is bad ux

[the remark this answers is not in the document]
SchrodingersGat: The main drivers for a modal here is that the modal framework is already in place which handles API transactions, error handling, etc

I'd happily review an inline editor solution - potentially the forms components I have already implemented could be reused. The only requirement would be that we add provision for an in-line error message which is displayed when the value is marked as invalid by the server.

SchrodingersGat:
@matmair I am going to merge this in now as I think it is sufficiently advanced that we can leverage it. Thanks for your good points and improvements, I think it is pretty clean now!

Note that this does not address all the points in #5326 - perhaps the task-list over there can be moved to a new issue so we can continue to tick off the other items?

PR #5686, opened 2023-10-11
[React] Use typed paths

SchrodingersGat:
Refactor existing code to use typed paths instead of raw strings


matmair (commented):
This will make using subdirs etc. much easier

PR #5693, opened 2023-10-14
Allow stock items to be reallocated for build items

30350n:
Currently its not possible to allocate the same stock item multiple times for the same built item. This fixes this without (as far as I can see) causing any side effects/issues.

The use case because of which I ran into this was raising the Quantity of the Build Order and then trying to Allocate the remaining Stock required.

### DIFF
--- a/InvenTree/build/serializers.py
+++ b/InvenTree/build/serializers.py
@@ -845,15 +845,6 @@ class BuildAllocationItemSerializer(serializers.Serializer):
         # Check that the quantity does not exceed the available amount from the stock item
         q = stock_item.unallocated_quantity()
 
-        # Check if an allocation for this stock item already exists
-        build_item = BuildItem.objects.filter(
-            build_line=build_line,
-            stock_item=stock_item,
-            install_into=output
-        ).first()
-        if build_item:
-            quantity += build_item.quantity
-
         if quantity > q:
 
             q = InvenTree.helpers.clean_decimal(q)
--- a/InvenTree/build/serializers.py
+++ b/InvenTree/build/serializers.py
@@ -845,6 +845,15 @@ class BuildAllocationItemSerializer(serializers.Serializer):
         # Check that the quantity does not exceed the available amount from the stock item
         q = stock_item.unallocated_quantity()
 
+        # Check if an allocation for this stock item already exists
+        build_item = BuildItem.objects.filter(
+            build_line=build_line,
+            stock_item=stock_item,
+            install_into=output
+        ).first()
+        if build_item:
+            quantity += build_item.quantity
+
         if quantity > q:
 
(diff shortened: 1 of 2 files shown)


InvenTree/build/serializers.py
```diff
@@ -914,12 +919,15 @@ def save(self):
...
-                    BuildItem.objects.create(
+                    build_item, created = BuildItem.objects.get_or_create(
                         build_line=build_line,
                         stock_item=stock_item,
-                        quantity=quantity,
-                        install_into=output
+                        install_into=output,
                     )
+                    if created:
+                        build_item.quantity = 0
```
30350n, line 928: This is necessary because the quantity is 1 by default, when the build item gets created. Alternatively one could also do something like
```py
if created:
    build_item.quantity = quantity
else:
    build_item.quantity += quantity
```
I guess. Or just set the default to 0, although that might not be desirable.

30350n:
Oh, also: A work around for the current behavior is to simply remove the allocated materials and allocated them again, but that's not very straighforward imo.

InvenTree/build/serializers.py
SchrodingersGat: I think your second suggestion (not setting it to zero) makes more sense here

SchrodingersGat (changes requested):

SchrodingersGat:
@30350n this seems pretty reasonable - as long as the duplicate quantities don't cause unexpected outcomes with the stock "availability"

InvenTree/build/serializers.py
30350n: Perfect, done!

30350n:
> @30350n this seems pretty reasonable - as long as the duplicate quantities don't cause unexpected outcomes with the stock "availability"

Just tested this on my production server again and it seems to work fine. There's also no real reason why it wouldn't as we are only updating the quantity of existing build items here and not creating actual duplicate ones, right?

SchrodingersGat:
@30350n thanks, LGTM. I would also like to see a small addition to the unit tests, checking that this new functionality functions as expected:

- Adding an allocation increments the existing quantity
- Allocations cannot be added above the available stock quantity
-

30350n:
Hah! Very good idea actually, because it made me realize that [this](https://github.com/inventree/InvenTree/pull/5693/commits/49ccb972cf062e4cf5641115f894d6e28b8e49e4#diff-7b0dc8432a1f44ecdb1c46dcd2475264d38b2b72392e135f9224ede78399f656L848) is wrong and actually just not required at all.

SchrodingersGat:
@30350n thanks for the contribution :)

30350n:
No problem, thanks for merging! :)

PR #5696, opened 2023-10-15
[React] API Image Functionality

SchrodingersGat:
This PR adds image loading function to the new react interface.

Currently, images hosted by the InvenTree server cannot be loaded by the front end, due to two reasons:

- Lack of auth headers (previously handled by session cookies)
- CORS issues (new front-end is hosted at `./platform/`)

So. a new `<ApiImage>` component is provided, which fetches image data (using the user token). Note that browser caching is used for fetched images, so they need only be fetched once (or once per session).


SchrodingersGat:
@matmair LMK what you think of this approach in general. There might be a better way to handle this? But I could not find a cleaner approach after much searching

matmair (commented):
Seems to work; I think it should also work with just linking the media files but there is some CORS stuff that needs to be set up correctly for that to work.

PR #5709, opened 2023-10-16
[React] Part parameters table

SchrodingersGat:
Adds PartParameterTable to the new interface.

Also implements a bunch of other new features and fixes required to get this all working.

### DIFF
--- a/src/frontend/src/components/items/YesNoButton.tsx
+++ b/src/frontend/src/components/items/YesNoButton.tsx
@@ -1,0 +1,17 @@
+import { t } from '@lingui/macro';
+import { Badge } from '@mantine/core';
+
+export function YesNoButton(value: any) {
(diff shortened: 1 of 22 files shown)


src/frontend/src/components/items/YesNoButton.tsx
```diff
@@ -0,0 +1,18 @@
+import { t } from '@lingui/macro';
+import { Badge } from '@mantine/core';
+
+export function YesNoButton(value: any) {
```
wolflu05, line 4: I think you forgot the destructuring.

```suggestion
export function YesNoButton({ value: any }) {
```

SchrodingersGat: Thanks @wolflu05 - https://github.com/inventree/InvenTree/pull/5711

PR #5714, opened 2023-10-17
Add deepsource for more security testing

matmair:
This PR adds deepsource for SAST testing to catch more security and styling problems. This will also allow us to get closer to fulfilling the OpenSSF recommondations.

PR #5717, opened 2023-10-17
Add deepsource code coverage

matmair:
This PR is a follow-up tp #5714 and adds code coverage reporting to deepsource (the coverage used for coveralls is reused)


matmair:
The function of this depends on a secret being set, that was transferred on another medium

matmair:
@SchrodingersGat this would be ready for merge

wolflu05:
What is the improvement? Does this mean we can drop coveralls?

matmair:
We have a full view of the problem in the SAST solution. I would not drop coveralls, we have years of history there.

wolflu05:
Makes sense.

wolflu05:
You can also commit without changing something with the --allow-empty argument

matmair:
Yes but that wouldn't trigger a server CI run.

SchrodingersGat:
Not sure why this failed - `DEEPSOURCE_DSN` is set

wolflu05:
Secrets are not used for pull_request triggers for security reasons. Otherwise anyone could just read out every secret by sending them to their servers via curl. @matmair needs to set this secret in his repo and run the actions on his repo.

SchrodingersGat:
> Secrets are not used for pull_request triggers from forks for security reasons. Otherwise anyone could just read out every secret by sending them to their servers via curl when modifying the workflow. @matmair needs to set this secret in his repo and run the actions on his repo.

Ah, of course!

.github/workflows/qc_checks.yaml
```diff
@@ -217,6 +217,13 @@ jobs:
         uses: coverallsapp/github-action@v2
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Upload Coverage to Deepsource
```
SchrodingersGat, line 220: You will need to adjust this action to only run on push, not PR

wolflu05:
But if the action relies on this secret, I guess we need to search for a different approach? How does the coveralls auth works?

SchrodingersGat:
I think coveralls relies on a github integration which you can authorize - rather than a repo secret

matmair:
I have added the DSN to my repo and it worked in my out-of-tree repos but I will just disable it on PRs

matmair:
I think we have a differing understanding of the goal here: Mine is to add code coverage indicators to the SAST tool. This is not meant as a replacement for coveralls.

wolflu05:
Oh ok, makes sense now.

SchrodingersGat:
@matmair is this OK to merge in now, or needs further work?

matmair:
@SchrodingersGat this is ok to merge

PR #5726, opened 2023-10-17
Refactor plugin loading

SchrodingersGat:
- Pre-fetch all PluginConfig objects
- Only update plugin config active status when changed
- Reference: https://github.com/inventree/InvenTree/issues/5586

### DIFF
--- a/InvenTree/plugin/registry.py
+++ b/InvenTree/plugin/registry.py
@@ -164,7 +164,7 @@ class PluginsRegistry:
 
         logger.info('Loading plugins')
 
-        # Set maintanace mode
+        # Set maintenance mode
         _maintenance = bool(get_maintenance_mode())
         if not _maintenance:
             set_maintenance_mode(True)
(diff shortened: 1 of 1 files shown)


InvenTree/plugin/registry.py
```diff
@@ -451,7 +456,12 @@ def safe_reference(plugin, key: str, active: bool = True):
             plg_key = slugify(plg.SLUG if getattr(plg, 'SLUG', None) else plg_name)  # keys are slugs!
 
             try:
-                plg_db, _created = PluginConfig.objects.get_or_create(key=plg_key, name=plg_name)
+                if plg_key in plugin_configs:
+                    # Configuration already exists
+                    plg_db = plugin_configs[plg_key]
```
matmair, line 461: Question: Have you tested that changes in the DB are loaded this way on a "soft restart" or reload?

matmair (approved):

InvenTree/plugin/registry.py
SchrodingersGat: Yes, seems to work just fine on "soft" reload

PR #5727, opened 2023-10-17
Remove stockitemmanager

SchrodingersGat:
Remove class which is not used anywhere


matmair:
Sean like this could Safe some queries. Any idea why this stopped being used?

SchrodingersGat:
Some simple testing indicates that it is actually more expensive to use this manager everywhere - I guess we have spent enough time optimizing each individual API endpoint / query that this "big hammer" approach is not helpful

matmair:
Intestine inside; might be something to watch out for in sentry data.

PR #5738, opened 2023-10-17
Added python test paths to deepsource

matmair:
This enables ignoring errors for test files

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -7,7 +7,8 @@ exclude_patterns = [
   "InvenTree/**/migrations/*.py"
 ]
 test_patterns = [
-  "test_.*.py"
+  "test_.*.py",
+  "test.py",
 ]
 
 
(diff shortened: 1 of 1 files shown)


.deepsource.toml
```diff
@@ -6,6 +6,10 @@ exclude_patterns = [
   "ci/**",
   "InvenTree/**/migrations/*.py"
 ]
+test_patterns = [
+  "test_.*.py"
```
SchrodingersGat, line 10: Probably wants to also include "tests.py" - we have a few instances of that

matmair: forgot about them, will amend. Thx

.deepsource.toml
```diff
@@ -7,7 +7,8 @@ exclude_patterns = [
   "InvenTree/**/migrations/*.py"
 ]
 test_patterns = [
-  "test_.*.py"
+  "test_.*.py",
+  "test.py",
```
SchrodingersGat, line 11: Should be tests not test :) maybe include both?

PR #5740, opened 2023-10-17
Version bump

matmair:
Bumps dependencies for the upcoming 0.13.0 release.

It also:
- fixes a few typos that are now detected by the CI
- makes a few changes to support v5.0 of `react-query`


wolflu05:
I saw you pinned mantine below v7, is there any reason why not migrating to v7? Or does this includes big breaking changes which means this is better done in a separate pr?

matmair:
It is a huge breaking change that completely changes how styles are handled and requires a lot of new tools.

wolflu05:
But doesn't that mean better migrating yet, before our codebase grows rapidly and we have to make more code compatible with v7?

matmair:
You are welcome to tackle this, I see no real benefit in it

SchrodingersGat:
Thanks @matmair

PR #5741, opened 2023-10-17
Fix news dashboard

matmair:
This PR fixes an error with the news dashboard item URL and converts the dashboard items to be typed - this should stop errors like this from sneaking in again.


SchrodingersGat:
Thanks for the fix!

PR #5749, opened 2023-10-18
[PUI] Raise translation/language support to CUI level

### LINKED ISSUE #5729: [PUI] Expose all language options
Currently language options for the react/platform interface are artificially limited. We need to expose all of the locale options

matmair:
This PR raises the theoretical language support of PUI to CUI levels.
Closes #5729

```[tasklist]
### Changes
- [x] Add all CUI languages to PUI
- [x] Add translation stumps
- [x] Exclude locales from deepsource checks
- [x] Changes the language selction to use "full" language names
```


src/frontend/src/contexts/LanguageContext.tsx
```diff
@@ -1,13 +1,76 @@
...
 import { I18nProvider } from '@lingui/react';
 import { useEffect } from 'react';
 
 import { api } from '../App';
 import { useLocalState } from '../states/LocalState';
 
 // Definitions
-export type Locales = 'en' | 'de' | 'hu' | 'pseudo-LOCALE';
-export const languages: Locales[] = ['en', 'de', 'hu'];
+export type Locales =
```
wolflu05, line 10: Why not use keyof languages instead of redecorating them at several different places? If a language is added now, we need to carefully search everywhere where we need to add the new language in be and fe at several places

matmair: I think I am not experienced enough in ts to understand what you mean. Could you provide a sample of how that would look?

wolflu05: Something like 
```ts
export type Locales = keyof languages
```
Typescript [documentation](https://www.typescriptlang.org/docs/handbook/2/keyof-types.html)

wolflu05: But this would require, that you remove the typing from the languages obj, so the object type is interfered

SchrodingersGat:
A question related to Crowdin, maybe you know @matmair - once we expose these new translations, will strings that have already been translated for the existing django interface be automatically translated? Or, will the community have to re-translate the same strings again?

If the second, perhaps we should write a script to scrape existing translations (where identical)

matmair:
@SchrodingersGat I already transferred all translated strings into the new PO files. They should be automatically mapped by Crowdin during the import that way.

src/frontend/src/contexts/LanguageContext.tsx
matmair: Implemented, thx for the suggestion

SchrodingersGat:
@matmair LGTM - happy to merge once checks pass

src/frontend/src/contexts/LanguageContext.tsx
```diff
@@ -1,13 +1,45 @@
...
 
 import { api } from '../App';
 import { useLocalState } from '../states/LocalState';
 
 // Definitions
-export type Locales = 'en' | 'de' | 'hu' | 'pseudo-LOCALE';
-export const languages: Locales[] = ['en', 'de', 'hu'];
+export type Locales = keyof typeof languages | 'pseudo-LOCALE';
+
+export const languages: Record<string, string> = {
```
wolflu05, line 12: You need to remove the typing here, so that autocompletion works for the languages obj.
```suggestion
export const languages = {
```

wolflu05:
Maybe we need to document somewhere what needs to be changed to add one new language, as this are now 3 places as far as I can see.

src/frontend/src/contexts/LanguageContext.tsx
matmair: How is autocomplete better then typesafety?

matmair:
> Maybe we need to document somewhere what needs to be changed to add one new language, as this are now 3 places as far as I can see.

I think that should be an extra PR.

src/frontend/src/contexts/LanguageContext.tsx
wolflu05: The type gets automatically interfered from the object definition, if you remove the typing. Basically like your definition, but without the more generic string typing for the keys but with all available keys.

matmair:
@SchrodingersGat is there anything else you would like to see implemented here?

SchrodingersGat:
Just tried out the netlify preview, getting some strange behaviour:

image

SchrodingersGat:
@matmair looks to be failing PUI build: https://app.netlify.com/sites/inventree-web-pui-preview/deploys/65360d6ce6ee7f0007f2a6f9

InvenTree/InvenTree/api_version.py
```diff
@@ -2,18 +2,11 @@
 
 
 # InvenTree API version
-INVENTREE_API_VERSION = 141
+INVENTREE_API_VERSION = 139
```
SchrodingersGat, line 5: @matmair looks like this PR is out of sync and will revert some other recent changes?

SchrodingersGat (changes requested):

InvenTree/InvenTree/api_version.py
matmair: All the different active branches are beginning to become a problem, cleaning up rn.

matmair:
@SchrodingersGat it has proven difficult to get the language selection itself translateable. I would merge it in this state and get the language selector itself tackled later.

SchrodingersGat (approved):

SchrodingersGat:
@matmair ok happy to merge. Sorry it looks like some conflicts from the translation merge have blocked this! If you can address, I'll merge ASAP

matmair:
@SchrodingersGat issued a fix, should work now.

PR #5759, opened 2023-10-20
[PUI] Render status labels

### LINKED ISSUE #5746: [PUI] Render "status" labels
Status values (i.e. StockItem.ATTENTION_NEEDED) are served via the API - can be cached by the client.

- Fetch from the server
- Use translated text in a "badge"

matmair:
This PR enables status label rendering for PUI as well as more generic status lookups.

Closes #5746 
```[tasklist]
### Tasks
- [x] Adds /generic/status/ endpoint to list all defined status models (including from plugins)
- [x] Adds /generic/status/<STATUSMODEL>/ endpoint to gather specific status models (this is a first step to possibly deprecate the per-model endpoints for status)
- [x] Adds global store for status labels
- [x] Adds render method for status labels
- [x] Resolves several TODOs for status label rendering
```


wolflu05:
How do you imagine this to work if custom statuses are implemented? I personally would implement the fetching from server now to seemlesly integrate custom statuses and in the future.

matmair:
That is nothing I have focused on till now @wolflu05. To work the type would need to be available when the UI is build so I do not think it would work.

wolflu05:
Why do you need the status codes as types? I mean you don't have a type for user groups too? And in the end, status codes won't be something different than just a user group, but coupled to some business logic.

matmair:
Not in the way I implemented it. You are welcome to look at the code (when it is finished) and make a PR if your use case is not covered.

matmair:
Not really sue what is happening here with the PUI test, I did not touch any part of that

SchrodingersGat:
> Not really sue what is happening here with the PUI test, I did not touch any part of that

that seems to be failing around 50% of the time

matmair:
It can not work the way a couple of things are working together up in the login flow right now, we have got to review changes to these parts better. The default hostlist is loading in an infinite loop and has been for a few days at least.

matmair:
@wolflu05 in fast testing this also seems to work with custom states. Depending on if/how we want to expose the UI components the labels should be very low effort for reusing.

matmair:
@SchrodingersGat this is ready for review - I think I should have found all status code placeholders, ping me if I missed one.

SchrodingersGat:
@matmair nice work, very clean implementation. It works well locally, but the labels do not render at all on the netlify preview. Might be related to the API query chaining?

I'm happy to merge this in now, but would be worth looking into potentially

matmair:
That is to be expected @SchrodingersGat, the preview runs against the demo server - ie. if there is an API change we can not preview it fully.

SchrodingersGat:
That makes sense! Should have thought of that

PR #5760, opened 2023-10-20
Fixed PlaygroundArea accordion behaviour

matmair:
All accordions opened when one was clicked when multiple playground areas were added.


SchrodingersGat:
Thanks 🙏

PR #5765, opened 2023-10-21
Frontend server settings

SchrodingersGat:
Adds more comprehensive settings method for setting frontend settings


InvenTree/web/templatetags/spa_helper.py
```diff
@@ -7,14 +8,10 @@
 from django.conf import settings
 from django.utils.safestring import mark_safe
 
-logger = getLogger("InvenTree")
+logger = getLogger("inventree")
```
matmair, line 11: Comment: This is breaking for ppl that use custom logging drivers

matmair (commented):
I think using „frontend“ has a potential to confuse users wether the old or new ui is meant. I get a feeling that the old UI will stick around for a bit so i think there should be a good distinction.

SchrodingersGat:
@matmair any comment on the currently failing `Deepsource:Python` issue (`mark_safe`)?

matmair:
Mark_safe is generally a security risk, I will review these specific instances

PR #5774, opened 2023-10-23
Part responsible owner

SchrodingersGat:
- Replaces https://github.com/inventree/InvenTree/pull/5060
- Fixes https://github.com/inventree/InvenTree/issues/5059


InvenTree/part/serializers.py
```diff
@@ -695,6 +696,11 @@ def get_starred(self, part):
         read_only=True,
     )
 
+    responsible = serializers.PrimaryKeyRelatedField(
+        queryset=users.models.Owner.objects.all(),
+        source='responsible_owner',
+    )
+
```
matmair, line 703: Comment: This should lead to a bumb of the API version, right?

matmair (approved):
lgtm

SchrodingersGat:
@miggland FYI this is how complex migrations should be handled. In separate sequential migrations:

1. Add the new field to the model
2. Perform custom data migration
3. Delete the old field

Data and schema migrations should not be performed in the same step. And it should be written in such a way that it is reversible if required :)

miggland:
Looks nice, thanks for fixing this. :+1:

SchrodingersGat:
No worries, it is a pretty tricky thing to get right!

PR #5776, opened 2023-10-23
Bump action versions

matmair:
Bump actions to remove the node12 version warnings in all recent workflow runs.


SchrodingersGat:
Thanks @matmair

PR #5777, opened 2023-10-24
Fix issues with running under Python 3.12

theacodes:
These two commit address issues encountered when running with Python 3.12.0. Notably, the changes remove the use of deprecated APIs related to import machinery.

(This is my first PR to this repo, my feelings won't be hurt if this doesn't get accepted or if you decide to make significant changes)

### DIFF
--- a/InvenTree/plugin/helpers.py
+++ b/InvenTree/plugin/helpers.py
@@ -7,6 +7,7 @@ import pathlib
 import pkgutil
 import sysconfig
 import traceback
+from importlib.util import module_from_spec
 from importlib.metadata import entry_points
 
 from django import template
@@ -156,9 +157,9 @@ def get_modules(pkg, path=None):
     elif type(path) is not list:
         path = [path]
 
-    for loader, name, _ in pkgutil.walk_packages(path):
+    for finder, name, _ in pkgutil.walk_packages(path):
         try:
-            module = loader.find_module(name).load_module(name)
+            module = module_from_spec(finder.find_spec(name))
             pkg_names = getattr(module, '__all__', None)
             for k, v in vars(module).items():
                 if not k.startswith('_') and (pkg_names is None or k in pkg_names):
--- a/InvenTree/plugin/helpers.py
+++ b/InvenTree/plugin/helpers.py
@@ -102,7 +102,7 @@ def handle_error(error, do_raise: bool = True, do_log: bool = True, log_name: st
 
 def get_entrypoints():
     """Returns list for entrypoints for InvenTree plugins."""
-    return entry_points().get('inventree_plugins', [])
+    return entry_points(group='inventree_plugins')
 # endregion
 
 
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@theacodes welcome! We have been following your kicanvas project for a while :) 

Have you done any specific testing with python 3.9, our lowest officially supported version?

theacodes:
Nope, not any specific testing on 3.9.

InvenTree/plugin/helpers.py
```diff
@@ -102,7 +102,7 @@ def handle_error(error, do_raise: bool = True, do_log: bool = True, log_name: st
 
 def get_entrypoints():
     """Returns list for entrypoints for InvenTree plugins."""
-    return entry_points().get('inventree_plugins', [])
+    return entry_points(group='inventree_plugins')
```
SchrodingersGat, line 105: This change is failing tests for 3.9 - either we'll need to check the python version or handle the excpetion.

Do you have a link to the change in 3.12 which breaks this behaviour?

SchrodingersGat:
> Nope, not any specific testing on 3.9.

Our CI has done the heavy lifting for you!

InvenTree/plugin/helpers.py
theacodes: Yeah, it's a little hard to find, but here's the relevant bit (found [here](https://docs.python.org/3/library/importlib.metadata.html#entry-points)):

> The “selectable” entry points were introduced in importlib_metadata 3.6 and Python 3.10. Prior to those changes, entry_points accepted no parameters and always returned a dictionary of entry points, keyed by group. With importlib_metadata 5.0 and Python 3.12, entry_points always returns an EntryPoints object. See [backports.entry_points_selectable](https://pypi.org/project/backports.entry-points-selectable) for compatibility options.

Parsing that out, it looks like there's two options:

1. The most preferable option is to require `importlib_metadata >= 3.6`. This shouldn't require any changes to the actual code, as `entry_points()` will now behave the same across versions.
2. Alternatively, require `backports.entry-points-selectable` and use `from backports.entry_points_selectable import entry_points`.

matmair:
Works on my machine, I have approved the CI again - we should know more in a few minutes.

InvenTree/plugin/helpers.py
matmair: @SchrodingersGat I think we can resolve this now?

matmair (approved):
LGTM

SchrodingersGat:
`TypeError: entry_points() got an unexpected keyword argument 'group'` - still  some compatibility issues

SchrodingersGat:
From the [django release notes](https://docs.djangoproject.com/en/5.0/releases/4.2/):

image

3.12 support has only been introduced in django 4.2.8 anyway...

SchrodingersGat:
Closed by https://github.com/inventree/InvenTree/issues/6424

PR #5781, opened 2023-10-24
Add CJK fonts to void garbled characters when generating labels

liukangxu:


wolflu05:
Can you add an example how you experimented a misbehavior, I never experienced anything bad.

Could you analyze how much this dependency influences the docker image size?

SchrodingersGat:
@liukangxu some screenshots of the garbled printing would be great here

PR #5785, opened 2023-10-25
Fixes for PUI CI run/login timeouts

matmair:
This fixes recent CI problems and login timeouts related to recent changes in that are. See https://github.com/inventree/InvenTree/pull/5749#issuecomment-1778328406 or any other recent PUI CI run

### DIFF
--- a/src/frontend/src/main.tsx
+++ b/src/frontend/src/main.tsx
@@ -25,26 +25,37 @@ export const IS_DEV = import.meta.env.DEV;
 export const IS_DEMO = import.meta.env.VITE_DEMO === 'true';
 export const IS_DEV_OR_DEMO = IS_DEV || IS_DEMO;
 
+// Filter out any settings that are not defined
+let loaded_vals = (window.INVENTREE_SETTINGS || {}) as any;
+Object.keys(loaded_vals).forEach((key) => {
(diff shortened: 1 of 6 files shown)


matmair:
@SchrodingersGat this fixes teh PUI CI issues - the problem is a mixture of long API timeouts and reloads caused by that.

src/frontend/src/main.tsx
```diff
@@ -25,26 +25,37 @@ export const IS_DEV = import.meta.env.DEV;
...
-          demo: {
+          'mantine-u56l5jt85': {
             host: 'https://demo.inventree.org/',
             name: 'InvenTree Demo'
           }
         }
       : {})
   },
-  default_server: IS_DEMO ? 'demo' : 'localhost',
+  default_server: IS_DEMO ? 'mantine-u56l5jt85' : 'mantine-cqj63coxn',
```
SchrodingersGat, line 54: Why the pseudo-random map keys here?

matmair: Way less likely to be used by the server_list provided by the backend. Was a problem for me already when debugging this on my Q enviroment.

SchrodingersGat:
Thanks for the fix @matmair

PR #5789, opened 2023-10-25
Readme sponsors

SchrodingersGat:
Adds github action to automatically add project sponsors to README.md

Ref: https://github.com/JamesIves/github-sponsors-readme-action


.github/workflows/sponsors.yml
matmair, line 1: Suggestion (non-blocking): I would prefer if we pinned the exact action we use for security reasons.

matmair (approved):
Lgtm

PR #5794, opened 2023-10-26
Update sponsors.yml

SchrodingersGat:
Fix target branch


wolflu05:
Why do we need this deployment to gh pages?

SchrodingersGat:
> Why do we need this deployment to gh pages?

We don't I think I am doing it wrong. I just am trying to auto-update README.md with the sponsors :|

wolflu05:
Then you need to use a commit action, to commit those changes to the main branch or open a pr like we do with crowdin

SchrodingersGat:
Yep, you are right. I blindly followed their demo example 👎

PR #5795, opened 2023-10-26
Check minimum python version before launching

SchrodingersGat:
We have been getting a lot of "error reports" from users running python below minimum version. This PR runs a python version check at the first entrypoint, to ensure that the correct python version is running


wolflu05 (approved):
This is a good addition. Hopefully people will now fix the issues themselves. Maybe you could add a link to the docs?

matmair (approved):

Issue #3814, opened 2022-10-20
[FR] Breadcrumb/Burger Menu navigation improvement


Zennate:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

The breadcrumb/burger menu at the top of the main menu needs precision to be navigated correctly.
Example:
![grafik](https://user-images.githubusercontent.com/92428456/196888650-4d4fab48-9ef1-4f38-8628-678773afcdf6.png)
Clicking on the field of Category A reveals Sub-category B.
Clicking on the names of Category A or B loads the appropriate name.
Clicking on the field of Category B has no effect, despite the changed cursor.

Additionally exiting this menu without switching can only be achieved by clicking the button again. This can be irritating when trying to traverse the menu quickly on larger displays.

### Suggested solution

Expanding the \<a\>-tag of categories that do not have a sub-category to cover the entire field would make it easier to select the desired category. The behaviour of Category A should probably not be changed, since these categories are most likely used for organizing sub-categories.

Exiting the menu could alternatively be achieved by clicking an empty space that is not part of the menu itself. I'm unsure if this would need a seperate FR, since it affects the same menu and functionality.

### Describe alternatives you've considered

An alternative to using this menu altogether would be a mixed view in the main window, where you could see sub-categories and parts contained in your selected category in one list, instead of switching between categories and parts using the left-hand menu. This would probably require more effort overall though.

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
@Zennate thanks for reporting. Would you like to contribute this enhancement yourself?

matmair:
@SchrodingersGat I would recommend moving this to 0.10.x and closing 0.9.x for new issues if not related to bugs.

Zennate:
@SchrodingersGat thanks for the quick response. I'm sadly out of my depth with the current codebase, so I'll stick to providing feedback for now.

matmair:
TBH any work done on this would probably be lost in the transition to a framework as it would probably be tailored to the current jquery way of doing things.
If somebody really wants that change they can send a patch - I have no objection to changing this.  Changing the tree menu while keeping it somewhat performant (the current solution is by no means fully optimised) could represent a significant amount of work as it took several iterations to come to the current one.

PR #5800, opened 2023-10-27
[React] Tree navigation

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/3814

This PR introduces a left-hand pop-out navigation tree nav drawer for:

- Part Categories
- Stock Locations

Currently, just a skeleton / placeholder, no content!

### Design Reference

image

### Working Example

image
image

### Replaces

image

(Gross! No one likes this)

@wolflu05 I believe you had a good react library for rendering tree data, perhaps?


wolflu05:
Yes, I found this library a while ago, this looks pretty promising. They have an example with about 2e6 nodes in the tree, and it works pretty smooth. Under the hood, they use react virtualized which is an library for virtualization, meaning not all elements are Dom elements, only the ones you are seeing and a bit more. And then if you scroll tree out of the window, they get removed and ok the other side appended. https://github.com/diogofcunha/react-virtualized-tree

SchrodingersGat:
@wolflu05 that library looks interesting! For now, (mostly as a POC) I've used a much simpler library that works directly with the existing API, and can handle a flat dataset.

wolflu05:
Do you checked how performant this library works when your tree has >1e4 nested locations, with also maybe >100 in a single location?

matmair (commented):
Simpler than I thought. In my short testing, I did not see performance problems but @wolflu05 point is valid, we had perf issues with the old implementation.

SchrodingersGat:
I have not tested with a large dataset. Would we need to paginate the API query in such a case? Or can we wait for loading a large initial query? I will fully support optimizing for large data.

I'm going to merge this in and keep moving but if we see performance or usability issues am happy to reassess

PR #5806, opened 2023-10-28
Improve deletion behaviour for InvenTreeTree model

SchrodingersGat:
This PR solves a long-running bug which I have only now tracked down, which can occur when deleting a tree element (such as stock location, or part category). Additionally it provides a significant improvement in behaviour, removing recursive function calls and running bulk database queries to reduce the O() number of database operations required

This means that updating a complex part or stock tree will be much faster when there is a significant number of descendant nodes

The "bug" being fixed here is that the recursive order of operations, combined with atomic transactions, means that the database records are not always updated in the right sequence, and can result in a thrown `IntegrityError`

- Remove recursive call to function
- Handle database operations as bulk queries
- Ensure child nodes have their pathstring updated correctly
- Remove old `receiver` hook
- Refactor StockLocation.delete method
- Refactor PartCategory.delete method
- Atomic transactions potentially problematic here

Existing unit tests cover tree element deletion, and should be mostly sufficient here, with some small tweaks

### DIFF
--- a/InvenTree/InvenTree/models.py
+++ b/InvenTree/InvenTree/models.py
@@ -580,6 +580,8 @@ class InvenTreeTree(MPTTModel):
         parent: The item immediately above this one. An item with a null parent is a top-level item
     """
 
+    # How items (not nodes) are hooked into the tree
+    # e.g. for StockLocation, this value is 'location'
     ITEM_PARENT_KEY = None
 
     class Meta:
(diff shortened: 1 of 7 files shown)


matmair (approved):
Looks fine to me. We should run this for a bit in the demo though to be sure no unforeseen errors occur.

PR #5819, opened 2023-10-30
Recursive delete fix fix

SchrodingersGat:
Attempting to fix bug introduced in recent PR which only seems to be an issue for mysql backend.

Ref: https://github.com/inventree/InvenTree/pull/5806

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -274,7 +274,6 @@ jobs:
     runs-on: ubuntu-20.04
 
     needs: [ 'pep_style', 'pre-commit' ]
-    if: github.event_name == 'push'
 
     env:
       # Database backend configuration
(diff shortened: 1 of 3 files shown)


.github/workflows/qc_checks.yaml
```diff
@@ -274,7 +274,6 @@ jobs:
     runs-on: ubuntu-20.04
 
     needs: [ 'pep_style', 'pre-commit' ]
-    if: github.event_name == 'push'
```
wolflu05, line 277: @SchrodingersGat I guess you don't wanted to remove that for merge?

SchrodingersGat: The various backend tests don't take that long any more, I think it is worth checking that the supported databases are tested on PR as well as push

wolflu05: Ok

PR #5821, opened 2023-10-30
Part table extra options

SchrodingersGat:
Add extra stock information as a drop-down in the part table


matmair:
@SchrodingersGat that renderer should probably be it's own function/component so it can be reused in other places

SchrodingersGat:
> @SchrodingersGat that renderer should probably be it's own function/component so it can be reused in other places

Would be good to factor this out into its own component

PR #5822, opened 2023-10-30
[PUI] Settings simplification and restructure

matmair:
This PR:
- merges the `Profile` page into the user settings
- adds a specific primary frontend route for server settings
- adds user account controls
- adds user security settings (Email, SSO, MFA)
- all the necessary API endpoints to view them

grafik

In a follow-up PR necessary API changes and a UI for user/group management (no permission managment is slated at this time) will follow; As we are not using the sessions framework for the new UI, I will not implement it here.


SchrodingersGat:
@matmair looks good :)

matmair:
@SchrodingersGat this is ready for review; I had to add a few new APIs to get it going, unit test for those are tbd

SchrodingersGat:
@matmair looking really great here.

Some small notes below.

image

Also, are we going to display (maybe somewhere under the system settings) which SSO backends are available, and allow configuration? Or will that be another PR you think

matmair:
grafik
The system checks if sso is enabled; I think that should be enough:
grafik

grafik

If do not have any plans to touch SSO backends. Supporting the 90+ backend options correctly seems like a lot of added work for a small one-time interaction per instance.

matmair:
The texts follow the current ones so that a simple 100% TM match from Crowdin should cover translations (if already present)

SchrodingersGat (approved):

SchrodingersGat:
Great :)

SchrodingersGat:
Thanks for this great update @matmair :)

PR #5848, opened 2023-11-03
Added example with user data input to the panel mixin doc

SergeoLacruz:
Hopefully without fails...


docs/docs/extend/plugins/panel.md
```diff
@@ -160,3 +161,93 @@ Please have a look at the css files for more options. The last line renders the
...
+define the data types. No regular expressions. The URL takes two parameters,
+layer and size, and passes them to the python function SelectCompany for further processing.
+Now the html template:
+
+```html
+{% raw %}
+<script>
+async function example_select(){
+    var layernumber = parseInt(document.getElementById("layer_number").value)
+    var size = document.getElementById("string").value
```
wolflu05, line 226: I would not recommend using var

docs/docs/extend/plugins/panel.md
```diff
@@ -160,3 +161,93 @@ Please have a look at the css files for more options. The last line renders the
...
+    response = await fetch("{% url 'plugin:examplepanel:transfer' '999999' 'Size' %}"
```
wolflu05, line 227: Following conventions, get requests shouldn't mutate anything. Get requests are for getting data requested by a client from the server. For triggering an action, post requests should be used.

docs/docs/extend/plugins/panel.md
```diff
@@ -160,3 +161,93 @@ Please have a look at the css files for more options. The last line renders the
...
+<form> <input id="layer_number" type="number" value="2"> </form>
+Size of Board in mm
+<form> <input id="string" type="text" value="100x160"> </form>
+
+<input type="submit" value="Save" onclick="example_select()" title='Save Data'>
+{% endraw %}
+```
+
+The HTML defines two forms for user input, one number and one string. Each
+form has an ID that is used in the java code to get the input of the form.
```
wolflu05, line 244: please use javascript or js. Java is something completely different.

docs/docs/extend/plugins/panel.md
```diff
@@ -160,3 +161,93 @@ Please have a look at the css files for more options. The last line renders the
...
+        return panels
+
+    def setup_urls(self):
+        return [
+                path("example/<int:layer>/<path:size>/",
+                    self.SelectCompany, name = 'transfer'),
+        ]
+
+# Define the function that will be called.
+    def SelectCompany(self, request, layer, size):
```
wolflu05, line 208: Following python conventions, variables and function names should be named in lower_snake_case. PascalCase is used for class names.

docs/docs/extend/plugins/panel.md
```diff
@@ -160,3 +161,93 @@ Please have a look at the css files for more options. The last line renders the
...
+                              .replace("999999", layernumber)
+                              .replace("Size", size)
+                          );
+}
+</script>
+
+Number of Layers
+<form> <input id="layer_number" type="number" value="2"> </form>
+Size of Board in mm
+<form> <input id="string" type="text" value="100x160"> </form>
```
wolflu05, line 237: Why do you use two forms here?

wolflu05 (commented):
Thank you for your pull request. Adding an example about how to handle user input that should be sent to the server is a great idea. But I'm not really liking the implementation that uses a GET request to send data. In my opinion this is bad practice and therefore we shouldn't recommend that to the users. I would suggest using a POST request for that. There is also already a [helper method](https://github.com/inventree/InvenTree/blob/ffda7002443d4c394bab37846703184584986e7d/InvenTree/templates/js/translated/api.js#L99-L133) in CUI for doing POST requests that has error handling already builtin.

docs/docs/extend/plugins/panel.md
SergeoLacruz: Thanks for the useful hints. JS is not my language...
I think a addressed all of them.

wolflu05:
Sorry, I haven't clearly said what I actually meant. I meant using post requests and sending the data in the post body instead of in the url. Imagine you have a description field, then the url would also need to be encoded and that gets pretty messy.

SergeoLacruz:
I understand. Transferring the data e.g. in a json container is a better way when there are more fields. I wanted to keep this example simple with just one or two parameters. The json stuff will follow  in the next chapter. But we can also put it here when you prefer.

SchrodingersGat:
@wolflu05 would you like to see any more changes on this?

wolflu05:
I actually don't care how you do it. Do it like you prefer, it looks really simple for beginners for sure. But if you don't use the body, you could actually also use get to keep it simple. But be aware that this should not work with special chars in the inputs, and the url can get pretty long for large inputs. Seems like the other team members are fine for now with this changes. This were just some suggestions how I would have done it.

PR #5852, opened 2023-11-03
Small size optimisation for devcontainer

matmair:
This PR cleans up a few unused files in the devcontainer.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -37,11 +37,11 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive && \
 # RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
 
 # Update pip
-RUN pip install --upgrade pip
+RUN pip install --no-cache-dir --upgrade pip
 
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -31,17 +31,18 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive && \
     default-libmysqlclient-dev mariadb-client \
     # LDAP support
     libldap2-dev libsasl2-dev && \
-    apt-get autoclean && apt-get autoremove
+    apt-get autoclean && apt-get autoremove && \
+    rm -rf /var/lib/apt/lists/*
```
wolflu05, line 35: Why do we actually remove the package lists? Doesn't that mean if we want to install some apt packages for testing, e.g. the cups headers or the ldap package I have to run apt update before?

matmair: In most use cases you won't need the lists and running an apt update before installing is always recommended (see recent deployment problems with our own packages with an old access token in the apt lists)

wolflu05: Ok. Makes sense

PR #5855, opened 2023-11-03
[PUI] Small nav additions

matmair:
This PR:
- Adds more nav items (everything that is in the header) 
- changes the action on the InvenTree logo to open nav instead of going home


SchrodingersGat:
Nice addition :) Can I request some small aesthetic tweaks:

- Make the elements larger, with more padding. they are hard to click!
- Add an icon to each one, similar to what we have on the nav bar in CUI

matmair:
I can add the padding but I do not really see the point with icons, we do not have them in the header nav either

SchrodingersGat:
@matmair happy to merge in now, can continue to tweak as we go :)

PR #5856, opened 2023-11-03
Use gpg for install script instead of apt-key

matmair:
This PR replaces the deprecated and warning-raising apt-key with the more modern way of using armoured gpg keys for the installer script.


SchrodingersGat (approved):

SchrodingersGat:
@matmair happy to merge?

matmair:
@SchrodingersGat yes

PR #5862, opened 2023-11-04
[PUI] Add user managment to frontend

### LINKED ISSUE #4962: [FR] Expose user / group management to "Settings" page
Currently user and group management needs to be done by the low-level "admin" page. 

In general, it would be good to move any required activity away from that page - including user / group management.

Reference: https://github.com/inventree/InvenTree/issues/3978

matmair:
This PR adds user managment to PUI. Included changes:
- Introduction of the "admin center" - a place where the Django admin fulfilled administrative functions can be mapped to; in a follow-up, I plan to add more permissions so that users that are not staff can be explicitly allowed to manage certain aspects
- AllowPanelGroups without collapsible control
- Adds user and group management section to admin center

Todo: 
- Adds API for user and group management

Closes #4962

### DIFF
--- a/src/frontend/src/components/nav/MainMenu.tsx
+++ b/src/frontend/src/components/nav/MainMenu.tsx
@@ -5,6 +5,7 @@ import {
   IconLogout,
   IconPlugConnected,
   IconSettings,
+  IconUserBolt,
   IconUserCog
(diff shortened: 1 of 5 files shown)


SchrodingersGat:
LGTM as a starting point :)

PR #5865, opened 2023-11-05
Expose API version descriptors for admins via API

matmair:
This PR adds a new endpoint to retrieve the API version texts and unit tests for it.


InvenTree/InvenTree/api.py
```diff
@@ -57,6 +58,18 @@ def get(self, request, *args, **kwargs):
...
+class VersionTextView(APIView):
+    """Simple JSON endpoint for InvenTree version text."""
+
+    permission_classes = [
+        permissions.IsAdminUser,
+    ]
+
+    def get(self, request, *args, **kwargs):
+        """Return information about the InvenTree server."""
+        return JsonResponse(inventreeApiText())
```
wolflu05, line 70: Question: I asked me this a while ago, shouldn't I use the Response from DRF, that automatically is a json response? Or should I use the response from Django is such situations?

matmair: You can use either

wolflu05:
Interesting, Shouldn't the coverage normally be 100% for strings that are defined on the global scope, if that file is imported somewhere?

matmair:
Yes @wolflu05 but there seems to be a long known problem/behaviour that only the first line is reported as covered - still thinking about how to tackle that if this solution is not working

wolflu05:
Ah thats bad. But this no cover comment in the first line doesn't look great, maybe we could deactivate coverage for that file, and factor everything else out?

matmair:
> Ah thats bad. But this no cover comment in the first line doesn't look great, maybe we could deactivate coverage for that file, and factor everything else out?

I find that suggestion by far less clean than excluding certain lines.

SchrodingersGat (approved):

matmair:
I have removed the coverage things altogether, some larger things should be investigated and solved consistently

matmair:

SchrodingersGat:
Thanks @matmair , LGTM

PR #5875, opened 2023-11-06
Added first UI components for user managment

matmair:
This PR adds a section for user managment in PUI. This does not add permission mangement for groups, still searching for a nice UI pattern for that. Also fixes the settings header refactor. 
Ref #4962
grafik


wolflu05:
I tried the preview now to see how the admin center goes. Looks already really promising. But one think I noticed, I need to scroll a bit to navigate on that page. Wouldn't it make sense to move the groups management onto another tab on the left in this really empty sidebar? I don't have a large instance, but I'm not sure how good this tiny user table is scaling for really large instances. Extracting everything group related onto a new panel would enable to make the tables larger in height.

matmair:
@wolflu05 groups and users are inheritly connected, spliting them up would be inefficient in practical usage. Those features are not all there yet so maybe it is not clear.
In my experience administering large user bases via webinterfaces I seldom need to see more than a few users. If I need to edit one, I use seach/filter. If I need to get a list, I download one or look at the total records.

wolflu05:
Using a drawer for the user detail is a good idea. I have the same struggle with machines. But wouldn't it make sense to somehow have the user id in the url for a detail page? Same would be helpful for plugins as well, as then it would be ease to link to a plugin from the machine detail page or share the link with others. I think such states would make sense to encode in the url path.

matmair:
@wolflu05 that is currently not planned from my side but would certainly be nice. I had something similar already implemented in the settings before they got refactored.

SchrodingersGat:
image

Getting some js errors

matmair:
@SchrodingersGat the API is changed so this will not work on the netlify preview

SchrodingersGat:
Of course... I'll check it locally :)

SchrodingersGat:
@matmair I am happy with this as a starting point :) Would be great to extend the available user actions to perform tasks such as:

- Change superuser / admin status
- Enable / disable each user
- Change which group(s) a user belongs to
- etc

matmair:
@SchrodingersGat click on a user line - you can do everything you asked in the user drawer. Those actions used to be row actions but that became to unclear with the latest refactor and removal of texts.

SchrodingersGat:
Oh I see, missed that one

Issue #4936, opened 2023-05-31
[BUG] Error in display of SVG when creating reports - Weasyprint will need to be upgraded


miggland:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When creating a report with an SVG embedded in it via a link, there is a display error. It's our logo, and half of it is missing.

When I run Weasyprint 59.0 with the same source code on a different computer, the PDF looks like it should.

I believe this is caused by a bug in Weasyprint )https://github.com/Kozea/WeasyPrint/issues/1864), which has been fixed in version 59.0.

At the moment, we can not update the version of Weasyprint to 59.0 because of an incompatibility with `django-weasyprint`, as described here: 
https://github.com/fdemmer/django-weasyprint/issues/71

Until [this PR](https://github.com/fdemmer/django-weasyprint/pull/70) is merged in `django-weasyprint`, weasyprint can not be updated. For now, there is not InvenTree can do.

Workarounds:
- Do no use <use ..:> tags in SVG
- Convert SVG to another format, eg PNG

### Steps to Reproduce

1. Make SVG with `<use ...>` tags
2. Use SVG in report template, eg with:
```
@top-left {
                width : 100%;
                height : 17mm;
                background-image: url("Logo.svg");
}
```
3. Create a report in InvenTree
4. Look for parts of SVG with `<use ...> ` in the PDF vain

### Expected behaviour

Same as with 59.0

### Deployment Method

- [ ] Docker
- [ ] Bare metal

### Version Information

Current master has:
`weasyprint==54.3`

Bug remains with
`weasyprint==58.0`

matmair:
Thanks for the report! We are pinned to 54.3 due to problems between django-weasyprint and newer weasyprint that popped up.

How much of an impact has this error? I personally do not use svgs that are affected but sounds like something pretty severe. Do you have any idea how widely used `use` is?
From a quick glance at django-weasyprint it seems like a monkey patch would be pretty simple/small.

miggland:
In my limited understanding, `use` is probably applied more often than necessary by drawing programs.

The impact is big if the SVG is affected, but since there is a workaround, it can be handled. Since there seems to be a fix on the way for `django-weasyprint`, I would wait for that and then check if the other problems are solved.

The issues with merging PDF pages should be fixed in the newest Weasyprint. The "good reason to update" may be this issue with SVGs.

Relevant issues, for reference:
[...]

PR #5885, opened 2023-11-08
Bump weasyprint

matmair:
This removes the pin as the issue that caused the pin seems to be resolved in django-waesyprint.
Might also close https://github.com/inventree/InvenTree/issues/4936, pending confirmation by @miggland 

Closed https://github.com/inventree/InvenTree/issues/4936


miggland:
This issue was blocking using an upgraded `django-weasyprint`: https://github.com/fdemmer/django-weasyprint/issues/71 . The fix was merged (https://github.com/fdemmer/django-weasyprint/pull/70), but not released?

Either way: if Weasyprint can be successfully upgraded, #4936 can be closed.

matmair:
You are right @miggland, we will have to push this

SchrodingersGat:
The fix in django-weasyprint has actually been merged, but the release was not pushed to GitHub:

https://pypi.org/project/django-weasyprint/

miggland:
Well spotted @SchrodingersGat  -  but as of 13 hours ago, that's been corrected as well it seems!

SchrodingersGat:
@miggland yes I reached out to the maintainer of the project :)

SchrodingersGat:
It appears that the rendered PDF does not contain the exact same information. If you are able to test that the generated PDF files are still good, we can adjust the single failing test

miggland:
I have tried out (again) with the dev branch: SVGs don't show up.
With this branch, they look correct! All of the PDFs I've created in my test setup look fine. I used the standard demo data and made a few different labels/reports (also on sheets, or concatenated reports). The only thing I changed was to add an example with the SVG.

So: this seems to fix #4936, and looks good otherwise.

The test: I guess a more robust test would be to try and open the PDF with some tool (eg PikePDF), eg:
`pdf = pikepdf.open(labelpath)`
which gives exception `pikepdf.PdfError` if there is something wrong with PDF:

Alternative: pdftotext:
```
with open(labelpath, "rb") as f:
    pdf = pdftotext.PDF(f)
    filetext = " ".join(pdf)  # Join for multiple pages
```
But that will only work if there is actually text in the PDF..
   
Of course, both require installation of extra packages..

matmair:
I am not sure adding a dependency is great but I also can not find a better solution, pretty much on the same page as @miggland.

I did not work with pikepdf - which is smaller than what I tried. I will adapt this PR to add a test based on it.

matmair:
@miggland I added a co-author credit for you as the change is basically what you suggested above

matmair:
The order of objects seems flaky between DBMS, a fix will follow shortly.

matmair:
@SchrodingersGat this would be ready for review

SchrodingersGat:
Thanks @matmair

PR #5889, opened 2023-11-08
Bumped docker base image

matmair:
Bumps Python from 3.10 to 3.11 - which brought a few performance improvements with it

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@
 # - Runs InvenTree web server under django development server
 # - Monitors source files for any changes, and live-reloads server
 
-FROM python:3.10-alpine3.18 as inventree_base
+FROM python:3.11-alpine3.18 as inventree_base
 
 # Build arguments for this image
 ARG commit_hash=""


wolflu05:
We already were on python 3.10, didn't know? 

1. have you checked that this is the newest alpine version?
2. wouldn't it make sense to bump the base for the devcontainer too?

matmair:
We have been for a while now, Python seems to get more effitient with every version.
1) The latest one available for this version
2) The devcontainer uses the debian variants and is configurable by the user so I do not think we need to change this. The benefits for the few dev instances are probably small

SchrodingersGat:
@matmair did you build this locally? Looks like there is an issue with this:

https://github.com/inventree/InvenTree/actions/runs/6844628610/job/18608743972

SchrodingersGat:
I think it would be worthwhile to update our CI workflow, to build the docker image on a PR *if* the dockerfile has been changed

SchrodingersGat:
Reverting in https://github.com/inventree/InvenTree/pull/5905 until we get this sorted out

matmair:
I have been running this branch in my CI setup for a few days. Seems like one of the libaries I yank out creates a problem

SchrodingersGat:
Might be related to setuptools - the docker build action currently is pinned at a specific version because of a previous issue

matmair:
Seems to be a more involved problem from what I see in #5909. Iam more and more glad that I only use pgsql, maria/mysql seems to be a constant source of problems with upgrades.

PR #5894, opened 2023-11-09
Model render refactor

SchrodingersGat:
Removes duplicate model rendering code in the new react interface


matmair:
To me it seems we lose functionality with this refactor. My renderers had the option to link to the object they were rendering while yours were just text/images. This greatly reduces the clicks needed to get to a specific object - I think that should be reintroduced.

SchrodingersGat:
@matmair happy to add that in. But not happy for two ways to render a single model! If you think this has rescued functionality please make a PR :)

matmair:
Yeah most of my WIP UI stuff seems to break after this, I will have to read the diff a bit deeper an reintroduce a few things

wolflu05:
Yes, the api changes will also break my WIP things.

matmair:
this removed so much of the interactivenes of renderes, I will revert this for my code base and work with the old variant for now to meet my deadlines

PR #5897, opened 2023-11-10
Make modals/forms more reactive

wolflu05:
This PR aims to make the modals and forms framework more reactive.

1. Separates ApiForms from modals so that they can be used on their own
2. Makes it possible to hook into the modal state and build some custom reactive things

This is already working, but it is a bit slow. I think this is caused by multiple reasons. The biggest is, that Form fields get passed every form state as props. That means they rerender on every change. But actually that is not needed. Because e.g. changing the value of a text input should not rerender the part category select input. The part category select input should only rerender if its definition changes. (This can be easily seen by placing `console.log("Rerender", definition.label)` inside of the ApiFormField)

**EDIT**: I now discovered, that the preview production build doesn't has that much of performance issues, and checking out the latest master in development has even more performance problems on my side.

```[tasklist]
### Todo
- [x] fix `onValueChange` issue
- [x] migrate to react-hook-form
- [x] improve initial data fetching query
- [x] Loading existing value into Related model field does not work
- [x] Remove unnecessary things from onValueChange callback parameters
- [x] use isDirty state for disabling the submit button
- [x] implement `nested object` field
- [x] ~~implement `dependent field` field~~ (Printing dialog needs to be implemented first)
- [x] validate that all fields work, especially a choices field
- [x] fix TS errors
- [x] Check TODO: comments
```


SchrodingersGat:
@wolflu05 nice, I'll check this out next week - I am away for the weekend :)

SchrodingersGat:
@wolflu05 very nice - I like the demo example in the "playground" area :)

wolflu05:
Thank you @SchrodingersGat that you like it. What do you suggest for going forward with this PR? Do you see some performance issues compared to the old implementation? (Make sure to only compare a dev version with a dev version and a build with a build, they have huge differences in my tests.)

SchrodingersGat:
I did not see any performance issues, I have not profiled each implementation. 

Do you have a set of features you still want to implement here?

wolflu05:
No, I have nothing else planned for this PR. Was just not sure if there is some more work required regarding performance. But if you say it's fine, then refactoring the field component to not rerender everytime would be another task.

What is your plan on using the new hook syntax? Should we refactor and remove the old implementation? Or should we just use the new hook implementation for the new forms, or only the ones require reactivity?

SchrodingersGat:
> But if you say it's fine, then refactoring the field component to not rerender everytime would be another task.

Would you be willing to take on this one? I think that you have a better idea where to look than I do.

> What is your plan on using the new hook syntax? Should we refactor and remove the old implementation? Or should we just use the new hook implementation for the new forms, or only the ones require reactivity?

Can you please submit a PR with an example of refactoring one of the existing forms? I would rather do them all "properly" and refactor what we have now, before we build out a lot more forms.

------------------

If you are happy with this, please mark as ready and I will merge

wolflu05:
> Would you be willing to take on this one? I think that you have a better idea where to look than I do.

I tried to eliminate unnecessary dependencies in my last PR, but it seems like the useForm hook has some issues from mantine, see this [discussion](https://github.com/orgs/mantinedev/discussions/1193). The best way to fix this issue and improve performance would be like they recommend in the discussion to switch to another form state library like `react-hook-form`. I previously already worked with that library.

> Can you please submit a PR with an example of refactoring one of the existing forms? I would rather do them all "properly" and refactor what we have now, before we build out a lot more forms.

Sure, I'll can refactor the old forms in this PR if you want.

wolflu05:
I now committed a first draft of switching to the react-hook-form lib. In my tests the performance now has drastically improved. I already planned the code so that the integration of the `nested object` and `dependent field` type become very easy. @SchrodingersGat Please let me know what you think about the move to another new form state lib. I attached two screen recordings which shown the performance improvements.

### Performance comparison

#### Old
https://github.com/inventree/InvenTree/assets/76838159/506a0a95-50a5-4411-9b83-59c7383479c1

#### New
https://github.com/inventree/InvenTree/assets/76838159/4d643332-3b0f-4774-853a-54a61504f253

SchrodingersGat:
@wolflu05 looking very nice :)

wolflu05:
@SchrodingersGat This would be ready for review now. I decided to implement the dependent field not yet, as 1. the diff is already large enough and 2. it would be better to build the print label dialog first. That would be the easiest think to test the dependent field then while implementing it. But that is a too big task for here.

I now switched to the already mentioned `react-hook-form` library which drastically improves performance. I already changed a few forms that required a touch because they were incompatible with the new approach of not passing the form  object down but rather using the powerful react features. For the rest of the forms I don't have the time now, because I want to finish the machine integration soon.

src/frontend/src/contexts/LanguageContext.tsx
```diff
@@ -45,11 +45,25 @@ export const languages: Record<string, string> = {
...
+      }
+    });
+
+    return () => {
+      isMounted.current = false;
+    };
   }, [language]);
 
-  return <I18nProvider i18n={i18n}>{children}</I18nProvider>;
+  return loaded && <I18nProvider i18n={i18n}>{children}</I18nProvider>;
```
matmair, line 66: Question: What is the state the page is left in if an error occurs while activating/mounting? Is it just an empty page then?

matmair (approved):
Amazing work; works very well on my laptop and iPad!

src/frontend/src/contexts/LanguageContext.tsx
wolflu05: Good catch, currently nothing will be shown. I'll change that.

SchrodingersGat:
After reviewing these changes in https://github.com/inventree/InvenTree/pull/5947 I am quite happy with this implementation - these are great forward steps :)

Please keep this great work coming @wolflu05

src/frontend/src/functions/forms.tsx
```diff
@@ -96,7 +187,28 @@ export function openModalApiForm(props: ApiFormProps) {
...
+        modals.close(modalId);
+      }
+    }
+  ];
+
+  const oldFormSuccess = props.onFormSuccess;
+  props.onFormSuccess = (data) => {
+    oldFormSuccess?.(data);
+    modals.close(modalId);
+  };
```
wolflu05, line 209: @SchrodingersGat To continue [this](https://github.com/inventree/InvenTree/pull/5947#discussion_r1399160566) conversation here. This is just a name, because I need to wrap the `onFormSuccess' function so that I can do something else and also call the callback provided in props. If I wouldn't save this in a variable, and just call it from inside, I would have an endless recursion.
image

PR #5902, opened 2023-11-12
[Doc] python example Part uploadImage correction

42CrMo4:
just a small correction to the python example.


SchrodingersGat:
Thanks for the fix!

PR #5907, opened 2023-11-13
Report image rendering fix

SchrodingersGat:
Adds ability to specify which image variation to use when rendering a "part_image" in a report.

If an uploaded image is very large this can cause the report size to bloat out.

- Can use "preview" image e.g. `{% part_image my_part preview=True %}`
- Can use "thumbnail" image  e.g. `{% part_image my_part thumbnail=True %}`
- Can also specify image width / height parameters

### DIFF
--- a/InvenTree/label/tests.py
+++ b/InvenTree/label/tests.py
@@ -154,8 +154,6 @@ class LabelTest(InvenTreeAPITestCase):
         self.assertIn(f"part: {part_pk} - {part_name}", content)
         self.assertIn(f'data: {{"part": {part_pk}}}', content)
         self.assertIn(f'http://testserver/part/{part_pk}/', content)
-        self.assertIn("img/blank_image.png", content)
-        self.assertIn("img/inventree.png", content)
 
     def test_metadata(self):
         """Unit tests for the metadata field."""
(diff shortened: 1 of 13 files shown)


InvenTree/label/tests.py
```diff
@@ -154,8 +154,6 @@ def test_print_part_label(self):
         self.assertIn(f"part: {part_pk} - {part_name}", content)
         self.assertIn(f'data: {{"part": {part_pk}}}', content)
         self.assertIn(f'http://testserver/part/{part_pk}/', content)
-        self.assertIn("img/blank_image.png", content)
-        self.assertIn("img/inventree.png", content)
```
matmair, line 158: Question: Why are those tests removed?

InvenTree/report/templatetags/report.py
```diff
@@ -184,13 +225,17 @@ def part_image(part: Part):
     Raises:
         TypeError if provided part is not a Part instance
     """
-    if type(part) is Part:
-        img = part.image.name
+    if type(part) is not Part:
+        raise TypeError("part_image tag requires a Part instance")
```
matmair, line 229: Suggestion: This error is not translated
```suggestion
        raise TypeError(_("part_image tag requires a Part instance"))
```

matmair (approved):
Looks good in general;

InvenTree/label/tests.py
SchrodingersGat: Replacing these now, they are outdated with the new approach

PR #5912, opened 2023-11-13
Added order cancel notifications

matmair:
This PR adds notifications when an order is canceld (additonally to the event already present).


SchrodingersGat:
@matmair nice addition. Can you please extend this to build orders too?

matmair:
I added it but IMO Orders and Builds should be handled seperate from each other, they are pretty different. Sales, Purchase and Return Orders could probably be substantailly reduced with some more generalisation.

SchrodingersGat:
@matmair nice work, thanks :)

PR #5922, opened 2023-11-15
Refactoring supplier barcode support

SchrodingersGat:
- Add a set of standard field name strings
- Map from custom fields to standard fields
- Helper functions for returning common field data
- Updated unit tests

### TODO

- [x] Get unit tests to pass still
- [x] Add unit tests for more barcodes (especially `30P` variant of Digikey Barcode)
- [x] Remove debug print statements


matmair (commented):
LGTM

PR #5837, opened 2023-11-02
Form improvements

wolflu05:
This PR adds a few things which improve the form framework. This is hopefully the last PR that is needed to continue #4824. 

Specifically the following is changed:
1. A new serializer field named `DependentField` is added. To test this new serializer field, the printing plugin code at the end can be used. This is actually a sample implementation of how I imagine the machine printing to work.
2. CUI now fully supports nested object serializer fields in the form framework.
3. CUI now supports the new `DependentField`

https://github.com/inventree/InvenTree/assets/76838159/cc0eae6e-ecb8-4922-a0bc-4b51026efc11

I actually tried to implement the nested serializer field in PUI, but I sadly have not found my way around the PUI form framework code, so this would be another task.

Please let me know what you think of that implementation and if there's anything I should change.
 
---

<details><summary>Sample printing plugin code for testing</summary>

```py
from django.utils.translation import ugettext_lazy as _
from InvenTree.serializers import DependentField
from label.models import LabelTemplate
from rest_framework import serializers

# InvenTree plugin libs
from plugin import InvenTreePlugin
from plugin.mixins import LabelPrintingMixin


class SettingsTestPlugin(LabelPrintingMixin, InvenTreePlugin):
    AUTHOR = "wolflu05"
    DESCRIPTION = "Label printing test plg"
    VERSION = "0.1.1"

    NAME = "LabelPrintingTest"
    SLUG = "labelprintingtest"
    TITLE = "Label printing test PLG"

    def print_label(self, **kwargs):
        return super().print_label(**kwargs)

    def print_labels(self, label: LabelTemplate, items: list, request, **kwargs):
        return super().print_labels(label, items, request, **kwargs)

    class PrintingOptionsSerializer(serializers.Serializer):
        machine = serializers.ChoiceField(choices=[
            ("a", "Machine A"),
            ("b", "Machine B"),
            ("c", "Machine C"),
        ])

        def get_machine_options(self, data):
            if data["machine"] == "b":
                return BMachineOptionsSerializer()

        machine_options = DependentField(depends_on=["machine"], field_serializer="get_machine_options")


class BMachineOptionsSerializer(serializers.Serializer):
    select_orientation_type = serializers.CharField()
    orientation = DependentField(depends_on=["select_orientation_type"], field_serializer="get_orientation_field")

    def get_orientation_field(self, fields):
        if fields["select_orientation_type"] == "1":
            return serializers.IntegerField()
        if fields["select_orientation_type"] == "2":
            return serializers.CharField()
[...]

PR #5924, opened 2023-11-15
CUI dependent/nested fields

wolflu05:
Sorry that my old code #5837, #5913 broke the forms. I really thought I checked it enough, but it appears otherwise. The was just one single issue with the code for forms that were using the old nested field implementation with the `__`. My plan was to now support both instead of refactoring all of them. The issue was that when using the old implementation the OPTIONS request returns a `nested object` field, but the field data doesn't contain it and that's why `optionsField` is undefined in L500. I thought that trying the part form is enough, but apparently it wasn't using any of those nested fields, only if I would have tried the duplicate part form. This PR includes the old code including the fix for this undefined problem. I now tested a bunch of other forms and they worked, hopefully all work now.


SchrodingersGat:
@wolflu05 thanks for the updates, it appears to work well now, with existing forms.

PR #5925, opened 2023-11-15
Factored typecasting out

matmair:
This PR factors the typecasting helper out so it can be used in other functions/plugins.


SchrodingersGat:
Thanks @matmair

PR #5927, opened 2023-11-16
Toggle full name display of users

### LINKED ISSUE #5222: [FR] Owner: Show User's full name instead of username

### Problem statement

Any Owner field tied to a user currently shows the username when printed in the GUI.

1. It's not always easily discernable who "PikachuLover5000" is
2. In conjunction with PO's, using "Attn: PikachuLover5000" in a report isn't very desirable either

### Suggested solution

By changing what value is returned based on Owner type, any "accidentally" bad usernames are filtered out of the GUI part, and only exposed to the "offending" user and admins.

I propose a very simple solution:
Return different model fields based on `owner_type`.

`owner.name` for Groups
`owner.full_name` for Users

Pretty much the above

### Examples of other systems

_No response_

- [X] I want to develop this.

LavissaWoW:
Would've closed #5222 if it was still open

### DIFF
--- a/InvenTree/users/models.py
+++ b/InvenTree/users/models.py
@@ -802,7 +802,6 @@ class Owner(models.Model):
 
     def __str__(self):
         """Defines the owner string representation."""
-        print("Setting is:", common_models.InvenTreeSetting.get_setting('DISPLAY_FULL_NAMES'))
         if self.owner_type.name == 'user' and common_models.InvenTreeSetting.get_setting('DISPLAY_FULL_NAMES'):
             display_name = self.owner.get_full_name()
         else:
--- a/InvenTree/users/models.py
+++ b/InvenTree/users/models.py
@@ -6,7 +6,7 @@ import logging
 from django.conf import settings
 from django.contrib import admin
 from django.contrib.auth import get_user_model
-from django.contrib.auth.models import Group, Permission
+from django.contrib.auth.models import Group, Permission, User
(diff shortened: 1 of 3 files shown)


InvenTree/users/models.py
```diff
@@ -785,10 +802,17 @@ def get_api_url():  # pragma: no cover
 
     def __str__(self):
         """Defines the owner string representation."""
-        return f'{self.owner} ({self.owner_type.name})'
+        print("Setting is:", common_models.InvenTreeSetting.get_setting('DISPLAY_FULL_NAMES'))
```
SchrodingersGat, line 805: Please remove this debug statement :)

LavissaWoW: Oops :>

SchrodingersGat:
@LavissaWoW thanks, looks good to me :)

wolflu05:
Wouldn't it make sense to also add this new setting to PUI so that they don't get out of sync?

SchrodingersGat:
@wolflu05 yep, that's a good point. @LavissaWoW would you be able to implement a similar change into the new react frontend?

matmair:
In this case, adding the key of the setting somewhere here should be enough BTW
https://github.com/inventree/InvenTree/blob/d30ff8a291d945c13984d6998dce67b4091839c4/src/frontend/src/pages/Index/Settings/SystemSettings.tsx#L47-L62

PR #5931, opened 2023-11-16
Only run sposnor action if on upstream repo

matmair:
The title says it all - currently the action is going to run on every fork till disabled.


SchrodingersGat:
Nice, thanks for that

matmair:
Bugged me on every one of my forks so more of a self-service ;-)

PR #5934, opened 2023-11-17
[PUI] Api state extension/cleanup

matmair:
This PR:
- adds the option to replace strings in the API lookup
- adds more usage of the pk parameter
- adds the pk to the user state (zustand) and fixes an error in the user account UI where that is needed

### DIFF
--- a/src/frontend/src/states/ApiState.tsx
+++ b/src/frontend/src/states/ApiState.tsx
@@ -192,7 +192,7 @@ export function apiEndpoint(path: ApiPaths): string {
 /**
  * Construct an API URL with an endpoint and (optional) pk value
  */
-export function apiUrl(path: ApiPaths, pk?: any, data?: any): string {
+export function apiUrl(path: ApiPaths, pk?: any, data?: Record<string, string | number>): string {
   let _url = apiEndpoint(path);
 
   // If the URL does not start with a '/', add the API prefix
--- a/src/frontend/src/states/ApiState.tsx
+++ b/src/frontend/src/states/ApiState.tsx
@@ -85,15 +85,15 @@ export function apiEndpoint(path: ApiPaths): string {
     case ApiPaths.user_sso:
       return 'auth/social/';
     case ApiPaths.user_sso_remove:
-      return 'auth/social/$id/disconnect/';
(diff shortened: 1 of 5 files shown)


wolflu05:
Thank you for crating this or, this was one of the things I wanted to push upstream next in a separate PR, I already implemented that in #4824. (But you could expect an other PR optimizing some types by adding generics and make that thing what you implemented work with the forms (an old implementation can already be found at #4824), but for that it would be good to base it on #5897)

src/frontend/src/states/ApiState.tsx
```diff
@@ -194,7 +192,7 @@ export function apiEndpoint(path: ApiPaths): string {
 /**
  * Construct an API URL with an endpoint and (optional) pk value
  */
-export function apiUrl(path: ApiPaths, pk?: any): string {
+export function apiUrl(path: ApiPaths, pk?: any, data?: any): string {
```
wolflu05, line 195: Why not type this as
```suggestion
export function apiUrl(path: ApiPaths, pk?: any, data?: Record<string, string | number>): string {
```
(Then the replacement value would need to be casted to a string using an f string, but that way we could also hand in some ids as numbers.)

matmair: Tbh my understanding of the typing system is still entry level

src/frontend/src/states/ApiState.tsx
```diff
@@ -206,5 +204,11 @@ export function apiUrl(path: ApiPaths, pk?: any): string {
     _url += `${pk}/`;
   }
 
+  if (_url && data) {
+    for (const key in data) {
+      _url = _url.replace(`:${key}`, data[key]);
```
wolflu05, line 209: To fix the issue you would need something like this, that was actually what I was trying to explain in my previous comment.
```suggestion
      _url = _url.replace(`:${key}`, `${data[key]}`);
```

matmair:
I think this is ready for merger; agree @wolflu05?

wolflu05 (approved):
Yes, LGTM

SchrodingersGat (approved):

SchrodingersGat:
Thanks @matmair

PR #5937, opened 2023-11-19
Refactor existing barcode API endpoints

SchrodingersGat:
- Expose fields using proper DRF serializers
- API endpoints are now self documenting
- Validation is handled by serializer models
- Serializers and endpoints are extensible
- Extended existing unit tests

@matmair this is a refactor of the existing barcode API endpoints to make use of the DRF serializers, for self-documenting endpoints. The functionality should be unchanged, however field validation is now handled by the serializer class(es) rather than being hidden inside the particular endpoint implementation.


matmair (approved):
LGTM

PR #5943, opened 2023-11-20
Fixes call to processField

SchrodingersGat:
- Previously used options.fields?
- Should use fields?

@wolflu05  can you please look over this? I believe that this is a bug introduced in your recent PR

### DIFF
--- a/InvenTree/templates/js/translated/forms.js
+++ b/InvenTree/templates/js/translated/forms.js
@@ -486,7 +486,15 @@ function extractNestedField(field_name, fields) {
  */
 function constructFormBody(fields, options) {
 
-    var html = '';
+    let html = '';
+
+    // Client must provide set of fields to be displayed,
+    // otherwise *all* fields will be displayed
+    const displayed_fields = options.fields || fields || {};
+
+    if(!options.fields) {
+        options.fields = displayed_fields;
+    }
 
     // add additional content as a header on top (provided as html by the caller)
     if (options.header_html) {
@@ -516,12 +524,10 @@ function constructFormBody(fields, options) {
     }
 
     for (const [k,v] of Object.entries(fields)) {
-        processField(k, v, fields[k]);
+        if (options.fields && k in options.fields) {
+            processField(k, v, options.fields[k]);
+        }
     }
-
-    // Client must provide set of fields to be displayed,
-    // otherwise *all* fields will be displayed
-    var displayed_fields = options.fields || fields;
 
     // Override default option values if a 'DELETE' form is specified
     if (options.method == 'DELETE') {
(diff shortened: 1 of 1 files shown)


wolflu05:
No, the idea was to assign the .child to the field in the form options which is passed everywhere in the form system. Because the .child was only available in the data got from the options request which is stored in fields. And because there are nested fields now we need to handle that recursively. Look at the Object.entries. This is taking the fields and looping over it.

SchrodingersGat:
> Do you experience some misbehavior?

Yes - if `options.fields` is not provided, this will crash. 

For example, in the "settings" window try to edit a setting such as `PART_NAME_FORMAT` - you will get an error.

You loop over the entries of `fields` but then apply to `options.fields` - one or the other needs to change here

wolflu05:
You're right if the field is empty, this won't work on the demo instance. Interestingly this also doesn't work entirely on my local dev instance no matter if the field is empty or not, maybe a caching instance. But I would fix this like the following. If there is no options.fields just set it to fields, because then every fields should be shown. And if we also have to deal with the case where neither `fields` nor `options.fields` is set, we set it to an empty object so it doesn't end with any errors. 
```diff
diff --git a/InvenTree/templates/js/translated/forms.js b/InvenTree/templates/js/translated/forms.js
index 652c744e1..dbc00f5c5 100644
--- a/InvenTree/templates/js/translated/forms.js
+++ b/InvenTree/templates/js/translated/forms.js
@@ -493,6 +493,14 @@ function constructFormBody(fields, options) {
         html += options.header_html;
     }
 
+    // Client must provide set of fields to be displayed,
+    // otherwise *all* fields will be displayed
+    const displayed_fields = options.fields || fields || {};
+    
+    if(!options.fields) {
+        options.fields = displayed_fields;
+    }
+
     // process every field by recursively walking down nested fields
     const processField = (name, field, optionsField) => {
         if (typeof optionsField !== "object") return;
@@ -519,10 +527,6 @@ function constructFormBody(fields, options) {
         processField(k, v, options.fields[k]);
     }
 
-    // Client must provide set of fields to be displayed,
-    // otherwise *all* fields will be displayed
-    var displayed_fields = options.fields || fields;
-
     // Override default option values if a 'DELETE' form is specified
     if (options.method == 'DELETE') {
         if (!('confirm' in options)) {
```

Sorry that this feature is causing so many issues now. I really hope this was the last issue, but luckily you spotted this issue before a release. For such things, typescript would help a lot to identify such issues if everything is typed correctly. The new frontend has chosen a good path :D .

SchrodingersGat:
@wolflu05 updated again, LMK what you think

wolflu05 (approved):
Yes, LGTM good extension with the added check. Thanks for fixing this.

Issue #5942, opened 2023-11-20
[Review] "full_name" for Part


SchrodingersGat:
After some simple profiling, I have found that the `full_name` Part field adds a significant overhead to the API. Even though it does not require many database hits, the rendering function for `full_name` does take a significant chunk of time to execute - and this time accumulates with the number of parts in the query.

### Query Results

Using the demo dataset, performed the following queries: 

| Query | Without `full_name` | With `full_name` |
| --- | --- | ---|
| `/api/part/?format=json` | 1.91s |  > 30 seconds |
| `/api/part/?format=json&limit=100` |  620ms | 1.34s |
| `/api/part/?format=json&limit=10` | 230ms | 350ms |

Note that in both cases the number of actual database hits was the same.

Clearly this needs to be optimized.

The original intent for `full_name` was to solve a specific problem on my end - it would be interesting to know how this field is *actually* being used. Perhaps removing this field is an option? (Although this would touch a *lot* of code. A cleaner option would be to improve the rendering efficiency)

matmair:
If I see it correctly there is a Template instantiation every time - maybe this can be solved by caching/reusing the Template?

We could write the full_name to the cache for example.

SchrodingersGat:
I think that caching the template caches the *compiled* template - including the contents? So for a list query set this does not help us...

Digging deeper, the default "part name" implementation uses the following template string:

```
{% raw %}
{{ part.IPN if part.IPN }}{{ ' | ' if part.IPN }}{{ part.name }}{{ ' | ' if part.revision }}{{ part.revision if part.revision }}
{% endraw %}
```

This takes around `5ms` to render.

If we used a simpler approach:

```python
entries = [self.IPN, self.name, self.revision]
return ' | '.join([e for e in entries if e])
```

This takes around `2us` to render.

I think that templates are just... slow?

I would propose that we consider replacing the "templated" implementation with a more pythonic approach, and then potentially expose this to the plugin system if people want more customisation?

wolflu05:
Maybe we need to improve the templating. I use jinja2 for my bulk plugin and hadn't any performance issues. What about trying that? The syntax is very similar, but if the template is compiled it is pretty fast.
```py
from jinja2 import Environment
import time

env = Environment(variable_start_string="{{", variable_end_string="}}")

def test(reps):
    t = env.from_string('{{a}} | {{b if b else ""}} part') # this compiles the template into a python function which takes the context as an argument and that's why it is pretty fast
[...]

PR #5946, opened 2023-11-20
Improvements for part.full_name

SchrodingersGat:
- Compile and cache the template
- Reduces typical render time from ~20ms to ~0.2ms

Closes https://github.com/inventree/InvenTree/issues/5942

### DIFF
--- a/InvenTree/part/helpers.py
+++ b/InvenTree/part/helpers.py
@@ -1,0 +1,64 @@
+"""Various helper functions for the part app"""
+
+import logging
+
+from jinja2 import Environment
+
+logger = logging.getLogger('inventree')
+
+
+# Compiled template for rendering the 'full_name' attribute of a Part
+_part_full_name_template = None
+_part_full_name_template_string = ''
+
+
+def compile_full_name_template(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


InvenTree/part/helpers.py
```diff
@@ -0,0 +1,64 @@
+"""Various helper functions for the part app"""
+
+import logging
+
+from jinja2 import Environment
```
matmair, line 5: Question: Is there a performance reason you are using Jinja2 instead of Djangos Template engine (which is based on Jinja2)? This will mean the syntax between reports and this field is different. I have not seen significant timing differences between Jinja2 and djangos template language when cached.

SchrodingersGat: @matmair this is specifically not to break current functionality. We have been using `jinja2` here for a long time - although I cannot remember *why*. 

However changing the rendering engine at this stage would break the current default template string - `jinja2` supports some tag attributes which the Django engine does not

SchrodingersGat: I'd be happy to look at switching to Django template engine in a separate PR

PR #5947, opened 2023-11-20
PUI general improvements

wolflu05:
This PR adds a few things I initially implemented in #4824, but want to push upstream before so that the diff is not that huge and not so much merge conflicts may exist:
- adds `pathParams` to `apiUrl`, forms, settings (I need them there because the machine settings url has path parameters)
- Show all setting keys if keys is not defined (required for machine-/plugin- settings)
- Make setting item list alternating shaded for better UX
   image
- Add type safety to tables using typescript generics
- Add `key` prop when rendering a list of components
- Fix `zustand/persist` deprecation warning for `getStorage`

Includes #5897


src/frontend/src/functions/forms.tsx
```diff
@@ -96,7 +191,28 @@ export function openModalApiForm(props: ApiFormProps) {
...
+    {
+      text: props.cancelText ?? t`Cancel`,
+      color: props.cancelColor ?? 'blue',
+      onClick: () => {
+        modals.close(modalId);
+      }
+    }
+  ];
+
+  const oldFormSuccess = props.onFormSuccess;
```
SchrodingersGat, line 209: Why is this called "old" here - is this intentional?

SchrodingersGat:
@wolflu05 some great improvements here :)

In particular the implementation of "nested" or "grouped" forms within the modals works really well!

wolflu05:
@SchrodingersGat all of the form stuff is from #5897 as this includes this branch, so it would be better to review #5897 first.

SchrodingersGat (approved):

SchrodingersGat:
LGTM, nice work again @wolflu05

src/frontend/src/functions/forms.tsx
wolflu05: I answered this on the other PR in this [comment](https://github.com/inventree/InvenTree/pull/5897#discussion_r1399180113)

SchrodingersGat:
@wolflu05 thanks

PR #5953, opened 2023-11-20
PUI: navigation/admin center improvements

wolflu05:
This PR improves the panels navigation and routers so that any selected panel is saved in the url.

```[tasklist]
### Todo
- [x] Save panel state in url
- [x] Admin center refactor
- [x] move System settings in Admin center (proposal)
- [x] User details drawer
- [x] Plugin details drawer
- [x] Notification details link from drawer
- [x] move InfoItem into own component and make it more generic
- [x] bump API version due to plugins settings API change
```


matmair:
I strongly think that settings and the admin centre should be separate. Settings are a set-once thing while the admin centre is meant to be an ergonomic place to manage operations of an instance (like user management).

src/frontend/src/components/tables/settings/UserTable.tsx
```diff
@@ -35,13 +38,88 @@ export interface UserDetailI {
...
+        ) : (
+          <Trans>An error occurred while fetching user details</Trans>
+        )}
+      </Text>
+    );
+  }
+
+  return (
+    <Stack>
+      <EditApiForm
```
matmair, line 76: I find this way of selecting the roles far less intuitive than the pills that signified no to all access from left to right. The texts are also not fitting as we use the roles in other ways.
grafik

wolflu05: I can change the texts, that's no problem. But look at the lines of code required for the user drawer now. That's by far less than the old implementation and easier to understand. And there is even more functionality than in the old implementation. Editing a user other than the permissions works now. I actually find it better to have some description instead of having no explanation.

wolflu05:
Why do you think that they are a set-once? I use them from time to time if I want to tweak something. E.g. manage the project codes, location types, ...

wolflu05:
I'm not to happy about having two separate pages as this can confuse users. I initially through something like this would look good, but it turns out that this seems not like a good solution in terms of UI/UX:
image

Then I thought about something like this, but that's up to you. I can also leave it like it is for now:
image

SchrodingersGat:
I am leaning in favour of the single area for "settings" - I appreciate that some are touched more than others, and that actions such as "user management" would clearly fall on one side of the line. But for other types of settings I am not so sure.

I like the latest screenshot by @wolflu05 above - priority is given to user management and plugins. @matmair would you be amenable to that?

I think that in combination with the detail "drawers" for complex settings management, it's pretty clean

src/frontend/src/components/tables/settings/UserTable.tsx
wolflu05: @SchrodingersGat what is your opinion on this one?

SchrodingersGat: I don't think that "active" / "staff" / "superuser" can be classified under the same banner as "rights".

I prefer the left hand screenshot here - but I'd like to see a comparison for group management too

wolflu05:
This would be ready for review now. I implemented a lot of plugin management functionality in this PR. Also the panels now all push their active panel state into the url, so that sharing a url works. The same behavior applies for detail panels that have opened e.g. a specific user.

I also move the system settings now into the admin center. Lets see what @matmair thinks about that implementation now.

src/frontend/src/components/tables/settings/UserTable.tsx
wolflu05: No, only the name of the group can be changed for now. Adjusting the rule set is not possible.

matmair: There are now 2 ways to change user data and group data and over-complicated info text with references to sites the user probably doe not need to know about - I do not see how this is an improvement in usability.

matmair:
The settings are a convolutet mixture of not-explained custom widgets and dense toggles. This does not feel intuitive and the admin centre was thought to be a new place for streamlined, re-designed controls for both Django admin controls and management tasks but I guess I will not be.
I have written what I think about this change now multiple times and I will change it in my deployment;
@SchrodingersGat is the maintainer, I will focus now on code bases where the things I invest time in are not removed within the week.

src/frontend/src/components/tables/settings/UserTable.tsx
wolflu05: What do you mean with two ways? And what empty detail page?

matmair: There are 2 ways to edit user details: the row action and the drawer; same applies to groups;
grafik
1 actionable field that is already exposed through row actions => empty

matmair:
grafik
Also, margins/shadows are jumping all over the place from left to right

SchrodingersGat:
@matmair I do appreciate that you are putting a lot of thought and effort into this. The fast pace of current development has resulted in some conflicts here. I would like to "slow down" to ensure that:

- We are all on the same page in terms of what and how we are developing
- The UX is consistent and user friendly
- We plan ahead for this development rather than re-inventing the wheel multiple times.

You have stated above that the current implementation does not feel streamlined and/or intuitive. If you can provide some clear screenshots of what you take issue with, and some comparison shots of your suggested alternative, let's look at those and work out how we can best proceed in a collaborative manner.

I did some playing around with the Azure "drawer" settings interface - that does work very well! I am not dedicated to reproducing our current UI, there are definitely things that could be improved there. I am also not going to try and dictate that things must be done my way "just because". If you feel that your improvements have been lost or overwritten, I apologize for that.

### Table Actions

One area where I would like some feedback / clarity is how we handle row actions in tables. Currently, these are presented at the right hand column of each row, with an ellipsis displayed if multiple actions are available:

image

This new approach of clicking on the row to show the "edit user" pop-out drawer is *different* and non-obvious. There is no button to click, and no visual indication of how the user accesses this information.

I am not against the drawer, but it should be a bit more accessible.

Also, it does not sit well with me that there are now multiple user interfaces for editing data:

edit-user-drawer
edit-user-modal

Let me know what you think @matmair @wolflu05 - my priority is ensuring we are all happy with our direction moving forward. I do not want to fracture frontend development or the whole thing will fall over.

wolflu05:
### Admin center / System settings
There is no reason to be upset, so calm down, I just proposed this change, this doesn't mean it have to be that way for me, its not written into stone, just because I pushed it into this PR, it was just a proposal, and I also marked it as such. I don't want there to be a fight about it, I reverted this change now. I actually just thought I need to improve the navigation system with the panels pushing the opened panel state into the url. Can you please explain a bit more about what you imagine to be in the admin center in the future next to the user management so that I can better understand your intents about it.

Do the plugins tab I moved in there fit with your ideas? Also the machines management table (#4824) which is in my opinion an administrative task, can there management panel be a tab in the admin center?

### User/Group drawer
Regarding the user/group drawer. I just took your approach with the drawers and wrapped it inside a new component `DetailDrawer` which also uses the url for the id state and is reusable. Which enables as to link between different drawers and pages. I already thought since you implemented that user detail drawer why all these inputs are disabled and how I can get them enabled to edit the user details, but now it seems like this was not intended from you, but I had the feeling that I can edit the InvenTree user with this detail drawer. 

And instead of just complying the 190 lines your details drawer had with those not working input fields I thought why not simplify things and use the new `OptionsApiForm` component which is now separated from the forms. And then rather use those lines to add some more functionality like not letting the user edit there own active, staff, superuser state so that he doesn't locks himself out which now is much simpler and takes only ~110 lines. In my opinion also those chip inputs provided bad UX, because if you click them accidentally an action is already performed.

I have no access to all of those azure things, so please don't blame me, but rather explain me where the differences between this already existing edit modal and the details drawers should be if I can use one only for changing the names and the other only for setting permissions, isn't that bad UX because I have to switch between different, also hidden (one may not know that clicking on a row opens the details drawer) places?

But that UX thing is like @SchrodingersGat already mentioned not great. What about removing the modal and do everything with the drawer? Then we could open the drawer by either clicking on the row actions edit button or directly on the table row.

### Margins/Shadows
I didn't changed anything regarding the margins. @matmair you actually created the admin center with the panels, quick actions, ... and everything else. What I did was just moving the plugins management into this center, adding a lot of functionality in the plugins tab, and making the panels/detail drawers push their active state/id into the url so that sharing a url works.

The only thing noticeable I did change regarding UI design was adding this shading to the setting items, but that was in one of my previous PRs, because in my opinion this was bad UX and I wanted to improve it instead of complain about it. If you have a different solution for helping the user to see what setting name on the left corresponds to what settings value on the right on wide screens feel free to propose a change. Adding shading was the easiest solution I could think of.

## Conclusion
I don't want to remove your code or that you have the feeling I did, I just wanted to improve the new Platform UI and add some more great user experiences and available options, and I think that's what we all want. So please take the time and and explain how I should continue now.

SchrodingersGat:
@wolflu05 thanks for your input :) I'll leave this one waiting here for a bit to allow @matmair a chance to respond.

> What about removing the modal and do everything with the drawer? Then we could open the drawer by either clicking on the row actions edit button or directly on the table row.

I think this is a good approach. But, do we continue to use modals everywhere else? How do we decide where to draw the line between a drawer and a modal?

matmair:
I think settings and admin centre should be different things. Mixing global settings and administrative controls leads to large blocks of mostly unused UI being displayed every day without any benefit to staff/superuser.
The more features we introduced the more spread out settings became and actionable tasks like management of users, plugins, project codes, test templates and parameters are more and more hidden in this long list. Watching new users deploy their instances and finding out about features months after installing because they never saw the button for it was eye-opening for me several times over the last few weeks.

I will step back from any larger development for now, feel free to do what you feel is the best.

SchrodingersGat:
I mostly agree with what @matmair is suggesting here in terms of separation of settings from "data"

### Settings Pages

- Configuration of `GlobalSettings` values
- Not often used, change behaviour of InvenTree installation

### Admin Section

- Changing "data" (not settings)
- e.g. tabulated data panels like "plugins" / "parameters" / "users"
- Expected to change over time

wolflu05:
@SchrodingersGat  This should be ready already, but as there are some api changes, you would need to run the server locally.

SchrodingersGat:
@wolflu05 one unintended consequence of this merge was the change to the panel group which adjusts the URL when the active panel changes.

Changing the URL causes the `id` value (extracted from the current URL in the `useId` hook) to change, which has unintended consequences.

For example, navigate to:

`http://localhost:5173/platform/part/category/`

and see that the URL is changed to:

`http://localhost:5173/platform/part/category/parts/parts`

This causes an unexpected URL pattern which can cause incorrect API queries to be called from the server (because the `useId` and `useInstance` hooks are receiving the wrong values)

I do not think that overriding the URL is a particularly clean solution here, but open to ideas.

PR #5972, opened 2023-11-22
[PUI/Feature] Integrate Part "Default Location" into UX

### LINKED ISSUE #5928: [FR] Larger extension of "Default location" functionality

### Problem statement

You can mark any part with a default location p.t.

However, there is no connection from a location to see which items have their default set to it.
There is no way to choose to transfer a stock item to its default location.
There is no way to easily add a newly arrived stock item to its default location from a Purchase Order

### Suggested solution

Integrate the concept of default locations more thoroughly throughout the Stock experience.
If a part has a default location, allow for any stock item to be transferred to this location with a simple click.
Allow all PO's to be added to the default location of the Part upon reception if a box has been ticked
Possible more ways to use default locations to smooth out workflows.

Opening tabs
Checking default locs
Typing out what you just read

### Examples of other systems

_No response_

- [X] I want to develop this.

### LINKED ISSUE #5938: "Default Location" of a category not propagated to parts inside the category

### Describe the bug*

The optional "Default Location" parameter filled in when creating new categories, is not propagated to parts in this category. No matter whether the part is newly created, or we assign it to the category afterward.
Contratry, the "Default Location" for a part, is properly transferred to newly created stock items.

The same behavior can be reproduced using web interface or admin panel (both import and manual entry)

It might be, that I'm somehow misunderstanding the "default location" concept here, however I did not find any reference in Inventree docs.

### Steps to Reproduce

1. Create a "New Location"
2. Create "New Category" and fill-in the location.
3. Create "New Part" under the category. The default location for a part is not set.
4.  Creat a stock item of the part. It is not assigned to any location

### Expected behaviour

The newly created parts should have "Default Location" prefilled based on the "Default location" for the category.
Same logic might be applied for new subcategories.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

InvenTree-Version: 0.12.8
Django Version: 3.2.19
Commit Hash: a7487ff
Commit Date: 2023-10-03

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.1.0-13-amd64-x86_64-with-glibc2.36
Installer: DOC

Active plugins: False

### Relevant log output

_No response_

LavissaWoW:
Closes #5928 
Closes #5938 

- [x] Add Default Parts filter to API
- [x] [PUI] Add Default Parts panel to Stock Location View
- [x] Add "Transfer to default location" functionality [PUI] [CUI?]
- [x] Add Default location checkbox when splitting stock items
- [x] Propagate default location from parent category tree to parts


LavissaWoW:
Still a DRAFT.
My mouse likes to double click lately...

LavissaWoW:
@SchrodingersGat @matmair @wolflu05 
Anyone currently implementing the following in PUI?
Stock Item Detail Panel
Stock Item Transfer Modal
Part Category Detail/List
Part Category Creation and Edit Modal

To do all the work I'm planning here, I'll have to add some elements to Stock Items, Part Categories, and additionally Purchase Orders. I don't want to overlap work others are doing :)
If these are TBD atm, I'll take a stab at implementing them, else I'll add the functionality to those when they are merged in.

src/frontend/src/pages/stock/LocationDetail.tsx
```diff
@@ -45,7 +46,7 @@ export default function Stock() {
       },
       {
         name: 'sublocations',
-        label: t`Stock Locations`,
+        label: t`Stock Locaons`,
```
matmair, line 49:
```suggestion
        label: t`Stock Locations`,
```

LavissaWoW: Snuck in accidentally. A small string debug change I made to check what I had to reload for PUI to update. Will be gone on the next commit 😀

SchrodingersGat:
@LavissaWoW I do not have any active progress on any of the items you list. We have not implemented any "detail" pages yet - I would be keen to see if we can develop a standard component or set of components for rendering a "detail" view - primarily for UI consistency. 

If you want to tackle a detail page in the new UI that's great 👍

InvenTree/part/api.py
```diff
@@ -1190,6 +1190,13 @@ def filter_queryset(self, queryset):
                 except (ValueError, PartCategory.DoesNotExist):
                     pass
 
+        # Does the user wish to filter by default location?
+        default_loc = params.get("default_location", None)
```
SchrodingersGat, line 1194: This should be removed now

SchrodingersGat (changes requested):

InvenTree/part/api.py
LavissaWoW: Yes, of course :)

SchrodingersGat:
@LavissaWoW not stale, right?

LavissaWoW:
Not stale indeed. Need to sort out priorities here though. I got some backend shenanigans planned, and several frontend changes.
Depending on the frontend change, prio would be on PUI.

LavissaWoW:
@SchrodingersGat I'm inclined to ignore CUI with this.
I'd rather not spend a lot of time digging into vanilla JS that we're phasing out soon. 
I.e. #5938 would be more easily implemented in PUI rather than CUI. (Update location field when changing category)
This sentiment does however mean that this PR will rely on some PUI work.

LavissaWoW:
@wolflu05 @SchrodingersGat 
I've implemented a form field type for rendering array-based values as tables. 
Relevant changes are in
`TableField.tsx`
`StockForms.tsx:stockTransferFields`
`StockForms.tsx:StockOperationsRow`
`ApiFormField.tsx (case: 'table')`

The use for this is forms where we pass arrays of objects ot the API. For this PR, I'll use it for Stock operations.
Let me know what you think.

wolflu05:
Nice idea to implement this as a field type. After reading array field, I first was thinking something different, but after reading the implementation the name `table` makes sense. I'm only concerned about that this is actually not a real form, you're just using it to display a modal? But I'm fine with it.

LavissaWoW:
It is a proper form for the /stock/transfer POST endpoint.
Looks like this with only 1 item in the table :)
image

LavissaWoW:
I'm feature creeping a bit here. This PR will implement all stock actions in PUI.
Implementing the ones relating to locations, and NOT others seem counter-productive.

LavissaWoW:
#6541 implements a form field that this PR will use for the `stock_receive_create` API endpoint. Will be ready when this is implemented.

LavissaWoW:
I should probably handle the category defaults for CUI too. Feel free to review, though. I'll add one last commit asap

LavissaWoW:
@SchrodingersGat All good, this is ready now

src/frontend/src/pages/stock/LocationDetail.tsx
```diff
@@ -139,10 +156,113 @@ export default function Stock() {
         label: t`Stock Locations`,
         icon: <IconSitemap />,
         content: <StockLocationTable parentId={id} />
+      },
+      {
+        name: 'default_parts',
```
SchrodingersGat, line 161: This should be hidden for the top-level category (i.e. no category specified)

SchrodingersGat:
@LavissaWoW this is great, such a lot of work here :)

Review below, there's a few points and suggestions but nothing major!

### Stock Table Actions

When no items are selected, either:

a) Disable the "actions" column, or
b) Select all stock items in the table by default

Otherwise, you get this "no entries available" message

image

In fact, this goes for other table forms too. If there are no available lines when opening the form, we need a "default selection" here.

### Receive Line Items

image

Really good! Perhaps the modal should be wider by default?

### Receive Line Items - Default Location

Looks like the "default location" does not propagate here?

image

This part does have a "default location" set:

image

Although it should also work if there is a "default category location"

### Category Detail Page

Should display the "default location" for the category (or upstream parent category as required). Currently not displayed:

image

src/frontend/src/pages/part/PartDetail.tsx
```diff
@@ -131,6 +136,13 @@ export default function PartDetail() {
         model: ModelType.stocklocation,
         hidden: !part.default_location
       },
+      {
+        type: 'link',
+        name: 'category_default_location',
+        label: t`Category Default Location`,
```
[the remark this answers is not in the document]
LavissaWoW: This is intentional, as it refers to the annotated field "category_default_location". I haven't removed the normal `default_location` field from the panel.
If you see the `hidden` value, it only displays if the part doesn't have its own default location. My logic here is that I don't want to populate the `default_location` field with the annotated data, as you easily would accidentally assign that location to the part on edit. This way, it's obvious that the location you see listed has been infered from the category tree.

src/frontend/src/pages/stock/LocationDetail.tsx
LavissaWoW: Good catch!

SchrodingersGat:
> or disable actions when selectedRecords is empty? This includes PO receive too.

This is probably the best approach for now, until (if) we can work out a different approach. At least visually the user should be aware that they have no records selected and thus cannot perform actions

LavissaWoW:
@SchrodingersGat I'm not able to reproduce the missing propagation of the default loc.
Here's a pic of how different pre-filled cases look like at my end:
image
Could it be that PUI data wasn't updated after you set a default on the part?

LavissaWoW:
@SchrodingersGat I was initially struggling to reproduce the PO line location, but it turns out I made a couple of mistakes, and it was even more broken than you experienced. The PO lines have a `destination` and a `destination_detail`. The latter can be set without the former being set. If the latter is set, and the former not, one of the following is true: 
1. The part has a default location
2. Some stock has already been received for this line

My logic was off in my assumptions here, and I found myself with test data where previously received stock had been placed at the default location. So some of my logic indicated default, while it was actually previously received that triggered it. Additionally, the queryset annotation I added to StockItem ofc. didn't magically carry over to POLineItems, so `category_default_location` never even existed on `line.part_detail`

Fix incoming shortly

SchrodingersGat:
@LavissaWoW I realize that part of the issue was that I was viewing the netlify preview - which talks to the demo server, which in turn does not have the latest API. So, this will explain the discrepencies.

SchrodingersGat (approved):

SchrodingersGat:
@LavissaWoW thanks again for the great work here!

PR #5990, opened 2023-11-27
Fix 'Maximimum' typo

mechanarchy:


SchrodingersGat:
@mechanarchy thanks for the fix :)

PR #6005, opened 2023-11-29
Fix reloading on viewport size change

### LINKED ISSUE #6004: [Bug/PUI] Window resizing triggers full page reload

### Describe the bug*

Any resizing of the browser window while using PUI will cause React to re-do all API requests.
It appears the cause is that PUI completely repaints all components, even the loading spinner.
This becomes extra apparent when resizing the dev tools panel of Chrome and looking at the API requests sent to the backend.

### Steps to Reproduce

1. Open a window with PUI
2. Perform any kind of resizing of the browser window
3. PUI will perform a full page repaint, including the "loading spinner"

### Expected behaviour

React repainting to be performed without having to re-fetch data from the backend.
No loading spinner on resize.

### Deployment Method

- [ ] Docker
- [ ] Bare metal

### Version Information

0.13.0 using devcontainer
(Not at the relevant computer atm)

### Relevant log output

_No response_

wolflu05:
This fixes the constant reloading if the viewport size changes. This has already bothered me a long time, and I knew it was reloading, because of the constant reimporting of the component , but didn't changed, I just popped out the dev tool in a separate window and avoided resizing till @LavissaWoW created that issue now :D . Also sometimes this can lead to a so high number of requests that the server temporarily get unavailable and the session logs out of PUI.

Fixes #6004

### DIFF
--- a/src/frontend/src/App.tsx
+++ b/src/frontend/src/App.tsx
@@ -25,15 +25,16 @@ function checkMobile() {
   return false;
 }
 
+const MobileAppView = Loadable(lazy(() => import('./views/MobileAppView')));
+const DesktopAppView = Loadable(lazy(() => import('./views/DesktopAppView')));
+
 // Main App
 export default function App() {
   // Check if mobile
   if (checkMobile()) {
-    const MobileAppView = Loadable(lazy(() => import('./views/MobileAppView')));
     return <MobileAppView />;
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@wolflu05 nice work, appears to work very smoothly :)

PR #6008, opened 2023-11-29
[PUI] Allow rendering of custom states

matmair:
This PR makes StatusRenderer and the Zustand it relies on less strict to allow usage of custom states that are defined using the generic StatusCode.

The lookup is now possible either via ModelType or a string. That string has to match the class name of the StatusCode enum that should be used; if no matching state is defined the renderer returns an empty node.
Provides full support for colors, keys, names and translated labels for custom states out of the box.


SchrodingersGat:
LGTM @matmair

PR #6012, opened 2023-11-30
[PUI] Enable passing of string urls to apiUrl

matmair:
This PR enables passing strings (preferably URLs) to API forms so they can be used with custom URLs / in plugins.

### DIFF
--- a/src/frontend/src/components/forms/ApiForm.tsx
+++ b/src/frontend/src/components/forms/ApiForm.tsx
@@ -60,7 +60,7 @@ export interface ApiFormAction {
  * @param onFormError : A callback function to call when the form is submitted with errors.
  */
 export interface ApiFormProps {
-  url: ApiPaths;
+  url: ApiPaths | string;
   pk?: number | string | undefined;
   pathParams?: PathParams;
   method?: 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE';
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
This will weaken the existing type checking checks which have already been useful in detection of mis-typed URLs. However I am not sure if there is a better approach here

wolflu05:
In my opinion the idea behind this function is good, but the issue with it is, that if adding an api path, I basically have to go to two different files and add something on two places. That slows down the quick development.  As the apiEndpoint function is just a huge switch case prefabricated building, why not just use an object with `key: url` and type the function accordingly. And then to allow strings, we add a second function that allows strings which gets called by the function which only allows ApiPaths. Because we cannot union `string` with choices. E.g. something like this:

```ts
export const API_PATHS = {
  "api-plugin-settings": "plugins/:plugin/settings/",
  // ...
}
export type API_PATH_TYPE = keyof typeof API_PATHS;

export function apiGenericUrl(path: string; ...) {
  // do  like before, but without the apiEndpoint call
} 

export function apiUrl(path: API_PATH_TYPE, ...args) {
  return apiGenericUrl(API_PATHS[path], ...args)
}
```

That way we would 
1. only need to change something at one place
2. only need to import the `apiUrl` function
3. have the possibility to use raw strings without mixing them

SchrodingersGat:
@wolflu05 I'd suggest making that as a separate issue :)

Issue #6009, opened 2023-11-29
[Bug/PUI] Table filters are ignored when initially loading a page URL


LavissaWoW:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

Tables are initially loaded with `null` as `pk` within PUI.
I've run this test on Part and Stock Item, and they both load an unfiltered table for their respective models.

### Steps to Reproduce

Based on the demo dataset

1. Make sure you have the "Parts" panel selected
2. Access http://localhost:5173/platform/part/category/13 (or your local version of this. category PK 13 at least)
3. This category has 57 items. but the total will be listed as 414.
4. Refresh, and it'll list it as 57.

I added this debug print to Stock Locations (/stock/location/7):
`console.log("Location:", location, "id:", id)`, and this is the output on hitting F5:

![image](https://github.com/inventree/InvenTree/assets/7506688/69bc4b42-f24d-4e97-8b0d-a928233c1adb)

The code passes `location.pk`, and that doesn't become available until after the first render of the table.

### Expected behaviour

API filters should be added on first load, not second. 

### Deployment Method

- [ ] Docker
- [ ] Bare metal

### Version Information

0.13.0 master commit f0f72a8742781495628ed09dfcf288061ff22071

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

SchrodingersGat:
Thanks for reporting. Probably an issue with the hooks when loading the table. Do you want to look into this one?

LavissaWoW:
@SchrodingersGat Yeah, I can have a look and find a fix.

LavissaWoW:
Hmm.

So, what's going on initially is that 

```js
params={{
  location: location.pk ?? null
}}
```
is populating no `null`, as 

```js
<PanelGroup pageKey="stocklocation" panels={locationPanels} />
```
is read before the `useInstance` resolves. 
This makes `InvenTreeTable` send an API request with no category/location/etc parameter in the request.
As made evident by the following log entry when accessing `/stock/location/7`: 
`[30/Nov/2023 19:18:51] "GET /api/stock/?part_detail=true&location_detail=true&supplier_part_detail=true&limit=25&offset=0 HTTP/1.1" 200 32476`
So the resulting table populates using the entire unfiltered list of whatever the table displays. 
I don't have to elaborate what a large performance hit this can have on large InvenTree instances.

[...]

PR #6014, opened 2023-11-30
[PUI] Table loading refactor

LavissaWoW:
Ties into #6009

Can eventually close said issue if the solution is acceptable.

Obviously disregard the `console.log`'s :>

### DIFF
--- a/src/frontend/src/components/tables/InvenTreeTable.tsx
+++ b/src/frontend/src/components/tables/InvenTreeTable.tsx
@@ -94,14 +94,12 @@ export function InvenTreeTable<T = any>({
   url,
   tableState,
   columns,
-  props,
-  wait = false
+  props
 }: {
   url: string;
   tableState: TableState;
   columns: TableColumn<T>[];
   props: InvenTreeTableProps<T>;
-  wait: boolean;
 }) {
   // Use the first part of the table key as the table name
   const tableName: string = useMemo(() => {
@@ -339,13 +337,6 @@ export function InvenTreeTable<T = any>({
   // Function to perform API query to fetch required data
   const fetchTableData = async () => {
     let queryParams = getTableFilters(true);
-
-    if (wait == true) {
-      console.log('Returned on wait');
-      return [];
-    }
-
-    console.log("Didn't wait");
 
     return api
       .get(`${url}`, {
--- a/src/frontend/src/components/tables/InvenTreeTable.tsx
+++ b/src/frontend/src/components/tables/InvenTreeTable.tsx
@@ -94,12 +94,14 @@ export function InvenTreeTable<T = any>({
   url,
   tableState,
   columns,
-  props
+  props,
+  wait = false
 }: {
   url: string;
   tableState: TableState;
   columns: TableColumn<T>[];
   props: InvenTreeTableProps<T>;
+  wait: boolean;
 }) {
   // Use the first part of the table key as the table name
   const tableName: string = useMemo(() => {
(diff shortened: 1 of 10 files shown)


LavissaWoW:
By introducing the `useInstance` object to the table's JSX, we're able to stop table generation when the proper parameters are not yet present. React will trigger a re-render when this parameter changes, and since it's treated as a first-level parameter of the underlying `InvenTreeTable`, we can be assured that the table will render properly.

As mentioned in the issue, even though the InvenTreeTable component gets an updated `pk` for the relevant table after `useInstance` finishes, this does not constitute a re-render from React's side for some reason.

LavissaWoW:
My initial thought was to simply use `id` rather than wait for the promise to resolve, as you suggest.
But I figured it's done this way intentionally, to allow for the browser given ID and `pk` to be different in the future? 

The first suggestion will solve the issue, given that the above is true.
I was initially thinking you'd want the panel to load even if the API was being slow. But thinking about it with a fresh head, if the instanceQuery is still fetching, the LoadingOverlay will be visible. So then that problem becomes moot.

wolflu05:
> My initial thought was to simply use id rather than wait for the promise to resolve, as you suggest.
But I figured it's done this way intentionally, to allow for the browser given ID and pk to be different in the future?

I'm not sure if this makes sense in the future, because the `location.pk` is the same as the id. And afaik that's the only way to get a unique instance. Given the alternative implementations complexity I would go for the `id` approach, what do you mean @SchrodingersGat ?

LavissaWoW:
@wolflu05 Ive read some issues asking for UUID based URLs. Basically asking for database PKs to be obfuscated. That's where I drew my reasoning from. But if @SchrodingersGat green-lights using the ID, I'll patch all the currently implemented tables in this manner and update the PR accordingly.

LavissaWoW:
@SchrodingersGat I've exchanged PK for ID only on panels that did not already rely on the objects returned via `instanceQuery`.
The logic being that if a panel uses the object for something other than PK, it's waiting for the object anyhow.

> Sometimes the id does not "resolve" correctly (could be a dependency issue)
The table renders multiple times, first time without the correct id - and so you get a "flash" of incorrect data into the table.

I saw a few oddities just now, but they did not seem to be related to problems with the ID. 
It looks to me as if data from old table objects are re-used in new table objects. 

To reproduce my experience:
1. Open BO-0015 (platform/build/15)
2. Open "Completed Outputs" panel
3. Go back to "Details" panel
4. Go to Build Index (platform/build)
5. Go to  BO-0016 (platform/build/16)
6. Open "Completed Outputs" panel
7. You'll see a brief flash of 3x Green Chair in the table.

SchrodingersGat:
@LavissaWoW I can definitely see the cached data appearing as you describe above. If you are confident that this is a different issue, I'm happy to merge this in when ready (but we really should make a new issue for the cached data behaviour).

LavissaWoW:
`GET http://localhost:5173/api/stock/?build=-1&is_building=false&part_detail=true&location_detail=true&supplier_part_detail=true&limit=25&offset=0 400 (Bad Request)`

@SchrodingersGat the `InvenTreeTable` component doesn't care about `-1`, and proceeds to ask the API for -1. I'll remove these so the console doesn't generate errors on these cases.

SchrodingersGat (approved):

SchrodingersGat:
@LavissaWoW LGTM - have you confirmed that the issue is now fixed?

SchrodingersGat:
Ok, great!

Can you please address these javascript issues:

https://app.deepsource.com/gh/inventree/InvenTree/run/450df5eb-93ae-4a53-9a86-04669e8c5477/javascript/JS-0356?listindex=0

LavissaWoW:
Of course. I suspect they were latent ones waiting to get triggered, as I didn't interact with those specifically. Just the file.
Commit inc. as soon as my slow af. devcontainer manages to do the pre-commit.

PR #6016, opened 2023-11-30
[API] Add default location filter to part endpoint

LavissaWoW:
A small PR in hope to make it into 0.13.0.

The purpose of this filter is mainly to extend the capabilities of label printing. It allows StockLocations down the road to access what Parts are set to have their default location there. 

With this, I can write a very simple temporary report plugin to apply this filter while I get the rest of the Default Part experience in place.

If you so desire, I'll implement this into the StockLocation Serializer before merging. That will anyway be a part of #5972


LavissaWoW:
@SchrodingersGat You have any pointers for that approach?
I'll admit, I met the so-called wall when trying toput this in the `PartFilter` class. That's why I put it in the Part class...
The incoming value is the PK of the StockLocation, and I didn't find/understand what filter-class to use for this case.

`ModelChoice` was at least the wrong choice as far as I understood...

SchrodingersGat:
Lgtm, thanks 👍

PR #6017, opened 2023-12-01
State transition support for generic states

matmair:
This WIP PR adds the option to define methods that should be called when objects transition between states.

Missing:
- [x] QoL improvements around developing classes
- [x] sample plugin & unit tests


SchrodingersGat:
Can you provide some end-goal examples of how this transition management would be used?

matmair:
@SchrodingersGat this would be ready for review; is the sample code added enough to understand the use case?
Without custom states #4969 this is not usable for full-state machines but still helpful to enhance the control flow via plugins.

SchrodingersGat:
@matmair this looks good to me. I would like to see this added to the `BuildOrder` and `StockItem` classes, but that can happen as a separate PR if you like. Happy to merge this now. thanks for the work!

Issue #5493, opened 2023-08-29
Adding supplier, error without information


Zontex:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

Tried to adding supplier got an error that there is an issue in the form. the form looks OK. I inspected network and saw that the backend responded url is wrong although url is ok. tried www.website.com also wrong. did https://website.com/ and it fixed the issue.

Image below:

<img width="1649" alt="Screen Shot 2023-08-29 at 12 37 54" src="https://github.com/inventree/InvenTree/assets/19342533/f514edff-561e-4425-9904-1a73457dd781">

Ont the demo website however, I get the errors properly but still without telling me why URL is wrong:

<img width="1175" alt="Screen Shot 2023-08-29 at 12 47 30" src="https://github.com/inventree/InvenTree/assets/19342533/eef7026f-6b85-4bbc-86b9-f41657d01bc5">

### Steps to Reproduce

1. add supplier.
2. fill in supplier information
3. on the supplier website write website.com (can be any website) or www.website.com.
4. get an error, see there is no information why the error happens or how to fix it.

### Expected behaviour

1. error is fine, the error should tell it expects format of https:// 
2. if error happens else where, give detailed explaination why and where the error happens.
3. backend gives u response, easer way will be to take from backend and show directly.

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.12.3
Django Version: 3.2.19
Commit Hash: 1f3a49b1
Commit Date: None
Commit Branch: stable
Database: sqlite3
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-5.15.0-82-generic-x86_64-with-glibc2.35
Installer: PKG
Target: ubuntu:20.04
Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}]

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

SchrodingersGat:
[...]

PR #6045, opened 2023-12-06
Url validation

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/5493

Allows URLs to be provided without an explicit schema (if setting is configured)


InvenTree/templates/InvenTree/settings/global.html
```diff
@@ -24,6 +24,7 @@
         {% include "InvenTree/settings/setting.html" with key="INVENTREE_DOWNLOAD_IMAGE_MAX_SIZE" icon="fa-server" %}
         {% include "InvenTree/settings/setting.html" with key="INVENTREE_DOWNLOAD_FROM_URL_USER_AGENT" icon="fa-server" %}
         {% include "InvenTree/settings/setting.html" with key="INVENTREE_REQUIRE_CONFIRM" icon="fa-check" %}
+        {% include "InvenTree/settings/setting.html" with key="INVENTREE_STRICT_URLS" icon="fa-link" %}
```
wolflu05, line 27: @SchrodingersGat we should add that to PUI too.

Issue #6013, opened 2023-11-30
Inserting serial numbers is getting slower when having high number of stock items


martonmiklos:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

I tried to migrate our 5 years of historical production data into the InvenTree with a python script using the InvenTree API.  I started to ran into sort of performance issues after inserting ~67K stock items (took a couple of day) what I thought worth to investigate before moving forward. A single request for creating a stock item with a single serial takes ~3s to complete now. It was not so slow at the beginning if I remember correctly, but now it started to be a bottleneck. 
(I am running InvenTree with MySQL on a notebook with reasonable performance (8th gen i9))

As a first idea I added some timing measurement here:
https://github.com/martonmiklos/InvenTree/blob/stockitem_save_performance_tracking/InvenTree/stock/models.py#L478

Which results the following for a single StockItem + serial insert:
```
Validate unique: 13.355768998735584 ms
Clean: 0.7472590077668428 ms
Update serial: 0.5881280085304752 ms
Super save: 837.7136269991752 ms
Validate unique: 15.880550010479055 ms
Clean: 0.6893499958096072 ms
Update serial: 0.5778909981017932 ms
Field update: 0.9629400010453537 ms
Super save: 624.8631559865316 ms
Tracking info: 1.3864210050087422 ms

```
The first thing what was strange for me that the StockItem::save got called twice from the API call
Here:
https://github.com/inventree/InvenTree/blob/master/InvenTree/stock/api.py#L898

And here:
https://github.com/inventree/InvenTree/blob/master/InvenTree/stock/api.py#L905

i do not have in depth knowledge on debugging such a problems with django+python (and maybe MySQL) so any hint/tip what I can look into/try would be welcome.

### Steps to Reproduce

-

### Expected behaviour

Adding a single serial number should be reasonably fast regardless to the stock item count in the system :)

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

Latest master

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

### Relevant log output

```shell
-
```

wolflu05:
[...]

PR #6051, opened 2023-12-07
Disable part prefetch on stock item save

martonmiklos:
As mentioned in the #6013 when saving the stock items the part pulls out all the stock items from the database due to the prefetch. As no related objects of the part is used in the stock items save I disabled the prefetch.


SchrodingersGat:
@martonmiklos have you profiled this to confirm that it reduces the time taken to create a new stock item?

PR #6054, opened 2023-12-07
[PUI] Fix Build Router

LavissaWoW:
Build entry of router.tsx is missing a `*` at the end, preventing sub-URLs from rendering.

Added a `*` at the build routing entry.

### DIFF
--- a/src/frontend/src/router.tsx
+++ b/src/frontend/src/router.tsx
@@ -130,7 +130,7 @@ export const routes = (
       </Route>
       <Route path="build/">
         <Route index element={<BuildIndex />} />
-        <Route path=":id/" element={<BuildDetail />} />
+        <Route path=":id/*" element={<BuildDetail />} />
       </Route>
       <Route path="purchasing/">
         <Route index element={<Navigate to="index/" />} />


src/frontend/src/router.tsx
```diff
@@ -130,7 +130,7 @@ export const routes = (
       </Route>
       <Route path="build/">
         <Route index element={<BuildIndex />} />
-        <Route path=":id/" element={<BuildDetail />} />
+        <Route path=":id/*" element={<BuildDetail />} />
       </Route>
```
wolflu05, line 134: I'm not sure if this is already sufficient, but if the `BuildIndex` also contains panels, `DetailDrawers` or anything that uses routes, we would need something like the purchasing section has:
```jsx
    <Route path="purchasing/">
        <Route index element={<Navigate to="index/" />} />
        <Route path="index/*" element={<PurchasingIndex />} />
        { /* ... */ }
      </Route>
```

LavissaWoW: The Build app seems to be a rare occurence of an index Page only containing its own index table. No panels, no detail drawer etc. Should we future-proof it then, I think becomes the relevant question?

SchrodingersGat: We can make this a page with only a single available panel (the "build orders" list) pending future additions. Would this make the routing easier? Or just more consistent?

PR #6055, opened 2023-12-08
Prevent check for low stock when creating new items

SchrodingersGat:
- Ref https://github.com/inventree/InvenTree/pull/6051
- Ref https://github.com/inventree/InvenTree/issues/6013


martonmiklos:
I think it would be useful to add check for the "dirtyness" of the quantity field and the negative quantity change too. I have not tried it out, but as I described in the #6013 during the creation of the stock items the save got called twice (one create, one update). I can check it on Monday.

PR #6067, opened 2023-12-10
fix(SSO): log error message when EMAIL_HOST is empty

### LINKED ISSUE #4553: EMail host is required for SSO self registration

### Describe the bug*

The [CustomSocialAccountAdapter](https://github.com/inventree/InvenTree/blob/c28c991591c9bd2c29be4c0046d500e3271feda2/InvenTree/InvenTree/forms.py#LL289C16-L289C16) extends the RegistratonMixin which overrides the `is_open_for_signup` method. In my opinion, the [check](https://github.com/inventree/InvenTree/blob/c28c991591c9bd2c29be4c0046d500e3271feda2/InvenTree/InvenTree/forms.py#L217) performed in the function should not be the intended behavior for Social Accounts.

### Steps to Reproduce

1. Configure a social account for your inventree instance
2. Make sure no email host is configured
3. Try to sign in for the first time with a social account user and experience the "Sign in Closed" page.

### Expected behaviour

In my opinion, it should not be required to configure an email host in order to enable SocialAccount self registration.

My personal use case:
I am currently running InvenTree on my local machine and played around with the OpenID config. It took me forever to figure out why my social login did not work. I finally solved it by specifying a dummy value `INVENTREE_EMAIL_HOST = localhost`.

Do you think it is possible to disable the email host check if a social account is used?

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

Master branch

### Relevant log output

_No response_

p-fruck:
fixes #4553. Actually the settings page already includes a warning stating there might be issue without the email host being configured, but I still think an error message should be printed as this is clearly a misconfiguration of the server admin and only happens after the user successfully signed in to their IdP.

### DIFF
--- a/InvenTree/InvenTree/forms.py
+++ b/InvenTree/InvenTree/forms.py
@@ -228,7 +228,12 @@ class CustomSignupForm(SignupForm):
 
 def registration_enabled():
     """Determine whether user registration is enabled."""
-    return settings.EMAIL_HOST and (InvenTreeSetting.get_setting('LOGIN_ENABLE_REG') or InvenTreeSetting.get_setting('LOGIN_ENABLE_SSO_REG'))
+    if InvenTreeSetting.get_setting('LOGIN_ENABLE_REG') or InvenTreeSetting.get_setting('LOGIN_ENABLE_SSO_REG'):
+        if settings.EMAIL_HOST:
+            return True
+        else:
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@p-fruck thanks for this fix :)

PR #6068, opened 2023-12-10
chore(docker): Remove intermediate `prebuild` stage

p-fruck:
Removes the `prebuild` stage from the final `production` image which solves the issue that build dependencies where not cleaned up properly. Saves a couple of MB in image size.


SchrodingersGat:
Cheers :)

Issue #6073, opened 2023-12-12
[PUI] URL changes when the active panel changes.


SchrodingersGat:
@wolflu05 one unintended consequence of this merge was the change to the panel group which adjusts the URL when the active panel changes.

Changing the URL causes the `id` value (extracted from the current URL in the `useId` hook) to change, which has unintended consequences.

For example, navigate to:

`http://localhost:5173/platform/part/category/`

and see that the URL is changed to:

`http://localhost:5173/platform/part/category/parts/parts`

This causes an unexpected URL pattern which can cause incorrect API queries to be called from the server (because the `useId` and `useInstance` hooks are receiving the wrong values)

I do not think that overriding the URL is a particularly clean solution here, but open to ideas.

_Originally posted by @SchrodingersGat in https://github.com/inventree/InvenTree/issues/5953#issuecomment-1852058964_

SchrodingersGat:
With the current setup, if you navigate to`/part/category/` page the URL re-write results in a query for *all* parts, which takes a significant amount of time to execute. This is how I found this issue in the first place

SchrodingersGat:
To reproduce:

1. Open developer console, so you can see the API queries
2. Navigate to `http://localhost:5173/platform/part/category/`
3. Select the `parts` tab
4. Note that the URL has now changed to `http://localhost:5173/platform/part/category/parts/parts`
5. See the *very long* API query which is requesting *all* parts from the database:

![image](https://github.com/inventree/InvenTree/assets/10080325/545fb1c9-3fda-4237-a558-76c9ebbcbbd6)

6. Look at query details:

![image](https://github.com/inventree/InvenTree/assets/10080325/bf92551f-e16c-4d5f-bc70-a5f32a01f2bf)

This is the *useInstance* query in the `<CategoryDetail>` page

PR #6078, opened 2023-12-13
[PUI] URl / panel fix

SchrodingersGat:
Fixes https://github.com/inventree/InvenTree/issues/6073

This PR provides two major improvements:

- Fixes routing issues for part and stock index pages
- Fixes issues with URL rewrites from panel changes

### Index Pages

Introduces explicit index pages for the `part` and `stock` views to fix issues related to undefined `useId` hooks. Overall much simpler / cleaner.

### URL Rewriting

Existing panel implementation rewrites page URLs which is problematic for a number of reasons:

- Confusing and convoluted code
- Messes with router implementation too
- Multiple "loops" - panel changes causing multiple component loads / slow
- The *page* does not actually change, so URL feels like the wrong approach here

@wolflu05 I missed the URL rewrite approach in your previous PR, sorry to revert it now without reviewing it properly previously. However currently is broken and causing a few issues. Happy to revisit if we still want to be able to "route" to a  particular panel via URL, but maybe using `#` anchor or `?` query param?

### DIFF
--- a/src/frontend/src/App.tsx
+++ b/src/frontend/src/App.tsx
@@ -15,8 +15,14 @@ export function setApiDefaults() {
   const token = useSessionState.getState().token;
 
   api.defaults.baseURL = host;
-  api.defaults.headers.common['Authorization'] = `Token ${token}`;
+
+  if (token) {
+    api.defaults.headers.common['Authorization'] = `Token ${token}`;
+  } else {
+    api.defaults.headers.common['Authorization'] = null;
+  }
 }
+
 export const queryClient = new QueryClient();
 
 function checkMobile() {
--- a/src/frontend/src/components/nav/PanelGroup.tsx
+++ b/src/frontend/src/components/nav/PanelGroup.tsx
@@ -10,15 +10,8 @@ import {
   IconLayoutSidebarLeftCollapse,
   IconLayoutSidebarRightCollapse
 } from '@tabler/icons-react';
-import { ReactNode, useMemo } from 'react';
+import { ReactNode, useCallback, useMemo } from 'react';
 import { useEffect, useState } from 'react';
-import {
-  Navigate,
-  Route,
-  Routes,
-  useNavigate,
-  useParams
-} from 'react-router-dom';
 
 import { useLocalState } from '../../states/LocalState';
 import { PlaceholderPanel } from '../items/Placeholder';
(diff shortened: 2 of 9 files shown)


wolflu05:
Hey @SchrodingersGat ,

pushing the active panel into the url was one of the main reason why I created #5953. I needed that so that we can link from a machine (#4824) to a plugin that provides that machine. I then implemented the link between a user group in the user drawer to the referenced group drawer just by clicking on a link as an POC.

Same applies if you just want to send something to a colleague and link to the correct panel. That feature is not only relevant to the admin center. E.g. your college has opened the part notes as last panel, but you want to show him something with the part stock items. Clicking on a link that contains the stock items panel key enables us to open the stock items panel directly on the colleges side.

> Messes with router implementation too

What do you mean by that? That's the way of how to use the router as far as I know. I used the `useNavigate` hook to navigate between pages instead of using the `window.location.href` directly.

> Multiple "loops" - panel changes causing multiple component loads / slow

That was not intended, and I haven't realized that this is the case. But instead of completely removing my solution why not should we focus on fixing the problem with it?

> The page does not actually change, so URL feels like the wrong approach here

Selecting a different panel changes more than 65% of the page, and that is significant, more than the half. But you're right, we are still in the admin center, but if we manage the users or the project codes make a lot of difference in my opinion. Or if we see the parts stock or the part notes is also significant.

Implementing this has cost me a lot of free time and I also tried to make sure everything works seamlessly. But now I'm not quite sure what to make of this if that was removed from the code in less than a day without me having the chance to say something about it. But I know, you are the maintainer, so you can do what you think is the best.

---
But leaving all that behind, if you're interested in actually solving the problem, we could do that in a few ways (I'm also open to other ways):
1. Create an own `useParamId` hook, which only returns the id if it's an actually id and not a string like `parts` (otherwise it returns `undefined`). That way we could add `hasPrimaryKey:true` to the `useInstance` hook and if the id is `parts` or any other string, the query will not run.
2. Similar to the first idea, but navigate to `platform/part/category/index` when clicking on the parts nav item. Then check if the id is `index`, then use undefined for the pk so that no query is made.
3. Don't use path params, use query params instead. This would create issues with the detail drawer implementation, as that is using the path params too, so we would need to look back to the detail drawer and use a query param for that too. But the path then would look like this `/platform/settings/admin?panel=1&plugin=2` .

SchrodingersGat:
> That was not intended, and I haven't realized that this is the case. But instead of completely removing my solution why not should we focus on fixing the problem with it?

@wolflu05 you are right I was too hasty in undoing the work you had done. There was a particular problem that was introduced, which I will elaborate on further.

For now, I have introduced https://github.com/inventree/InvenTree/pull/6088 to revert these changes, and I'll open a new issue to discuss the particular problems with you further before moving on.

I'm sure that we can improve the current implementation (using URL paths) to work around the issues I have come up against.

Issue #6086, opened 2023-12-14
Tables with nothing checked assume everything is selected.


CodingPupper3033:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

In all of the tables I tested (Stock Items, Part List, Sub Locations), the custom_actions table buttons will apply to all items if none have been selected. (The warning about no items selected does not appear).

I believe I have found the issue; give me a few days to try figuring out the proper git commands and all that! If you're curious/I don't respond, the issue is a rouge exclamation point in the [tables.js](https://github.com/inventree/InvenTree/blob/master/InvenTree/templates/js/translated/tables.js) file, line 153, character 29 (it should be if there are no selections AND we do allow empty [return list]). Also, it appears the set parts category on the parts table doesn't have any sort of protection against no items selected, so I'll try to fix that as well!

### Steps to Reproduce

1. Go to any table (Stock Item, Part List, Sub Locations, etc)
2. Don't select anything
3. Click any of the action buttons (like transfer stock or something)
4. See it update everything

### Expected behaviour

It should warn you (and there is code to do so)

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

0.12.10

### Please verify if you can reproduce this bug on the demo site.

- [X] I can reproduce this bug on the demo site.

SchrodingersGat:
@CodingPupper3033 this behaviour is intentional - by default, an action will apply to all items in the table, or only specific items if the user has not selected any. There is no use case where it makes sense to apply an action to "zero selections".

Perhaps this is a bit obtuse? But definitely intetional

CodingPupper3033:
I see now, thanks! I would suggest needing at least one box checked and showing an error if not; however, I can see how this may make actions quicker for the user.

PR #6087, opened 2023-12-14
Fixes issue #6086

CodingPupper3033:
Fixed tables.js treating nothing selected as everything selected. Also added warning no selection warning to the Parts page set category button


SchrodingersGat:
@CodingPupper3033 as discussed in the related issue, this behaviour is intended. We are also currently moving the front-end to use react, so minor UI changes like this will not be implemented for the legacy UI

Issue #5866, opened 2023-11-05
Microsoft SSO fails with "Sign up is currently closed."


mittkkoo:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

I configured SSO with single-tenant Azure AD (Microsoft) as described [in the docs](https://docs.inventree.org/en/stable/settings/SSO/#enable-sso-settings)

This is in config.yaml:
```
# Add SSO login-backends (see examples below)
social_backends:
  - 'allauth.socialaccount.providers.microsoft'

social_providers:
  microsoft:
    TENANT: 'copied from https://entra.microsoft.com/#view/Microsoft_AAD_IAM/TenantOverview.ReactView'
``` 

This is the Settings / Login screen:

![image](https://github.com/inventree/InvenTree/assets/76665232/e2bcd40e-56f4-46a7-8484-e2a488922fe4)

The Button Microsoft Graph appers on the login screen and the Microsoft process is started, but afterwards inventree reports "Sign up is currently closed."

### Steps to Reproduce

1. Setup Microsoft SSO as in the docs
2. Add a social application through the UI
3. Change Settings / Login as in the screenshot

### Expected behaviour

Expect SSO to work.

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.12.8
Django Version: 3.2.19
Commit Hash: a7487ff8
Commit Date: None
Commit Branch: stable
Database: postgresql
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-5.4.0-166-generic-x86_64-with-glibc2.31
Installer: PKG
Target: ubuntu:20.04
Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}]

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

SchrodingersGat:
Can you share a (redacted) screenshot of the SSO application you setup in the admin interface? Ensure that the name of the application is `microsoft` (case sensitive)

mittkkoo:
Yes, microsoft
[...]

Issue #6037, opened 2023-12-05
[BUG] OIDC provider does not appear on login screen


WERTBON1789:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

It seems that the allauth openid_connect provider doesn't show up as a login option, but others (like microsoft) do

### Steps to Reproduce

configure social_backends in config.yaml:
```
social_backends:
  - 'allauth.socialaccount.providers.openid_connect'
  - 'allauth.socialaccount.providers.microsoft'
```

and even configure openid_connect, like this
```
social_providers:
  openid_connect:
    APPS: [
      {
        "provider_id": "authelia",
        "name": "Authelia",
        "client_id": "inventree",
        "secret": "<secret>",
        "settings": {
          "server_url": "https://sso.example.com"
        }
      }
    ]
```
or like this (not quite sure which if there is a problem in this config)
```
social_providers:
  openid_connect:
    APPS:
      - provider_id: authelia
        name: Authelia
        client_id: inventree
        secret: '<secret>'
        settings:
          server_url: https://sso.example.com
```

### Expected behaviour

The openid_connect provider should show up like the microsoft one does

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

Docker: inventree/inventree:latest (pulled today)
Inventree: 0.13.0dev
Python: 3.10.13

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

SchrodingersGat:
Seems related to https://github.com/inventree/InvenTree/discussions/5966

Callout to anyone who has got this working?

SchrodingersGat:
@WERTBON1789 this should have been fixed in recent updates. Can you possess try with latest version and see if the issue has been rectified?

mp-strachan:
@SchrodingersGat Hi Oliver, Tried this today and can confirm that this is now working.

SchrodingersGat:
Thanks @mp-strachan that's great news

Issue #4715, opened 2023-04-27
[FR] Make OIDC SSO configurable via API/while running


matmair:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

Currently, you need to reach into config files to add/change OIDC settings. Especially with regards to #4562 / https://github.com/p-fruck/inventree-federation-plugin this makes the config harder.

### Suggested solution

Add a way to change these settings
- on the fly / with the database
- expose it to the API

### Describe alternatives you've considered

The current way works but get tedious.

### Examples of other systems

You can change SSO settings in nextcloud on the fly.

### Do you want to develop this?

- [X] I want to develop this.

matmair:
This would benefit from #2326 being available security wise; that can be added later without problems

matmair:
still relevant

matmair:
still relevant

matmair:
still relevant

SchrodingersGat:
Not stale

matmair:
Still open; important if we want to have click and go docker-compose deployments

matmair:
Still important, no specific timeline yet available

matmair:
still important

matmair:
Still required

matmair:
Working on this now

SchrodingersGat:
@matmair do you think this will be available for 1.0.0?

matmair:
Yes, the required prerequisites are now there

SchrodingersGat:
Do you have a development strategy in mind here? Should we delay for this, or hold it back for 1.1.0?

matmair:
Not important, moving to 1.1

matmair:
fixed by https://github.com/inventree/InvenTree/pull/12799

PR #6099, opened 2023-12-15
Update django-allauth version

SchrodingersGat:
### Overview

This PR provides updates to the django-allauth dependency - recent library updates have fixed some significant issues with social authentication

Closes https://github.com/inventree/InvenTree/issues/5866

### References

- https://github.com/inventree/InvenTree/discussions/5966#discussioncomment-7787934
- https://github.com/inventree/InvenTree/issues/6037
- https://docs.allauth.org/en/latest/release-notes/recent.html
- 
### Further Reading

From the [django-allauth documentation](https://docs.allauth.org/en/latest/socialaccount/provider_configuration.html#provider-configuration) it appaers that our current documentation is incorrect. We suggest a mix of configuring provider info in `settings.py` (via `config.yaml`) and also adjusting settings in the database (using the `SocialApp` model). The docs advise to use either one or the other.

It looks like the `SocialApp` approach is "friendlier" (can be done via the web interface) but the `config.yaml` approach is better from a security perspective - as there are no "leaks" of secrets such as client secrets into the database.

### DIFF
--- a/InvenTree/InvenTree/forms.py
+++ b/InvenTree/InvenTree/forms.py
@@ -359,6 +359,13 @@ class CustomSocialAccountAdapter(CustomUrlMixin, RegistratonMixin, DefaultSocial
         # Otherwise defer to the original allauth adapter.
         return super().login(request, user)
 
+    def authentication_error(self, request, provider_id, error=None, exception=None, extra_context=None):
+        """Callback method for authentication errors."""
+
+        # Log the error to the database
+        log_error(request.path if request else 'sso')
+        logger.error("SSO error for provider '%s' - check admin error log", provider_id)
+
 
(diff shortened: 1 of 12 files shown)


mianumar:
Now it's look fine

SchrodingersGat:
@mianumar did you pull down these changes and confirm?

matmair:
@SchrodingersGat does this also address https://github.com/inventree/InvenTree/issues/4715 / could it be adapted to do that?

SchrodingersGat:
@matmair I'm not sure I fully understand the problem in #4715 

Is it just that the OIDC settings need to be editable via the database? I think that if the admin sets up an SSO app on the db then it should be possible right?

Issue #6092, opened 2023-12-15
Receipt of Return Order Line does not clear SO Allocation


mcollins-sol:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

After a serialized stock item is received against a Return Order line, it can't be allocated to another SO even after it has been inspected/repaired and Status=OK.  The allocation against the original SO isn't cleared when the Return Order line is received.
Adding **_stock_item.clearAllocations()_** under the Update Stock Items section of function **_InvenTree/order/models.py  ReturnOrder.receive_line_item_**   solves this issue.  I made this change at current line# 1929 and verified in our test environment.

Curiously, although **_StockItem.clearAllocations_** function is defined, it doesn't appear to be called anywhere yet.

### Steps to Reproduce

1. Create a SO #1
2. Add SO Line of a Serialized Part P1
3. Issue SO #1
4. Create Pending Shipment
5. Allocate a serialized stock item to the SO Line
6. Complete Shipment
7. Complete SO #1
8. Create a Return Order for same customer
9. Issue Return Order
10. Receive Return Order Line into a Stock Location
11. Stock Tracking history, status, quantity, location all look fine, however Stock Item is still allocated to SO #1071 
12. Edit Stock Item, set Status=OK to indicate it has been inspected/repaired and make it an Available Status
13. Create SO #2
14. Add SO Line of the same Serialized Part P1
15. Issue SO
16. Create Pending Shipment
17. Attempt to Allocate the returned serialized stock item to the SO Line; it is not in the list of Stock Items available to allocate
18. Via Admin interface, find and delete the specific SO Allocation
19. Repeat Allocation of the returned serialized stock item to the SO Line; it is now available to allocate

### Expected behaviour

Receipt of a Return Order Line should clear SO Allocations.  It is in a Stock Location with Status=Quarantined and should not be allocated to the SO.  Once the returned Stock Item Status Code is changed to one that's in the Available list, it should be Available for allocation on other SO Lines.

### Deployment Method

- [ ] Docker
- [ ] Bare metal

### Version Information

0.12.8

### Please verify if you can reproduce this bug on the demo site.

- [X] I can reproduce this bug on the demo site.

SchrodingersGat:
@mcollins-sol your suggested solution seems reasonable - can you submit a PR to address this?

mcollins-DL:
@SchrodingersGat  I would love to help - I'm in awe of what you've accomplished.
[...]

PR #6117, opened 2023-12-18
Fix to clear SO Allocations on receipt of Return Order Line Item

mcollins-DL:
Added stock_item.clearAllocations() to order/models.py ReturnOrder.receive_line_item
Fixes issue #6092


SchrodingersGat:
@mcollins-DL please address style issues :) You should also check out [our contributing guide](https://github.com/inventree/InvenTree/blob/master/CONTRIBUTING.md#setup-devtools) - you can setup so that most of these style issues are fixed automatically on each git commit

SchrodingersGat:
Thanks @mcollins-DL for the fix

PR #6126, opened 2023-12-21
Enable existing_image on PATCH requests

### LINKED ISSUE #6121: [bug] existing_image field is ignored for API patch requests

### Describe the bug*

Hi,

I am very thankful for the implementation of the existing_image endpoint, as it saves us a lot of storage space. However I noticed a problem while I tried to update existing parts: It seems like the existing_image field is simply ignored for patch requests. I tried it with the python API as well as with direct requests and neither seems to do anything. An example request would look like that:

```
 payload = {'params': {'format': 'json'}, 'timeout': 10, 'headers': {'AUTHORIZATION': 'my_token'}, 'auth': None, 'proxies': {}, 'json': {'existing_image': '2881_thumbnail.jpeg'}}
requests.patch('https://inventree-test/api/part/2882/', **payload)
```

The request does not fail, it returns a `200` response with the part data unchanged.

### Steps to Reproduce

Try to change a image with the `existing_image` field with an API path request.

### Expected behaviour

The part image should change as according to https://github.com/inventree/InvenTree/pull/6021#issue-2021905841 the `existing_image` field should work for existing parts as well as for new ones (where it does work).

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.13.0
Django Version: 3.2.23
Commit Hash: 301afaa
Commit Date: 2023-12-15

Database: postgresql
Debug-Mode: True
Deployed using Docker: False
Platform: Linux-5.15.0-91-generic-x86_64-with-glibc2.35
Installer: GIT

Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}, {'name': 'InvenTreeCurrencyExchange', 'slug': 'inventreecurrencyexchange', 'version': '1.0.0'}, {'name': 'InvenTreeLabel', 'slug': 'inventreelabel', 'version': '1.0.0'}, {'name': 'InvenTreeLabelSheet', 'slug': 'inventreelabelsheet', 'version': '1.0.0'}, {'name': 'DigiKeyPlugin', 'slug': 'digikeyplugin', 'version': '1.0.0'}, {'name': 'LCSCPlugin', 'slug': 'lcscplugin', 'version': '1.0.0'}, {'name': 'MouserPlugin', 'slug': 'mouserplugin', 'version': '1.0.0'}, {'name': 'TMEPlugin', 'slug': 'tmeplugin', 'version': '1.0.0'}, {'name': 'SupplierCart', 'slug': 'suppliercart', 'version': '0.1.0'}]

### Relevant log output

_No response_

LavissaWoW:
Closes #6121

### DIFF
--- a/InvenTree/part/serializers.py
+++ b/InvenTree/part/serializers.py
@@ -604,7 +604,7 @@ class PartSerializer(InvenTree.serializers.RemoteImageMixin, InvenTree.serialize
         return reverse_lazy('api-part-list')
 
     def skip_create_fields(self):
-        """Skip these fields when instantiating a new Part instance"""
+        """Skip these fields when changing an existing Part instance"""
         fields = super().skip_create_fields()
 
         fields += [
@@ -612,7 +612,6 @@ class PartSerializer(InvenTree.serializers.RemoteImageMixin, InvenTree.serialize
             'initial_stock',
             'initial_supplier',
             'copy_category_parameters',
-            'existing_image',
         ]
 
         return fields
--- a/InvenTree/part/serializers.py
+++ b/InvenTree/part/serializers.py
@@ -592,7 +592,10 @@ class PartSerializer(InvenTree.serializers.RemoteImageMixin, InvenTree.serialize
 
         if not create:
             # These fields are only used for the LIST API endpoint
-            for f in self.skip_create_fields()[2:]:
+            for f in self.skip_create_fields():
+                # Fields required for certain operations, but are not part of the model
+                if f in ["remote_image", "existing_image"]:
+                    continue
                 self.fields.pop(f)
 
         if not pricing:
@@ -607,15 +610,12 @@ class PartSerializer(InvenTree.serializers.RemoteImageMixin, InvenTree.serialize
         """Skip these fields when instantiating a new Part instance"""
         fields = super().skip_create_fields()
 
-        # Placed by itself to prevent changing of order in array
-        # This to prevent being popped by call in self.__init__
-        fields += ['existing_image']
-
         fields += [
             'duplicate',
             'initial_stock',
             'initial_supplier',
             'copy_category_parameters',
+            'existing_image',
         ]
 
         return fields
--- a/InvenTree/part/serializers.py
+++ b/InvenTree/part/serializers.py
@@ -607,9 +607,12 @@ class PartSerializer(InvenTree.serializers.RemoteImageMixin, InvenTree.serialize
         """Skip these fields when instantiating a new Part instance"""
         fields = super().skip_create_fields()
 
+        # Placed by itself to prevent changing of order in array
+        # This to prevent being popped by call in self.__init__
+        fields += ['existing_image']
+
         fields += [
             'duplicate',
-            'existing_image',
             'initial_stock',
             'initial_supplier',
             'copy_category_parameters',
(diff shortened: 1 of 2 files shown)


LavissaWoW:
@SchrodingersGat Due to the docstring anomaly, it looks like the tests you made for this didn't run properly. 
I'll look into it. No need to merge before all CI tests are green :)
\

SchrodingersGat:
Ok, the issue here is the reason that "skip_create_fields" was added in the first place. The "existing _image" field does not exist on the "Part" model, so it needs to be removed from the dataset before the model is saved. So this field will need to be manually popped in the custom save method

LavissaWoW:
Nope. That doesn't work either.

LavissaWoW:
@SchrodingersGat The latest commit solves the issue, but I'm not sure I like it.

```python
for f in self.skip_create_fields()[2:]:
    self.fields.pop(f)
```
is very prone to breaking if something is altered with the `skip_create_fields` method. I.e. changing the order of fields, adding new fields "in-between" etc. Even adding fields to `super()` calls will cause this to "bug" out.

Perhaps it would be better to whitelist fields we want to keep inside the for loop?

```python
for f in self.skip_create_fields():
    if f not in some_whitelist:
        self.fields.pop(f)
```

InvenTree/part/serializers.py
```diff
@@ -592,7 +592,7 @@ def __init__(self, *args, **kwargs):
 
         if not create:
             # These fields are only used for the LIST API endpoint
-            for f in self.skip_create_fields()[1:]:
+            for f in self.skip_create_fields()[2:]:
```
SchrodingersGat, line 595: Instead of using the index offset here (`[2:]`) I would suggest doing a string comparison:

```python
for f in self.skip_create_fields():
    if f == 'existing_image':
        # existing_image field is also support for PATCH requests, but does not exist on the model
        continue
    self.fields.pop(f)
```

This way we do not have to enforce list ordering, and it is clear *why* we are doing this custom operation

SchrodingersGat (changes requested):

SchrodingersGat:
In addition to the change above, can you also please add a unit test to ensure that the new functionality works:

1. Create a part with an image (Part A)
2. Update another part (Part B) to reference the image of Part A
3. Update another part (Part C) with an *invalid* `existing_image` field value, to ensure that it fails correctly

LavissaWoW:
Can do!

InvenTree/part/serializers.py
LavissaWoW: Looking at the indexing issue, the original index of `[:1]` was put there to assure that the `remote_image` field of the `RemoteImageMixin` isn't removed either. It's due to this that I proposed a potential whitelist concept.

Then, you could formulate the loop like this:

```python
for f in self.skip_create_fields():
    if f in ["remote_image", "existing_image"]:
        continue
    self.fields.pop(f)
```

SchrodingersGat: That makes sense. Much cleaner and obvious outcome!

LavissaWoW:
This should suffice :)

SchrodingersGat (approved):

Issue #5995, opened 2023-11-27
[FR] Test statistics


martonmiklos:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

It would be great if statistics could be made from test results easily.
I am not only  thinking about from the part perspective view like suggested in the #821, but from other, higher perspective (per test templates for e.g.)  or in general (see later).

What I am thinking about:

Test failure type analysis:
As I read about the test reporting documentation I realized that we use somewhat differently the test results concepts what it was originally designed:
We run test sequences (composed from several test steps). If any of the test steps failing the sequence terminated immediately and get logged in the following manner:  I store the failed step name in the value field. In the case if all steps passing a passed result will be logged w/o any value. 
I do not know if this InvenTree usage approach should be follow-able, but logging each step as a separate test result would not make much sense in our case (we have looped test steps with a few K iterations for e.g.).
Anyway: it would be good if the test results count could be grouped and counted by the failure reason. If we use the value for this calculation or we introduce a failure reason field for the test run is up for discussion. The calculation could be filtered by test run date (time interval), part type or test template.

Retest statistics
Sometimes it is important to know how "smoothly" going the testing at the manufacturing floor. It would be great if the test results could include a retest field which would be zero for the first test results saved for a specific stock item for the given test template, and would be set to true in the case if the test is executed again for the same stock item. From this retest flag we can make some statistics about the retest metrics to see how stable the testing goes on the floor.

Test equipment statistics
It would be great if the test equipment information (test station number for e.g.) could be logged to each test result and some metrics could be created to see how much the specific stations are utilizeed. (See next point.)

Execution time statistics
There are tests which require operator intervention which contributes to the testing processes capacity. It would be great if the test execution time (start/stop) could be stored for each test result, and some statistics (min, max, average, distribution of the test time) could be created per test template (with filtering capability on time intervals, or specific build orders). It would be also useful to accumulate the test times grouped by test stations to see the given stations utilization.

### Suggested solution

First I would like to discuss if storing more data (test time, station info, failure step) makes sense for the fellow developers.  I know that our "InvenTree test adoptation" was went in a somewhat different way as InvenTree used at different places, but I think these additions would makes sense at many other users.
If we agree upon something in this area I think this is something what I can help to move forward.

[...]

PR #6149, opened 2024-01-02
Add additional test fields

martonmiklos:
As we discussed in the #5995 I would like to add the following extra fields to the test results:
- test station name
- test start/finish time (to handle use cases where the test execution is asynchronous from the  logging)
- retest (for improving the performance of the statistics creation)

I have an another branch where I built some statistics UI on the top of this:
kép

Once we iron out this I will PR those changes as well.


SchrodingersGat:
In general I am happy with these additions, except for the "retest" field. I think that this could be easily calculated dynamically by counting the number of test results (for a single stock item) which have the same test key...

martonmiklos:
> In general I am happy with these additions, except for the "retest" field. I think that this could be easily calculated dynamically by counting the number of test results (for a single stock item) which have the same test key...

My intention with the retest "caching" is the fact that we are mainly interested in the retesting percentage per part/test type/build order. In these queries we could have a plenty of test results: sometimes we manufacture 1K Build orders atm. I have 270K serial numbers imported in InvenTree.

martonmiklos:
> In general I am happy with these additions,

One more question came into my mind the (old) UI:
The new test station, test started/finished date columns are hidden by default:

kép

I would rather sort the results by the Test finished date (which is identical if no other date passed) and would rename the "Test date" to "Result save date" or something similar. This would always allow the test execution process to be overviewed in the "natural flow" regardless if the test results were uploaded asynchronously or mixed orders.

The test "Start date" and the "Result save date" columns could be hidden by default to spare "screen estate".

martonmiklos:
> In general I am happy with these additions, except for the "retest" field. I think that this could be easily calculated dynamically by counting the number of test results (for a single stock item) which have the same test key...

@SchrodingersGat 
May I ask you reflect my comments above? I would like to push this through even without the retest field if you really against that.

SchrodingersGat:
@martonmiklos I would like to see this move forward too. After another chance to look at it:

### Timestamping

We already have the `date` field (which records the datetime timestamp when the test was recorded). Is this sufficient for recording the "test end time"? Otherwise we are potentially duplicating this data. Would like to hear your thoughts on this.

### Retest Field

Would you be opposed to a dynamic calculation which calculates percentage of failed tests on the fly? I'm still not 100% sure on your proposed usage for this field

martonmiklos:
> @martonmiklos I would like to see this move forward too. After another chance to look at it:
> ### Timestamping
> 
> We already have the `date` field (which records the datetime timestamp when the test was recorded). Is this sufficient for recording the "test end time"? Otherwise we are potentially duplicating this data. Would like to hear your thoughts on this.

The date field currently holds the date when the test was uploaded to InvenTree, which is not necessary the same as the execution start date. (In the case if the test result uploading is done asynchronously.) Storing the end date might also help making statistics (i.e. some operators performing testing way slower than others etc.)

> ### Retest Field
> 
> Would you be opposed to a dynamic calculation which calculates percentage of failed tests on the fly? I'm still not 100% sure on your proposed usage for this field

You are right, I will remove the field and do the calculations on the fly.

SchrodingersGat:
@martonmiklos ok, happy if you keep the new date fields, and remove the "retest" field. Please address conflicts in the current PR.

Can you also update the new testresulttable in the react interface?

https://github.com/inventree/InvenTree/blob/master/src/frontend/src/tables/stock/StockItemTestResultTable.tsx

martonmiklos:
> @martonmiklos ok, happy if you keep the new date fields, and remove the "retest" field. Please address conflicts in the current PR.
> 
> Can you also update the new testresulttable in the react interface?
> 
> https://github.com/inventree/InvenTree/blob/master/src/frontend/src/tables/stock/StockItemTestResultTable.tsx

@SchrodingersGat 

I managed to rebase to master. I originally planned to add the UI changes in a separate PR (of course I will update the new UI too). What do you think?

SchrodingersGat (approved):

SchrodingersGat:
@martonmiklos thanks for the contribution! Please submit a separate PR to add these new fields to the test result table in the react interface!

martonmiklos:
> @martonmiklos thanks for the contribution! Please submit a separate PR to add these new fields to the test result table in the react interface!

Thank you for your patience and the merging!

PR #6158, opened 2024-01-05
Fix supplier barcode order numbers

30350n:
Seems like the logic for retrieving a purchase order from an order number from a supplier barcode got broken when refactoring. This fixes the logic and also some supplier specific details.

### DIFF
--- a/InvenTree/plugin/base/barcodes/mixins.py
+++ b/InvenTree/plugin/base/barcodes/mixins.py
@@ -351,10 +351,11 @@ class SupplierBarcodeMixin(BarcodeMixin):
         supplier_reference_filter = Q(supplier_reference__iexact=supplier_order_number)
 
         orders_union = orders.filter(reference_filter | supplier_reference_filter)
-        if orders_union.count() == 1:
+        orders_intersection = orders.filter(reference_filter & supplier_reference_filter)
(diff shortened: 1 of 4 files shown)


30350n:
(I'll add some test cases later to confirm this keeps working in later versions.)

SchrodingersGat:
@30350n can you please provide some more specific information about how the current implementation is "broken"?

30350n:
Sure!

1. The current implementation always checks for **both** the PO and CPO matching, **if** both are present in the barcode. Having a match on one of them should be sufficient.
2. LCSCs and TMEs barcode POs were incorrectly labeled as CPOs. TME actually has a CPO field though, which hasn't been used.
3. Mouser uses the ecia barcode 'K' field for **both**, POs and CPOs interchangeably (yeah I know, don't ask me why).

30350n:
Actuall the implementation I came up is kind of convoluted I guess, I'll try to refactor it a bit.

SchrodingersGat:
@30350n if you have the data on hand to test that this works, I'm satisfied with that. please keep an eye on whether this continues to work

30350n:
Thanks, will definitely do! I actually also wanted to confirm everythings actually working correctly with some tests but didn't find time to implement them yet. I'll try to get that down the coming days and create another PR 👍

PR #6162, opened 2024-01-06
Fix reassigning supplier barcodes

30350n:
Because of the way the current `BarcodeAssign` view is implemented, it's not possible to reassign 3rd party supplier barcodes anymore, if they get successfully matched.

To mitigate this there's two possible approaches:

#### 1. Only check the internal InvenTreeBarcode plugin in `BarcodeAssign`

I can't really think of a reason why *any* barcode recognized by a 3rd party plugin shouldn't be able to be reassigned to an InvenTree object. The only problematic case is the same barcode data being directly assigned to 2 different objects (which is something only the internal barcode plugin should be able to do I'd guess?).
(This is what I commited here).

#### 2. Add some kind of marker to `SupplierBarcode` plugins so they can get ignored in `BarcodeAssign`

I guess the simplest form of this would just be getting all of them via `registry.with_mixin('supplierbarcode')`.

*Note: Looking at the code again I guess this would also fix itself if the supplier barcode plugins wouldn't be builtin ...*


SchrodingersGat:
@30350n thanks for the fix to address this - I think that your approach here is good.

I'm not sure based on your comments above if you have any more work to do, or happy for me to merge in?

30350n:
Sure, if you don't see any obvious problem with that approach too, then it should be good to go ^^

SchrodingersGat:
If you can fix the conflicting files I will merge!

30350n:
Done! (urgh, single quotes ...)

30350n:
Uhhh, now pre-commit is complaining about stuff I didn't even touch. Guess that's no bueno?

SchrodingersGat:
We just merged in some changes to our style checks which are now a bit more strict. It looks like only a few adjustments needed?

30350n:
Well I already fixed up all the code this PR touches like I said. Anything else shouldn't be part of it.

SchrodingersGat:
Once https://github.com/inventree/InvenTree/pull/6203 is merged you should be able to simply merge master back into this PR and then it will pass

30350n:
Perfect, that seemed to work, thanks you!

SchrodingersGat:
Thanks @30350n :)

PR #6165, opened 2024-01-06
Added update notification

matmair:
Added notification to superusers when possible InvenTree updates are detected


matmair:
@SchrodingersGat thoughts?

SchrodingersGat:
LGTM @matmair

PR #6167, opened 2024-01-07
CI updates

matmair:
This targets to make the CI faster and simpler:
- replaced flake8 with ruff (mostly for speed improvements but also autoformatting)
- moves djlint config to pyproject
- moves isort config to pyproject
- removes flake8 from CI and requirements (not necesarry anymore due to ruff)


SchrodingersGat:
@matmair LGTM - any idea why the `Style [Python]` CI check is not completing?

matmair:
@SchrodingersGat that step was a duplicate and is being removed with this PR so it is not runinnig -> thus not passing

matmair:
You would need to adapt the protection rules and remove that one (or replace it with the pre-commit step) - I do not have access to those settings

SchrodingersGat:
Ok, fair enough. If you can fix the conflicting files I'm happy to merge

matmair:
@SchrodingersGat done

SchrodingersGat:
Nice, thanks @matmair

PR #6173, opened 2024-01-07
Bump to Dj 4.x

matmair:
Close https://github.com/inventree/InvenTree/issues/2853
Follow up to https://github.com/inventree/InvenTree/pull/4584; Switch the code base to Django 4.2 (LTS) as 3.x is EOL in 2 months.


matmair:
See https://endoflife.date/django

matmair:
@inventree/triage any idea why the postgres stuff is failing?

SchrodingersGat:
There seem to be a lot of errors like:

```
psycopg.errors.GroupingError: column "u0.tree_id" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: ...id" = ("stock_stocklocation"."tree_id")) ORDER BY U0."tree_i...
```

As far as I can see, they are related to the `tree_id` column (which is a *MPTT* column). 

May be related to this ticket:

https://code.djangoproject.com/ticket/34551

SchrodingersGat:
@matmair a critical error message in the logs:

```
2024-01-14T23:09:29.0582382Z Error rebuilding PartCategory objects 'Manager' object has no attribute 'rebuild'
...
2024-01-14T23:09:30.2364607Z Error rebuilding Part objects 'Manager' object has no attribute 'rebuild'
...
2024-01-14T23:11:55.8170298Z Creating test database for alias 'default'...
2024-01-14T23:11:59.5715187Z Rebuilding PartCategory objects
2024-01-14T23:11:59.5716479Z Error rebuilding PartCategory objects 'Manager' object has no attribute 'rebuild'
2024-01-14T23:12:01.0836518Z Rebuilding Part objects
2024-01-14T23:12:01.0837432Z Error rebuilding Part objects 'Manager' object has no attribute 'rebuild'
2024-01-14T23:12:06.2310149Z Rebuilding StockLocation objects
2024-01-14T23:12:06.2311235Z Error rebuilding StockLocation objects 'Manager' object has no attribute 'rebuild'
2024-01-14T23:12:07.4533116Z Rebuilding StockItem objects
2024-01-14T23:12:07.4534569Z Error rebuilding StockItem objects 'Manager' object has no attribute 'rebuild'
2024-01-14T23:12:10.2241460Z Rebuilding Build objects
2024-01-14T23:12:10.2242441Z Error rebuilding Build objects 'Manager' object has no attribute 'rebuild'
2024-01-14T23:12:40.9884304Z Added field checkbox to model partparametertemplate
```

It appears that the MPPT manager method to rebuild a tree does not function correctly for django-4.x - we will need to address this!

matmair:
@SchrodingersGat there is a bunch of erros left, especially with Postgres. The manager is accessible after the migrations, just not during them. Still trying to figure out what changed there

matmair:
I can not figure out these wired issues with postgres and it seems like the maintainer of django-mptt is not really happy that he still has to use his libary (https://github.com/django-mptt/django-mptt/issues/833) so either we try fixing the postgres stuff or we switch to something like https://github.com/django-treebeard/django-treebeard.
I am not sure what is more prudent - we really need to upgrade away from Django 3 before the EOL hits us. Thoughts @SchrodingersGat / @wolflu05

SchrodingersGat:
I do not think there is a good alternative out there to `django-mptt`. It is a *great* library. Migrating to treebeard would take a *lot* of effort and I'm not even sure it would work how we want it to. I would much rather support `django-mptt` if we can:

image

It looks like he is actually making new tags: https://github.com/django-mptt/django-mptt/releases/tag/0.16

And it looks like the latest code *does* (or *should*) support django-4.2:

https://github.com/django-mptt/django-mptt/commit/f70628be6df71417e42af382e6e27b443c4bfa6d

SchrodingersGat:
@matmair I have spent some time looking into this. I do not have a solution yet but I have tracked down at least one of the core errors which pops up if running tests in postgresql.

Note that the error only occurs in the unit test runner - not if just running the code "normally". Not sure what that is about

This error:

image

Can be traced to this function:

https://github.com/inventree/InvenTree/blob/676bb02f6ed5d723c08e2a16b157037ec1a47fbf/InvenTree/part/filters.py#L237

The following references make some suggestions as to how this might be tackled, but I'm not sure how to apply it here:

- https://copyprogramming.com/howto/column-users-id-must-appear-in-the-group-by-clause-or-be-used-in-an-aggregate-function
- https://stackoverflow.com/questions/52969103/django-programmingerror-must-appear-in-the-group-by-clause-or-be-used-in-an-aggr
- https://stackoverflow.com/questions/75666366/programmingerror-column-must-appear-in-the-group-by-clause-or-be-used-in-an
- https://learnsql.com/blog/must-appear-in-group-by-clause/
- https://stackoverflow.com/questions/66065987/postgres-sql-column-must-appear-in-the-group-by-clause-or-be-used-in-an-aggrega

matmair:
@SchrodingersGat are you sure we can remove the rebuilds safely?

SchrodingersGat:
Only 1 error now

matmair:
IMO the flag for the custom manger support is just not present on django-mptt. Adding it adds a bunch of migrations so I did not commit that.
If we get it to run in this state I think we should just merge and not care about it. The affected migrations are pretty old

SchrodingersGat:
> The affected migrations are pretty old

We should look seriously into squashing migrations before a certain date, back a few major releases.

matmair:
@SchrodingersGat I think we should do stuff like that with the 1.0. Maybe make a 0.99 release before and add a check that everything was updated to that and have some breaking changes in the transition to 1.0 like removing old migrations, moving config files out of the code source, move plugins dir etc.

matmair:
FYI Debian 11 (Python 3.9, current min) is EOL 2024-08-15 so I think after that we could switch to Python 3.11 (default for Debian 12) or at least Python 3.10 (default for Ubuntu 22.04 which was released April 2022)

SchrodingersGat:
> FYI Debian 11 (Python 3.9, current min) is EOL 2024-08-15 so I think after that we could switch to Python 3.11 (default for Debian 12) or at least Python 3.10 (default for Ubuntu 22.04 which was released April 2022)

@matmair happy with that. At some point we need to consider supporting python 3.12 - but that has only just been supported officially by django

matmair:
@SchrodingersGat can you check the TZ addition in the notification age? Seems to work for me but want to be sure. Should be ready for review otherwise.
No idea why Deepsource is not working again - maybe we should switch to [SonarCloud](https://sonarcloud.io/project/overview?id=matmair_InvenTree)

SchrodingersGat (approved):

SchrodingersGat:
@matmair LGTM - glad that we finally got this one worked out :)

matmair:
I was getting nervous about the End of Life approaching; 0.14 is shaping up to be a good release.

SchrodingersGat:
I've actually put a release data against `0.14.0` (end of Feb) so that we have something to work to :)

SchrodingersGat:
Note (https://github.com/inventree/InvenTree/pull/6453) that this PR requires some changes to the setup configuration if running behind a proxy

PR #6178, opened 2024-01-09
Allow orders' creation date edit via API

### LINKED ISSUE #6159: Question: Importing Data with Creation Dates
I have a relatively simple question, but one that is really important to my use case. I've spent some time writing import scripts to bring data over from a legacy system into InvenTree. I'm doing so via Python using InvenTree's API to create objects.

However, when importing Purchase Order information, I noticed there is no documented way to create a Purchase Order with a custom creation date. It obviously would be ideal if I could preserve dates during my import, both for POs and other types of objects. Is there a recommended way of doing so?

I'm using the internal API documentation at `/api-doc/#tag/order/operation/order_po_.*_create`

Thank you in advance

martonmiklos:
Fixes #6159


SchrodingersGat:
Thanks for tackling this one @martonmiklos :)

martonmiklos:
> @martonmiklos could you merge in master and check if it works now? Would love to assist in debugging this if the error persists

Done, let's see how the CI will progress this time.

matmair:
From my POV this is ready for a review, If you agree please mark it as ready @martonmiklos

martonmiklos:
> From my POV this is ready for a review, If you agree please mark it as ready @martonmiklos

Yes, I have no idea why the CI failed with timeout on DeepSource.
It was a lesson learned for me about the behavior of the serializers :)

matmair:
Deepsource is mainly code quality/security stuff and can be ignored if the rest of the CI passes. It seems like there is some kind of issue on their server end.

matmair:
@SchrodingersGat this is ready for review and merge; no idea why deepsource is timing out but this should not be blocking IMO.

SchrodingersGat:
@martonmiklos thanks for the contribution :)

Issue #6213, opened 2024-01-12
PATCH method on BuildReport does not work


miggland:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

API Documentation states that endpoint
`/api/report/build/{id}/`
supports PATCH method, as do all other Report methods (ReportBoM,  ReportPurchaseOrder, ReportSalesOrder, ReportReturnOrder, ReportStockLocation, ReportTest)

### Steps to Reproduce

```
$ curl --basic -u admin:inventree -X PATCH -F name="Changed title for test2" https://demo.inventree.org/api/report/build/1/
{"detail":"Method \"PATCH\" not allowed."}
```

Other report methods:
**ReportBoM**
```
$ curl --basic -u admin:inventree -X PATCH -F name="Changed title for test2" https://demo.inventree.org/api/report/bom/1/
{"pk":1,"name":"Changed title for test2","description":"Bill of Materials report","template":"/media/report/inventree/bom/inventree_bill_of_materials_report.html","filters":"","enabled":true}
```
**ReportPurchaseOrder**
```
$ curl --basic -u admin:inventree -X PATCH -F name="Changed title for test2" https://demo.inventree.org/api/report/po/1/
{"pk":1,"name":"Changed title for test2","description":"Purchase Order example report","template":"/media/report/inventree/purchaseorder/inventree_po_report.html","filters":"","enabled":true}
```
**ReportSalesOrder**
```
$ curl --basic -u admin:inventree -X PATCH -F name="Changed title for test2" https://demo.inventree.org/api/report/so/1/
{"pk":1,"name":"Changed title for test2","description":"New description","template":"/media/report/report_template/salesorder/dummytemplate.html","filters":"","enabled":true}
```
**ReportReturnOrder**
```
$ curl --basic -u admin:inventree -X PATCH -F name="Changed title for test2" https://demo.inventree.org/api/report/ro/1/
{"pk":1,"name":"Changed title for test2","description":"Return Order example report","template":"/media/report/inventree/returnorder/inventree_return_order_report.html","filters":"","enabled":true}
```
**ReportStockLocation**
```
$ curl --basic -u admin:inventree -X PATCH -F name="Changed title for test2" https://demo.inventree.org/api/report/slr/1/
{"pk":1,"name":"Changed title for test2","description":"Stock Location example report","template":"/media/report/inventree/slr/inventree_slr_report.html","filters":"","enabled":true}
```
[...]

PR #6214, opened 2024-01-12
Add tests for report API

miggland:
Adds more tests for report and label API.

Prompted by https://github.com/inventree/InvenTree/issues/6213


miggland:
Currently stuck with 403 results

SchrodingersGat:
Something strange is going on with the permissions. The API endpoints should determine required permissions based on the model type. You can tweak the permissions assigned to the test user here:

https://github.com/inventree/InvenTree/blob/9b1a310ffe2416ed8e7ffd282ec36103e301e844/InvenTree/InvenTree/unit_test.py#L111

Note that when I run these tests with `superuser = True` added to the test class, the 403 errors go away.

miggland:
I was expecting the PATCH on build to fail, since I have not merged the the fix from #6218 into this branch yet. However, it seems to work through the API test methods.

The data-export failure is another surprise though.. manually running `export-records` and then `import-records` seems to work fine.

SchrodingersGat:
@miggland please note I am working on a fix for the data importing errors:

https://github.com/inventree/InvenTree/pull/6253

miggland:
I added similar tests for labels as well, since those api tests were limited (and marked with "TODO".. )

SchrodingersGat:
Thanks @miggland !!

Issue #6071, opened 2023-12-11
News Feed task doesn't work behind proxy, impacting performance


LavissaWoW:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

The `update_news_feed` task attempts to fetch the RSS/Atom feed once daily. This, however, doesn't work behind a proxy server.

The result is that these tasks occupy workers all the time, and never complete.
Each worker is terminated roughly every 90 seconds due to this.

### Steps to Reproduce

1. Put the InvenTree backend on a network unable to reach `INVENTREE_NEWS_URL`
2. Trigger the task
3. Task will lead to continuous timeout termination of workers

### Expected behaviour

Task should finish with no new News entries added if URL is unreachable.

### Deployment Method

- [ ] Docker
- [X] Bare metal

### Version Information

0.12.10

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

SchrodingersGat:
@LavissaWoW we have been discussing removing the news feed entirely - it is very rarely updated anyway. Happy to run with an interim fix until we make that change though. 

In this case do you know what is blocking? Is it simply a connection timeout to the remote server?

LavissaWoW:
@SchrodingersGat My guess is a network timeout. My prod instance sits behind a proxy, and from what I see, feedparser doesn't really account for timeouts when trying to use it for requests as well.

I can't really mess a whole lot around and experiment with my prod server, so not too much done on that front. Looking at the feedparser package though, it seems that the intended use of it is only parsing returned data from I.e the requests library. Not running the request itself. Saw some discussions on their GH page about removing the ability for it to do HTTP request by itself.

On my prod, I've deleted that task schedule from my admin view as a workaround. But I suppose that can pop back into existence after a reboot.

SchrodingersGat:
@LavissaWoW do you see the same problem in the `check_for_updates` task?

LavissaWoW:
![image](https://github.com/inventree/InvenTree/assets/7506688/6fb28de1-3411-4ba6-b5e9-2d12189d7b75)

Only the news_feed is problematic here, And I think the issue lies in how it's implemented. 
As I mentioned above, the maintainer for the `feedparser` package has stated that the ability to do requests with the library _will_ be deprecated in due time.

[...]

PR #6225, opened 2024-01-12
Remove support for RSS news feed

SchrodingersGat:
The news feed is no longer maintained. In preparation for the new react interface, which does not / will not support this, removing the functionality entirely.

- Closes https://github.com/inventree/InvenTree/issues/6071

@matmair this was your thing - are you happy for this to be removed?


matmair:
I think it is a mistake to remove the only way we have pushing info to users but go ahead if you think we do not need it.

SchrodingersGat:
@matmair your preference to keep it is enough for me to keep it. I think that in this case we should be making more use of it - to announce new blog posts, or features, or even things like your C3 talk.

PR #6236, opened 2024-01-14
Add tests for get_purchase_orders logic

30350n:
Follow up to #6158.


30350n:
This seems to be working as expecting for me locally (🤞 the CI also passes) so I'm advocating for also backporting #6158 + this to 0.13.x

SchrodingersGat (approved):

SchrodingersGat:
Thanks for the additional tests :) Let's see if the backport works

Issue #6127, opened 2023-12-21
Server migration/JSON import results in "Key (key)=(BUILDORDER_REFERENCE_PATTERN) already exists"


n0xlf:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [X] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

As the title states, I'm trying to migrate a docker server to another docker server and getting the error "Key (key)=(BUILDORDER_REFERENCE_PATTERN) already exists" during the JSON import.

Thanks!

### Steps to Reproduce

1.  Update source server with **_docker-compose pull_**
2.  Update source server installation with **_docker-compose run inventree-server invoke update_**
3.  Export data on source server with **_docker-compose run inventree-server invoke export-records -f /home/inventree/data/export.json_**
4. Clean install on destination server, running **_docker-compose run inventree-server invoke update_** to initialize the DB
5. Copy data/media folder from source server to destination server
6. Copy esport.json from source to destination, in /data
7. Import data on destination server with **_docker-compose run inventree-server invoke import-records -f /home/inventree/data/export.json_**
8. See error

### Relevant log output

```bash
django.db.utils.IntegrityError: Problem installing fixture '/home/inventree/data/export.json.tmp.json': Could not load common.InvenTreeSetting(pk=1): duplicate key value violates unique constraint "common_inventreesetting_key_key"
DETAIL:  Key (key)=(BUILDORDER_REFERENCE_PATTERN) already exists.
```

n0xlf:
Importing with the "-c" option fixed the above error, but now I'm getting:

django.db.utils.IntegrityError: Problem installing fixtures: insert or update on table "build_build" violates foreign key constraint "build_build_responsible_id_49af6fa1_fk_users_owner_id"
DETAIL:  Key (responsible_id)=(16) is not present in table "users_owner".

SchrodingersGat:
The -c option will definitely be required to clear our existing data, as you have found. 

The next error may be due to how the "owner" data is constructed. Can you check in the exported data (text editor) to see if the "users_owner" data is there, and there is an entry for record 16?

n0xlf:
I don't see anything for users_owner.  The users themselves seem to be there under the auth.user model.  I can change the responsible fields to null but that doesn't help obviously.

[...]

PR #6244, opened 2024-01-15
Updates to data import/export steps

SchrodingersGat:
- Allow import/export of users.owner model
- Allow export of API tokens (optional, default=False)
- Fixes https://github.com/inventree/InvenTree/issues/6127

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -58,11 +58,12 @@ def apps():
     ]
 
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
This has now broken unit tests for postgresql. I guess that is why we exclude `users.owner` from the exported dataset.. will need a patch to fix

PR #6251, opened 2024-01-15
Move more paths to basic path

matmair:
Refactor to remove unnecessary usage of re_path

### DIFF
--- a/InvenTree/InvenTree/middleware.py
+++ b/InvenTree/InvenTree/middleware.py
@@ -7,7 +7,7 @@ from django.conf import settings
 from django.contrib.auth.middleware import PersistentRemoteUserMiddleware
 from django.http import HttpResponse
 from django.shortcuts import redirect
-from django.urls import Resolver404, include, re_path, resolve, reverse_lazy
+from django.urls import Resolver404, include, path, resolve, reverse_lazy
 
 from allauth_2fa.middleware import AllauthTwoFactorMiddleware, BaseRequire2FAMiddleware
 from error_report.middleware import ExceptionProcessor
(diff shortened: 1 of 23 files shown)


wolflu05:
Does this help to reduce the open api spec generation warnings? And how much regex paths do we use now?

matmair:
That was not the goal, I have not looked into the effects on schema generation. After merge there will be 22 re_paths left. All of those use some form of regex matching.

SchrodingersGat (approved):

wolflu05:
Ok, thanks for checking. I like the change, but I'm curious what your intention then is behind? 

I now also checked the impact on schema generation warning, but it seems like it doesn't change much, it adds a few more warnings, but that has to be fixed all differently.

matmair:
To make it harder to introduce errors with paths. Re_path where it is not needed always bothers me.

matmair:
I had to roll back a few paths (label/report printing routes) as I do not want to touch the API methods/behaviour in this PR. This would be ready for a review now.

Note: The PR lowers the coverage a bit as the total LoC is lowered

wolflu05:
Is there any option so that one don't need the / at the end? I think that is really weird, and I always forgot about that and wonder why the api route doesn't work. Is this actually a standard? I never seen this on other REST apis. The only thing I can find about it is this [quote](https://stackoverflow.com/a/76750369).

matmair:
Trailing slash is the default behaviour in DRF applications. No idea why but I have no intention to change it with this PR as that would be a new API iteration and a breaking change.

SchrodingersGat:
Thanks @matmair

PR #6257, opened 2024-01-16
[React] UI Translation Updates

SchrodingersGat:
A few updates / tweaks / fixes for locale translation in the react UI:

- Expose "default_locale" (of the server) to the root API endpoint
- Ignore compiled `messages.ts` files in .gitignore
- Move locale information into new python file
- Improve `Accept-Language` header


SchrodingersGat:
@matmair running into a weird issue here, which I'm stuck on. It is a discrepancy between running the "live" react pages (using `invoke frontend-dev`) and the "compiled" files (e.g. after `invoke frontend-compile` and `invoke static`)



### Live Server

When running under `invoke frontend-dev` we get a nice list of "languages" to select:

image

### Static Server

When hosting the statically compiled files, the names of the languages are displaying as the translation "keys":

image

--------------------

Any idea why this might be happening?

matmair:
@SchrodingersGat nothing jumps out but maybe it has to do with the double loading that react does in dev mode

wolflu05:
To exclude the "double loading", try commenting out the script mode in main.tsx

Issue #6303, opened 2024-01-21
[FR] Improve Docker deployment 


ChristianSchindler:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

I want to suggest to simply the deployment of Docker.

1) I would like that you only need no longer a `Server` and `worker` but only use 1
2) invoke updates of the db when you start a newer version of the container and updates are done 
3) many include also Nginx 

### Suggested solution

For the first two make a script with first checks if we are in first created, update or reboot and run `invoke update ` then start both process 

### Describe alternatives you've considered

Implent it directly in the docker file but this makes it more complex 

### Do you want to develop this?

- [X] I want to develop this.

matmair:
I am not sure how scalable and reliable the system would be with the suggested changes, after all we have deployments with 100k+ items and hundreds of background tasks a minute so this is certainly something that would need extensive validation with power users.

2. Would be a no for me - invoke updates does more than just DB updates and only running those can lead to problems (we have automigrations as an option in the config file but it presents problems form time to time and is off by default). Running a full update cycle can take minutes on larger deployments - probably leading to a lot of confusion if InvenTree suddenly just isn't comming up
3. I do not understand this point, could you expand this a bit?

matmair:
1. I do not think we will support this approach
2. invoke update has to be run after every upgrade - this is essential for a functioning instance. I do not see a value in this beeing controlable
3. We do not ship a static file hoster (the debug server is not designed to handle this) so a webserver is a must

ChristianSchindler:
If we run `invoke update` on every update it makes even more sense to automate it because then it is less likely to get user errors

SchrodingersGat:
@ChristianSchindler I am happy to consider a PR which can run `invoke update` on container start, if it needs to be updated - but I am not keen to change the overall construct of the container services (i.e. combining two containers into one).

ChristianSchindler:
@SchrodingersGat I think also that this is the most important part. What happens when the worker is running and `invoke update` simultaneously? Does this crash something?

PR #6305, opened 2024-01-21
first implementation of improved docker

ChristianSchindler:
Based on: https://github.com/inventree/InvenTree/issues/6303

### Improve Docker Image:

1) Run `invoke update` Automatically on image updates
2) Combine Server and Worker 1 container


SchrodingersGat:
I have some concerns that this goes against [docker best practice](https://testdriven.io/tips/59de3279-4a2d-4556-9cd0-b444249ed31e/#:~:text=Docker%20best%20practice%3A,needed%20to%20handle%20more%20traffic.)

### Process Reliability

If one of the processes (web worker, background worker) fails - this may be obscured by this update. The docker container generally manages process state, and allows auto-restart of the process due to an error

### Logging

Again, running a background process in the docker container may obscure logs - which are crucial for debugging InvenTree performance.

### Scalability

While "simplifying" the docker container setup, this reduces the flexibility of the setup.

------------------

I'd like to hear your thoughts on the potential advantages. Maybe I am missing something here, but the `docker compose` file takes care of all of this abstraction anyway...

ChristianSchindler:
@SchrodingersGat 
> Process Reliability
>
>If one of the processes (web worker, background worker) fails - this may be obscured by this update. The docker container generally manages process state, and allows auto-restart of the process due to an error

If I can do it so that if one fails, the complete container restarts, is this ok?

>Logging
>
>Again, running a background process in the docker container may obscure logs - which are crucial for debugging InvenTree performance.

I know that's why I made it so that both still print out and also add in front `Sever` or `Worker` to identify where the problem lays

>Scalability
>
>While "simplifying" the docker container setup, this reduces the flexibility of the setup.

If you want to scale, this is still entirely possible. Then you can use it without the script similar to it is now, but most will not need it, and that's a way for it to be as easy as possible.

ChristianSchindler:
@matmair 
>Looks intresting @ChristianSchindler - could you expand the description of the PR a bit more?
Seems like a notable shift in the deployment strategy that would need concesus within @inventree/triage @inventree/maintainer and docs updates + release notes.


In gennerla ther are two changes:
1) Automatic updates if there is a new db version or on the first install
2) In general, you can remove the worker, and then it is entirely the same as before.

matmair:
Please ammend your explaination to the PR description to make it easier to find

matmair:
@ChristianSchindler  We will not merge this with a combined worker and server container in the compose due to the resons listed by @SchrodingersGat above.
If you are open to revert that change we are willing to review and merge the auto-update, which seems valueable.

ChristianSchindler:
> @ChristianSchindler We will not merge this with a combined worker and server container in the compose due to the resons listed by @SchrodingersGat above. If you are open to revert that change we are willing to review and merge the auto-update, which seems valueable.

Sorry had no time will look at it next week

PR #6309, opened 2024-01-21
[PUI] Registration

matmair:
Adds registration (for built in auth, SSO will be part of #5753) via PUI.
Closes #6282

Issue #5753, opened 2023-10-19
[PUI] SSO Support


SchrodingersGat:
Add SSO for the new interface:

```[tasklist]
### Tasks
- [x] Expose SSO information to the API
- [x] Add SSO options to login screen
- [x] SSO login returns a token (and then normal operations after that)
- [x] Add SSO configuration to system settings pages
- [x] Add login taint
```

matmair:
@SchrodingersGat I would split out settigns into a seperate issue, they are very different APIs

PR #6314, opened 2024-01-22
[PUI] Auth behaviour controls

matmair:
Follow up to #6309.
This PR:
- Adds registration and passwort reset to /auth/providers api (requires API version bump)
- [PUI] UI not respects settings for registration and password forgotten
- [PUI] Extends API zustand with information regarding Auth behaviour (SSO, MFA, feature control)

This should be the last thing required to implement #5753, all the state is already there. Missing is UI

### DIFF
--- a/InvenTree/InvenTree/api_version.py
+++ b/InvenTree/InvenTree/api_version.py
@@ -1,10 +1,13 @@
 """InvenTree API version information."""
 
 # InvenTree API version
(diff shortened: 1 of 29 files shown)


matmair:
Ready for review

SchrodingersGat:
@matmair LGTM :) Thanks for implementing this

PR #6324, opened 2024-01-23
Specify empty OIDC prefix

SchrodingersGat:
- Ref: https://github.com/inventree/InvenTree/discussions/6273

Specifies explicitly empty prefix for OIDC providers to maintain backwards compatibility

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -1000,6 +1000,10 @@ SOCIALACCOUNT_PROVIDERS = get_setting(
 
 SOCIALACCOUNT_STORE_TOKENS = True
 
+# Explicitly set empty URL prefix for OIDC
+# Ref: https://docs.allauth.org/en/latest/release-notes/recent.html#backwards-incompatible-changes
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@p-fruck thoughts on this?

InvenTree/InvenTree/settings.py
```diff
@@ -1000,6 +1000,11 @@
 
 SOCIALACCOUNT_STORE_TOKENS = True
 
+# Explicitly set empty URL prefix for OIDC
+# The SOCIALACCOUNT_OPENID_CONNECT_URL_PREFIX setting was introduced in v0.60.0
+# Ref: https://docs.allauth.org/en/latest/release-notes/recent.html#backwards-incompatible-changes
```
p-fruck, line 1005:
```suggestion
# Ref: https://github.com/pennersr/django-allauth/blob/0.60.0/ChangeLog.rst#backwards-incompatible-changes
```

I would prefer to use a permanent reference, as the release notes on docs.allauth.org do not allow to point to the changes of a specific release

p-fruck (changes requested):

PR #6330, opened 2024-01-23
fix(docker): SELinux volume labels

p-fruck:
When mounting volumes into containers with SELinux enabled on the host the z option must be specified. I already submitted the same change for the [dev setup](https://github.com/inventree/InvenTree/pull/4354) but forgot to upstream the changes made for the production deployment


SchrodingersGat (approved):

SchrodingersGat:
@p-fruck thanks for submitting this change

PR #6333, opened 2024-01-23
[PUI] SSO Support

matmair:
Adds SSO (login, registration) support to PUI
Closes https://github.com/inventree/InvenTree/issues/5753

### DIFF
--- a/src/frontend/src/components/buttons/SooButton.tsx
+++ b/src/frontend/src/components/buttons/SooButton.tsx
@@ -1,22 +1,47 @@
 import { Button } from '@mantine/core';
 import {
+  IconBrandAzure,
+  IconBrandBitbucket,
+  IconBrandDiscord,
+  IconBrandFacebook,
+  IconBrandFlickr,
   IconBrandGithub,
+  IconBrandGitlab,
   IconBrandGoogle,
+  IconBrandReddit,
+  IconBrandTwitch,
+  IconBrandTwitter,
   IconLogin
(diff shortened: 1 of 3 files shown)


matmair:
This is done UI wise; during implementation I came across the need for a way to tell the backend if CUI or PUI should be used/redirected to. Working name is ui_taint - I will add this to the session and add some initial logic to redirect to the correct page after SSO usage

matmair:
@SchrodingersGat this is ready for review; for testing it needs to be pulled locally as there are API changes

SchrodingersGat:
@matmair on reviewing this I found some issues which needed addressing, not relating to your changes but just in general. 

Working on fixing these in https://github.com/inventree/InvenTree/pull/6368

Once that is merged, then your branch should be much easier to review

matmair:
Thank for reviewing @SchrodingersGat

SchrodingersGat:
@matmair AWESOME work - at least with my testing on Azure, this works very cleanly.

One edge case where it can fail (minor, but took me a while to work out):

1. Run the `frontend-dev` server at `localhost:5173/platform`
2. Login via SSO from `localhost:5173`
3. The server after login redirects to `localhost:8000/platform` (not `localhost:5173/platform`)

This can cause some strange issues with the dev server. However if you start from `localhost:8000/platform` it works *ok*.

SchrodingersGat:
I am happy to merge as is, but any comments on the edge case above?

PR #6353, opened 2024-01-29
Tracing improvements

SchrodingersGat:
Update code to ensure that the "tracing" is only enabled for the server process (not the worker process!)

### DIFF
--- a/InvenTree/InvenTree/ready.py
+++ b/InvenTree/InvenTree/ready.py
@@ -24,6 +24,20 @@ def isInWorkerThread():
     return 'qcluster' in sys.argv
 
 
+def isInServerThread():
+    """Returns True if the current thread is a server thread."""
+    if isInWorkerThread():
+        return False
+
+    if 'runserver' in sys.argv:
+        return True
+
+    if 'gunicorn' in sys.argv[0]:
+        return True
+
+    return False
+
+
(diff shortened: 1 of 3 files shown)


matmair:
I do not understand the value of this change; having spans of all server actions seems valueable. Can you be more specific what the benefit is here?

SchrodingersGat:
The tracing only monitors external API function calls, right? Or do internal processes (e.g. the background worker) get traced too?

Aside from this question, I think it makes sense to disable tracing for other "temporary" processes like importing data or running migrations, yes?

matmair:
@SchrodingersGat every general process get's traced - so background workers should show up too. That might depend on the data backend though.

wolflu05:
(Even if we close that pr, I think we should keep the added invoke task (already would needed that a while ago) and the two guard functions.)

SchrodingersGat:
> @SchrodingersGat every general process get's traced - so background workers should show up too. That might depend on the data backend though. If tracing is enabled I would trace everything possible to be able to find performance bottlenecks. The overhead is not too high on performant machines (esp with fast network).

Ok, fair enough. What are your thoughts on disabling tracing for operations such as "import" or "migration"? My justification is to simplify the running code in these parts, as this is where we get a lot of problems.

PR #6365, opened 2024-01-30
Forms initial data

SchrodingersGat:
Replaces https://github.com/inventree/InvenTree/pull/6332

This PR is working towards a complete refactoring of how we handle modal forms. The intent is to remove the duplicate pathways for creating forms, and settle on a single, consistent framework (as implemented by @wolflu05 )

So, in this PR we refactor / rework a number of existing forms, add some new features, add some new forms, and fix some bugs along the way.

#### Important Notes

- DateInput has been fixed, and works correctly now
- Initial data can be easily passed into a form
- The adjustFilters callback is now passed the complete form valueset, so it can make decisions without extra hooks


src/frontend/src/components/forms/ApiForm.tsx
```diff
@@ -163,13 +172,23 @@ export function OptionsApiForm({
...
+      return field.value ?? field.default ?? undefined;
+    });
+
+    // If the user has specified initial data, use that instead
+    if (props.initialData) {
+      defaultValuesMap = {
+        ...defaultValuesMap,
+        ...props.initialData
+      };
+    }
```
wolflu05, line 186: Does this work recursively for nested forms?

src/frontend/src/components/forms/ApiForm.tsx
```diff
@@ -390,16 +410,18 @@ export function ApiForm({ id, props }: { id: string; props: ApiFormProps }) {
...
-          {Object.entries(props.fields ?? {}).map(([fieldName, field]) => (
-            <ApiFormField
-              key={fieldName}
-              fieldName={fieldName}
-              definition={field}
-              control={form.control}
-            />
-          ))}
-        </Stack>
+        <FormProvider {...form}>
```
wolflu05, line 413: Why do we need that form provider here? The form ref gets already passed to the fields via the control ref.

src/frontend/src/components/forms/fields/RelatedModelField.tsx
```diff
@@ -71,28 +79,49 @@ export function RelatedModelField({
...
 
-      let filters = definition.filters ?? {};
+      let _filters = definition.filters ?? {};
 
       if (definition.adjustFilters) {
-        filters = definition.adjustFilters(filters);
+        _filters =
+          definition.adjustFilters({
+            filters: _filters,
+            data: form.getValues()
```
wolflu05, line 113: Why do we need that data here? Can't we just hook the required values to a state and use them in the filter function?

src/frontend/src/components/tables/company/AddressTable.tsx
```diff
@@ -108,6 +108,52 @@ export function AddressTable({
     ];
   }, []);
 
+  const addressFields: ApiFormFieldSet = {
```
wolflu05, line 111: Not sure about the performance, have you tested if there is a difference if you put this in a usememo? I think then the entire form doesn't get rerendered when any other state eg for the table on this component changes.

src/frontend/src/main.tsx
```diff
@@ -42,6 +42,10 @@ window.INVENTREE_SETTINGS = {
       host: `${window.location.origin}/`,
       name: 'Current Server'
     },
+    'mantine-2j5j5j5j5': {
+      hots: 'http://localhost:8000',
+      name: 'Localhost'
+    },
```
wolflu05, line 48: Wouldn't it make sense to only add this in dev mode?

src/frontend/src/components/forms/fields/RelatedModelField.tsx
```diff
@@ -32,6 +36,8 @@ export function RelatedModelField({
     fieldState: { error }
   } = controller;
 
+  const form = useFormContext();
```
wolflu05, line 39: Does this have any drawback on performance? Eg does this component now gets rerendered more often as without that hook?

src/frontend/src/components/forms/ApiForm.tsx
SchrodingersGat: Probably not - at least I have not tested it with that. Do you have any examples of a nested form already implemented I could test it with?

wolflu05:
As you asked on the other, now closed pr, if there is any reason to keep two ways, no absolutely fine to refactor to the more reactive one. Thank you that you did that now.

The only thing which I sometimes experienced and not sure about yet how to fix it are some performance problems. Mostly in react dev mode with safari on macOS. If I open a large form with lots of fields and type in something it lags a bit behind to how fast I'm typing. Have you seen this too?

src/frontend/src/components/forms/ApiForm.tsx
SchrodingersGat: As far as I could tell, the control ref is only for a specific field, not the whole form? I need the whole form dataset to pass through to `adjustFilters` for related fields (see below)

src/frontend/src/components/forms/fields/RelatedModelField.tsx
SchrodingersGat: By passing the `data` object through, we ensure that we get a "fresh" set of data on every call, without needing to setup individual hooks from each other field we are interested in. Say that a related field wants to adjust the API filters based on the values of three other fields - that's three additional hooks, and a lot of additional boilerplate code.

This approach feels much cleaner and more consistent to me. There may be a performance tradeoff, but at a significant advantage in terms of reduced code complexity.

src/frontend/src/forms/PurchaseOrderForms.tsx
```diff
@@ -66,3 +62,52 @@ export function purchaseOrderLineItemFields({
...
+    order_currency: {
+      icon: <IconCoins />
+    },
+    target_date: {
+      icon: <IconCalendar />
+    },
+    link: {},
+    contact: {
+      icon: <IconUser />,
+      adjustFilters: (value: ApiFormAdjustFilterType) => {
```
SchrodingersGat, line 93: @wolflu05 see how much simpler it is now to adjust the filters of one field based on the value of another field?

src/frontend/src/components/tables/company/AddressTable.tsx
SchrodingersGat: If there were anything in `addressFields` which was dynamic, I would put it in a `useMemo()` - but I would assume that as this is a static variable, tagging it as `const` is sufficient? Happy to be told otherwise :)

src/frontend/src/main.tsx
SchrodingersGat: Yah, good point. Will look into this.

SchrodingersGat: Has been addressed now

matmair:
Looks fine; regarding performance Lukas is probably a better discussion partner, I am not knowledgable in that area.

src/frontend/src/components/forms/ApiForm.tsx
wolflu05: Yes, the part form:

https://github.com/inventree/InvenTree/blob/master/src/frontend/src/forms/PartForms.tsx#L72-L84

src/frontend/src/components/forms/fields/RelatedModelField.tsx
wolflu05: Ok, makes sense.

src/frontend/src/components/tables/company/AddressTable.tsx
wolflu05: That's only partially correct. If we pass the fields inside the form component anywhere into a dependency array, this will constantly rerun, as fields will be different every component render. However if we put it into a useMemo with an empty dependency array, it always returns the same object, so the fields is not responsible for a rerun of the hook. That's also the reason to why we sometimes need a useCallback.

src/frontend/src/components/forms/ApiForm.tsx
SchrodingersGat: Ok, yep works just fine for nested. I'll add to the playground example.

src/frontend/src/components/tables/company/AddressTable.tsx
SchrodingersGat: Ah, ok.. is it worth the extra boilerplate to wrap this in a `useMemo` for the performance boost, in your opinion?

src/frontend/src/components/forms/ApiForm.tsx
wolflu05: But not if you only supply just one value and get the other from the fields default, because the spread syntax doesn't do deepmerging. But given the fact that one probability doesn't use this like this it's fine for me.

src/frontend/src/components/tables/company/AddressTable.tsx
wolflu05: In my opinion yes, but it's not that much to add, just two lines.

SchrodingersGat:
@wolflu05 thanks for reporting - I will look into it

PR #6375, opened 2024-01-31
Request column names from server when loading table

SchrodingersGat:
This PR adds facility for tables to fetch the translated column names from the server, using the existing metadata endpoint. This allows us to reduce duplication of translated strings between the server and client.

As each table is constructed from data provided by a defined endpoint, and (most) API fields have localized `label` attributes, we can use those to render the localized table columns.

Note that the table column titles can be manually overridden if required.

If this is deemed a good idea, following improvements can be made:

- Cache the column names based on the table key
- Update existing tables to not specify column names where not required.


matmair:
Looks/Sounds intresting and relativley simple to implement. Maybe caching all table column names similar to states would be a good idea?

matmair (approved):

wolflu05:
Nice idea, LGTM now after caching is implemented.

src/frontend/src/tables/InvenTreeTable.tsx
```diff
@@ -137,9 +182,12 @@ export function InvenTreeTable<T = any>({
         hidden = tableState.hiddenColumns.includes(col.accessor);
       }
 
+      console.log(col.accessor, col.title, fieldNames[col.accessor]);
```
matmair, line 185: Question: Is this left in inventionally?

src/frontend/src/tables/InvenTreeTable.tsx
```diff
@@ -513,40 +562,48 @@ export function InvenTreeTable<T = any>({
...
+            fetching={isFetching}
+            noRecordsText={missingRecordsText}
+            records={data}
+            columns={dataColumns}
+            onRowClick={tableProps.onRowClick}
+            defaultColumnProps={{
+              noWrap: true,
+              textAlignment: 'left',
+              cellsStyle: {
+                // TODO: Need a better way of handling "wide" cells,
```
matmair, line 601:
```suggestion
                // TODO @SchrodingersGat : Need a better way of handling "wide" cells,
```

src/frontend/src/tables/sales/SalesOrderTable.tsx
```diff
@@ -90,7 +90,7 @@ export function SalesOrderTable({
     return [
       {
         accessor: 'reference',
-        title: t`Sales Order`,
+        // title: t`Sales Order`,
```
matmair, line 93: Question: Ist this left in intentionally?

matmair (approved):

src/frontend/src/functions/string.tsx
```diff
@@ -0,0 +1,9 @@
+/**
+ * Convert an input string to title case.
+ */
+export function toTitleCase(str: string): string {
+  return str
+    .split(' ')
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase())
+    .join(' ');
+}
```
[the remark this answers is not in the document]
SchrodingersGat: Ok, fair point. I've removed this and we will rely on the server-side translations being correctly capitalized.

src/frontend/src/tables/ColumnRenderers.tsx
```diff
@@ -112,8 +114,8 @@ export function ResponsibleColumn(): TableColumn {
 export function TargetDateColumn(): TableColumn {
   return {
     accessor: 'target_date',
-    title: t`Target Date`,
     sortable: true,
+    title: t`Target Date`,
```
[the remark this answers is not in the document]
SchrodingersGat: Some of the fields are not introspecting properly yet - still working out this one.

src/frontend/src/tables/InvenTreeTable.tsx
SchrodingersGat: Oops, thanks!

src/frontend/src/tables/sales/SalesOrderTable.tsx
SchrodingersGat: Fixed!

wolflu05:
This has now broken my machine integration tables. E.g. this API endpoint is used in a table:

https://github.com/wolflu05/InvenTree/blob/34d730febf68dc5e0851837955797e1866042f19/InvenTree/machine/serializers.py#L110
https://github.com/wolflu05/InvenTree/blob/34d730febf68dc5e0851837955797e1866042f19/InvenTree/machine/api.py#L157-L172

And PUI always shows a notification on load:

image

After investigation, it seems like that this is coming from:

https://github.com/inventree/InvenTree/blob/15f58b965ededab9928f7d372a5ef1cdfaf88ab8/src/frontend/src/functions/forms.tsx#L56-L63

Because when using an `APIView` for non model api endpoints this results in no actions added to the OPTIONS response.

Same things applies for the Plugin page in the Admin center.

image

How should we fix this?

SchrodingersGat:
@wolflu05 you found that so quickly! I had just been looking into this. I think that we should 

a) Remove the "form error - response did not contain action data" - it is a very "nerdy" error which I had in there for debugging. Probably not very useful
b) Where is the "permission denied" warning coming from?

wolflu05:
Yes, was trying to resolve the merge conflicts on the machines branch and refactor the tables, ...

Yes, I'm no board with just removing this notification. The enduser doest know what that means. Do you create an PR for that?

Regarding the permissions error, I guess that is coming from here:
https://github.com/inventree/InvenTree/blob/ec2a66e7a51ac2a5cc88ceb01568da35694e947e/src/frontend/src/functions/forms.tsx#L72-L76

PR #6379, opened 2024-01-31
Fix for invoke update

SchrodingersGat:
- If frontend option was skipped during update, then "static" step would not be performed


SchrodingersGat:
On review, a backport is not needed here, as this bug has only been introduced since `0.13.x`

PR #6383, opened 2024-02-01
Fix urls loading for plugin testing

wolflu05:
#5648 has broken my plugin testing CI setup. 

https://github.com/inventree/InvenTree/pull/5648/files#diff-5b51ddf4fb2cbf458278a6627285ebdd20c82e3e4276be8cd74fb63bc6d25abfL14-R19

Basically, I set `INVENTREE_PLUGIN_TESTING: True` and `INVENTREE_PLUGIN_TESTING_SETUP: True` . This already sets up and loads my plugin that was installed via `pip install -e /path/to/plugin` before and the tests run fine. However now the urls are not activated which results in failing tests. This PR proposes a fix for this problem.

### DIFF
--- a/InvenTree/plugin/urls.py
+++ b/InvenTree/plugin/urls.py
@@ -1,5 +1,6 @@
 """URL lookup for plugin app."""
 
+from django.conf import settings
 from django.urls import include, re_path
 
 PLUGIN_BASE = 'plugin'  # Constant for links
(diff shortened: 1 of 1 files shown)


SchrodingersGat (approved):

SchrodingersGat:
Seems like a pretty reasonable change :)

wolflu05:
The inventree tests itself rely on the urls being disabled if not specially enabled. However as I understand `INVENTREE_PLUGIN_TESTING_SETUP` is used if external tests are done which is not used by inventree itself, but for testing external plugins. So I just changed to variable to that and the tests run fine now.

wolflu05:
Thanks for that quick merge 👍

PR #6384, opened 2024-02-01
Backport: Fix urls loading for plugin testing (#6383)

wolflu05:
This is a manual backport of #6383

(hope its ok to manually do it instead of using that backport tool)


SchrodingersGat:
Thanks @wolflu05 there were conflicts which caused the backport tool to fail

Issue #6133, opened 2023-12-25
Email confirmation key missing domain


HammyHavoc:
### Deployment Method

- [ ] Installer
- [X] Docker Development
- [ ] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

The URL in the email to confirm an address consists of `/accounts/confirm-email/` followed by the key, but it isn't prefixed by the base URL.

### Steps to Reproduce

1. Send an email confirmation.

SchrodingersGat:
@HammyHavoc did you find a solution?

HammyHavoc:
Yes! :- ) Seems that for whatever reason, `base_url` in `config.yaml` wasn't respected and wasn't visible in `/settings` ("No value set"), so added it via the Server Settings page, and worked like a charm.

HammyHavoc:
If it's useful, my `base_url` in `config.yaml` was set as `base_url: 'https://inventree.mydomain.suffix'`.

SchrodingersGat:
Ok, I will reopen this to determine why the setting in the config file was not being respected

SchrodingersGat:
Ok, can confirm that the "base_url" config setting is not being "respected" as one might expect. This is a real issue

HammyHavoc:
> Ok, can confirm that the "base_url" config setting is not being "respected" as one might expect. This is a real issue

Just wanted to say I appreciate you and what you do and give to the world in both solutions, and your own time and energy. I hope you have the greatest year yet. 🖖 You're awesome.

Looking forward to getting more involved with contributing to Inventree however I can too.

SchrodingersGat:
@HammyHavoc FYI - see the updates in https://github.com/inventree/InvenTree/pull/6390

The issue on your end is that it should be called `site_url` (and not `base_url`) in the settings file :)

Issue #5585, opened 2023-09-21
Cannot define hostname as base url


lippoliv:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

I host InvenTree (currently) on a NAS in my local network. But I cannot define the hostname as a base url, probably because it has no TLD.

_(Good news / workaround: I was able to workaround that using the tld "local", so I used "http://some-nas.local" and it worked)_

### Steps to Reproduce

1. Go to "Settings" section
2. Go to "Server settings"
3. Edit "Base url"
4. Enter "http://some-nas" (you can combine it with port or not, doesn't matter)
5. Save

### Expected behaviour

Since it is a valid URL, I would expect it to save. But it doesn't. It Complains about invalid value.

<img width="1042" alt="image" src="https://github.com/inventree/InvenTree/assets/2561859/24312592-8b25-4bef-b50d-0de24250dcc2">

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

0.12.7

### Please verify if you can reproduce this bug on the demo site.

- [X] I can reproduce this bug on the demo site.

lippoliv:
@matmair here's the Bug which I mentioned in #5577 yesterday. Turns out, it's unrelated to port AND I found a workaround. So probably prio is not too high.

SchrodingersGat:
Not sure if this is a "bug" here or not.. We would have to presumably change the way that URL validation is working under the hood, to allow specification of URLs with no TLD

lippoliv:
Since it's a valid URL, I personally would classify it as a bug. On the other hand, this type of software most likely runs on companies who have internal servers with DNS configured, so most users will probably not hit this issue.

You can maybe "work arround" by mentioning that ".local" can be used as a TLD for local network hosting? Regarding the RFC it seems to be established world wide, so maybe this is a low-hanging-fruit to "fix" this?

matmair:
A host without any domain comes with interesting security/usability implications. some-nas could be a (wired) gTLD, I would strongly recommend using a TLD. Recommending .local is a double-edged sword. People should really be using proper DNS, self-announced hostnames/mDNS are pretty easily exploited.

lippoliv:
I know that, and it's a fair point. But the argument is: It's still a valid base url and InvenTree is saying it's not.

I think either it is a accepted value by InvenTree or in this case InvenTree informs about "only hostname" base urls are permitted.

[...]

PR #6390, opened 2024-02-02
[WIP] Site ID Fixes

SchrodingersGat:
This PR is a work in process to fix / simplify the site URL / multi-site support for InvenTree.

This fixes a few bugs and addresses some shortcomings of how the "site_url" parameter is configured.

Additionally, multi-site support is now disabled by default (a "breaking" change but most users would not use this feature at all). Disabling multi-site support makes a lot of things much simpler...

- Fixes https://github.com/inventree/InvenTree/issues/6133

### TODO

- [x] Add documentation around enabling multi-site support
- [x] Update SSO documentation (much simpler with multi-site support disabled)

### References

- https://github.com/inventree/InvenTree/issues/6133
- https://github.com/inventree/InvenTree/issues/5585


matmair:
@SchrodingersGat great improvement. Do you think adding site control to the admin view might be in-scope for this?

SchrodingersGat:
> Do you think adding site control to the admin view might be in-scope for this?

Do you mean the PUI admin view? Site control is already available in the Django Admin view

SchrodingersGat:
I do not see the value here - it would be a very small user set (if any) who are using multi-site support. There is still always the Django Admin site available if absolutely required to setup?

matmair:
I think you are right, we can add it later

PR #6396, opened 2024-02-03
Fix for filter adjustment in related field

SchrodingersGat:
- Cannot compare two objects directly
- Ref: https://github.com/inventree/InvenTree/pull/6365#issuecomment-1924220520

### DIFF
--- a/src/frontend/src/components/forms/fields/RelatedModelField.tsx
+++ b/src/frontend/src/components/forms/fields/RelatedModelField.tsx
@@ -115,7 +115,7 @@ export function RelatedModelField({
       }
 
       // If the filters have changed, clear the data
-      if (_filters != filters) {
+      if (JSON.stringify(_filters) !== JSON.stringify(filters)) {
         resetSearch();
         setFilters(_filters);
       }


src/frontend/src/components/forms/fields/RelatedModelField.tsx
```diff
@@ -115,7 +115,7 @@ export function RelatedModelField({
       }
 
       // If the filters have changed, clear the data
-      if (_filters != filters) {
+      if (JSON.stringify(_filters) !== JSON.stringify(filters)) {
```
wolflu05, line 118: Thanks for that quick fix, but does this work if the object keys are returned in a different order, or is that case irrelevant here?

SchrodingersGat: That would indeed cause the query to be recalculated, and thus result in the same behaviour we are seeing here. I think that should be a relatively low-surface-area vector, but if you have a suggestion for a clean way for a bettery comparison, I'd happily merge it in :)

wolflu05: That would indeed be a more complex solution, we would need to write a recursive function to compare the two objects. Something like this: https://stackoverflow.com/a/16788517 , but you're right, that not worth it currently.

PR #6398, opened 2024-02-03
Remove devcontainer Git Feature

LavissaWoW:
Out Alpine image has never really been supported by `devcontainer/features/git`.
In release 1.2.0, a distro check was added, and building the image now fails.

InvenTree installs git, so removing the git feature from devcontainer.json


SchrodingersGat:
Thanks @LavissaWoW ; 👍

Issue #6391, opened 2024-02-02
[PUI] Use sessions backend for login


matmair:
@matmair this review process has made me question - is it a good idea to just use CSRF session auth for PUI (as we do for CUI) and *not* use API tokens at all? As far as I can see there are two major advantages:

### Session Management

Using a single *session* (based on CSRF token) means that CUI and PUI are in sync. No need to manage two separate login methods (session token / API token) as we do now.

### File Downloads

Moving to session auth would allow us to download files / reports / etc in the browser - a problem which is currently *hard to solve* with API tokens - see https://github.com/inventree/InvenTree/issues/5697

So, move to session token and this problem goes away (because the *web session* has the same token and we don't have to work out how to pass an API token to a file download).

-----------------

Thoughts on this? Now that we are deep in the weeds on this is the right time to do it. I'm just not sure how to get the CSRF token into the session when you login from PUI - we have to replicate the CUI behaviour...

_Originally posted by @SchrodingersGat in https://github.com/inventree/InvenTree/issues/6368#issuecomment-1920885602_

matmair:
login is working, looking at detecting if we logged out on CUI to log out of PUI

Issue #5697, opened 2023-10-15
[PUI] How to download attachments and reports


SchrodingersGat:
Related issues:

- https://github.com/inventree/InvenTree/pull/5696
- https://github.com/inventree/InvenTree/issues/5048

The new user interface will be driven entirely by the API. Thus, downloading files (like attachments, reports, etc) no longer relies on session cookies.

so, how do we provide "file download" for the user? Note that any media files served / hosted by InvenTree require user authentication (e.g. via auth token).

So, we need a way to fetch the file and present to the user for download, via the web client. 

Many of the ways I have found are very clunky, requiring first downloading the file to browser RAM, and then making a hidden button and "clicking" the button (in software). Or, similar tricks using hidden web forms.

Surely there is a better way? Any suggestions how we can cleanly do this via the API, without any of these hacks tricks?

wolflu05:
Also related #5531

matmair:
I do not really see why it is not working with basic links; I will investigate tonight.
If required I think we should change the backend for this so we can just link to a download endpoint instead of having to work around in the frontend with it.
Maybe [this](https://stackoverflow.com/a/56763856/17860466) is a solution?

SchrodingersGat:
@matmair as far as I can tell, it is because the browser does not have a session cookie. Have a look at the request headers when you download a file in the current UI, vs the react UI.

Here is the request to do a "simple download" of the file in the new UI, which only has login via token auth - no browser session auth:

<img width="796" alt="Screenshot 2023-10-16 at 7 11 25 pm" src="https://github.com/inventree/InvenTree/assets/10080325/fb45c480-88ac-4305-b08a-b7b017de6b7c">

This is the same reason that I have been working on [this PR for images](https://github.com/inventree/InvenTree/pull/5696)

One option I have considered:

1. At login, request a "session cookie" from the server
2. Save the cookie to the session
3. Any "media file links" can then be downloaded without any extra issues

SchrodingersGat:
> If required I think we should change the backend for this so we can just link to a download endpoint instead of having to work around in the frontend with it.

I would prefer this too - the required changes to the frontend would be prohibitive

SchrodingersGat:
> I think adding a session token should be a (somewhat) easy feat.

I hope so! I have not yet found any good examples of this

matmair:
[...]

PR #6399, opened 2024-02-03
[PUI] Use sessions backend for login

matmair:
Removes the usage of API tokens for PUI - switches to CSRF and session token.

Fixes https://github.com/inventree/InvenTree/issues/6391
Ref https://github.com/inventree/InvenTree/issues/5697


src/frontend/src/components/images/ApiImage.tsx
```diff
@@ -1,6 +1,5 @@
 /**
- * Component for loading an image from the InvenTree server,
- * using the API's token authentication.
+ * Component for loading an image from the InvenTree server using the CSRF cookie
```
SchrodingersGat, line 2: If the browser session has this cookie now, do we still need the `ApiImage` component? Or could we replace it entirely with the `<Image>` component?

SchrodingersGat:
netlify preview does not seem to want to login currently - is that expected?

src/frontend/src/components/images/ApiImage.tsx
matmair: Not sure, it is working on my deployment but not on netlify

matmair:
Seems like the CRSF stuff only works on the same domain, have to look deeper into it

SchrodingersGat:
> Seems like the CRSF stuff only works on the same domain, have to look deeper into it

Have you set the `INVENTREE_TRUSTED_ORIGINS` setting? - https://docs.inventree.org/en/latest/start/config/#server-access

What errors / feedback are you getting here?

src/frontend/src/components/images/ApiImage.tsx
matmair: this seems resolved now

matmair:
Seems to fix #5697 put I would like to see this confirmed after the demo server runs this version before closing that

SchrodingersGat:
@matmair does not seem to work locally for me. I have pulled down the code, and running in a private browser session. Stuck at the "login" screen:

image


image


The "token" request is being made correctly, and returns:

- api token
- CSRF token
- session token

image

But none of these tokens get sent in the next request to `/api/user/me/`:

image

And so the server rejects it of course:

image

src/frontend/src/functions/auth.tsx
```diff
@@ -7,49 +7,36 @@ import { api, setApiDefaults } from '../App';
...
+  // Login to the server
   await axios
     .get(apiUrl(ApiEndpoints.user_token), {
       auth: { username, password },
       baseURL: host,
-      timeout: 2000,
-      params: {
-        name: tokenName
-      }
+      timeout: 2000
```
SchrodingersGat, line 28: You need to have `withCredentials: true` here, so that the cookies are accepted from the response header. Otherwise the `set-cookie` response headers are ignored.

Ref: https://stackoverflow.com/questions/52549079/does-axios-support-set-cookie-is-it-possible-to-authenticate-through-axios-http

SchrodingersGat:
@matmair it seems that this approach does not let the user upload files?

matmair:
It seems to work localy but we would need to change CSRF to lax - not sure if it is worth it

SchrodingersGat:
> It seems to work localy but we would need to change CSRF to lax - not sure if it is worth it

Can you expand on this? What adjustments would we need to make to CSRF?

matmair:
Replaced by https://github.com/inventree/InvenTree/pull/6970

Issue #5804, opened 2023-10-28
Updating Inventree to 0.12.8. fails at datatbase migration


MatjazSi:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [X] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

Hi,
When trying to update Inventree to the latest version (0.12.8) I get an error in the database migration step (see log output section)

### Steps to Reproduce

Followed instructions on this [page](https://docs.inventree.org/en/stable/start/migrate/#migrating-data-to-newer-version)

1. Stop server:
sudo docker-compose down

2. Backup DB:
sudo docker-compose run inventree-server invoke backup

3. Download new version:
sudo docker-compose pull

Migtrate DB to new version:
sudo docker-compose run inventree-server invoke update

This is whwere error happens

### Relevant log output

```bash
Running InvenTree database migrations...
========================================
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
No changes detected
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
Operations to perform:
  Apply all migrations: InvenTree, account, admin, auth, authtoken, build, common, company, contenttypes, django_q, error_report, exchange, flags, label, order, otp_static, otp_totp, part, plugin, report, sites, socialaccount, stock, taggit, user_sessions, users
Running migrations:
  Applying InvenTree.0001_initial...Traceback (most recent call last):
  File "/root/.local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "django_migrations_pkey"
DETAIL:  Key (id)=(431) already exists.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/inventree/InvenTree/manage.py", line 23, in <module>
    execute_from_command_line(sys.argv)
  File "/root/.local/lib/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
[...]

PR #6407, opened 2024-02-05
[Docker] Pin postgresql version

SchrodingersGat:
- Our docker compose setup specifies postgresql:13
- Pin postgresql libs to v13
- Should prevent version issues between InvenTree image and db image
- Ref: https://github.com/inventree/InvenTree/issues/5804#issuecomment-1907751541


matmair (approved):
Lgtm

PR #6410, opened 2024-02-05
[Plugin] Enhanced custom validation

SchrodingersGat:
This PR provides a significant improvement to the `ValidationMixin` plugin - allowing greater flexibility in custom validation of database models

- Adds a `validate_model_instance` method to the `ValidationMixin` plugin
- Exposes "deltas" to the method - allowing plugins to validate based on which fields have been changed
- Exposes *most* database models to the this method using the `PluginValidationMixin` model mixin class
- Updates docs and adds more clear examples

### DIFF
--- a/InvenTree/part/models.py
+++ b/InvenTree/part/models.py
@@ -3367,9 +3367,7 @@ class PartCategoryStar(models.Model):
     )
 
 
-class PartTestTemplate(
-    InvenTree.models.MetadataMixin, InvenTree.models.InvenTreeModelBase
-):
+class PartTestTemplate(InvenTree.models.InvenTreeMetadataModel):
     """A PartTestTemplate defines a 'template' for a test which is required to be run against a StockItem (an instance of the Part).
 
     The test template applies "recursively" to part variants, allowing tests to be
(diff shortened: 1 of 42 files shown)


InvenTree/plugin/helpers.py
```diff
@@ -220,6 +220,37 @@ def get_plugins(pkg, baseclass, path=None):
...
+
+    if name and cfg.name != name:
+        # Update the name if it has changed
+        try:
+            cfg.name = name
+            cfg.save()
+        except Exception as e:
+            logger.exception(f'Failed to update plugin name: {e}')
+
+    return cfg
```
matmair, line 251: Why is this moved out of the registry? IMO it makes more sense there as the registry is always made available under a constant import path

InvenTree/stock/models.py
```diff
@@ -26,20 +26,13 @@
...
 from InvenTree.fields import InvenTreeModelMoneyField, InvenTreeURLField
-from InvenTree.models import (
-    InvenTreeAttachment,
-    InvenTreeBarcodeMixin,
-    InvenTreeNotesMixin,
-    InvenTreeTree,
-    MetadataMixin,
-    extract_int,
-)
 from InvenTree.status_codes import (
```
matmair, line 36: Question: Why is this import changed?

InvenTree/part/models.py
```diff
@@ -37,6 +37,7 @@
 import common.settings
 import InvenTree.conversion
 import InvenTree.fields
+import InvenTree.models
```
matmair, line 40: Question: Why is this import changed?

InvenTree/plugin/helpers.py
SchrodingersGat: Fair enough!

PR #6418, opened 2024-02-05
Support custom user model

matmair:
This PR refactors the usage of the build in User model to support custom auth models

- [x] refactors all uses of User
- [x] Add migrations/refactor migrations
- [ ] Docs
- ~~[ ] unit tests to ensure custom user models work~~


wolflu05:
Does this also break related model fields in plugins that use the auth.user model?

matmair:
@wolflu05 yes, that is the main "breaking" thing

matmair:
This is targeted for 0.16.0 now and will be revisited once 0.15.0 is out.

I think it is already advisable for plugin deps to either target Owners or the django.cong.settings auth model - a list of best practices is on my personal roadmap for 0.16.0

SchrodingersGat (approved):

SchrodingersGat:
@matmair LGTM

- Any documentation needed around this? 
- What happens to the migrations if  a custom user model is actually used? Do we need to edit existing migrations?

matmair:
Needs docs


> * What happens to the migrations if  a custom user model is actually used? Do we need to edit existing migrations?

I have tested that it works but I will try to add a CI test for that too

SchrodingersGat:
@matmair do you want to target 0.16.x or bump to 0.17.x?

matmair:
@SchrodingersGat let's move this, I am focusing on getting user custom states in for 0.16.x

matmair:
Closing this in favour of an upcoming PR that adds a custom user model

PR #6435, opened 2024-02-06
Add parent to the filterable fields of the StockLocations and PartCat…

martonmiklos:
…egories

### DIFF
--- a/InvenTree/InvenTree/api_version.py
+++ b/InvenTree/InvenTree/api_version.py
@@ -1,10 +1,13 @@
 """InvenTree API version information."""
 
 # InvenTree API version
-INVENTREE_API_VERSION = 166
+INVENTREE_API_VERSION = 167
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
+
+v167 -> 2024-02-06 : https://github.com/inventree/InvenTree/pull/6431
+    - Add parent to the filterable fields of the StockLocations and PartCategories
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@martonmiklos LGTM. FYI you can already filter by `category` (for `PartCategory` API) and `location` (for `StockLocation` API). But I think a generic parent filter is a good addition

martonmiklos:
> @martonmiklos LGTM. 

Ahm I put back to WIP because some tests are failing and I want fix them.
 

> FYI you can already filter by category (for PartCategory API) and location (for StockLocation API). But I think a generic parent filter is a good addition

Well I made this change to support some tool development based on the API.
Under the Stock location list I do not see the location filter:
kép

Am I missing something locally?

SchrodingersGat:
> Am I missing something locally?

No. This is a very "old" part of the code which does not get introspected properly by the API generation - something we need to fix!

Actually, upon further inspection I see that the "parent" filter *is* actually already there - it just is a *hidden* field...

image

So what we need to do is:

- Move the "parent" filter into the appropriate `FilterSet` class (so that it gets exposed to the API docs)
- Work out how to support the "cascade" option (this is why it currently sits in the current `filter_queryset` method

Note that any custom filtering in the `filter_queryset` method does not get exposed to the API docs

SchrodingersGat:
@martonmiklos I'm replacing this with https://github.com/inventree/InvenTree/pull/6536 - please refer to the code there :)

Issue #6434, opened 2024-02-06
[FR] Sales Order Additional Status


altispeedkschmidt:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

The issue I am running into is that when using the sales order function in the program there is no status to differentiate between a "shipped" sales order and a "Complete" sales order. As it stands when you hit the "Ship Items" button in the program it dose not change the status to "Shipped" it keeps the status of  "In Progress" until after you have then hit "Complete Order" 

The workflow I am trying to accomplish is the following.

![image](https://github.com/inventree/InvenTree/assets/120754288/e9f45862-9673-446a-8533-3dd0e5590f59)

### Suggested solution

Add an additional status for "Completed" to the sales orders function in Inventree.

### Describe alternatives you've considered

Currently if I want to see if an order has been shipped but not marked as completed I have to open each individual sales order instead of at a glance being able to see the status of all orders at one glance from one page.

### Do you want to develop this?

- [ ] I want to develop this.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/inventree) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/inventree/InvenTree/issues/6434">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/inventree/InvenTree/issues/6434/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/inventree/InvenTree/issues/6434/pledge.svg">
</picture>
</a>

SchrodingersGat:
@altispeedkschmidt this makes sense, and would be a good addition. Should not be too difficult to implement.

Given our current development workload, would you be willing to implement this yourself, or provide a small amount of sponsorship to see it prioritized?

tsimonq2:
I'll take this from the Altispeed side, feel free to assign the issue to me.

Thanks!

altispeedkschmidt:
@SchrodingersGat I have spoken with @tsimonq2 one of our software developers at Altispeed Technologies and he has offered to take on this project and commit a pull request with the features described in this feature request.

SchrodingersGat:
Awesome, looking forward to it :)

PR #6449, opened 2024-02-07
Add support for a Completed status (Fixes: #6434)

tsimonq2:
Per #6434 (filed by my colleague at Altispeed), this is the user interface and backend state. This does piggyback off of some existing `shipped` functionality, but does work.


SchrodingersGat:
@tsimonq2 looks like you know what you are doing! LMK if you need any specific pointers :)

SchrodingersGat:
@tsimonq2 any progress on this? I'd like to see this integrated and not get lost to the sands of time :)

altispeedkschmidt:
> @tsimonq2 any progress on this? I'd like to see this integrated and not get lost to the sands of time :)

I know this is still on @tsimonq2  list of dev projects. He is currently assigned a pretty big project of developing a parcel tracking software so this may take a while to get to but we will make sure it gets added as once we get this feature added to Inventree we will be implementing it at our company.

tsimonq2:
@SchrodingersGat How does this look? It works :smile:

InvenTree/order/migrations/0099_alter_salesorder_status.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('order', '0098_auto_20231024_1844'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='salesorder',
+            name='status',
+            field=models.PositiveIntegerField(choices=[(10, 'Pending'), (15, 'In Progress'), (20, 'Shipped'), (30, 'Complete'), (40, 'Cancelled'), (50, 'Lost'), (60, 'Returned')], default=10, help_text='Purchase order status', verbose_name='Status'),
```
matmair, line 16: This should be adapted to work like below-linked line to ensure we do not need migrations on states in the future.

https://github.com/matmair/InvenTree/blob/eecda20b83e42b10b43b12e78ebb4c38f26916c7/InvenTree/stock/migrations/0102_alter_stockitem_status.py#L18

InvenTree/order/models.py
```diff
@@ -1086,7 +1089,7 @@ def issue_order(self):
         )
 
     @transaction.atomic
-    def complete_order(self, user, **kwargs):
+    def ship_order(self, user, **kwargs):
```
matmair, line 1092: Not sure if this is not breaking backwards combat for plugins - thoughts @SchrodingersGat @wolflu05?

matmair (approved):
LGTM in general;
@SchrodingersGat do we also want the new button/modal implemented in PUI or would this be a separate issue?

InvenTree/order/migrations/0099_alter_salesorder_status.py
SchrodingersGat: Agreed - please make the change as suggested here @tsimonq2

InvenTree/order/models.py
SchrodingersGat: It could - but any plugin which calls this method will need to know about the new status code too. I'm not sure if this is something we should account for (e.g. with a deprecation warning) or just leave

matmair: At least a warning in the release notes and a breaking label is waranted

SchrodingersGat: I think that should be sufficient.

InvenTree/order/migrations/0099_alter_salesorder_status.py
tsimonq2: How does my recently-pushed commit look for this? @SchrodingersGat @matmair

matmair: Looks good @tsimonq2 !

SchrodingersGat:
@tsimonq2 are you able to address the failing QI issues as above?

SchrodingersGat:
@tsimonq2 if you were able to get back to this soon, it could still make the next release cycle. Please let me know.

tsimonq2:
@SchrodingersGat Should be all set now :)

SchrodingersGat:
@tsimonq2 many thanks for the contribution :)

matmair:
@SchrodingersGat this needs to be called out as breaking for plugins that call complete_order in the release notes as the effect changes

PR #6422, opened 2024-02-06
Maintenance Mode Fix

SchrodingersGat:
There are ongoing issues with using the StaticFiles backend for `django-maintentance-mode`.

This PR provides a custom backend which utilizes the `InvenTreeSetting` DB storage class. Hopefully should provide more reliability.

- Fixes https://github.com/inventree/InvenTree/issues/6352
- Fixes https://github.com/inventree/InvenTree/issues/6156

PR #6451, opened 2024-02-08
Maintenance Mode Improvements

SchrodingersGat:
Follow up to https://github.com/inventree/InvenTree/pull/6422

Provides some enhancements to maintenance mode support.

- Force maintenance mode when running management commands which change database data
- Custom "migrate" command (which wraps db migrations in management mode)


InvenTree/common/models.py
```diff
@@ -703,6 +703,9 @@ def set_setting(cls, key, value, change_user=None, create=True, **kwargs):
                     attempts=attempts - 1,
                     **kwargs,
                 )
+        except (OperationalError, ProgrammingError):
+            # Likely the DB is locked - not much we can do here
+            pass
```
matmair, line 708: I think we should at least log this

tasks.py
```diff
@@ -369,10 +369,8 @@ def migrate(c):
     print('Running InvenTree database migrations...')
     print('========================================')
 
-    manage(c, 'makemigrations')
-    manage(c, 'migrate --noinput')
-    manage(c, 'migrate --run-syncdb')
```
matmair, line 374: Question: are you sure we do not issue the sync command anymore?

matmair (approved):
LGTM

tasks.py
SchrodingersGat: Added it back in :)

PR #6452, opened 2024-02-08
(Yet another) Fix for dockerfile

SchrodingersGat:
Small tweak to the docker build, following on from new postgresql lib requirements.

Removing the `psycopg-c` python library, it requires `postgresql-libs` (150MB) to be installed at the system level. It is not required for functionality, and current throws a warning message in the logs.

---------------

Hopefully this is the final in a series of changes required to get the docker image working reliably with the new version of django


SchrodingersGat:
As discovered in https://github.com/inventree/InvenTree/pull/6453 - some of the issues here were to do with incorrect django settings - not the database backend.

PR #6453, opened 2024-02-08
Update Proxy support for django 4.2

SchrodingersGat:
Improve support for django configuration, in particular around host / proxy settings.

Updating to django 4.2 has changed the way that some core settings are specified, and these now need to be explicitly configured for InvenTree to run properly behind a proxy.

Ref : https://github.com/inventree/InvenTree/pull/6452 (turns out the issue is proxy, not db)


matmair:
@SchrodingersGat this will need to be in the release notes

PR #6456, opened 2024-02-08
[WIP] Docusaurus Docs

SchrodingersGat:
This PR represents a work in progress to update the documentation to a modern platform - specifically [docusaurus.io](https://docusaurus.io)

### Major Goals

- Update docs to a modern framework which is easier to develop in and supports some new features that we want
- Ensure that the docs and website use the same tooling and are visually consistent - see https://github.com/inventree/inventree-website/pull/177
- Framework that is more consistent with the InvenTree development process (react support)

### Tech Stack Goals

With docusaurus we can more easily achieve the following running goals:

- Include release notes in documentation
- Include versioned API schemas in documentation - [docusaurus-openapi-docs](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs)
- Improve support for [versioned API documentation](https://docusaurus.io/docs/versioning)
- Add code documentation support
- Improved support for diagrams / mermaid / etc

### TODO List

- [ ] Migrate existing documentation
- [ ] Update / transfer "admonitions" (different styling in docusaurus)
- [ ] Change pages from `.md` to `.mdx` (only where required for extra features)
- [ ] Configure scripting for making releases
- [ ] Configure generation of (versioned) API documentation (Ref: https://github.com/inventree/InvenTree/pull/6319)
- [ ] Configure generation of (versioned) code documentation
- [ ] Get build preview working with netlify
- [ ] Build and host site
- [ ] Update support for image integration / click-to-zoom etc

-------------------


matmair:
This is an improvement in search of a problem IMO. Using a fast-moving thing like a React for docs  will leave us with a lot more maintenance work to just keep it going.

I have a lot of problems with that framework but here are a few important ones:
- Nowadays you can easily create PDFs or EPUBs from the docs, that is not possible/hard with MDX - I do not see what usesfull functionallity MDX brings for that huge downside
- Their version support is a joke - duplicating files for each version instead of using git. Instead of fixing the architectural problem they just [suggest](https://docusaurus.io/docs/versioning#keep-the-number-of-versions-small) removing old version
- They have a huge list of dependencies leading to a lot of breaking changes. Take a look at their [recent major release](https://docusaurus.io/docs/versioning#keep-the-number-of-versions-small) - I can not remember us dealing with anything the large with mkdocs

We can achive all tech stacks goals with the current doc stack. It is not a lack of tools but time that has prevented it form beeing done and I do not see how this change would improve that, I just see added maintenance burden.

matmair:
I think the new website has the potential to look nice but the stack doe not make sense.

### Rant about current/propsed stack
The beauty of the current doc (and to some limited degree website) stack is that it closely follows long-held standards in the python/static website generator standards like GFM, easily portable content and separation between logic, theming and content. That keeps it more or less evergreen. Most of the features in Docusaurus seem to be geared towards fast-moving frontend devs/designers that want to have something "professional" looking up and running in a few hours. You get pretty results in a few minutes but once you look into technical details everything is a naive 80% solution.

### Further development suggestions

I agree with 1,2,5  but would like to discuss 3 and 4.
It probably makes sense to discuss the long-term goals of all the components in the (eco)system and lay out what (information, code) goes where, how things should work (tech stack, update cycles) and where we document things like this. I think something akin to a ADR process ([full-scope docs](https://docs.aws.amazon.com/prescriptive-guidance/latest/architectural-decision-records/adr-process.html) by AWS, [lighter](https://adr.github.io/#lightweight-adrs-should-be-adopted) version) makes sense - Pyhton has PIPs for that for example.

wolflu05:
I would actually had supported this idea, but if @matmair is not happy if maybe that's not the right direction and and closing this PR is the best option. 

But IMO we need to improve the look and feel of the website and the docs (to also align with the new PUI refactor), if we want to gain more users (what @matmair calls, do better marketing). Modern users like modern websites more than old school style sites.

matmair:
Design and tech stack are different things - trying to solve one with the other is a fools errant.

@wolflu05  The website was created in Feb 2022 with a focus on being accessible, working on every device and not needing unnecessary tech. There has been very little feedback on its design so it stayed like it was proposed in the demo. If the website is keeping a user from using InvenTree they won't like the software in it's current state either.

SchrodingersGat:
Closing this out - targeting API docs in https://github.com/inventree/InvenTree/pull/6463

Issue #6350, opened 2024-01-28
[FR]Add servers section to the OpenAPI schema


martonmiklos:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

As I mentioned in the #4587 some of the openapi-codegen generators (Qt/C++ client generator for sure) expects the servers section to be present in the schema descriptor yaml file, because they generating the API endpoints to the code.

The InvenTree generated schema lacks it so it has to be manually extended before generating code.

### Suggested solution

It would be straightforward to add the servers section into the SPECTACULAR_SETTINGS in the [settings.py](https://github.com/inventree/InvenTree/blob/master/InvenTree/InvenTree/settings.py#L489), however I have no idea how can I 'smuggle' the re the base url. 

### Describe alternatives you've considered

Adding manually.

### Examples of other systems

-

### Do you want to develop this?

- [ ] I want to develop this.

matmair:
There is an active PR for openApi generation, maybe feedback regarding the schema should be placed there: https://github.com/inventree/InvenTree/pull/6319

SchrodingersGat:
@martonmiklos you are asking for the ability to have the local instance base URL in the servers section? as in, when you run `invoke schema` locally, the "server" will include *your* server URL?

martonmiklos:
@SchrodingersGat well I was not aware of the local `invoke server` command I have used the /schema/ endpoint so far. It would be good if via that endpoint the servers section would be filled with the same url base through what I downloading the schema.

SchrodingersGat:
Check out this PR - https://github.com/inventree/InvenTree/pull/6463 - I have made a change which allows this to be filled out, *if* you have a globally defined SITE URL

SchrodingersGat:
Closed by #6463

Issue #6269, opened 2024-01-17
[FR] API schema tracking in QC


matmair:
I would like to create a new org repo that holds all API schemas from now on and maybe a few versions back(starting from the point where it was possible to create OpenAPI schemas).
Additionally, I would like to add a CI check that generates the schema of the current PR and checks it against the latest one from the new repo to ensure API changes do not go unnoticed.
Thoughts? @inventree/triage

_Originally posted by @matmair in https://github.com/inventree/InvenTree/discussions/5861_

SchrodingersGat:
@matmair with reference to https://github.com/inventree/InvenTree/pull/6456

Would it be a better approach to encode the API schema into the documentation itself?

Your [API change detection PR](https://github.com/inventree/InvenTree/pull/6440) can be used to ensure that the API schema documentation is always up-to-date. We can potentially use [something like this](https://www.neoteroi.dev/mkdocs-plugins/web/oad/) to generate docs pages from the schema file. Additionally the schema file can be made available for download.

This means that the "latest" API is always accessible on the docs site.

Then whenever we make a versioned release, that release has the pinned API docs available too.

If we can get this workflow up and running, is there an advantage to a separate repo where we track these API changes also?

matmair:
@SchrodingersGat I think we should include as little generated content as possible in the source repo (inventree/inventree) to keep the git repo small/useable. The user can always create the schema purely from the source code using the schema command (that is reproducible).
w
Generating API docs is certainly something to shoot for but should take place in the build step. As a comparison: We also do not commit the HTML for the docs or the rendered language js files to the main repo.
A separate repo keeps the main history cleaner and leaves us more flexibility in regard to repo layout.

SchrodingersGat:
Check out https://github.com/inventree/InvenTree/pull/6463 - we can generate the API as part of the docs generation, and provide:

- A documented schema in the docs
- A copy of the schema file for download

In this manner, we get versioned snapshots of the API for "free" without having to track another external repo - readthedocs does this for us

matmair:
How do you reference specific API versions (for automatic client generation/backtesting between versions) with that? Are we releasing docs for every API version and does this mean we will have 2 different version numbers in releases? That does not seem easier than having another org repo similar to the Python client or app.

SchrodingersGat:
[...]

PR #6440, opened 2024-02-07
Add API change detection

matmair:
This PR is a follow-up to https://github.com/inventree/InvenTree/pull/6319 and adds a CI step that checks if the API schema was changed. If the api_version file was not changed but the diff between the published version of the schema and the worktree version of the schema is not clean the Job fails.
If the api_version file was changed and the CI runs on master the new schema is pushed to the reference repo (it needs a PAT for that).

Closes #6269
Closes #5861

PR #6463, opened 2024-02-10
Render API docs

SchrodingersGat:
Render API schema as part of documentation.

- A snapshot of each schema will be available at each release, and also for "latest" (master)
- API schema file available for download (`.yaml` file)

### References

- https://github.com/inventree/InvenTree/pull/6319
- https://github.com/inventree/InvenTree/issues/6350
- https://github.com/inventree/InvenTree/issues/6269


docs/docs/api/schema.md
```diff
@@ -0,0 +1,9 @@
+---
+title: API Schema
+---
+
+The InvenTree API is implemented using the [Django REST framework](https://www.django-rest-framework.org). The API schema as documented below is generated using the [drf-spectactular](https://github.com/tfranzel/drf-spectacular/) extensions.
+
+## Schama
```
matmair, line 7:
```suggestion
## Schema
```

docs/docs/api/schema.md
```diff
@@ -0,0 +1,9 @@
+---
+title: API Schema
+---
+
+The InvenTree API is implemented using the [Django REST framework](https://www.django-rest-framework.org). The API schema as documented below is generated using the [drf-spectactular](https://github.com/tfranzel/drf-spectacular/) extensions.
```
matmair, line 5:
```suggestion
The InvenTree API is implemented using the [Django REST framework](https://www.django-rest-framework.org). The API schema as documented below is generated using the [drf-spectactular](https://github.com/tfranzel/drf-spectacular/) extension.
```

matmair (commented):
Seems reasonable. Is the api supposed to be rendered already?

SchrodingersGat:
> Seems reasonable. Is the api supposed to be rendered already?

It did not seem to be rendering reliably with that library. I have tried a new one - check it out now and see what you think

InvenTree/InvenTree/settings.py
```diff
@@ -1177,3 +1162,23 @@
...
+        'description': 'More information about InvenTree in the official docs',
+        'url': 'https://docs.inventree.org',
+    },
+    'VERSION': str(inventreeApiVersion()),
+    'SERVE_INCLUDE_SCHEMA': False,
+    'SCHEMA_PATH_PREFIX': '/api/',
+}
+
+if SITE_URL:
+    SPECTACULAR_SETTINGS['SERVERS'] = [SITE_URL]
```
martonmiklos, line 1184: The url subkey is missing here:

> # Optional list of servers.
>     # Each entry MUST contain "url", MAY contain "description", "variables"
>     # e.g. [{'url': 'https://example.com/v1', 'description': 'Text'}, ...]
>     'SERVERS': [],

wolflu05: Maybe we can also use this then to auto generate a ts client (fully
Typesafe) for PUI (and also publish as npm package for the users) and a python client to eventually replace the InvenTree-Python package to reduce maintainance tasks there.

matmair: The stuff needed for that is in https://github.com/inventree/InvenTree/pull/6440

docs/api.yml
matmair, line 1: Having an auto-generated file here will create a lot of fun with merge conflicts and blow up the repo size, we have had this discussion multiple times in the last year with frontend resources and I pointed this out [2 days ago](https://github.com/inventree/InvenTree/issues/6269#issuecomment-1934625949).

SchrodingersGat: This is only a WIP to ensure that we are all happy with the generated output. Once that is all clear, I'll setup a build pipeline to render this file on docs build, rather than actually committing the file itself

InvenTree/InvenTree/settings.py
SchrodingersGat: > The url subkey is missing here:
> 
> > # Optional list of servers.
> > ```
> > # Each entry MUST contain "url", MAY contain "description", "variables"
> > # e.g. [{'url': 'https://example.com/v1', 'description': 'Text'}, ...]
> > 'SERVERS': [],
> > ```
> 
> Apart from this it is working fine: I had been able to generate code with openapi-codegen without any hacking now!

Thanks, should be fixed now :)

SchrodingersGat:
The generated schema file (`.html`) is `16MB` - which might take some time to load. Do we think this is an issue?

wolflu05:
This is really laggy to open with a browser. How about splitting it in multiple pages categorized by the Django apps, so each apps api has a sub page. This reduces the download size and also the page size.

SchrodingersGat:
@wolflu05 nice idea - any idea how to achieve that?

wolflu05:
Not really sure about how to support that, the mkdocs plugin seems like it's not. So we would need to either find a way to export the schema for each app individually or write a script to divide the schema into multiple pieces and then generate multiple pages that use the OAD plugin.

SchrodingersGat:
>  or write a script to divide the schema into multiple pieces and then generate multiple pages that use the OAD plugin.

Looking into this approach, doesn't seem too bad...

SchrodingersGat:
Ok @matmair @wolflu05 I have a pipeline working - please check out the rendered docs and LMK what you think of the current state

wolflu05:
This already looks pretty good. Awesome work.

But I would adjust a few things:
1. at the bottom of each page are all models listed, maybe only list the required models for each sub page when splitting the schema
2. maybe improve the navigation between the sub pages, so that I don't need to first click on the Extend > InvenTree API > API schema and then select to correct sub page (eg plugin) on the api schema page, but instead maybe Extend > InvenTree API Schema > Plugin or Extend > InvenTree API Schema > Overview for the index page
3. is it correct that if a new Django app is added, a new api markdown file needs to be created and the schema splitting tool needs some adjustments to the list of sub pages?

SchrodingersGat:
@wolflu05 can you suggest how to implement any of those? Otherwise I'd be keen to merge this as is, and work on improvements in a later PR.. Thoughts?

wolflu05:
1. while splitting the schema, analyze all referenced models and only output those in the splitted schema. This drastically will improve page performance and load speed
2. Already explained how we could do that, by moving the api schema documentation one level more to the top of the navigation like we did with plugins and plugin mixins.
3. We could maybe auto discover the available apps, but I think the current solution with adding it manual is sufficient. (I guess the machine API needs to be added since #4824 is merged?)
4. That can't be easily solved here, we would need to improve the used plugin, but I think that's just a nice to have, so keeping it like now is fine

SchrodingersGat:
@wolflu05 point 1) should be implemented now, 3) and 4) I think would be not possible without digging into the plugin some more. LMK what you think as now

wolflu05:
Performance and load speed is much better now, thank you really much. Maybe it's worth putting the machine api endpoints into its own page, but I'm happy with it now.

matmair:
@wolflu05 seems like the machine API is throwing an error and not building https://github.com/inventree/InvenTree/actions/runs/7959068634/job/21725283385?pr=6463#step:4:40

SchrodingersGat:
> @wolflu05 seems like the machine API is throwing an error and not building https://github.com/inventree/InvenTree/actions/runs/7959068634/job/21725283385?pr=6463#step:4:40

@matmair why is the CI passing then - the API schema errors should cause CI to fail...

SchrodingersGat:
@wolflu05 I have split `machines` out into their own path now

wolflu05:
Thank you. 

This error is interesting, because there is actually a correct schema annotation with extend_schema. Maybe this error caused no ci fail, because only the OAD plugin which generates the md fails and not while exporting the schema?

matmair:
@SchrodingersGat as far as I can read the CI runs the API schema task failed and threw an error

wolflu05:
Yes, but not the generation step. That passed. The difference checking step is the step that failed. Has that something todo with the ongoing gh actions beta where there are bugs from time to time like this?

matmair:
It fails within the CI job, I think that is enough

wolflu05:
I see, but why does the ci does not fail on master then? The machine api is already merged in there.

SchrodingersGat:
@matmair @wolflu05 any ideas how we should proceed here? This is the final blocker for the `0.14.0` release

matmair:
AFAIK there is a schema issue and a problem with translations; I have just restarted the docs CI - maybe it runs now

SchrodingersGat:
@matmair any chance you can have a look at this? I am still a bit confused by the CI process here

wolflu05:
@SchrodingersGat The issue occurs while exporting the schema. A requests serializer is missing for introspection in one of the machine API serializers. However this doesn't throw an error, but instead the `schema.yml` file is not created, so a later step expecting this file failed because it cannot find it. 

Regarding the machine api error it seems like this can be fixed with this patch:
```diff
diff --git a/InvenTree/machine/api.py b/InvenTree/machine/api.py
index 9448d2ea9..2fec1b4cc 100644
--- a/InvenTree/machine/api.py
+++ b/InvenTree/machine/api.py
@@ -144,7 +144,7 @@ class MachineRestart(APIView):
 
     permission_classes = [permissions.IsAuthenticated]
 
-    @extend_schema(responses={200: MachineSerializers.MachineRestartSerializer()})
+    @extend_schema(request=None, responses={200: MachineSerializers.MachineRestartSerializer()})
     def post(self, request, pk):
         """Restart machine by pk."""
         machine = get_machine(pk)
```

And to already fail in the export step, we may could check for the existence of the file and if not exit 1.

PR #6467, opened 2024-02-11
Fix uncaught Company.DoesNotExist error in SupplierBarcodeMixin

30350n:


SchrodingersGat:
Thanks for the fix @30350n

PR #6468, opened 2024-02-11
Fix APICallMixin url_args handling

30350n:
The current implementation of `APICallMixin.api_call` handles `url_args` very weirdly. This improves the behavior, so that non single character string args will be handled correctly.

Before:
```py
>>> api_build_url_args({'a': 'abc123'})
'?a=a,b,c,1,2,3'
```

After:
```py
>>> api_build_url_args({'a': 'abc123'})
'?a=abc123'
```


SchrodingersGat:
Nice catch :)

PR #6472, opened 2024-02-12
Fixing Header to rely on location and not parameter.

### LINKED ISSUE #6471: Platform Header Tabs vanish on refresh
### Discussed in https://github.com/inventree/InvenTree/discussions/6470

<div type='discussions-op-text'>

<sup>Originally posted by **CodingPupper3033** February 12, 2024</sup>
Hello! I've been trying to wrap my head around the new ui's source code, and I can't figure something out. 

In Headers.tsx, function NavTabs, we have the variable tabValue to tell the tabs which one it's on. It uses parameters, so I looked at the router.tsx file, and I don't see any `:tabValue,` so how does the value get set?

Also, I think there's a minor bug based on this where if you use the browser bar to navigate to a page and not a tab, the header (and just the header) will think you're on the default home page. It's not a critical bug, but it exists. </div>

Issue:
The header tabs value is set to be a react-router dom parameter; however, the parameter is not set in the router. This means that on a refresh or navigating to the URL directly, the tab will be incorrectly labeled as the Home tab (default tab).

I still can't figure out how to assign labels and the issue to myself, but I have a fix. I'll try uploading it!

CodingPupper3033:
Closes #6471
Fixes issue #6471 

Let me check the commit test server thing to see if it's working before it's merged.

### DIFF
--- a/src/frontend/src/components/nav/Header.tsx
+++ b/src/frontend/src/components/nav/Header.tsx
@@ -3,7 +3,7 @@ import { useDisclosure } from '@mantine/hooks';
 import { IconBell, IconSearch } from '@tabler/icons-react';
 import { useQuery } from '@tanstack/react-query';
 import { useState } from 'react';
-import { useNavigate, useLocation } from 'react-router-dom';
+import { useLocation, useNavigate } from 'react-router-dom';
 
 import { api } from '../../App';
 import { navTabs as mainNavTabs } from '../../defaults/links';
--- a/src/frontend/src/components/nav/Header.tsx
+++ b/src/frontend/src/components/nav/Header.tsx
@@ -3,7 +3,7 @@ import { useDisclosure } from '@mantine/hooks';
 import { IconBell, IconSearch } from '@tabler/icons-react';
 import { useQuery } from '@tanstack/react-query';
 import { useState } from 'react';
-import { useNavigate, useParams } from 'react-router-dom';
+import { useNavigate, useLocation } from 'react-router-dom';
 
 import { api } from '../../App';
 import { navTabs as mainNavTabs } from '../../defaults/links';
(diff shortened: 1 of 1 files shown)


CodingPupper3033:
^ The `commit test server thing` translates to `Deploy Preview.` It appears to work; however, I am slightly confused as to what the Style error is telling me.

CodingPupper3033:
I see; it was telling me what to change, not what the issue was. Got it!

src/frontend/src/components/nav/Header.tsx
```diff
@@ -100,12 +100,13 @@ export function Header() {
 
 function NavTabs() {
   const { classes } = InvenTreeStyle();
-  const { tabValue } = useParams();
+  const defaultTabValue = 'home';
+  const tabValue = useLocation().pathname.split('/')[1] || defaultTabValue;
```
matmair, line 104: Not sure that is a great sulution, this is indirectly hardcoded to the current schema - which might change at any point. Thoughts @wolflu05 ?

matmair (changes requested):

src/frontend/src/components/nav/Header.tsx
wolflu05: Yes, this would also be my first concern, as we have the feature to configure a relative path where PUI is hosted. Maybe we can use that setting to somehow strip the prefix or find a solution to only get the url from after the prefix. 

And secondly it would be better to wrap the .split and parsing of the panel name into a useMemo hook, so this is in a state and does not cause a rerendering.

SchrodingersGat:
How important is it that the current "top level tab" is indicated in this fashion? There are many pages (e.g. settings / notifications / etc) which do not even fit under this scheme.

The [<Tabs> component](https://v6.mantine.dev/core/tabs/) makes sense for switching between content within a single page, but maybe feels like the wrong solution for global path navigation...

Thoughts?

wolflu05:
Personally I really like this navigation to switch between the core parts of InvenTree and would miss it if it would be removed.

SchrodingersGat:
To be clear I am not talking about removing the tabs themselves - just the "current tab" indicator (unless we can work out a good way to implement)

SchrodingersGat:
I am bumping this to `0.15.0` until we have some consensus on a sensible solution

matmair:
@CodingPupper3033 could you adress the merge conflict so we can get this into 0.16.0?

SchrodingersGat:
I am going to close this one due to lack of response from @CodingPupper3033

PR #6477, opened 2024-02-13
Add  BOM creation to the sofa example

SergeoLacruz:


SchrodingersGat:
Thanks @SergeoLacruz for the addition! I have cleaned up the wording a bit but otherwise looks good :)

PR #6480, opened 2024-02-14
Pint update

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/5527#issuecomment-1941104515


SchrodingersGat:
Unit tests are still broken as of `pint==0.23`.

As there is otherwise no urgent reason to change, closing this for now.

PR #6483, opened 2024-02-14
Added various PO fixes

wolflu05:
This PR fixes various issues with POs.

## Changes
- [x] If a new PO is created through the "Order Part" btn in CUI the newly created PO will now be selected correctly (`supplier_detail=true` was missing)
- [x] If there are multiple SP when ordering a part in CUI, the PO select will now be filtered based on the selected SP 
- [x] When adding or updating a new PO Line item, pricing is automatically recalculated based on the supplier part price break data, if available when the `auto_pricing` option is set (Added this field to CUI and PUI)
- [x] When adding a new PO Line item and the `merge_items` option is set, an existing lines quantity will be updated instead of adding a new PO line. (Added this field to CUI and PUI)
- [x] Adds tests for the new `auto_pricing` and `merge_items` functionality

---

I now got the idea, why shouldn't we fix this on the backend, by automatically using the current price break data if existing from on the SP? Doing it that way, PUI can also benefit from this change. So I would suggest overriding the `PurchaseOrder.save` method and implement the logic there?

While trying to fix this, I also noticed that there is a [`PurchaseOrder.add_line_item`](https://github.com/InvenTree/InvenTree/blob/658b0624eae3f6d30e34053a763b2c3be7ae6107/InvenTree/order/models.py#L489-L562) function which tries to increase the quantity of an existing line and also updates the price break accordingly.  But it seems like this is not used anywhere in the code. Any ideas?


SchrodingersGat:
Generally happy with the changes you have made here. Are you going to include any PUI stuff in this PR?

wolflu05:
> Are you going to include any PUI stuff in this PR?

I don't think so, I just wanted to fix a few things that bothered me since a while.

@SchrodingersGat Can you take a look at my questions in the initial comment?

SchrodingersGat:
> I now got the idea, why shouldn't we fix this on the backend, by automatically using the current price break data if existing from on the SP? Doing it that way, PUI can also benefit from this change. So I would suggest overriding the PurchaseOrder.save method and implement the logic there?

So, filling in the pricing information based on the available supplier price break data if available? Makes sense!

> While trying to fix this, I also noticed that there is a [PurchaseOrder.add_line_item](https://github.com/InvenTree/InvenTree/blob/658b0624eae3f6d30e34053a763b2c3be7ae6107/InvenTree/order/models.py#L489-L562) function which tries to increase the quantity of an existing line and also updates the price break accordingly. But it seems like this is not used anywhere in the code. Any ideas?

I'm not sure about this - a vague memory perhaps. When was it last touched?

wolflu05:
> So, filling in the pricing information based on the available supplier price break data if available? Makes sense!

Yes, but on the backend if the frontend does not provide pricing data. 
1. It would be breaking, but what do you think about also removing the automatic pricing recalculation when adding a new line item in CUI to a PO. Because then the user leaves the pricing field empty it gets filled out by the backend.
2. Do we want to also automatically merge line items then like the add_line_item function would do, or is that something that is not wanted?

> I'm not sure about this - a vague memory perhaps. When was it last touched?

I think that was like 4-5 years ago, where I didn't know anything about InvenTree.

LavissaWoW:
@wolflu05 

> Do we want to also automatically merge line items then like the add_line_item function would do, or is that something that is not wanted?

Looking from a purchaser's standpoint: Most of the time, you're ordering X amount of SKU Y. For accounting, reception checks, etc, you'd rather avoid having the same SKU on multiple lines, unless you so intended that when setting up the PO. 
In my experience at least, having several lines of the same SKU is the edge case while having "one SKU, one line" is the normal outcome. 
Given that the edge case exists, I don't know if "enforced" automatic merging is the route to go down. It is what will be most applicable for normal day-to-day use, but having to fight the system automation the few times you need to do custom things can be frustrating.

One place where auto merging could make sense is when using the "Order Part" buttons on the part pages. It's already the shortcut amongst shortcuts, and you could add a switch in the form to say if you want to merge with possibly existing lines or not.

wolflu05:
Adding a default true checkbox (auto-merge) is a good idea, definitely will add that to 2 if that should be implemented.

wolflu05:
@SchrodingersGat this would be ready for review now. I now decided to add two fields to the serializer to control the merge line items and automatically set the purchase price based on SP price break data, functionality. Those are both now supported in PUI and also CUI. Additionally there are some fixes regarding POs with CUI included (for details see the PR description).

SchrodingersGat:
@wolflu05 thanks for the implementation and updates based on the notes above. I'm happy that this is working as intended.

PR #6488, opened 2024-02-14
Added pdf2image kwargs

wolflu05:
While investigating some print quality issues, I noticed that there are some issues with the rendered png. So I first looked into to intermediately generated pdf, but everything looks fine there. So I suspect this has something todo with the png rendering. After a lot of playing with different options I found out that the pdf2image library has an optional setting (`use_pdftocairo=True`) to use `pdftocairo` instead of `pdftoppm` which is used by default. 

| pdftoppm (current) | pdftocairo |
|--------|--------|
| image | image | 
| No sharp edges are rendered, which results in not always quadratical QR code pixels | |

So this PR adds an option to let the developer specify some custom options to pass over to `pdf2image.convert_from_bytes(...)`. Not sure if we should use Cairo by default, as this may can break some printing plugins?

### DIFF
--- a/InvenTree/machine/machine_types/label_printer.py
+++ b/InvenTree/machine/machine_types/label_printer.py
@@ -183,7 +183,7 @@ class LabelPrinterBaseDriver(BaseDriver):
             use_cairo (bool): Whether to use the pdftocairo backend for rendering which provides better results in tests,
                 see [#6488](https://github.com/inventree/InvenTree/pull/6488) for details. If False, pdftoppm is used (default: True)
             pdf2image_kwargs (dict): Additional keyword arguments to pass to the
-                [`pdf2image.convert_from_bytes`](https://pdf2image.readthedocs.io/en/latest/reference.html#pdf2image.pdf2image.convert_from_bytes) method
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
I would suggest that we add a specific optional argument such as `use_cairo=False` rather than making the plugin developers discover and pass the `use_pdftocairo` argument themselves. I'd also be on board with making it default to `True` given the better fidelity in your results

wolflu05:
I now added the direct `use_cairo` kwarg which is True by default now. But I still think it can have value to have all [pdf2image arguments](https://pdf2image.readthedocs.io/en/latest/reference.html#pdf2image.pdf2image.convert_from_bytes) exposed to the developer, so I would keep that, what do you think?

SchrodingersGat:
> I now added the direct `use_cairo` kwarg which is True by default now. But I still think it can have value to have all [pdf2image arguments](https://pdf2image.readthedocs.io/en/latest/reference.html#pdf2image.pdf2image.convert_from_bytes) exposed to the developer, so I would keep that, what do you think?

I agree, I am happy with the implementation as it stands

PR #6496, opened 2024-02-15
Plugin context fix

SchrodingersGat:
- Provide a *copy* of the page context to pass to plugin
- Prevents plugins from overriding page context when rendering

----------------

This fixes a very subtle (and hard to track down) bug where a broken plugin was corrupting the page rendering context data.


matmair:
Oh, this looks very annoying to debug/find - congrats on finding it @SchrodingersGat!

PR #6499, opened 2024-02-15
Test UV performance

matmair:
Uv, a new tool by the creators of ruff could speed up the pre-commit step considerably


matmair:
@SchrodingersGat yes—around 20 seconds on the pipeline and 15 on my device. The pre-commit hooks become nearly unnoticable.

matmair:
We can not merge right now though as there are still some unresolved issues upstream due to it being this new

matmair:
OK with 0.1.8 uv is functionally the same as pip tools. I think this could be reviewed/merged.

Changes:
- Switches internal and invoke install commands to use uv (there is a new optional switch `--nouv` that disables that)
- The pre-commit compile hook is very fast now ;-)
- The pre-commit compile hook now has a target Python version set - ie. you can run your dev setup on a newer, faster version of Python than 3.9 and will not get different requirements due to different standard libraries etc.

There are 3 changes in the requirement files as visible [in this diff](https://github.com/inventree/InvenTree/pull/6499/commits/01b075357f0ec2326e39c5c87e5b0a1410bb1af8):
- the tool header changes to uv
- there are no `via ....` references anymore - they are observed, just not listed
- extras (ie `toml` in this sample `coverage[toml]`)  are not listed but observed (there seems to be some discussions with pip-tools too if those should be in the lockfile)

All of those are cosmetic

SchrodingersGat:
@matmair looks good to me - thanks for putting the time into this!

Issue #6403, opened 2024-02-04
[PUI] Locale Name Translation


SchrodingersGat:
In the new react interface, the language selection options are not correctly translated:

![image](https://github.com/inventree/InvenTree/assets/10080325/b0872efe-f415-4d31-a654-4b5f047a3e35)

*Note: This issue does not exhibit when using the vite development server...*

PR #6528, opened 2024-02-20
Locales fix

SchrodingersGat:
Fix for translated locale names.

Closes https://github.com/inventree/InvenTree/issues/6403


wolflu05:
Good catch, makes sense now why it doesn't work. The context is not available and also cannot change one language state change. Putting it into a function acts like a hook and therefore it is context aware.

PR #6543, opened 2024-02-21
Seperate CUI url paths and tests

matmair:
This PR creates a clearer separation between CUI and PUI URLs, enabling users to fully disable CUI content serving.  It also tags CUI test to make it possible to disable running them temporarily.


InvenTree/order/test_views.py
```diff
@@ -1,5 +1,7 @@
 """Unit tests for Order views (see views.py)."""
 
+from django.conf.locale import ta
```
SchrodingersGat, line 3: Is this line an error?

InvenTree/stock/models.py
```diff
@@ -2304,7 +2309,9 @@ def get_api_url():
 
     def get_absolute_url(self):
         """Return url for instance."""
-        return f'/stock/track/{self.id}'
+        if settings.ENABLE_CLASSIC_FRONTEND:
+            return f'/stock/track/{self.id}'
+        return InvenTree.helpers.pui_url(f'/stock/id/{self.item.id}')
```
SchrodingersGat, line 2314: I think this pui_url is incorrect?

SchrodingersGat (commented):
Overall pretty reasonable, a bit of code to prune back out later on but should be OK

InvenTree/order/test_views.py
matmair: yes, fixed now

InvenTree/stock/models.py
matmair: You are right, fixed now

matmair:
@SchrodingersGat I hope to just remove most code that is tagged at that point ;-)

SchrodingersGat:
Sounds good. Happy to merge this when it passes?

matmair:
Yes please @SchrodingersGat

PR #6551, opened 2024-02-22
[WIP] docker / Caddy

SchrodingersGat:
Refactoring docker compose setup to use caddy instead of nginx.

### Motivation

- much simpler config
- better handling of auth redirects
- built in ssl
- improved localhost serving
- improved logging
- persistent storage of cert files, etc

### Todo

- [x] update install guide docs
- [x] add docs on persistant files for caddy (certs / etc)
- [x] test on demo server 
- [x] rebase onto correct remote... Oops..

### Further Work

The whole docker / docker-compose structure needs improving / simplifying:

- [ ] rework folder structure
- [ ] remove "dev" docker compose file (just use devcontainer)
- [ ] Simplify docker install / setup guide


docker/Caddyfile
```diff
@@ -0,0 +1,36 @@
+
+(log_common) {
+    log {
+        output file /var/log/caddy/{args.0}.access.log
+    }
+}
+
+# Change the host to your domain (this will serve at inventree.localhost)
+inventree.localhost {
```
matmair, line 9:
```suggestion
{$INVENTREE_SITE_URL:inventree.localhost} {
```

See for context https://caddyserver.com/docs/caddyfile-tutorial#environment-variables

docker/docker-compose.yml
```diff
@@ -82,51 +85,38 @@ services:
...
         depends_on:
             - inventree-server
-        env_file:
-            - .env
         ports:
-            # Default web port is 1337 (can be changed in the .env file)
-            - ${INVENTREE_WEB_PORT:-1337}:80
+            - 80:80
+            - 443:443
         volumes:
```
matmair, line 117:
```suggestion
        env_file:
            - .env
        volumes:
```

matmair (approved):
LGTM; I would suggest using the environment variable in the Caddyfile

SchrodingersGat:
@matmair nice suggestions :)

30350n:
From my testing in #6555 the auth uri should probably be changed to `uri /auth/` with a trailing slash, otherwise 301 redirects might cause problems here.

SchrodingersGat:
@30350n cheers

wolflu05:
Can we actually also use caddy without ssl behind another reverse proxy that manages a bunch of other services too?

docker/docker-compose.yml
```diff
@@ -99,30 +103,28 @@ services:
...
     inventree-proxy:
-        image: nginx:stable
+        container_name: inventree-proxy
+        image: caddy:alpine
+        restart: always
         depends_on:
             - inventree-server
+        ports:
+            - 80:80
+            - 443:443
```
matmair, line 116: I think it would be beneficial to be able to set the web port via environment file in case another server is used to proxy for inventree

30350n:
> Can we actually also use caddy without ssl behind another reverse proxy that manages a bunch of other services too?

Yes, you might have to set `auto_https off` though.

30350n:
Why are requests to `/api/*` being handled separately in the Caddyfile btw?

SchrodingersGat:
> Why are requests to `/api/*` being handled separately in the Caddyfile btw?

That was a holdover from some earlier testing - has been fixed now, thanks

docker/docker-compose.yml
SchrodingersGat: Done

SchrodingersGat:
@matmair currently `uv` breaks the docker setup process, due to some issue with trying to re-build and re-install packages from `requirements.txt` during `invoke install`. Note that these packges are already available in the base python directory, from when the image was built.

For now, I am changing the logic for `invoke install` / `invoke update` / `invoke plugins` to *not* use `uv` by default - it can however still be used with the optional `--uv` flag.

I'd like to work out why in a future PR, but I feel that perhaps this is a `uv` issue as it is still very new...

wolflu05:
and from where do we get the https certs? Letsencrypt? Don't we need to provide an email and accept the TOS somewhere?

30350n:
> and from where do we get the https certs? Letsencrypt? Don't we need to provide an email and accept the TOS somewhere?

Caddy seems to be using Letsencrypt https://caddyserver.com/docs/automatic-https#overview

SchrodingersGat:
> and from where do we get the https certs? Letsencrypt? Don't we need to provide an email and accept the TOS somewhere?

It *just works*. I setup a new demo server (`https://demo2.inventree.org`) (down now) and it auto-certified with letsencrypt.

The Caddyfile can be adjusted by users as required, though

PR #6556, opened 2024-02-22
Move get_scheduled_tasks call into register_tasks function

30350n:
The documentation of the `get_scheduled_tasks`:
https://github.com/inventree/InvenTree/blob/5dbd3030d1e983a63812dfe32f6951d977496fbc/InvenTree/plugin/base/integration/ScheduleMixin.py#L105-L109 function in the `ScheduleMixin` states that you can override it, if you want the scheduled tasks to be "dynamic" i.e. influenced by plugin settings.

Doing this breaks settings initialization pretty badly though, because the `get_scheduled_tasks` function gets called in `__init__.py`. That means any `get_setting` calls in it, will be resolved **before** any Mixins (most importantly the `SettingsMixin`) are actually loaded. The setting will still be created, but it won't be initialized correctly, thus it's `default` value won't be set which can then result in errors in the `ScheduleMixin` code which essentially completely lock up the server.

Moving the relevant code to the `register_tasks` function resolves this.

### DIFF
--- a/InvenTree/plugin/base/integration/ScheduleMixin.py
+++ b/InvenTree/plugin/base/integration/ScheduleMixin.py
@@ -50,8 +50,6 @@ class ScheduleMixin:
     def __init__(self):
         """Register mixin."""
         super().__init__()
-        self.scheduled_tasks = self.get_scheduled_tasks()
-        self.validate_scheduled_tasks()
 
         self.add_mixin('schedule', 'has_scheduled_tasks', __class__)
 
@@ -156,6 +154,9 @@ class ScheduleMixin:
 
     def register_tasks(self):
         """Register the tasks with the database."""
+        self.scheduled_tasks = self.get_scheduled_tasks()
+        self.validate_scheduled_tasks()
+
         try:
             from django_q.models import Schedule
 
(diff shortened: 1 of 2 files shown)


30350n:
Also sorry for the rambling, but I feel like the [`BaseInvenTreeSetting`](https://github.com/inventree/InvenTree/blob/5dbd3030d1e983a63812dfe32f6951d977496fbc/InvenTree/common/models.py#L183) class could use some refactoring. The whole passing-around-`**kwargs`-everywhere thing makes it some of the most opaque spaghetti code I've ever seen, took me quite a few hours to trace the bug through that mess.

matmair:
If you find an easier way to handle all the different setting types ( that can also be defined by plugins so changing the python API is not great) I am happy to review a PR. I think kwargs is one of the better ways to handle this as we have to support an arbitrary amount of filters in the base class.

SchrodingersGat:
@30350n the code is admittedly very complex. As @matmair says, if you have a cleaner way of implementing this, please submit a PR :)

SchrodingersGat (approved):

30350n:
(Tests still need fixing, do not merge yet please.)

30350n:
> If you find an easier way to handle all the different setting types ( that can also be defined by plugins so changing the python API is not great) I am happy to review a PR. I think kwargs is one of the better ways to handle this as we have to support an arbitrary amount of filters in the base class.

Yeah I'll try to look into that.

Will adjust the tests.

SchrodingersGat:
Thanks for the contribution @30350n

PR #6562, opened 2024-02-23
Refactor docker CI for speed increases

### LINKED ISSUE #6344: Slow docker build times on master
## Problem

We are currently occuring around 2.5 hours of build times on docker. This makes integrated CI/CD workflows like canary deployments very lon running - not ideal for fast paced development

## Possible solutions

I played around with this [on my fork](https://github.com/matmair/InvenTree/tree/docker-tests) and here are some possible solutions:
1. Split the build and push action per plattform (amd64, arm64) (via a matrix build) -> Faster availability of the images that are already done
2. Create a custom Pypi index with all the missing wheels (like sentry for example) -> Wheel builds are one of the mayor hold ups
3. Switch to python 3.11 or 3.12 for the image (breaking!) -> better availabilty of new wheels, performance improvements
4. Use Arm-based builders (either GitHubs MacOS machines, 3rd party machines or self-hosted runners on our sponsor DigitalOcean)

matmair:
This PR targets to make Docker CI times faster by:
- switching to Python11 on the image
- using matrix builds (failing fast)
- switching to uv inside of the container

Speed improves especially in the final image build step from [Old](https://github.com/inventree/InvenTree/actions/runs/8013835461/job/21891491969#step:15:1) 2h12m to [New](https://github.com/matmair/InvenTree/actions/runs/8013053275/job/21889468972#step:11:1) 2h5m.

We would see significantly better results if we were using arm64 build systems. There are macOS M1 runners coming online from different vendors but we would exhaust their free tiers soon (GitHub) or the offerings are from very new companies.

Closes #6344


matmair:
OK GitHub support M1 for free (to us) now: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

SchrodingersGat:
@matmair nice work. Happy with this to merge?

matmair:
I would wait for a review by @wolflu05 but should be ok for merge; Building on GitHubs macos-14 runners is broken right now so that will be a separate PR

Dockerfile
```diff
@@ -46,6 +46,8 @@ ENV INVENTREE_BACKGROUND_WORKERS="4"
 ENV INVENTREE_WEB_ADDR=0.0.0.0
 ENV INVENTREE_WEB_PORT=8000
 
+ENV VIRTUAL_ENV=/usr/local
```
wolflu05, line 49: Does this mean we now use a venv again in the docker image?

Dockerfile
```diff
@@ -90,7 +92,7 @@ FROM inventree_base as prebuild
 
 ENV PATH=/root/.local/bin:$PATH
 RUN ./install_build_packages.sh --no-cache --virtual .build-deps && \
-    pip install --user -r base_requirements.txt -r requirements.txt --no-cache-dir && \
+    pip install --user uv --no-cache-dir && uv pip install -r base_requirements.txt -r requirements.txt --no-cache && \
```
wolflu05, line 95: Is there another user now in the docker image called uv?

Dockerfile
matmair: No, this just sets the venv variable to the python path

Dockerfile
matmair: No - see the cli docs for pip https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-user

wolflu05: Oh you're right, I missed that this is already the next argument and not a value for the user argument

Dockerfile
wolflu05: Ok

matmair:
@SchrodingersGat this should be ready now!

SchrodingersGat:
@matmair this has unfortunately broken our docker tags - see here - https://hub.docker.com/r/inventree/inventree/tags

Three days ago when this was merged, our tags switched from "latest" to "master".

With the imminent release, I am concerned that we won't get the "stable" and "0.14.0" tags working properly, so I will make a PR to revert these changes. We can look at it again after the "0.14.0" release.

Edit: A full revert is not necessary - see https://github.com/inventree/InvenTree/pull/6606

matmair:
@SchrodingersGat no idea why this is a problem, works great on my fork. Maybe we once pushed a wrong tag?

SchrodingersGat:
Fixed now, is pushing against `latest` tag:

image

However, it is only pushing `linux/amd64` now. Previously it was pushing `linux/arm64` too:

image

Any ideas?

SchrodingersGat:
Ok, and now only the "arm64" tag is available:

image

I think that this "matrix" approach means that there's a race condition to see *which* image tag gets uploaded. Rather than both being uploaded as per before.

This has broken the demo server now, as the "amd64" image is not available.

@matmair is there any way to remove the "matrix" approach here, and upload *both* tags? Otherwise, we will have to probably revert this. Action needed soon as we need to get the `0.14.0` release out.

PR #6569, opened 2024-02-25
Fix postgresql isolation level

30350n:
So after fixing my deployment again, I was getting a whole bunch of `django.db.utils.OperationalError: could not serialize access due to concurrent update` errors which also made the interface feel noticeably sluggish (things would straight up not load ~50% of the time).

Looking at the existing issues regarding that error, this seems to be caused by the postgresql isolation level setting being set to `ISOLATION_LEVEL_SERIALIZABLE`, which seems to have been messed up during the switch to Django 4 in #6173.

I tested this patch on my production server and by clicking through random pages furiously for 2 minutes and didn't get any further errors (before there were sometimes multiple when visiting a single page).

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -571,6 +571,11 @@ db_options = db_config.get('OPTIONS', db_config.get('options', {}))
 
 # Specific options for postgres backend
 if 'postgres' in db_engine:  # pragma: no cover
+    from psycopg2.extensions import (
+        ISOLATION_LEVEL_READ_COMMITTED,
+        ISOLATION_LEVEL_SERIALIZABLE,
+    )
+
     # Connection timeout
     if 'connect_timeout' not in db_options:
         # The DB server is in the same data center, it should not take very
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@30350n thanks for giving attention to this :)

SchrodingersGat:
@30350n ok, something has broken here...

https://github.com/inventree/InvenTree/actions/runs/8041284048/job/21960366000

SchrodingersGat:
Ok, we recently moved away from `psycopg2` module. It is now `psycopg` that needs to be imported. Not sure how the initial tests passed here

30350n:
Hmm, I checked and it wasn't part of the dependencies before #6173 either.

30350n:
And the [deployment docs](https://docs.inventree.org/en/latest/start/install/#install-python-bindings) mention that you should install it, so I figured that'd be the way to go.

30350n:
Also seems like Django 4 [**requires**](https://docs.djangoproject.com/en/4.0/ref/databases/#postgresql-notes) `psycopg2`.

matmair:
@30350n they call out that they are switching away in the release notes https://docs.djangoproject.com/en/5.0/releases/4.2/#what-s-new-in-django-4-2 so depending on v2 instead of v3 will probably be a problem in the future

matmair:
we are on 4.2 so these are the supported/recommended engines https://docs.djangoproject.com/en/4.2/ref/databases/#postgresql-notes

30350n:
Ah yeah, I was looking at the docs for 4.0 I see, my bad. So I'll make another PR to switch to `psycopg` that also fixes the docs?

SchrodingersGat:
Thanks for addressing this so speedily :)

30350n:
No Problem ^^

PR #6571, opened 2024-02-25
Create tests in seperate folder & cleanup before image build

matmair:
We are currently shipping our docker images with files left over from testing in the CI flow.

This PR places all testfiles in a separate folder and deletes them after the CI tests so that they are not shipped with the image.


InvenTree/_testfolder/.gitignore
wolflu05, line 1: Can't we ignore the complete _testfolder in the global .gitignore?

wolflu05 (approved):
Good idea to put temporary test data into a separate folder. I haven't even noticed to files in the docker image, but you're right, they shouldn't be there.

InvenTree/_testfolder/.gitignore
matmair: I did this to make it easier to find files that should no be in the directory, we can also globally ignore - not really married to this change

wolflu05: Good point too.

SchrodingersGat:
@matmair LGTM, nice idea thanks

matmair:
@SchrodingersGat currently looking into the image more so I found a few things we should not/do not need to ship

PR #6573, opened 2024-02-25
Switch tests, docs, database settings from psycopg2 to psycopg

30350n:
Follow up to #6569.


SchrodingersGat:
LGTM @30350n - thanks for the quick follow-up!

PR #6579, opened 2024-02-26
Bump frontend dependencies

matmair:
This PR:
- migrates to yarn 4.1
- updates all dependencies (keeps mantine below 7 for now)


src/frontend/.yarn/releases/yarn-4.1.0.cjs
wolflu05, line 1: what is the use of this file?

.pre-commit-config.yaml
```diff
@@ -5,7 +5,8 @@ exclude: |
         InvenTree/InvenTree/static/.*|
         InvenTree/locale/.*|
         src/frontend/src/locales/.*|
-        .*/migrations/.*
+        .*/migrations/.* |
+        src/frontend/.yarn/.*
```
wolflu05, line 9: can we also ignore the `yarn.lock` file? I recently had the following error:
```
codespell................................................................Failed
- hook id: codespell
- exit code: 65

src/frontend/yarn.lock:2152: OT ==> TO, OF, OR, NOT
```

src/frontend/package.json
```diff
@@ -27,41 +27,43 @@
...
+        "@vitejs/plugin-react": "^4.2.1",
         "babel-plugin-macros": "^3.1.0",
-        "typescript": "^5.2.2",
-        "vite": "^4.5.2",
+        "typescript": "^5.3.3",
+        "vite": "^5.1.4",
         "vite-plugin-babel-macros": "^1.0.6"
-    }
+    },
+    "packageManager": "yarn@4.1.0"
```
wolflu05, line 68: Yarn@v4 is very new to me, but I read a bit about it now. Seems like we were accidentally using yarn@v1 the whole time. And if we upgrade we need to also add how to enable core pack to the docs and test if the docker container build still works. And secondly as far as I read, core pack is currently experimental.

.github/workflows/qc_checks.yaml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   python_version: 3.9
-  node_version: 16
+  node_version: 18
```
wolflu05, line 13: is there a reason to not directly upgrade to node@v20 which is the current LTS?

src/frontend/.yarn/releases/yarn-4.1.0.cjs
matmair: this ensures runs are reproducible; see https://yarnpkg.com/getting-started/qa

src/frontend/package.json
matmair: I do not think I have enabled corepack (i just ran yarn set version) but validating it runs every without problems is certainly something that has to be done before merge.

.pre-commit-config.yaml
matmair: Probably

src/frontend/package.json
wolflu05: When I tried it out on macOS i got a message, that I first have to run `corepack enable`. I then removed the version constrain and the error disappeared.

SchrodingersGat:
@matmair are you returning to this one any time soon?

matmair:
I was planning on waiting for feedback on the other refactoring PRs before finishing it but I can bump it if necessary

src/frontend/package.json
matmair: I have switched back to v1 now

matmair:
I have updated and cleaned this up; the switch to yarn4 is removed for now

PR #6478, opened 2024-02-13
Move media operations to storages backend

matmair:
This PR is part of https://github.com/inventree/InvenTree/issues/5703. Changes all media  (label, attachments, reports) to use Django's storage API.
The following files are not covered: config.yaml, plugins.txt, secret.txt, locale_stats.json as they are used in management commands / during startup and possible solutions felt unintuitive/made config stuff much more complicated.

This allows the usage of alternative storage backends like S3. This PR does **not** include the dependencies, config, docs, etc. for that. This PR is an internal refactor in preparation for the new File model.

Part of adressing https://github.com/inventree/InvenTree/issues/2585.

Issue #5841, opened 2023-11-02
Refactor tables and API for reports and labels


SchrodingersGat:
- Move all reports into single table (add "report type" column)
- Move all labels into single table (add "label type" column)
- Consolidate API endpoints for reports
- Consolidate API endpoints for labels
- Ensure report context is determined based on report type

matmair:
Supporting generic model relations would be nice, that way plugins could use reports easily.

wolflu05:
Still relevant

matmair:
Related https://github.com/inventree/InvenTree/issues/5884 https://github.com/inventree/InvenTree/issues/2777

matmair:
Maybe this could be structured as a generic similar to states - consistent code docstrings, fully covered with tests within the module. I would also remove the backend distinction between labels and reports and move everything into one table to reduce duplications.

matmair:
Could also be https://github.com/inventree/InvenTree/issues/6417

SchrodingersGat:
Regarding comments on other threads: I would like to see reports and labels kept separate. They are functionally different enough to warrant separate models and handling 

- Labels can be sent to specific external printers 
- Labels use width and height parameters
- Reports specify page size i.e. A4 / Letter
- Labels can be grouped together when printing (e.g. array onto a single sheet)

Happy to look at merging existing label tables together, as with merging existing report tables

matmair:
@SchrodingersGat do you think this should also apply to the code/database level? We have a lot of duplication in this area currently. Nearly half of our Python code-duplication is between the label and report app and the differences in the API and database make it more difficult than it needs to be. All the things you describe could also be flags on the merged definition - that would also make it possible to use the templating functions more flexibly and document contexts, variables etc dynamically. The amount of support we have around this area is significant.

wolflu05:
I would be for having all in one generic table and adding flags. Having different apis that have different interfaces leads to a lot of codeduplication and messy code on the frontend. 

While implementing the template editor in the frontend I came up with a collection of things that we should consider now if we refactor that.

1. I think we should call them from now on templates that exist in different types (label, report), each having some different variants (e.g. labels: stock item, location, ... and report: bom, location, ...)
1. template attributes like width, height for labels and page size, orientation for reports (also at frontend level, how do we want to expose them via the api?)
1. it would be nice to have a generic interface that can later be extended by plugins to add custom types/variants
1. how do we handle template assets
[...]

PR #6582, opened 2024-02-27
Refactor label/report template copying

matmair:
This PR is a first step towards https://github.com/inventree/InvenTree/issues/5841 and will help with #6478.
It factors out the `ready()` function of the `report` and `label` apps into a generic function that can also be used by plugins to ship their own default templating content.
This does not change the functionality around templating and is only a refactor to get functionality into one place.


SchrodingersGat:
@matmair nice work, this will clean things up nicely for our future work

wolflu05:
@matmair Does this also fixes the bug that if I change an InvenTree provided template, it gets crated again and I end up with duplicate templates?

matmair:
@wolflu05 no, it is a pure internal refactor. The behaviour you describe is a more nuanced problem and will have to be discussed as the current way templates are handled (file based) has some problems when interacting with an API.

wolflu05:
Ok. Yeah that's right.

PR #6593, opened 2024-02-27
Changelog in repo + autorelease

matmair:
Placeholder PR; will sketch further out soon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,72 @@
+# Changelog
+
+All notable changes to this project will be documented in this file (starting with 0.15.0).
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). InvenTree is currently considered to be in a pre-1.0 state, bumps to the minor version number signify new features. The [roadmap to 1.0](https://github.com/inventree/InvenTree/milestone/17) is public.
+
+## [Unreleased]
+
+### Added
+
+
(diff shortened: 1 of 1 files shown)


LavissaWoW:
I like it! I'm a big fan on user-friendly update notes. You don't need to tell everyone that you changed the return of fn(x) from int to str. Let the users kn ow what they're going to notice. At work, we use the term "politically correct" writing, and I think this fits that bill perfectly!

SchrodingersGat:
@matmair sorry did not get to look at this yet, but let's make it a focus for `0.16.0`

matmair:
@SchrodingersGat what is your general position on keeping a dedicated CHANGELOG file? Should this target 0.16.0?

SchrodingersGat:
@matmair my main concern is the extra effort it would take to maintain. Could we limit it to major features / fixes / breaking changes?

matmair:
I think everything that is breaking and any major new feature (everything we would put into the release notes/blog) should be included

SchrodingersGat:
@matmair this is the last outstanding PR for 0.16.x - is it in a state ready for review / merge?

SchrodingersGat:
Can you explain the workflow that would cause `.github/workflows/do_release.yaml` to be run - and what the effects would be?

matmair:
Let's move this to 0.17.x

matmair:
.github/workflows/do_release.yaml would be run by manually dispatching it. You could select if you wanted to release from master or stable and the action would then gather the auto-generated changelog, add the manual changelog above and create a release. After that the normal release things would start

PR #6614, opened 2024-03-01
docker: push both multiple image tags

SchrodingersGat:
Fix for currently broken docker tags

Ref: https://github.com/inventree/InvenTree/pull/6562#issuecomment-1974010882

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -58,15 +58,7 @@ jobs:
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       python_version: "3.11"
-    strategy:
-      matrix:
-        platform: ["linux/amd64", "linux/arm64"]
-        include:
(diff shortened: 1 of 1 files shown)


wolflu05 (commented):
I think you now need to disable one of the matrix entries so that the docker image is only built and uploaded once.

SchrodingersGat:
> I think you now need to disable one of the matrix entries so that the docker image is only built and uploaded once.

Yes, the matrix strategy does not seem appropriate here. Waiting for @matmair to comment on this

wolflu05:
I think at the end the matrix strategy is not a bad choice (especially when we can build on m1 arm runners), we just would need a way to combine both builds and upload them at once under one tag.

SchrodingersGat:
> I think at the end the matrix strategy is not a bad choice (especially when we can build on m1 arm runners), we just would need a way to combine both builds and upload them at once under one tag.

That sounds like a good idea for the next release

Issue #6615, opened 2024-03-01
Devcontainer volume permissions issue


LavissaWoW:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

The newly updated devcontainer setup mounts the db as user `root`, and the server as user `vscode`.

The Docker-compose file is structured such that the server has the db as a dependency.
The Postgres DB mounts in `../dev/database/`
The Server sets its data dir to be `../dev/`

Because the Server depends on the db, postgres gets installed first.
Running as root, it creates the `../dev/` dir as `root:root 755`.
This leaves the server unable to write to it, leading to the `postCreateCommand.sh` script to fail, no venv setup, and no InvenTree install being performed.

Pre-creating the `dev/` folder before building the containers will solve this issue. But that's a manual step that you need to "know" about.

### Steps to Reproduce

1. Delete `dev/` _including_ the directory itself
2. Delete the previously used containers (Not sure if needed, but reduces chances of errors)
3. Open and build the devcontainer

This error message appears: 
```
[7925 ms] Start: Run in container: /bin/sh -c .devcontainer/postCreateCommand.sh
Error: [Errno 13] Permission denied: '/home/inventree/dev/venv'
.devcontainer/postCreateCommand.sh: line 10: /home/inventree/dev/venv/bin/activate: No such file or directory
```
Then this:
```
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/usr/bin/pip'
Consider using the `--user` option or check the permissions.
```

### Expected behaviour

I haven't found any way to control what permissions are set when creating the host folder, and it appears like that's something Docker doesn't do. 

One (albeit undesirabla) fix is to change `  "containerUser"` in `devcontainer.json` to `root`. Then the Server will mount as root too. Along with all terminals run in VSCode...

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

0.14.0-dev

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

SchrodingersGat:
Another option would be to not store the database data locally at all - just in a docker volume?

SchrodingersGat:
It looks like this is potentially an issue with the production docker image as well, but does not exhibit in the same way when running outside of the devcontainer context

PR #6616, opened 2024-03-01
devcontainer volume permission fix

SchrodingersGat:
- Do not bind externally
- Fixes https://github.com/inventree/InvenTree/issues/6615

@LavissaWoW thoughts on this one?

I have tested as per your steps, and the folder permissions look correct now:

image

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     expose:
       - 5432/tcp
     volumes:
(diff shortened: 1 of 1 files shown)


LavissaWoW:
@SchrodingersGat Can confirm.
The `dev/` folder is now created with UID:GID 1000:1000.
Currently, its running migrations on a fresh devcontainer.

SchrodingersGat:
Ok, thanks for confirmation :)

PR #6660, opened 2024-03-07
Bugfix/tasks function signatures

oscgonfer:
Small PR to fix function signatures error in python 3.10. Argument types are not allowed otherwise.


oscgonfer:
Apologies, this branch had other local changes. Will create the right one. Apologies!

Issue #5948, opened 2023-11-20
PUI CORS issues only with the default part image 


wolflu05:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

If I open a part with an black image in PUI I have the following errors in console:
![image](https://github.com/inventree/InvenTree/assets/76838159/bcb7bdb1-413e-42e8-b392-3346b7b78253)

### Steps to Reproduce

1. Run PUI on another port than the dev server in dev mode
2. open the parts list table which renders parts without a part image
3. See errors in browser console

### Expected behaviour

no errors, `/static/img/blank_image.png` should load as the other images.

### Deployment Method

- [ ] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.13.0 dev
Django Version: 3.2.23
Commit Hash: 53bd75e
Commit Date: 2023-11-16
Commit Branch: pui-improvements
Database: sqlite3
Debug-Mode: True
Deployed using Docker: False
Platform: Linux-5.15.49-linuxkit-pr-x86_64-with-glibc2.31
Installer: GIT

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

SchrodingersGat:
I believe this is related to how Django handles static files in particular, with the "development" server.

Ref: https://stackoverflow.com/questions/55751033/django-cors-on-static-asset

wolflu05:
Should we then use this `--nostatic` method for development by default and setup the static handlers ourselves?

SchrodingersGat:
@wolflu05 this *may* have been fixed now with latest CORS fixes - can you please confirm when you have a chance?

wolflu05:
I still have the error, do I need to do something specific or just pull the newest code? (I tested with d502d933801c62ce8ecd522c8d874a9d7985db6f)

wolflu05:
Tried now, still the same behavior. 
So to clarify, what I did:
1. run inventree server on port 8000
2. run pui via `npm run dev`
3. connect to `http://localhost:5137`, add a new server pointing to `http://localhost:8000`
4. open the parts page where a lot of parts that have no image are shown
5. See the errors in the dev console

SchrodingersGat:
I have just retried locally, with the server pointing at both `localhost:5173` and `localhost:8000` - both seem to work on my end. Do the tables load? Just the images are not coming through?

[...]

PR #6662, opened 2024-03-07
Use whitenoise for static file handling

SchrodingersGat:
- Fixes https://github.com/inventree/InvenTree/issues/5948
- Uses [whitenoise](https://whitenoise.readthedocs.io/en/latest/django.html) as a replacement for django's built-in static file server
- Still really only useful for "development" - as whitenoise is not a recommended solution for serving media files

### Example

- Running local dev server at `http://localhost:8000`
- vite server running at `http://localhost:5173`
- allows CORS requests for static files

image

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -20,7 +20,6 @@ env:
   INVENTREE_MEDIA_ROOT: ../test_inventree_media
   INVENTREE_STATIC_ROOT: ../test_inventree_static
   INVENTREE_BACKUP_DIR: ../test_inventree_backup
-  INVENTREE_SITE_URL: http://localhost:8000
 
 jobs:
   paths-filter:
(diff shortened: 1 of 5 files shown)


SchrodingersGat:
@LavissaWoW @wolflu05 @matmair after much reading, this seems like the only reliably way around this issue. Even though it is a very small target for the issue (static file CORS in dev mode, from the vite server), this seems like a reasonable solution.

I looked into whether this would be a good solution for serving static / media files in general - however it does not handle media files. So, stuck with Caddy (or similar) solution for now

LavissaWoW:
@SchrodingersGat Probably unrelated, but should we fix the `favicon` 404 too?

SchrodingersGat:
@LavissaWoW I've forgotten the favicon issue, can you remind me?

LavissaWoW:
@SchrodingersGat I'm not able to reliably reproduce the console error message.
But any tab showing PUI lacks the favicon on its tab icon. And I regularly get a 404 error in the console for `/favicon.ico`

SchrodingersGat:
Ok if you see it again please raise an issue :)

Issue #6668, opened 2024-03-10
Docker INVENTREE_CUSTOMIZE environment variable not working


E-Hain:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

In the [documentation](https://docs.inventree.org/en/0.14.1/start/config/#customization-options) it says that you can use INVENTREE_CUSTOMIZE to edit the customize.login_message and customize.navbar_message. However when adding INVENTREE_CUSTOMIZE into a docker-compose .env file it causes an error.  You can get around this by editing the config.yaml directly but this should still be addressed.

### Steps to Reproduce

1. Use docker
2. Add INVENTREE_CUSTOMIZE as an environment variable in .env file
3. Launch Inventree and find "internal server error"

### Expected behaviour

I expected it to use the variable value to edit the Login message and/or the navbar message

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.14.1
Django Version: 4.2.10
Commit Hash: df0bb34
Commit Date: Mar 05 2024

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.5.0-25-generic-x86_64-with
Installer: DOC

Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}, {'name': 'InvenTreeCurrencyExchange', 'slug': 'inventreecurrencyexchange', 'version': '1.0.0'}, {'name': 'InvenTreeLabel', 'slug': 'inventreelabel', 'version': '1.0.0'}, {'name': 'InvenTreeLabelMachine', 'slug': 'inventreelabelmachine', 'version': '1.0.0'}, {'name': 'InvenTreeLabelSheet', 'slug': 'inventreelabelsheet', 'version': '1.0.0'}, {'name': 'DigiKeyPlugin', 'slug': 'digikeyplugin', 'version': '1.0.0'}, {'name': 'LCSCPlugin', 'slug': 'lcscplugin', 'version': '1.0.0'}, {'name': 'MouserPlugin', 'slug': 'mouserplugin', 'version': '1.0.0'}, {'name': 'TMEPlugin', 'slug': 'tmeplugin', 'version': '1.0.0'}]

[...]

PR #6676, opened 2024-03-12
Fix typecasting in settings.py

SchrodingersGat:
- Properly cast dict values
- Fixes https://github.com/inventree/InvenTree/issues/6668

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -296,7 +296,10 @@ if LDAP_AUTH:
 
     # get global options from dict and use ldap.OPT_* as keys and values
     global_options_dict = get_setting(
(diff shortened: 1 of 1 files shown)


InvenTree/InvenTree/settings.py
```diff
@@ -723,7 +732,10 @@
         logger.info('OpenTelemetry tracing enabled')
 
         _t_resources = get_setting(
-            'INVENTREE_TRACING_RESOURCES', 'tracing.resources', {}, dict
+            'INVENTREE_TRACING_RESOURCES',
+            'tracing.resources',
+            default_value={},
```
matmair, line 737: @SchrodingersGat you changed this to none in other places but {} here - is there a reason for that?

SchrodingersGat: @matmair no reason, I just missed it. Will issue a small patch

PR #6694, opened 2024-03-13
Fix undefined tabValue (#6470)

XanderLuciano:
#6470 Points out that the new UI nav bar does not correctly set the `;tabValue` on initial page load.

This is because `useParams` will return the params from the route as defined in the router. Our router does not define `:tabValue/*` for all routes so the object returned from `useParams` does not have a `tabValue` property and is therefore undefined on initial page load.

This PR uses `useMatch` to create a new URL match for `:tabName/*` and sets the `tabValue` variable correctly.

The result, is a navbar that always has the correct tab shown as active.

### DIFF
--- a/src/frontend/src/components/nav/Header.tsx
+++ b/src/frontend/src/components/nav/Header.tsx
@@ -3,7 +3,7 @@ import { useDisclosure } from '@mantine/hooks';
 import { IconBell, IconSearch } from '@tabler/icons-react';
 import { useQuery } from '@tanstack/react-query';
 import { useState } from 'react';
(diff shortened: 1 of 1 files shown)


XanderLuciano:
Also fixes #6471

SchrodingersGat:
@XanderLuciano nice work, seems to work pretty well :)

SchrodingersGat (approved):

XanderLuciano:
Thank you! I'm Looking forward to submitting additional contributions towards the new react UI 🥳

PR #6699, opened 2024-03-13
Make modal footer fixed (#6696)

### LINKED ISSUE #6696: [PUI] Make "footer" (buttons etc) stick to the bottom of the modal (no scroll)
> Make "footer" (buttons etc) stick to the bottom of the modal (no scroll)

Big +1 on this, the ability to click "submit" without having to scroll to the bottom is a killer advantage of the old UI vs. the new UI currently

Also, I know the old UI also handles "enter" as a modal form submission. This behavior also should be kept / added to the new UI.

This is a task I'm interested in submitting a PR for.

EDIT: Related Issue: https://github.com/orgs/mantinedev/discussions/1141

_Originally posted by @XanderLuciano in https://github.com/inventree/InvenTree/issues/5354#issuecomment-1993793221_

XanderLuciano:
Quick fix for #6696 

Does not work great on low vertical resolution screens, need to probably calculate max view size in pixels for device and use that to fix div height. I was having trouble making the outer modal container fill to viewport height AND scroll as expected.

Another option is to add a ton of padding / margin to the bottom of the form and make the bottom footer fixed.

There's a few ways to approach this in a more "proper" way. But wanted to get an early and semi-functional scroll in case someone else wants to suggest some improvements that build on this.

Otherwise, this is a short term fix to make the bottom bar a little nicer to use again.

Closes #6696


XanderLuciano:
image

SchrodingersGat:
@XanderLuciano looks great! I'm happy to merge this as-is, unless you are going to work on the "improvements" right away?

wolflu05:
Thanks for that fix, that is looking really promising, I remember I was trying this too, but did not succeed.

But when using the APIForm without a height restricted modal, e.g. in the Admin center in the user modal we have a scroll bar too at the wrong place. Instead if it overflows we should be able scroll in the drawer, otherwise we would end up with two scroll bars if e.g. the groups list gets to long.

| Before | Now |
|--------|--------|
| image | image |

matmair (approved):
LGTM; thank you @XanderLuciano

SchrodingersGat:
@XanderLuciano can you please address the comments by @wolflu05 here? :)

SchrodingersGat:
One thing I have noticed here (not introduced by your PR, but just with forms in general). 

When loading the form initially, the form displays with a big empty gap between the header and the footer. It would be nice to display a `<Skeleton />` or `<Loader />` in the middle of the form whilst the initial content is fetching? At the very least, get rid of the ugly "gap"

XanderLuciano:
Originally was going to say I was happy enough with this PR as-is, but the issues pointed out by @wolflu05 are regressions I would like to avoid. 

I think I can come up with a better solution that improves the behavior in more places, and solves the empty modal appearance on initial load. 

Good comments all around detailing the small issues. I’d like to refine this a little further before we merge this. :)

XanderLuciano:
@wolflu05 @SchrodingersGat I've (mostly) fixed the issues with the original commit, and also improved the behavior in a few other places as well at the same time.

The modal forms will now shrink if the form is short, and I made the user groups display partially hidden into a spoiler so it should avoid the double scrollbar initially (some screen resolutions may still have double scroll bars).

I'm happy with how this behaves now though :) What do you guys think?

SchrodingersGat:
@XanderLuciano looking pretty nice :) I would still like to see the "loading" overlay fixed, so there is not a large gap in the middle of the modal form initially

XanderLuciano:
image

image

@SchrodingersGat I'm unable to reproduce the issue mentioned about the modal appearing as a large empty block upon first load (I simulated by throttling the network connection in dev tools). The modal is small until the API request for the fields list completes.

Unless I am misunderstanding and the issue is the desire is to have a loading overlay appear while the API request for fields is in progress?

SchrodingersGat:
@XanderLuciano sorry I realise that the way I described it was misleading. 

> The modal is small until the API request for the fields list completes.

Correct, this is a much better description than mine!

> Unless I am misunderstanding and the issue is the desire is to have a loading overlay appear while the API request for fields is in progress?

This is what I was looking for. Do you think you can tack it on to this PR? No worries if not, we can make it a separate thing

XanderLuciano:
Shouldn't be too much trouble, and at this point I'm quite familiar with this single component so let's include it in the scope of this PR 😄 I'll mess around with some ideas tonight!

src/frontend/src/tables/settings/UserTable.tsx
```diff
@@ -12,7 +19,10 @@ import {
   DetailDrawerLink
 } from '../../components/nav/DetailDrawer';
 import { ApiEndpoints } from '../../enums/ApiEndpoints';
-import { openCreateApiForm, openDeleteApiForm } from '../../functions/forms';
+import {
+  useCreateApiFormModal,
+  useDeleteApiFormModal
+} from '../../hooks/UseForm';
```
XanderLuciano, line 25: Converted from old `openCreateApiForm` and `openDeleteApiForm` to newer `useCreateApiFormModal` and `useDeleteApiFormModal`

I believe this was leftover from the improvements done in https://github.com/inventree/InvenTree/pull/5897

This fixes an issue with the create new user button and the create new group button having different behavior on the admin page.

src/frontend/src/components/forms/ApiForm.tsx
```diff
@@ -139,6 +139,12 @@ export function OptionsApiForm({
   const formProps: ApiFormProps = useMemo(() => {
     const _props = { ...props };
 
+    // This forcefully overrides initial data
+    // Currently, most modals do not get pre-loaded correctly
```
matmair, line 143: @XanderLuciano what do you mean by that? Do you have a sample?

src/frontend/src/components/forms/ApiForm.tsx
```diff
@@ -222,41 +224,46 @@ export function ApiForm({ id, props }: { id: string; props: ApiFormProps }) {
...
-                res[k] = dataValue;
-              }
+      try {
+        // Await API call
+        let response = await api.get(url);
+        // Define function to process API response
+        const processFields = (fields: ApiFormFieldSet, data: NestedDict) => {
+          const res: NestedDict = {};
+
+          // TODO: replace with .map()
```
matmair, line 234: Question (non blocking): can this be addressed now?

matmair (approved):
Lgtm

src/frontend/src/components/forms/ApiForm.tsx
XanderLuciano: image
image
image

If you look at the difference in the values for the `fields` object, there is a note able difference, and some of that data is required for the ApiFormField component to correctly appear, otherwise it shows the error message in the first photo. After the API call is made to the endpoint using OPTIONS and gets all the values for each field, it renders correctly / normally.

XanderLuciano: I think it's primarily missing the `type` attribute, which would need to be added everywhere a model form is called. 

e.g. here 
image

but that I think should be scoped out to a new issue.

wolflu05: The idea behind that is that one can define which fields to show and which to hide. Only fields added through `filename: { /* ... */}` should be shown. And then one could add some customization to the defaults requested via the OPTIONS request. E.g. hook up a value state or make the disabled state react to some other condition.

Does this still work?

src/frontend/src/components/forms/ApiForm.tsx
XanderLuciano: Don't think I understand the logic of that loop well enough to refactor that properly yet

src/frontend/src/components/forms/ApiForm.tsx
XanderLuciano: @wolflu05 oh! Huh, surprisingly it does actually still work! Here is me removing the first name field on the edit user form
image
image

SchrodingersGat:
@XanderLuciano nice work the loader looks really good!

@wolflu05 are you also happy to merge this in?

src/frontend/src/components/forms/ApiForm.tsx
wolflu05: ok, got it, you only override if data is not defined. I now also tested if the reactiveness (hooking up a state and use that for the disabled state (see playground)) still works, seems so.

XanderLuciano: Yes! I re-used the idea of not rendering the entire ApiForm component if data was not available. Instead of not rending the component, I just override the fields temporarily until the data arrives. 

Now that I know the intended behavior, the method I chose actually seems quite fitting IMO.

Accidentally worked just right 😄 that's rare!

wolflu05:
Yes, LGTM too.

SchrodingersGat:
@XanderLuciano great work here! Sorry for the complications, but this has turned out very nicely indeed.

XanderLuciano:
Thank you! I appreciate the quick response and feedback along the way.

src/frontend/src/components/forms/ApiForm.tsx
SchrodingersGat: @XanderLuciano @wolflu05 after testing the new code a bit more, I have found that this change breaks loading an "edit" form. The "initialData" query often returns first, before the "OPTIONS" (which tells us which fields are available).

Commenting out this line: `_props.fields = undefined` seems to result in correct behaviour again.

We need to review this ASAP as edit forms are really broken now

SchrodingersGat:
@XanderLuciano please see my inline review comment above - this PR has introduced an unfortunate new behaviour

PR #6709, opened 2024-03-14
handle report previewing errors

### LINKED ISSUE #6705: [PUI] Report editor not rendering error message correctly
When template rendering throws an error, the error is not displayed as expected. Instead it just shows `[object Blob]`

image

@wolflu05 any thoughts on this one? I have tried to fix but no luck.

wolflu05:
When requesting the preview of a label template, the response type always is json (either with an error or with the path to a rendered pdf), but when requesting the preview of a report the axios response type is set to `blob`, because the PDF normally is directly returned as a blob. However if there are some errors, the server returns json format, but axios still parses this as a blob. This PR ensures that the error blob is first parsed as JSON and then the correct error is thrown to be handled by the TemplateEditor. Of course, this mess has to be changed along with some other things if we refactor the label/report API.

fixes #6705

### DIFF
--- a/src/frontend/src/components/editors/TemplateEditor/PdfPreview/PdfPreview.tsx
+++ b/src/frontend/src/components/editors/TemplateEditor/PdfPreview/PdfPreview.tsx
@@ -1,4 +1,4 @@
-import { Trans } from '@lingui/macro';
+import { Trans, t } from '@lingui/macro';
 import { forwardRef, useImperativeHandle, useState } from 'react';
 
 import { api } from '../../../../App';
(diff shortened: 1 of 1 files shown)


matmair:
Can you please add a todo in the section we should refactor and a note on. The refactor issue? Currently on mobile

wolflu05:
@matmair there is already one, a bit above. Basically the complete function after this comment needs to be refactored then. And if there is previewing without saving on the API available the complete `PdfPreview` component needs a refactor.
image

SchrodingersGat:
@wolflu05 LGTM!

Also I *finally* fixed the CORS issues so that the netlify preview works well with the demo server again. It was an absolute nightmare to get running after the update to django 4.2...

Ref: https://github.com/inventree/InvenTree/pull/6712

PR #6733, opened 2024-03-16
Worker thread fix

SchrodingersGat:
- Closes https://github.com/inventree/inventree-brother-plugin/issues/35
- Fixes critical bug which prevented worker thread from loading plugins


wolflu05:
Why do we skip the initialization of the registry if we are NOT in a worker thread?

SchrodingersGat:
@wolflu05 - https://github.com/inventree/InvenTree/pull/5005

wolflu05:
Sorry I don't get it. Maybe I understand something wrong, but doesn't this change mean that if we run invoke server, the plugin registry will not be initialized?

SchrodingersGat:
@wolflu05 I can confirm that the plugins *do* get loaded / installed correctly  from the following entry points:

- `invoke server`
- `invoke gunicorn`
- `invoke worker`

Issue #6715, opened 2024-03-14
[PUI] Details Field Text Length


SchrodingersGat:
Text in a "details field" should clip, but a very long value causes rendering issues:

![image](https://github.com/inventree/InvenTree/assets/10080325/e79007a5-2a66-4eb0-8289-17adf353b3df)

@wolflu05 @LavissaWoW any thoughts on this one?

wolflu05:
Maybe we should cut that, and add some kind of accordion component to show the full text on click?

SchrodingersGat:
That sounds suitable, I think

wolflu05:
Or isn't there a backend restriction? If it's not too long, we may could also simply show all by like breaking?

XanderLuciano:
I think we could wrap that data in a [`<Spoiler>`](https://mantine.dev/core/spoiler/) to clip the text, but allow expansion if desired

SchrodingersGat:
It looks like `<Spoiler>` is based on height, really what we want is to break the text into multiple lines so it does not exceed the table width. I think this can be achieved with some simple CSS?

LavissaWoW:
@SchrodingersGat I Actually tinkered with something like this just yesterday. You can define a max width, and instead of pixels, use `ch` for characters. Additionally, this can be used with the CSS `clamp(min, optimal, max)` function

wolflu05:
https://stackoverflow.com/questions/499137/how-can-i-force-a-long-string-without-any-blank-to-be-wrapped this may be useful

PR #6739, opened 2024-03-18
Fix CSS for long details field

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/6715
- Wraps field text correctly

image

### DIFF
--- a/src/frontend/src/components/details/Details.tsx
+++ b/src/frontend/src/components/details/Details.tsx
@@ -198,7 +198,13 @@ function TableStringValue(props: FieldProps) {
   }
 
   return (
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/details/Details.tsx
```diff
@@ -198,7 +198,13 @@ function TableStringValue(props: FieldProps) {
   }
 
   return (
-    <div style={{ display: 'flex', justifyContent: 'space-between' }}>
+    <div
+      style={{
+        display: 'flex',
+        justifyContent: 'space-between',
+        wordBreak: 'break-word'
```
wolflu05, line 205: I would add an alignItems: flex-start too, so that the icon, field name starts at the top of the container.

PR #6743, opened 2024-03-18
Details CSS tweaks (again)

SchrodingersGat:
Follow up to https://github.com/inventree/InvenTree/pull/6739

@wolflu05

### DIFF
--- a/src/frontend/src/components/details/Details.tsx
+++ b/src/frontend/src/components/details/Details.tsx
@@ -367,7 +367,7 @@ export function DetailsTableField({
   const FieldType: any = getFieldType(field.type);
 
   return (
-    <tr>
+    <tr style={{ verticalAlign: 'top' }}>
       <td
         style={{
           gap: '20px',
--- a/src/frontend/src/components/details/Details.tsx
+++ b/src/frontend/src/components/details/Details.tsx
@@ -202,7 +202,8 @@ function TableStringValue(props: FieldProps) {
       style={{
         display: 'flex',
         justifyContent: 'space-between',
-        wordBreak: 'break-word'
+        wordBreak: 'break-word',
+        alignItems: 'flex-start'
       }}
     >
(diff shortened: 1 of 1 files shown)


wolflu05:
Not sure how you prefer it, it was just a suggestion, but what I actually tried to explain was this, but no strong feelings here.

image

If you like that, you need `vertical-align: top` on the `tr`. Sorry, didn't saw that this is not a flex element but a table where I initially commented. But feel free to do like you think its the best.

wolflu05:
And maybe we need wordwrap in the heading too:

image

LavissaWoW:
Pics from @wolflu05 look good.
I wrote the table fields to be able to have a theoretical "unlimited" amount of flex children, but realistically, it's going to be the field, maybe a user badge, and maybe a copy button. So the calculations for width in the code probably won't cause any issues.

SchrodingersGat:
@wolflu05 I have made the fix to the vertical alignment, thanks! Where / what would you suggest I change for the page header?

wolflu05:
If the part description is too long it doesn't break but instead moves the control buttons outside of the card (the arrow should indicate the card end, you can see the shadow from the paper component there).

image

I now suggest we should just clip the word with `...` in that case so that the buttons fit there, because the description actually is there in the Part details tab (where we had this wrapping issue too, but you have fixed it in the last PR).

wolflu05:
Maybe we should generally think about clipping the text in the whole UI at some points a bit more. E.g. in tables we have the same behavior, and I thing it would make sense to clip the text, but maybe show it on hover. And I bet, we still have this at other places too.

https://github.com/inventree/InvenTree/assets/76838159/c43a6699-0d5f-46eb-9fbf-37d04e641284

SchrodingersGat:
@wolflu05 good point about table rows - we should consider that too.

I'm going to merge this as-is, we can think about tables and headers separately.

Issue #6727, opened 2024-03-16
Cannot login .../platform


SergeoLacruz:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [ ] Docker Production
- [X] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

According to 
https://github.com/wolflu05/inventree-dymo-plugin/issues/2
Login to mysite/platform is not possible. My Inventree username and password does not work. The reset password. function seems
to do nothing. 
![login](https://github.com/inventree/InvenTree/assets/10836099/f06d3865-4968-4c27-a2d0-033145ab5e1a)

Installed version: 0.15.0 dev 
Comit hash 9ff09c8 
Commit Date 2024-03-15 
commit Branch master

What am I missing?

### Steps to Reproduce

Try to login mysite/platform

### Relevant log output

```bash
16/Mar/2024 08:28:43] "GET /platform HTTP/1.1" 200 636
[16/Mar/2024 08:28:44] "GET /static/web/assets/index-19438820.js HTTP/1.1" 200 3172
[16/Mar/2024 08:28:44] "GET /static/web/assets/MobileAppView-ba0ea19c.js HTTP/1.1" 200 491
[16/Mar/2024 08:28:44] "GET /static/web/assets/DesktopAppView-747a2232.js HTTP/1.1" 200 57459
[16/Mar/2024 08:28:44] "GET /static/web/assets/vendor-c50a8390.js HTTP/1.1" 304 0
[16/Mar/2024 08:28:44] "GET /static/web/assets/BaseContext-a18d7fff.js HTTP/1.1" 304 0
[16/Mar/2024 08:28:44] "GET /static/web/assets/links-4c07aa67.js HTTP/1.1" 304 0
[16/Mar/2024 08:28:44] "GET /assets/MobileAppView-ba0ea19c.js HTTP/1.1" 301 0
[16/Mar/2024 08:28:44] "GET /assets/BaseContext-a18d7fff.js HTTP/1.1" 301 0
[16/Mar/2024 08:28:44] "GET /assets/links-4c07aa67.js HTTP/1.1" 301 0
[16/Mar/2024 08:28:44] "GET /assets/vendor-c50a8390.js HTTP/1.1" 301 0
[...]

PR #6754, opened 2024-03-19
Fix PUI build workflow

wolflu05:
As per #6727 it seems like we missed to include the compiled translations the whole time in the artifact builds. This PR adds the lingui compile step to the artifact and release build.

I suggest we could also back port this change to 0.14.x

---
**EDIT:** Tested the build produced by [this run](https://github.com/inventree/InvenTree/actions/runs/8341836937/job/22828730734?pr=6754), works now.

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -507,7 +507,7 @@ jobs:
       - name: Install dependencies
         run: cd src/frontend && yarn install
       - name: Build frontend
-        run: cd src/frontend && npm run build
+        run: cd src/frontend && npm run compile && npm run build
       - name: Zip frontend
         run: |
           cd InvenTree/web/static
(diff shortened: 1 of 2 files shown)


matmair (approved):
LGTM

matmair:
I think this should be backported too

SchrodingersGat:
@wolflu05 do the translations get included in the compiled docker image

wolflu05:
Yes, see 

https://github.com/inventree/InvenTree/blob/master/Dockerfile#L110

and if the wont be included, we would actually already have noticed with the docker image because that's used on demo and so much prod instances.

The fact that this issue actually existed for so long time shows how rare such a bare metal, from code setup is.

SchrodingersGat:
Thanks for confirming. And thanks for the fix!

PR #6764, opened 2024-03-20
[PUI] table fixes

SchrodingersGat:
Refactoring / improvements for PUI tables

- Add ability to "update" a single row without reloading entire table
- Refactor / simplify click-through of table rows

### DIFF
--- a/src/frontend/src/tables/InvenTreeTable.tsx
+++ b/src/frontend/src/tables/InvenTreeTable.tsx
@@ -21,12 +21,15 @@ import {
   DataTableSortStatus
 } from 'mantine-datatable';
 import { Fragment, useCallback, useEffect, useMemo, useState } from 'react';
+import { useNavigate } from 'react-router-dom';
 
 import { api } from '../App';
 import { ActionButton } from '../components/buttons/ActionButton';
 import { ButtonMenu } from '../components/buttons/ButtonMenu';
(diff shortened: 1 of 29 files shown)


src/frontend/src/tables/InvenTreeTable.tsx
```diff
@@ -501,6 +508,30 @@ export function InvenTreeTable<T = any>({
...
+        props.onRowClick(record, index, event);
+      } else if (tableProps.modelType && record?.pk) {
+        // If a model type is provided, navigate to the detail view for that model
+        let url = getDetailUrl(tableProps.modelType, record.pk);
+
+        // Should it be opened in a new tab?
+        if (event?.ctrlKey || event?.shiftKey || event?.buttons & 0x04) {
+          // Open in a new tab
+          // TODO: Remove /platform/ prefix later on
+          window.open('/platform' + url, '_blank', 'noreferrer');
```
wolflu05, line 525: I think we have a variable for the relative path somewhere we could use here, just import it from `main.tsx`:

https://github.com/inventree/InvenTree/blob/08c0953eb73741d1754fd9cc8169793e553a00d2/src/frontend/src/main.tsx#L82

SchrodingersGat: Oh nice, thanks!

PR #6772, opened 2024-03-20
Fix machine request pickeling and settings

wolflu05:
This PR should fix three things when using the machine printing feature in a production setup with workers:

## Cannot print a label with a machine using workers (Request pickeling issue)

⚠️  This PR is kind of breaking. Kind of, because actually the machine print interface could not have worked before in production mode with workers, so I guess its ok to just barely change that before there is some real logic that relies on this later.

<details><summary>Details</summary>

This fixes the bug discovered in https://github.com/wolflu05/inventree-dymo-plugin/issues/2#issuecomment-2010271005

Basically requests can't be picked (needed when they are handed to a worker), so I decided to remove the request from the machine print method signature. Because I really think that the request should not be needed to render a label template. For now I added a dummy method to emulate that there is a request. This can be removed once #7351 is resolved.

</details> 

## Machine state is not shared across workers and main thread

When changing the e.g. printer state to no media after trying to start a print job from within a worker. That state is not set for the machine instance in the main thread, therefore this issue does not show up to the user in the admin center. 

## Settings definition not available in worker

When defining settings with e.g. an `int` validator, these are not correctly used when getting settings values in a worker. 

<details><summary>Details</summary>

The issue here is, that the workers do not initialize the machine registry and therefore the getter function for `.machine` in the `MachineConfig`  class does not return the machine instance because the registry is empty in the workers.

</details> 

## TODO:

- [x] remove request entirely from machine printing functions and replace with dummy request for now
- [x] move class state to a shared class state mixin which uses the Redis cache as a storage
- [x] clear class state storage on server start
- [x] when deleting/adding a machine update the registry in the workers somehow
- [x] update documentation (how the state is managed, and that a Redis cache is required)
- [x] adjust dymo and cups drivers
- [x] test in prod setup

Closes wolflu05/inventree-dymo-plugin#2
Closes wolflu05/inventree-cups-plugin#10


matmair:
A possible solution would be to implement a proxy model for machines when running in a worker is detected. This could take multiple forms. If a shared cache is available that could be used to send lookup requests to server instances websockets would also be possible via django channels.

wolflu05:
Do you have some examples for this?

SchrodingersGat:
@wolflu05 I do not fully follow the problem here, but are you saying that the worker thread does not load the machine registry? Surely this causes issues with the worker not being able to "use" the machines?

Note that last year we implemented a "plugin registry checksum" to ensure that the worker and server threads keep the plugins in sync (when a new plugin is installed for the server it is updated in the worker too). Maybe something similar is needed here?

wolflu05:
>  that the worker thread does not load the machine registry

Yes, and that's actually intended. Because there should only live one instance of a machine. Otherwise this can lead to side effects. E.g. changing the state on the machine instance in the main thread doesn't affect the state in the instance of the worker thread. They need to be synced.

I guess that's what @matmair suggested with backchannels.

SchrodingersGat:
@wolflu05 how is the progress on this one?

wolflu05:
Regarding removing the request object from the printing functions (first issue) my plan was to just wait until #7074 is done now. That's why I suggested this change and actually re-thinked about why we have the request at all in the rendering functions.

And regarding the the worker issue, sorry for letting this hang around, but I have no time currently to do bigger development. To fix this, we would need some kind of communication channel between the workers and the main process. Because machine state is managed in a class that exists in the main process. And it should by design only be lived on one place. If the background worker needs to get a machine class instance, we would need to send a request to the main process and initialize the class somehow. Also for state changes, if we change the machine status in the background worker, we would need to send that to the main thread and change the state there.

Mawiguk0:
@SchrodingersGat is this plugin the only component of inventree that has Problems with the request pickeling?

I will need a lot of time to get to a point where I could contribute to this issue.

Is there a way to work around this issue, that could be adopted. Or is solving this issue just not that high priority, beside the whole Frontend refactoring going on?

wolflu05:
Depends on #7351

wolflu05:
This PR is now Ready for review now. Basically it fixes the 3 described issues in the initial comment, so that the machine printing feature can finally be used in a production setup:
1. Request objects cannot be pickeled -> I removed the request from all machine printing functions and replaced it with a dummy request for now until #7351 is solved
2. Machine state is not shared across workers and the main process -> I used the Redis cache as a communication layer. Now the state is stored in the global Redis cache rather than in the class instance
3. Setting definition not available in workers (numbers get not parsed correctly) -> Machines get loaded in workers too, but the initialize function is not called, that way the machines class instances exist in the worker containers too.

I have now also tested this in a docker production setup without any issues with inventree-dymo-plugin.

I guess the failing CI is not related to my PR, as I encountered the flaky PUI tests a couple of times now in my other PR too.

SchrodingersGat (approved):

SchrodingersGat:
@wolflu05 nice work, I think it is reasonable to require the shared cache here. And nice that we have got rid of the legacy modal form calls now :)

wolflu05:
> And nice that we have got rid of the legacy modal form calls now :)

Yep, the last usage where in the machine forms and I experienced a bug that the forms couldn't basically be used due to the legacy implementation, so I replaced them and as this was the last usage I finally dropped the complete implementation.

PR #6790, opened 2024-03-20
Python dependency bump

matmair:
Updating most python deps in one go, will close most dependabot python things automatically after merged


matmair:
@SchrodingersGat this bumps most open python deps. Missing is drf - the update there is breaking a lot of stuff, I need to look into that later

SchrodingersGat:
Thanks @matmair :)

PR #6791, opened 2024-03-21
Fix for initial form data

SchrodingersGat:
- Ref: https://github.com/inventree/InvenTree/pull/6699

@XanderLuciano @wolflu05 please review. This needs urgent assessment to fix this bug!


matmair:
Seems to work again but there is a flash of error messages in the modal before the data loads - at least on netlify

matmair (approved):

SchrodingersGat:
> Seems to work again but there is a flash of error messages in the modal before the data loads - at least on netlify

Yep, need to work that one out.

XanderLuciano:
While annoying and unwanted, it's probably better to have the error flash initially than for editing not to work.

I think a full solution would be one of two options

1. We pass 1 object with the fields we want to display and what to display with each field, and then another object holds the results from the OPTION api request. We hide behind a loading panel until that OPTION object is loaded

2. We track the loading progress of the OPTION request and only remove the loading overlay when the call succeeds.

I'm more in favor of option 1, passing an object to configure the fields we want to display separate from the OPTION request.

I will experiment tonight with better solutions and report back what I learn.

XanderLuciano:
Or, we can add an extra field to the response object in intitialData to prevent it from being overwritten if it does return first.

SchrodingersGat:
I like the idea of not displaying any of the fields until the OPTIONS request has succeeded. As you have been digging around in here recently do you have a suggestion as to where that could be coded?

matmair:
The errors flashing is very bad UX, I think we should display a loading screen till the OPTIONS call has concluded

Issue #3101, opened 2022-05-30
[FR] Import tool refactor


matmair:
**Is your feature request the result of a bug?**
#2341
#3098
#3097 
#6321

**Problem**
We have several ways to do imports - we should refactor imports out.

**Suggested solution**
Two principles:
- API first
- backend-run first

Backend:
- [x] Switch to [data-wizard](https://github.com/wq/django-data-wizard)
- [x] Refactor all import code into a seperate app
- [x] Clean up / limit exposed APIs from data-wizard
- [x] Create backend worker for django-q
- [x] Create frontend that consumes the new APIs to build up an import flow

Refactor code base
- [x] Refactor BOM importer #3097
- [x] Refactor PO importer #3098
- [x] Refactor SO importer
- [x] Refactor Part importer

Unit tests:
- [ ] E2E Tests for import flow
- [ ] E2E for all refactored importers based on static import files that are in git

**Describe alternatives you've considered**
We should fix this, not sure what release to target though.

**Do you want to develop this?**
Yes - this will be something for the whole team I think.

-----------------------------------------------------

### Outstanding Tasks

- [x] Support "default values" when importing dataset
- [ ] Allow user to navigate back to previous data steps
- [ ] Add unit test for cleanup backup task
- [x] Ensure user has correct permissions before creating a new import session
- [x] Expose import / export functions to plugins
- [x] Allow plugins to define custom export fields / processes
- [ ] Allow bulk data export action to django admin interface
- [x] Replace existing "BOM Import" tool with new importer
- [x] Replace existing "BOM export" tool with new exporter
- [x] Replace existing "part import" tool with new importer
- [x] Replace existing "purchase order import" tool with new importer
- [x] Remove `django-import-export` framework entirely
- [ ] Allow bulk "update" via import (override current records, not create new ones)
- [x] Remove all custom import UI template code (html templates, legacy)
- [x] Check that all existing "django-import-export" models are covered by new framework
- [x] Show "original row data" for each row when importing data
- [ ] Allow user to create new related model "in place"
- [x] Add "context specific" importers for particular tables, rather than just in the admin center
- [x] Improve efficiency of data import process (currently takes ~1 minute for 1,000 records)
- [x] Impove user interface for data import - in particular, react hook efficiency
- [x] Enable per-column specification of "default" values in the admin portal
- [x] Documentation for new data import process

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/inventree) so you can upvote and help fund this issue.
[...]

PR #5125, opened 2023-06-30
Data import API

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/3101

Provides an API-based data import tool using [django-data-wizard](https://github.com/wq/django-data-wizard)

### Tasks

- [ ] Run data import using background worker
- [ ] Preview imported data and allow user to confirm
- [ ] Auto-assign columns based on previous import runs
- [ ] API endpoints
- [ ] Significant unit testing
- [ ] Remove old clunky import code
- [ ] Remove django-import-export integration entirely

PR #6792, opened 2024-03-21
Data Import Wizard

SchrodingersGat:
A fresh start on the data import wizard config, which uses existing django-rest-framework serializers to import data from an external file.

- Replaces https://github.com/inventree/InvenTree/pull/5125
- Closes https://github.com/inventree/InvenTree/issues/3101

## Roadmap

- [ ] Implement data import for existing serialises using `data-wizard`
- [ ] Expose custom API endpoints to handle this process via new react interface
- [ ] Implement front-end user interface elements for data import managing
- [ ] Remove existing custom import logic (lots of custom forms e.g. for BOM data import)
- [ ] Remove django-import-export entirely
- [ ] Use existing "validation" of rest serializers
- [ ] Process final import task using background worker (if possible)
- [ ] Add custom framework for "exporting" data using serializers
- [ ] Allow manual override of field values (see below)

## Field Override

For some import processes, we have important "context" information we want to provide. For example, if we are importing a list of "purchase order line items" for a specific order, we do not want to specify the "order" for each row. Instead, we want to override the imported rows with a specific value for the "order" ID. So, we need to be able to override this data somehow...

## References

- https://github.com/wq/django-data-wizard/issues/44

### DIFF
--- a/InvenTree/InvenTree/settings.py
+++ b/InvenTree/InvenTree/settings.py
@@ -188,6 +188,9 @@ INVENTREE_ADMIN_URL = get_setting(
 INSTALLED_APPS = [
     # Admin site integration
     'django.contrib.admin',
+    # django data wizard
+    'data_wizard',
+    'data_wizard.sources',
     # InvenTree apps
     'build.apps.BuildConfig',
     'common.apps.CommonConfig',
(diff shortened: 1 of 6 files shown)


matmair:
Very exciting, let me know if I can support in any way!

SchrodingersGat:
Replaced by https://github.com/inventree/InvenTree/pull/6911

PR #6812, opened 2024-03-21
Document security measures

matmair:
This PR adds a document stating the security measures to keep development and the code base secure.


matmair:
@SchrodingersGat @inventree/triage this would be ready for review and scrutinisation. I think it provides a fair overview of the current security posture. Having this stuff written down makes it easier to point at when questions arise regarding security posture.

SchrodingersGat:
I am happy with this, thanks for taking the time to write this down.

PR #6814, opened 2024-03-21
Fix new manifest path

matmair:
Follow up to https://github.com/inventree/InvenTree/pull/6579 as the path of the manifest file changed.


matmair:
@SchrodingersGat this fixes a bug in PUI that was introduced by me recently https://github.com/inventree/InvenTree/actions/runs/8382277676/job/22955810396

SchrodingersGat:
Nice, looks like this is currently blocking https://github.com/inventree/InvenTree/pull/6815 also

matmair:
@SchrodingersGat this fix now works

PR #6835, opened 2024-03-24
CI-Actions: Security fixes

matmair:
This PR:
- updates the shared actions to run on a current version of node
- defines the workflow permission for actions on the top-level to ensure OIDC tokens are issued with the lowest possible permissions


SchrodingersGat (approved):

SchrodingersGat:
@matmair I believe that this PR has broken the [translations workflow](https://github.com/inventree/InvenTree/blob/master/.github/workflows/translations.yml)

You can see that the workflow [has been broken since then](https://github.com/inventree/InvenTree/actions/workflows/translations.yml)

I assume it is due to the permissions change. Should we revert this for this file, or add an additional permission?

matmair:
You are right, will create a PR to address this

SchrodingersGat:
@matmair great, much appreciated :)

matmair:
PR created, dumb error

PR #6836, opened 2024-03-24
Fix a few warnings in api generation

matmair:
This addresses most „real“ warnings, nearly all remaining are operationId warnings - which I think we can ignore on a functional level.

It:
- Removes the /plugin/activate/ endpoint - which does not seem to serve any purpose as there is a detail endpoint to active a plugin
- Updates a few interpolated fields to typed definitions
- fixes a lot of ModelChiceFilters with custom methods where the type could not be auto-asserted from string to integer (as they are pk-based)
- Updates docstrings to reflect that a bunch of list endpoints also support (bulk) delete actions

### DIFF
--- a/InvenTree/InvenTree/api.py
+++ b/InvenTree/InvenTree/api.py
@@ -91,7 +91,7 @@ class VersionView(APIView):
         })
 
 
-class VersionSerializer(serializers.Serializer):
+class VersionInformationSerializer(serializers.Serializer):
     """Serializer for a single version."""
 
     version = serializers.CharField()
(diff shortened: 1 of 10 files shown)


SchrodingersGat:
Cheers @matmair

SchrodingersGat:
@matmair looks like this is failing in master now:

image

Any ideas?

matmair:
Seems like a version miss-match - should be very easy to solve. Will create follow up PR

SchrodingersGat:
Hey @matmair for some reason this API schema file did not upload - see missing `185` here: https://github.com/inventree/schema/tree/main/export

This is causing subsequent PRs to fail:

https://github.com/inventree/InvenTree/actions/runs/8445857532/job/23133713488?pr=6866

Any ideas? Should we manually generate and upload version `185`?

matmair:
I have manually uploaded 185 - will look into the CI history why that did not update

matmair:
The schema was not uploaded because of the bug fixed in #6837 - should work again as a fix for that is already merged

matmair:
i have manually fixed it - we have some issues in API generation

PR #6837, opened 2024-03-25
[CI] Bump upload to v4 to ensure schema-push still works

matmair:
See https://github.com/inventree/InvenTree/pull/6836#issuecomment-2016977467 for details. Upload and download action need to be in the same major version.


SchrodingersGat:
@matmair thanks for the quick fix :)

PR #6855, opened 2024-03-26
[PUI] Add licenses texts to PUI

matmair:
This PR will:
- [x] Add license extraction steps to frontend and backend builds
- [x] Adds an API that exposes them
- [x] Adds a licenses screen to show frontend and backend licenses (closing PLH component tin nav)
grafik


SchrodingersGat:
@matmair I believe that this requires some further consideration regarding generation of the license file on systems which do not have the node packages installed.

For example in a docker based setup, this would error out as the frontend packages are discarded from the final build image. Maybe we need to update the docker image to pre-compile this file and COPY it across?

Additionally, this means that the `invoke install` step fails unless the required node packages are on the system. As this PR introduces a *new* package requirement, the initial testing on my system failed. 

```bash
git checkout -b matmair-pui-licenses-text master
git pull git@github.com:matmair/InvenTree.git pui-licenses-text
invoke install
```

results in:

```bash
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'rollup-plugin-license' imported from /home/inventree/src/frontend/vite.config.ts.timestamp-1711495260776-0c5a4685b33bb.mjs
```

But after running `invoke frontend-install` the process seems to work (however I do get the following warning message):

image

Running the server, (`invoke server` and `invoke frontend-dev`) I get the following error when trying to view the licences:

image

After running:

```bash
invoke frontend-compile
invoke static -f
```

And then, we have success!

image

SchrodingersGat:
With regard to the actual implementation - It is very hard to look at a particular license because you have to scroll through the entire text. Can we split it up via package, and the user can click on a particular package to see the license available for that package?

This would likely require rendering the file as JSON (which is better anyway IMO) so that the frontend can split by package name.

SchrodingersGat:
Can we potentially make the license file generated via github actions, and remove the need for local generation?

matmair:
The license file for the frontend is part of the PUI static bundle (in the .vite like the manifest required to render the right entry points) and is automatically included so it is not an issue on packaged or docker installs.

When updating an install we recommend running inv update - which will install the new frontend package. What I do not understand is how to invoke install triggers any frontend compilation.

This PR is mainly intended to fulfil the requirement to distribute copies of the licenses visible for the end user for some of the dependencies, I do not think it needs to be pretty to fulfil this legal obligation.

The Python dependencies will always need to be generated on the backend, we do not know in advance what plugins with what dependencies are installed on the server.

SchrodingersGat:
> When updating an install we recommend running inv update - which will install the new frontend package. What I do not understand is how to invoke install triggers any frontend compilation.

After attempting again, I may have got my order of operations wrong in the first comment - I cannot reproduce this on calling `invoke install`.

SchrodingersGat:
@matmair I have added some error handling and split the licenses up so they render like this:

image

Having tinkered a bit more I understand the implementation better. With the addition of error handling (so that dev setups don't break for example) I am happy with this.

wolflu05:
Very nice @SchrodingersGat that you have split up by package now.

Maybe as a last thing it would be good to get line wrapping working, for that a simple `white-space: pre-line` css attribute on the license body container should be sufficient.

src/frontend/src/components/modals/LicenseModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+          </Tabs.List>
+
+          <Tabs.Panel value="backend">
+            {LicenceView(data?.backend ?? [])}
+          </Tabs.Panel>
+
+          <Tabs.Panel value="frontend">
+            {LicenceView(data?.frontend ?? [])}
+          </Tabs.Panel>
+        </Tabs>
```
matmair, line 93: I would much prefer if we would not hard-code the sections here as many deployers will need to add additional dependency sections for their infra

InvenTree/InvenTree/settings.py
```diff
@@ -1015,7 +1015,10 @@
 # Ensure that the ALLOWED_HOSTS do not contain any scheme info
 for i, host in enumerate(ALLOWED_HOSTS):
     if '://' in host:
-        ALLOWED_HOSTS[i] = host.split('://')[1]
+        ALLOWED_HOSTS[i] = host = host.split('://')[1]
+
+    if ':' in host:
+        ALLOWED_HOSTS[i] = host = host.split(':')[0]
```
matmair, line 1021: Question: What is the reason for this?

SchrodingersGat: Looks like this has bled through from https://github.com/inventree/InvenTree/pull/6871 - which is now merged.

Apparently the format of `ALLOWED_HOSTS` does not include protocol *or* port

src/frontend/src/components/modals/LicenseModal.tsx
SchrodingersGat: Would you like to make this extensibility change now, or merge as is with future work to make it extensible? This would require backend API changes too but not much

matmair: My initial PR just had support in the frontend, that should be enough. You can always monkey-patch the backend with a plugin/custom settings.py; I will amend this PR today

src/frontend/src/components/modals/LicenseModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+            <Tabs.Tab value="backend">
+              <Trans>Backend Packages</Trans>
+            </Tabs.Tab>
+            <Tabs.Tab value="frontend">
+              <Trans>Frontend Packages</Trans>
+            </Tabs.Tab>
+          </Tabs.List>
+
+          <Tabs.Panel value="backend">
+            {LicenceView(data?.backend ?? [])}
```
wolflu05, line 87: Is there a reason why we don't use a normal component here, but a function that returns jsx?

```suggestion
            <LicenceView data={data?.backend ?? []} />
```

And then:

```tsx
type LicenseType = {
  name: string;
  license: string;
  version: string;
  licensetext: string;
}

export function LicenceView({ data }: { data: LicenseType[] }) {
// ...
    {entries?.map((entry, index) => (
```

src/frontend/src/components/modals/LicenseModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+            <Accordion.Item key={index} value={`entry-${index}`}>
+              <Accordion.Control>
+                <Group position="apart" grow>
+                  <Text>{entry.name}</Text>
+                  <Text>{entry.license}</Text>
+                  <Space />
+                  <Text>{entry.version}</Text>
+                </Group>
+              </Accordion.Control>
+              <Accordion.Panel>
```
wolflu05, line 37: Make the license text wrapping on `\n`:
```suggestion
              <Accordion.Panel style={{ whiteSpace: "pre-line" }}>
```

src/frontend/src/components/modals/LicenseModal.tsx
matmair: No idea, I have not coded this but I will not type the response - see above comment

wolflu05: I know that you potentially want to monkey patch this, but this has no effect on it as far as I understand you correct. This just types the expected format of one package's license. Providing some different format through patching the backend source will fail on the frontend.

But nevertheless, typing was just the cherry on top I tried to provide, while suggesting using a real react component  instead of a function that returns jsx to stick with the react syntax.

src/frontend/src/components/modals/LicenseModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+            <Accordion.Item key={index} value={`entry-${index}`}>
+              <Accordion.Control>
+                <Group position="apart" grow>
+                  <Text>{entry.name}</Text>
+                  <Text>{entry.license}</Text>
+                  <Space />
+                  <Text>{entry.version}</Text>
+                </Group>
+              </Accordion.Control>
+              <Accordion.Panel style={{ whiteSpace: "pre-line" }}>
```
wolflu05, line 37: Sorry, this was my fault with the `"`, I used to use `"` instead of `'` in a lot of other projects.
```suggestion
              <Accordion.Panel style={{ whiteSpace: 'pre-line' }}>
```

SchrodingersGat:
@matmair happy to merge?

matmair:
@SchrodingersGat I will try to fix this up soon

matmair:
This is now ready for merge from my POV

SchrodingersGat:
@matmair please address conflicting files and then I will review again

matmair:
@SchrodingersGat merge is in!

matmair:
@SchrodingersGat this is ready now, the 4 errors in docker seem to be related to https://github.com/inventree/InvenTree/issues/6943

SchrodingersGat (approved):

matmair:
@SchrodingersGat CI is successful now

SchrodingersGat:
Any idea why the migrations checks were skipped here? They should be running due to the change in `requirements.txt`...

matmair:
Probably because the diffset does not contain the file but not sure

matmair:
Seems like that change is not working. Can we either roll back #6942 or merge this so that master is building again and try to figure out the migrations in a dedicated PR?

SchrodingersGat:
Yep, I'll merge this in.

SchrodingersGat:
@matmair cheers :)

PR #6865, opened 2024-03-26
Fix: add write permissions to translation workflow

matmair:
Addresses https://github.com/inventree/InvenTree/pull/6835#issuecomment-2018915630 - in this specific case we want to allow writes to the code base


SchrodingersGat:
Thanks for the fix @matmair

PR #6866, opened 2024-03-27
Order responsible requirement

SchrodingersGat:
Adds a new setting (a bunch of new settings) to control whether the "responsible" field is required for various order types.


SchrodingersGat:
Note the failed API schema test - which does not apply here - but needs to be sorted out.

PR #6872, opened 2024-03-27
Implement URL redirect for PUI

SchrodingersGat:
- Return from login page to redirect URL
- Return from auth check to redirect URL


matmair:
@SchrodingersGat I took the liberty to fix the test

PR #6883, opened 2024-03-27
Add test start, end date and test station to the next gen UI

martonmiklos:
Following upon this:
https://github.com/inventree/InvenTree/pull/6149#issuecomment-2002637842

Here is a PR which adds the new test fields to the React UI.

### DIFF
--- a/src/frontend/src/components/forms/fields/ApiFormField.tsx
+++ b/src/frontend/src/components/forms/fields/ApiFormField.tsx
@@ -216,7 +216,6 @@ export function ApiFormField({
           />
         );
       case 'date':
-        return <DateField controller={controller} definition={definition} />;
       case 'datetime':
         return <DateField controller={controller} definition={definition} />;
       case 'integer':
--- a/src/frontend/src/components/forms/fields/ApiFormField.tsx
+++ b/src/frontend/src/components/forms/fields/ApiFormField.tsx
@@ -64,6 +64,7 @@ export type ApiFormFieldType = {
     | 'string'
     | 'boolean'
     | 'date'
+    | 'datetime'
     | 'integer'
     | 'decimal'
     | 'float'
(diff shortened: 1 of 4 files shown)


martonmiklos:
> While you are likely still working on this:
> 

Dauh, my bad. I will fix it tomorrow...

src/frontend/src/components/forms/fields/ApiFormField.tsx
```diff
@@ -216,6 +217,8 @@ export function ApiFormField({
         );
       case 'date':
         return <DateField controller={controller} definition={definition} />;
+      case 'datetime':
+        return <DateField controller={controller} definition={definition} />;
```
wolflu05, line 221: This could be simplified to using two case blocks with one return as both have the same arguments like done with int, float, ...

src/frontend/src/components/forms/fields/DateField.tsx
```diff
@@ -11,20 +13,25 @@ export default function DateField({
   controller: UseControllerReturn<FieldValues, any>;
   definition: ApiFormFieldType;
 }) {
+  dayjs.extend(customParseFormat);
```
wolflu05, line 16: Wouldn't this make sense to move this outside of the component into the global scope of this file, as this function is called a lot of times (on every change of the value)?

wolflu05:
BTW, it would be great if you could stop force pushing as this may break review comments and makes it really hard to follow along the changes. Just push new changes as new commits, they are squash merged anyway.

src/frontend/src/components/forms/fields/DateField.tsx
martonmiklos: I am new to the PUI and the typescipt thing, where is exactly  the global scope of this file defined?

wolflu05: Just outside the function directly in the file after the imports. Because this line doesn't seem to be dependent on any react stuff.

martonmiklos: Ah thanks, I thought that it should be moved to some more global file.

SchrodingersGat:
LGTM - @wolflu05 are you happy with this also?

wolflu05 (approved):
Yes, LGTM

Issue #5748, opened 2023-10-18
[React] User Interface for new barcode scanning


SchrodingersGat:
https://github.com/inventree/InvenTree/pull/5509/ adds new barcode scanning features, to check-in items from a purchase order.

The UI elements should be added, but only to the *new* user interface (new large UI features are no longer accepted on the old UI)

@30350n FYI for future work :)

matmair:
I'd be happy to add this to the new scanning experience but @30350n has first dips ofc.

30350n:
Uhhh i'm not really much of a frontend dev tbh (I can count the number of times I've written JS on one hand), but I guess I could provide a general outline of how I'd expect to be using this from the UI.

So for the app, I've added the functionality to a few different places:
1. Purchase Order List:
    - Basically the main place to access the functionality
    - no assumptions: build order has to be able to be concluded from the barcode, location either has to be already set (with a default), or entered manually
2. Purchase Order Detail:
    - purchase order is already selected, useful if the barcode doesn't contain a reference or it's somehow invalid
3. Stock Location Detail:
    - purchase order has to be concluded from the barcode, but stock location is explicitly set

All of these can also still require additional information to be filled in via a form, if for example no quantity is provided in the barcode.

matmair:
I will just implement it in the scanning section for now, we can see where we want it once we have more of the UI ported over.

XanderLuciano:
I'd like to hop in potentially. I have a need for the ability to easily and quickly scan stock items in/out to track quantity (not using PO's / Sales Orders / Builds).

I am imagining a system that is basically a page that acts like a self checkout stand at a store. 

For my needs, a user needs to be able to grab a stock item from an stock-location, scan the barcode, and adjust the stock-location quantity.

Not sure if that functionality fits in well to the vision @matmair has for this page, but open to adding a "check out" / "check in" action to this page

matmair:
That would be a possible action. The main thing missing here is implementing all the actions for all the models

XanderLuciano:
![BetterCameraLoadingBehavior](https://github.com/inventree/InvenTree/assets/13877593/17281597-5a20-4e23-8341-6da143786d00)

Work in progress preview

XanderLuciano:
![image](https://github.com/inventree/InvenTree/assets/13877593/94167997-75d3-40f2-abaf-947195f35bb5)

Additional UI Progress, scoping feature to removing items from a stock location first, then will work towards checking in new stock. 

[...]

PR #6884, opened 2024-03-28
Refactor Scan Page

XanderLuciano:
Refactors Scan Page into more discrete components. The components on the page are still fairly tightly coupled, but separating into separate component files makes the separation of functions clearer and improves maintainability of each individual aspect.

Along with the refactor, I also included a few improvements of existing functions. I have changed some synchronous functions into Async/Await promises that are no longer blocking. I've also added some new UX features to communicate to the user what the current state is for the camera (stopped vs. loading vs. scanning).

This also improves the UX of the scanning with camera feature by preventing the dom element from resizing when the camera starts/stops.

Barcodes are now automatically searched upon successful scan. 

I have more planned for this page, but felt the current state of this PR is at a good state to push upstream while I work on new features.

Example Functionality:
ScanPage-1


XanderLuciano:
tracks towards: #5748

src/frontend/src/pages/scan/BarcodeInputManual.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+// Region input stuff
+enum InputMethod {
+  Manual = 'manually',
+  ImageBarcode = 'imageBarcode'
+}
+
+interface inputProps {
+  action: (items: ScanItem) => void;
+}
```
matmair, line 30: Blocking: This is defined in multiple files, there should be one shared interface

matmair (changes requested):
I see multiple problems in a quick test:
1) Constant errors that no QR code was found making notifications unusable
2) Unknown codes are just disregarded - and do not show up in history; this is a blocking bug

XanderLuciano:
> Constant errors that no QR code was found

Hmm, I encountered this at one point and thought the string comparison would suffice, but probably need to revert to just dump the errors to console instead of the pop-up notification. I was hoping to pass meaningful errors up to the user, but if it results in an error notification on every frame, that's bad.

> Unknown codes are just disregarded - and do not show up in history

What is the intended behavior for scanning invalid QR codes? Just showing the raw QR code string value in the history?

Does showing invalid data in the history provide meaningful information to the user? Or would a popup notification with an error message and the QR code string also provide sufficient feedback to the user?

src/frontend/src/pages/scan/BarcodeInputManual.tsx
XanderLuciano: Moved into common interface file for reuse

PR #6900, opened 2024-03-30
Fix TME supplier order number splitting

30350n:
Very late follow up to #6158, just noticed while scanning some TME barcodes that the Order Numbers don't get parsed correctly anymore.


30350n:
(Also seems like they recently changed their DataMatrix code formatting, which makes those codes straight up unparsable ...

SchrodingersGat:
@30350n thanks for the patch

PR #6901, opened 2024-04-01
Allow for non-square logos in Navbar

psuet:
Some Logos are not "squareable", so they currently look either distorted (without user intervention) or way too small (by manually squaring it)

By loosening the width restriction, most logos, square or rectangular (max ratio 2:1), will not be distorted and scaled automatically.

### DIFF
--- a/InvenTree/templates/navbar.html
+++ b/InvenTree/templates/navbar.html
@@ -14,7 +14,7 @@
 <nav class="navbar {% if sticky %}fixed-top{% endif %} navbar-expand-lg navbar-light">
   <div class="container-fluid">
     <div class="navbar-header clearfix content-heading">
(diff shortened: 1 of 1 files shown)


matmair:
At a minimum, there should be max-width set to enforce a reasonable/usable layout

SchrodingersGat:
@psuet can you please address the `max-width` requirement as above

SchrodingersGat:
Hi @psuet if you can fix the conflicts here I'm happy to merge this in, don't want it sitting here forever :)

matmair (approved):
LGTM @SchrodingersGat

SchrodingersGat:
@psuet can you please address the conflicts here?

Issue #6589, opened 2024-02-27
[Docker] Container permissions


SchrodingersGat:
currently the docker / docker compose setup runs with `root` permissions - this means that any external files are created under the `root` user.

- How to force run as a different user?
- Do we need to make any changes to the `Dockerfile` to address this?
- What about other containers which store data (postgres / caddy)

oddstr13:
Setting the `user` in the compose file is not sufficient, attempting to access file `//.gitconfig` fails – I did not investigate any further.

Caddy will be needing at least capabilities `NET_BIND_SERVICE` in order to bind to the privileged ports 80 and 443, [at least for older docker versions](https://stackoverflow.com/a/66892807/1035647)

```
inventree-server  | error: could not lock config file //.gitconfig: Permission denied
inventree-worker  | error: could not lock config file //.gitconfig: Permission denied
```

PR #6905, opened 2024-04-02
Dockerfile fixes

SchrodingersGat:
- Change "as" to "AS"
- Exposed by sonarcloud analysis


SchrodingersGat:
Note that sonarcloud has raised the "root user" issue - https://github.com/inventree/InvenTree/issues/6589

matmair:
@SchrodingersGat sonarcloud is pretty good about detecting these kind of issues

PR #6921, opened 2024-04-02
Fix TME supplier order number splitting (Backport #6900)

30350n:
Backport for #6900.

* Fix TME supplier order number splitting

* Fix order number not split for QR codes, handle standard ecia codes


SchrodingersGat:
@30350n thanks for the backport :)

PR #5582, opened 2023-09-21
Code structure refactor

matmair:
This PR implements #5578 and reorganises the code base more scalable. This is the first PR in a string of PRs needed, see #5578.
- [x] move container things into contrib/docker
- [x] move Django stuff into src/backend
- [x] add docs about structure to CONTRIBUTING
- [x] move misc things out
- [x] test docker build
- [x] test package build
- [x] update docs for bare metal git install

closes https://github.com/inventree/InvenTree/issues/5578

### Breaking Change

This is a significant breaking change as it rearranges a lot of source files. Any developers who are using the devcontainer setup will need to do a complete rebuild (without cache) of the devcontainer.

PR #6945, opened 2024-04-03
Fix test coverage

matmair:
It seems that #5582 has reduced coverage due to leaving some apps tests untouched.

This PR:
- removes unnecessary artefact steps to preserve resources
- adds a mechanism that forces almost all ci steps to run if the PR has either the `dependency` or `full-run` label
- sets up coverage source and apps correctly
- re-adds migration testing coverage


matmair:
@SchrodingersGat this is ready for merge!

PR #6947, opened 2024-04-03
Use xml for coverage info

matmair:
This PR improve python coverage accuracy for coveralls to keep if comparable with codecov; especially coverage on multilines. Follow up to #6945


matmair:
this messes with codecov settings - do not merge

SchrodingersGat:
So many commits for such a small overall change :)

matmair:
The smaller the final diff the happier I am ;-)
This revealed a lot of holes in coverage where coveralls falsely assumed statements were covered but codecov showed they were not. Now the score on coveralls is far more acurate

PR #6948, opened 2024-04-04
Add support for django-query-count

SchrodingersGat:
In `DEBUG` mode (and with a particular env var set) this PR enables support for query-count reporting on the `/api/` and `/plugin/` endpoints.

This is a setup I use a lot for debugging inefficient database queries, and it is very useful!

Ref: https://github.com/bradmontgomery/django-querycount


matmair:
Cool package!

PR #6965, opened 2024-04-05
Tests: remove dead paths in test files

matmair:
This PR removes some dead testing paths found during work on #6947


matmair:
Submitting now as the testing PRs #6945 #6947 might take longer than anticipated to finish

PR #6970, opened 2024-04-07
[PUI] Session authentication

SchrodingersGat:
- Replaces https://github.com/inventree/InvenTree/pull/6399
- Closes https://github.com/inventree/InvenTree/issues/6391
- Closes https://github.com/inventree/InvenTree/issues/5697

This PR provides session based authentication for the react frontend. A number of backend changes are also required for this to work correctly. A lot of testing has been performed to ensure that it covers various login / logout edge cases.

### Major Features

- Remove API token support from frontend entirely
- ApiImage now just uses regular `` tag (and works!)
- Attachments can now be downloaded from PUI frontend
- Attachments can now be uploaded successfully in PUI frontend
- Opens up pathway for label and report printing in PUI frontend
- Custom implementation of `/api/auth/login/` to prevent login failure with stale `sessionid` cookie
- Fix backend issue resulting from our custom token implementation
- Refactor / simplify frontend authentication implementation


SchrodingersGat:
@matmair I was playing around with https://github.com/inventree/InvenTree/pull/6399 - trying to get it to work reliably. In the end (and after reading way too many docs on cookies / cross site stuff) I have discovered that we need some backend changes too. So, I have written this new implementation which replaces #6399

This has been tested with attachment upload / download and handles it all quite cleanly (as far as I have been able to test)

Let me know what you think!

SchrodingersGat:
@matmair can you give some insight into the playwright tests failing?

wolflu05 (commented):
I never used session authentication with a SPA. I always used api tokens. How does this session cookie work when the backend is hosted somewhere else than the frontend? E.g. like the netlify previews with the demo server. Do there need to be something configured specifically?

SchrodingersGat:
> I never used session authentication with a SPA. I always used api tokens. How does this session cookie work when the backend is hosted somewhere else than the frontend? E.g. like the netlify previews with the demo server. Do there need to be something configured specifically?

The *host* needs to have the correct **CORS** and **CSRF** settings configured. I have tested that this works on the devcontainer setup (which is effectively a CORS request from `localhost:5173` to `localhost:8000`). It works well there and also in the "compiled" version when the javascript files are served statically from `localhost:8000`.

The current netlify previews won't work because the demo server code needs updating (based on this PR). But once that is updated it will work fine

wolflu05:
So you need to explicitly configure the netlify domain in the demo server then? Or is there a wildcard allowed currently on the demo server?

SchrodingersGat:
The demo server already allows cross origin requests from the netlify preview site

matmair:
To me, the frontend tests make it look like the username field are not available. @SchrodingersGat  Have you checked that the tests run on your dev setup`

SchrodingersGat:
@matmair I cannot run these tests locally, I believe that it is due to incompatibility with the alpine devcontainer. I get errors like:

```
Error: browserType.launch: Failed to launch: Error: spawn /home/vscode/.cache/ms-playwright/firefox-1440/firefox/firefox ENOENT
```

matmair:
It seems to run on my dev setup, I will try to fix this CI on the weekend

SchrodingersGat:
@matmair I have pulled in your recent test changes but still no luck with these tests. If you do have any time to look into this I'd appreciate it :)

SchrodingersGat:
@matmair reviewing the playwright traces it appears that the login screen is being bypassed, and the homescreen is being loaded:

image

SchrodingersGat:
Still not wrapping my head around these tests :|

matmair:
@SchrodingersGat it seems like there are problems with login on tests that do not specifically cover logins. Maybe we can bypass logging in there similar to https://github.com/inventree/InvenTree/pull/7022 to solve this? Testing login once specifically should be enough.

SchrodingersGat:
Woo! It works, finally :) Thanks for the help with the tests @matmair

matmair:
Seeing that coveralls coverage for PUI was disabled here - should we discuss #6986?

PR #6973, opened 2024-04-07
[BUG] Include .vite subdir in PUI releases

matmair:
This PR:
- fixes a wrong path in the license export
- includes the .vite directory in the ziped releases
- uses yarn as a runner instead of npm for consistency


matmair:
Thanks for the report @wolflu05

wolflu05:
Thanks for the quick fix 👍

PR #5992, opened 2023-11-27
[PUI] Part pricing

SchrodingersGat:
Implement part pricing detail screens in the new interface

PR #6985, opened 2024-04-09
[PUI] pricing tab

SchrodingersGat:
- Replaces https://github.com/inventree/InvenTree/pull/5992
- Implement part pricing views for PUI
- Use [recharts](https://recharts.org/en-US/) for various chart elements

### Future Work

Ideas for improvements after this is merged:

- [ ] click on a line in the overview and jump to the matching accordion (that should also open)
- [ ] The data is displayed as decimal, not in local currency formatting. Can that be changed?
- [ ] use a colour scheme generator like CUI - those red and green colours especially are a bit difficult for colour-blind users
- [ ] if iv. is implemented as a donut: a switch to render about and percental rendering
- [ ] improve rendering of tooltip label data
- [ ] Improve rendering of legend label data


SchrodingersGat:
@matmair @wolffam LMK what you think. I quite like the recharts library, once this is merged, I will work on scheduling and stocktake panels with the same charts

wolffam:
> @matmair @wolffam LMK what you think. I quite like the recharts library, once this is merged, I will work on scheduling and stocktake panels with the same charts

tagging @wolflu05 instead ;)

matmair:
Very nice overall @SchrodingersGat; I noticed a few things:
- 1) If pricing is available it feels redundant to be able to open it. We could simply disable the "empty" accordion and add a tooltip
- 2) It would be nice if it would be possible to click on a line in the overview and jump to the matching accordion (that should also open)
- 3) there is a lot of empty space here, could the table and graph be the same height?
grafik
- 4) on my real-world data the BOM section is very unreadable because there are 20+ items - could we switch back to a donut again?
grafik
- 5) There seems to be horizontal padding in the accordion tabs. Can that be removed/reduced?
grafik
grafik

 - 6) The data is displayed as decimal, not in local currency formatting. Can that be changed?
 - 7) At initial loading you shortly see all possible pricing ranges - they then disappear. Could that be changed to first show nothing and then render applicable elements?

Ideas for future iterations:
- use a colour scheme generator like CUI - those red and green colours especially are a bit difficult for colour-blind users
- if iv. is implemented as a donut: a switch to render about and percental rendering
- hover cards for BOM items to show min max pricing in the different categories, maybe history

src/frontend/src/defaults/formatters.tsx
```diff
@@ -21,13 +43,21 @@ interface formatCurrencyOptionsType {
...
-  value: number | null,
+  value: number | string | null | undefined,
   options: formatCurrencyOptionsType = {}
 ) {
-  if (value == null) {
+  if (value == null || value == undefined) {
     return null;
   }
 
+  value = parseFloat(value.toString());
```
matmair, line 53: Question: Is there a way to only convert the value if it is not a number? Seems redundant to do it this way but not sure how much overhead this creates

src/frontend/src/functions/urls.tsx
```diff
@@ -1,18 +1,30 @@
...
   const modelInfo = ModelInformationDict[model];
 
   if (pk === undefined || pk === null) {
     return '';
   }
 
   if (!!pk && modelInfo && modelInfo.url_detail) {
-    return modelInfo.url_detail.replace(':pk', pk.toString());
+    let url = modelInfo.url_detail.replace(':pk', pk.toString());
+    let base = base_url;
```
matmair, line 21: Question: Is there a reason those are let and not const?

src/frontend/src/pages/part/PartPricingPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+    instanceQuery
+  } = useInstance({
+    pk: part?.pk,
+    hasPrimaryKey: true,
+    endpoint: ApiEndpoints.part_pricing_get,
+    defaultValue: {}
+  });
+
+  // TODO: Do we display internal price? This is a global setting
+  const internalPricing = true;
```
matmair, line 33: Comment: Settings can be accessed via useGlobalSettingsState zustand

matmair (approved):
LGTM

src/frontend/src/functions/urls.tsx
SchrodingersGat: Fixed

src/frontend/src/pages/part/PartPricingPanel.tsx
SchrodingersGat: Done now, that was a placeholder for myself :)

SchrodingersGat:
@wolflu05 as far as I can tell, mantine uses recharts internally, which is why I chose it for the upgrade path

PR #6986, opened 2024-04-09
Remove/Replace all references to coveralls

### LINKED ISSUE #6966: Switch coverage tracking provider?
We are experiencing some weird behaviour around the UI with coveralls and there is some doubt about how active they still work on their infra/coverage uploaders. Their behaviour around flags/carryforward and closing flags is also causing problems

CodeCov would be a known, open-source (ish?) competitor by Sentry, which we already use without any problems. I would propose to test-drive CodeCov parallel to Coveralls and decide if it is better after a few weeks/months. Especially regarding the disrubtions the usage of DeepSource caused. Thoughts @inventree/triage @inventree/maintainer ?

matmair:
Remove references to Coveralls and replace where appropriate with CodeCov
Closes #6966

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -286,16 +286,6 @@ jobs:
         run: python3 .github/scripts/check_migration_files.py
       - name: Coverage Tests
         run: invoke test --coverage
-      - name: Upload Coverage Report to Coveralls
-        if: always()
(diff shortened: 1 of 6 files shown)


matmair:
This is in preparation for once we are confident in CodeCov - only submitting it now as I am still very comfortable with this part of the code base

SchrodingersGat (approved):

Issue #5888, opened 2023-11-08
[PUI] Quick commands pallet


matmair:
Add a quick command pallet like GitHub provides:
- Navigation
- Actions (logout, switch theme)
- Maybe content-dependent actions

PR #6987, opened 2024-04-09
[PUI] Quick commands pallet

matmair:
This adds a quick command palette with (at least now) simple actions to navigate to a few parts of the UI without the mouse.

https://github.com/inventree/InvenTree/assets/66015116/fad52d2b-86b3-4c4a-bfd1-726f2f4368b1

Clsoes #5888


matmair:
This is now feature complete and ready for review; I still have to figure out why the keyboard-based tests keep failing but that is probably something with the testing framework

matmair:
OK the tests are solved (stupid issue on my side, is documented by playwright very clearly) - I will try to get coverage of new lines to 100% and than it should be ready

matmair:
Coverage of new code is at 90%+, this is ready for review @SchrodingersGat

SchrodingersGat (approved):

SchrodingersGat:
@matmair really happy with this, it looks great. In the future I would love to see some custom actions for various context (i.e. "create a new build order").

Additionally, should we update "search" panel to the same visual style (spotlight vs drawer)? Or keep them visually separate.

In any case, happy to merge, nice work :)

matmair:
I think search works fine as is, I would leave it till we have the minimal function set of CUI reached in PUI. Flashing out this while there is nearly no usage seems like a waste.

There could be integrated search in the command line, context actions, maybe plugins. At some point I would like this to become similar to command + shift + P in vs code or Command + K in Linear but we will probably have to shift libraries for that.

PR #6989, opened 2024-04-10
Purchaser order approvals and "ready" state.

LavissaWoW:
Adds approvals to Purchase Orders.

Also introduces a "ready" status and ability to limit order issuance to members of a group.


LavissaWoW:
image

LavissaWoW:
@SchrodingersGat Have a look at the comments here.
Any input?

src/backend/InvenTree/InvenTree/templatetags/inventree_extras.py
```diff
@@ -663,3 +663,43 @@ def admin_url(user, table, pk):
             pass
 
     return url
+
+
+@register.simple_tag()
+def approval_allowed(user, order):
```
SchrodingersGat, line 669: I would not put this code into "inventree_extras" as a template tag. This should be a method on the `PurchaseOrder` model, such as:

```python
def approval_allowed(self, user):
    ...
```

And then *call* that from the template tag. That will make the code far more re-usable.

src/backend/InvenTree/order/models.py
```diff
@@ -577,13 +608,181 @@ def add_line_item(
...
+            raise ValidationError('Ready state setting not active')
+
+        if self.is_pending or (self.is_pending_approval and self.approved_by):
+            self.status = PurchaseOrderStatus.READY.value
+            self.save()
+
+            user = kwargs.pop('user', None)
+
+            purchaser_group_name = getSetting('PURCHASE_ORDER_PURCHASER_GROUP')
+            if purchaser_group_name:
```
SchrodingersGat, line 746: You could potentially use the aforementioned `approval_allowed` method here.

SchrodingersGat:
@LavissaWoW thanks for submitting. @matmair I'd be interested in your thoughts on this, obviously you have put a lot of work into a very similar plugin.

I'd like to see this integrated into the new interface before merging. This will require the "purchase order actions" (submit / cancel / approve / etc) to be added to the PurcahseOrderDetail page.

LavissaWoW:
@SchrodingersGat working on it 👌

matmair:
I have a conflict of interest and will not comment or review anything relating to this.

LavissaWoW:
Some obvious Mantine v7 changes needed. I'm going to be mostly off the next 10 days or so, but I'll try to patch this up.

LavissaWoW:
Using the PO model property `is_complete` in a serializer results in an O(n) increase in queries made.
Just if anyone was wondering :>

src/backend/InvenTree/InvenTree/templatetags/inventree_extras.py
LavissaWoW: Done

src/backend/InvenTree/common/models.py
```diff
@@ -2926,7 +2959,7 @@ class Meta:
     @classmethod
     def check_recent(cls, key: str, uid: int, delta: timedelta):
         """Test if a particular notification has been sent in the specified time period."""
-        since = InvenTree.helpers.current_date() - delta
+        since = InvenTree.helpers.current_time() - delta
```
matmair, line 2962: Is this intentionally changed? This is breaking

src/backend/InvenTree/order/api.py
```diff
@@ -1617,6 +1659,30 @@ def item_link(self, item):
...
+                    path(
+                        'can_issue/',
+                        PurchaseOrderIssueAllowed.as_view(),
+                        name='api-po-issue-allowed',
+                    ),
+                    path(
+                        'pending/',
+                        PurchaseOrderPending.as_view(),
+                        name='api-po-pending',
+                    ),
```
matmair, line 1685: Why are we adding this many API endpoints for this? At least `can_approve`, `can_issue`, `pending` could be either merged or just added to the PO detail endpoint

src/backend/InvenTree/order/models.py
```diff
@@ -523,6 +523,37 @@ def company(self):
...
+
+    approved_date = models.DateField(
+        blank=True,
+        null=True,
+        verbose_name=_('Approve Date'),
+        help_text=_('Date order was approved'),
+    )
+
+    reject_reason = models.CharField(
+        max_length=128,
```
matmair, line 542: 128 chars is short and very specific for freeform text - is there a technical reason?

src/backend/InvenTree/order/models.py
```diff
@@ -523,6 +523,37 @@ def company(self):
...
+        blank=True,
+        verbose_name=_('Reason for rejection'),
+        help_text=_('The reason for rejecting this order'),
+    )
+
+    placed_by = models.ForeignKey(
+        User,
+        related_name='orders_placed',
```
matmair, line 550: This field is only present in purchase orders, the related name could be confusing to plugin creators

src/backend/InvenTree/order/models.py
```diff
@@ -618,6 +817,8 @@ def _action_place(self, *args, **kwargs):
                 exclude=self.created_by,
                 content=InvenTreeNotificationBodies.NewOrder,
             )
+        else:
+            raise ValidationError('Order must be Pending or Ready')
```
matmair, line 821: This error is not translated

src/backend/InvenTree/order/models.py
```diff
@@ -638,10 +839,58 @@ def _action_complete(self, *args, **kwargs):
...
+                PurchaseOrderStatus.READY,
+                self,
+                self._action_ready,
+                user=user,
+            )
+        else:
+            raise ValidationError('Ready state setting not active')
+
+    @transaction.atomic
+    def place_order(self, user):
```
matmair, line 886: This is breaking for plugin devs as a user is required and not optional

src/backend/InvenTree/order/models.py
```diff
@@ -696,6 +965,35 @@ def _action_cancel(self, *args, **kwargs):
                 content=InvenTreeNotificationBodies.OrderCanceled,
             )
 
+    def approval_allowed(self, user):
+        """Check that the given user is allowed to approve the order."""
+        active = getSetting('ENABLE_PURCHASE_ORDER_APPROVAL')
+        masters = getSetting('PURCHASE_ORDER_APPROVE_ALL_GROUP')
```
matmair, line 971: Masters is an intresting word choice

src/backend/InvenTree/order/serializers.py
```diff
@@ -325,6 +336,107 @@ def save(self):
...
+class PurchaseOrderRequestApprovalSerializer(serializers.Serializer):
+    """Serializer for requesting approval of a purchase order by eligible users."""
+
+    class Meta:
+        """Metaclass options."""
+
+        fields = []
+
+    def save(self):
+        """Save the serializer to 'approve' the order."""
```
matmair, line 348: 1 of these docstrings is wrong

src/backend/InvenTree/order/serializers.py
```diff
@@ -325,6 +336,107 @@ def save(self):
...
+class PurchaseOrderReadySerializer(serializers.Serializer):
+    """Serializer for approving a purchase order."""
+
+    class Meta:
+        """Metaclass options."""
+
+        fields = []
+
+    def save(self):
+        """Save the serializer to 'approve' the order."""
```
matmair, line 381: These docstrings and varnames do not match

src/backend/InvenTree/order/serializers.py
```diff
@@ -325,6 +336,107 @@ def save(self):
...
+
+
+class PurchaseOrderPendingSerializer(serializers.Serializer):
+    """Serializer for rejecting a purchase order."""
+
+    class Meta:
+        """Metaclass options."""
+
+    def save(self):
+        """Save the serializer to 'reject' the order."""
```
matmair, line 435: These docstring do not match the class name

src/backend/InvenTree/order/templates/order/order_base.html
```diff
@@ -53,7 +55,7 @@
         <li><a class='dropdown-item' href='#' id='export-order'><span class='fas fa-file-download'></span> {% trans "Export order to file" %}</a></li>
     </ul>
 </div>
-{% if roles.purchase_order.change %}
+{% if roles.purchase_order.add %}
```
matmair, line 58: Why is this role-check changed?

src/frontend/src/states/UserState.tsx
```diff
@@ -167,5 +168,8 @@ export const useUserState = create<UserStateProps>((set, get) => ({
   },
   hasViewRole: (role: UserRoles) => {
     return get().checkUserRole(role, UserPermissions.view);
+  },
+  canApproveOrder: (order: number) => {
+    return false;
```
matmair, line 173: Is this a placeholder?

src/backend/InvenTree/InvenTree/api_version.py
```diff
@@ -1,12 +1,26 @@
...
 
+v202 - 2024-05-23 : https://github.com/inventree/InvenTree/pull/6989
+    - Adds new API endpoints for PurchaseOrder state transitions
+        - Request Approval
+        - User allowed to Approve
+        - Reject order
+        - Order Ready to issue
+        - User allowed to Issue
+        - Recall order to Pending
+    - Adds fields to PurchaseOrder PAI calls:
```
matmair, line 18: Typo

src/backend/InvenTree/common/models.py
LavissaWoW: Is it? Did I miss some day-to-time conversions? As it stands, with current date, no notification will be sent twice within the same date. So I changed it to time, and only altered it if an override is supplied to the notification method.

src/backend/InvenTree/order/api.py
LavissaWoW: Pending is a state transition like others. But if you know a better way to assert if a user is able to approve/issue an order through other means, I'm all ears :)

src/backend/InvenTree/order/models.py
LavissaWoW: Nope. Chosen based on other values for strings within the codebase. Happy to change it if it's short.

src/backend/InvenTree/order/models.py
LavissaWoW: Fair point. I'll rename it.

src/backend/InvenTree/order/models.py
LavissaWoW: Can make it optional. Will mean that placed_by won't be populated, but I presume that's okay for authors using that endpoint.

src/backend/InvenTree/order/models.py
LavissaWoW: Well, I named it a master group in the settings, so. It just matched that choice 😅

src/backend/InvenTree/order/models.py
matmair: All fields that use 128 chars are md5 checksums. Our default for short free text fields is 250 chars

src/backend/InvenTree/order/templates/order/order_base.html
LavissaWoW: This is directly related to the issue I raised regarding roles. This is where I discovered the discrepancy. A change role doesn't allow you to actually do state transitions. So I changed the role to reflect what you actually need to perform the action.

src/frontend/src/states/UserState.tsx
LavissaWoW: This looks like leftovers from an experiment they slipped through the cracks. I'll remove it if I don't re-discover why I added it.

src/backend/InvenTree/order/models.py
LavissaWoW: Aight. Will fix 😊

src/backend/InvenTree/common/models.py
matmair: current_date returns a date object that is not aware of the specific time, current_time a datetime object so the behaviour changes and this is a public method (ie plugins like mine are using this already). The PR needs to be marked breaking and mentioned in the release note

LavissaWoW:
@matmair Thank you for the review. You caught several oversights from my side.

matmair:
I just do not want to have breakages in master, I will never use any of this

src/backend/InvenTree/common/models.py
LavissaWoW: Hmm. That's definitely something. I could alter it so current_date is used if no delta override is defined? Then it won't be breaking?

LavissaWoW:
Still, I appreciate it!

src/backend/InvenTree/order/models.py
matmair: I would prefer that, sourcegraph shows at least 20 source-available plugins using this function, probably more

LavissaWoW: Consider it done 👌

src/backend/InvenTree/order/api.py
matmair: then it should be renamed - the name `pending` is not communicating that it is changing something, a verb  like `reject` or an action like `request_approval` would be better

src/backend/InvenTree/common/models.py
matmair: the problem is that a date is always 0hrs 0min while datetime is the real time so the value of since will be different

src/backend/InvenTree/order/api.py
LavissaWoW: Hmm, yeah. I named it for its state, but I've later called it "recall" on buttons. So maybe recall might work better than pending. Rejecting uses the same endpoint, but from its own backend. What do you think about changing it to "Recall", "retract" or similar?

src/backend/InvenTree/common/models.py
LavissaWoW: I'll look into it. I'll implement it such that if you do not supply a delta, nothing's changed.

src/backend/InvenTree/order/api.py
matmair: recall or retract would both make sense

src/backend/InvenTree/order/templates/order/order_base.html
matmair: @SchrodingersGat do we want to change the perm or this assignment?

LavissaWoW: Just to make sure we're on the same page here:
With only the change role, you get 403 errors when pushing the buttons. So, essentially, without the add role you can't actually perform the action you're allowed to see
 State transitions are POST requests, and change roles are only permitted to use PUT/PATCH.

LavissaWoW:
Changes made after review:

- Helper function `current_time` only used if `trigger_notification` received a kwarg called "delta", else `current_date` is used. (Prevents possibly breaking change to notification timers)
- PO Pending endpoint renamed to Recall (To better reflect what the transition achieves)
- Merged endpoints `PurchaseOrderApprovalAllowed` and `PurchaseOrderIssueAllowed` into `PurchaseOrderStatePermissions`
  - The two former endpoints have been deleted
  - `useQuery` hooks have been given identical `queryKey` to prevent duplicate requests if the user does not reload the page between approve and issue state transitions.
- `po.reject_reason` field length increased to standard 250.
- `related_name` for `placed_by` changed to `purchase_orders_placed`
- Made `user` argument optional for `po.place_order` to prevent breaking changes.
- Changed some funny wording
- Fixed some Serializer and API docstrings
- Removed UserState expermient code

src/backend/InvenTree/common/models.py
LavissaWoW: Let me know if the revised solution is okay

src/backend/InvenTree/order/templates/order/order_base.html
LavissaWoW: firefox_aJ6GW5JBmV

LavissaWoW: @matmair @SchrodingersGat This is what a user without Add gets

LavissaWoW:
@SchrodingersGat I consider this one ready if @matmair is happy with my mitigations to avoid breaking changes.

LavissaWoW:
@SchrodingersGat  I'm not able to fix the conflicts until next week, but please let me know if I need to change anything else. Looking at the requested changes, I'm going to interpret silence as agreement and resolve the last two review points unless I hear anything by monday.

SchrodingersGat:
@LavissaWoW I will review this now. Apologies for the radio silence on this, have been very busy at work and also focusing on the installer related issues.

src/backend/InvenTree/order/models.py
```diff
@@ -599,13 +630,181 @@ def add_line_item(
...
+        if not self.requires_approval and not self.can_be_ready:
+            raise ValidationError(_('Ready state setting not active'))
+
+        if self.is_pending or (self.is_pending_approval and self.approved_by):
+            self.status = PurchaseOrderStatus.READY.value
+            self.save()
+
+            user = kwargs.pop('user', None)
+
+            purchaser_group_name = getSetting('PURCHASE_ORDER_PURCHASER_GROUP')
```
SchrodingersGat, line 767: Once you have merged master in please use the new `common.settings.get_global_setting` method here - `getSetting` has been deprecated

src/backend/InvenTree/order/models.py
```diff
@@ -599,13 +630,181 @@ def add_line_item(
...
         """
-        if self.is_pending:
+        purchaser_group = getSetting('PURCHASE_ORDER_PURCHASER_GROUP')
+        if self.is_pending or self.is_ready:
+            user = kwargs.pop('user', None)
+            if (
+                purchaser_group
+                and not user.groups.filter(name=purchaser_group).exists()
+            ):
+                raise PermissionDenied()
```
SchrodingersGat, line 804: Can we add a custom message to the `PermissionDenied` error?

src/backend/InvenTree/order/serializers.py
```diff
@@ -325,6 +336,99 @@ def save(self):
...
+        help_text=_('Message outlining the reason for rejecting the order'),
+        default='',
+        required=False,
+        allow_blank=True,
+    )
+
+    def save(self):
+        """Save the serializer to 'reject' the order."""
+        order = self.context['order']
+        reason = self.data['reject_reason']
```
SchrodingersGat, line 414: What happens here if 'reject_reason' is not supplied? Do you need to use `self.data.get('reject_reason', '')` instead?

src/backend/InvenTree/order/templates/order/order_base.html
SchrodingersGat: I would suggest that we look into custom permissions on the API endpoints, rather than requiring `add` permission here. 

Logically this is a "change" action, right? However the natural link for a "change" action is a `PUT` or `PATCH` method. But, the API views can have custom role requirements which override this default association.

src/backend/InvenTree/order/templates/order/order_base.html
```diff
@@ -75,10 +77,38 @@
         {% endif %}
     </ul>
 </div>
-{% if order.is_pending %}
+{% if order.is_pending and order.requires_approval %}
```
SchrodingersGat, line 80: This templating is a complex mess... not sure if there's a way around this though. Let's expedite the move to platform!

SchrodingersGat (commented):
@LavissaWoW first round review done with a few comments. I'd like to have another look when the conflicts are fixed. Looks reasonable, although I am concerned about the added complexity. It will definitely need to be "unchanged behaviour" for migrating users.

Additionally before merging this will need:

- Comprehensive documentation
- Platform UI testing (playwright)
- Further functional review

Depending on how long this all takes, we may get it in this release cycle. But I don't want it to go in right before release, it will need time for the "bleeding edge" users to test it.

src/backend/InvenTree/order/templates/order/order_base.html
LavissaWoW: Logically, it's a change action, yeah. As the functionality stands, it would tectnically be considered a bug when only possessing a Change role.

I can revert to "change" in the CUI template if altering API endpoints is more fruitful than changing this line. As illustrated above, trying to do state transitions while having only "change"-permissions in CUI will only lead to permission denied atm. The example is of one of the new states, but it goes for all states.

MY view was that then, using "add" more accurately reflects that you actually have the permissions required to click these buttons.

src/backend/InvenTree/order/serializers.py
LavissaWoW: I haven't tried not supplying it, actually. It defaults to the value of the field in both CUI and PUI. (empty string if no reason is typed in)
I'll attempt some pure API calls and see what happens

src/backend/InvenTree/order/models.py
LavissaWoW: Sure thing

src/backend/InvenTree/order/models.py
LavissaWoW: Will do

LavissaWoW:
@SchrodingersGat I agree that it's reasonable ho have it on bleeding-edge for a bit before commiting to a release. 
Any new settings are set to false by default, so there should be no change for migrating users. 

I'll get on with docs sometime this week, and look into Playwright testing.

src/backend/InvenTree/order/models.py
SchrodingersGat: @LavissaWoW for reference this will required https://github.com/inventree/InvenTree/pull/7404 to be merged in first

LavissaWoW:
@SchrodingersGat Playwright tests seem to change status from run to run, and it's getting to become quite frustrating...
The new tests pass locally, and they passed the QC check on the previous commit.

LavissaWoW:
Note: It's impossible with our current playwright config to run "global setting"-altering tests.
Even with `test.describe.configure({ mode: 'serial' })`, Chromium and Firefox tests will run in parallell.
In my ridiculous amount of variations of tests, it's clear that not even global file-scope is shared between workers.
There is literally no way to isolate Firefox and Chromium runs to prevent them from changing settings and messing up each others runtime.

LavissaWoW:
@SchrodingersGat I'm just gonna be blunt: The tests you want require me to write tests that do this:
```
while(setting_not_set(setting)) {
  sleep(1000)
  try_to_set_setting(setting)
}

await check_some_thing_depending_on_the_setting_hoping_that_the_setting_wasnt_changed_by_the_other_browser()
```

Unless I'm missing some huge element of Playwright, I'm putting my hair pulling frustrations on hold until I get some feedback.

SchrodingersGat:
@wolflu05 any thoughts on the tests that @LavissaWoW wants to run here? Maybe there's a way to reset states between test runs?

@LavissaWoW the difference between the frontend tests and backend (django) tests is that the underlying database is reset to a known state before every test run. I'm not sure how we could accomplish a similar thing here

LavissaWoW:
```
Running 8 tests using 2 workers

     1 [Chromium] PUI - Pages - Purchasing - Pending transitions
     2 [Firefox] PUI - Pages - Purchasing - Pending transitions
```

So it's not even considered "between" runs. They run in parallell, and read from the same test DB, at the same time. 
I'm at a loss here.

wolflu05:
Sorry, I have really not much experience with frontend testing. I do understand the problem, but the only thing I could think of is to spin up multiple inventree backend servers on different ports which are used for the different browsers. That way when each browser runs the tests sequentially, there is no race condition. And regarding the data reset after each test, maybe there is some possibility to reset the database to the migrated state after each test by using transactions or save points somehow that are send to the database.

SchrodingersGat:
@LavissaWoW perhaps if there is sufficient API testing via python unit tests, you can simplify the tests for the frontend, and do a "read only" set of playwright tests?

LavissaWoW:
@SchrodingersGat That would exclude the setting-based states.
One suggestion I see thrown around online is to split the Playwright testing into two steps where you run non-destructive tests in parallell, then afterwards run destructive tests with only 1 worker. This will force the problem away, but will mean modifying the QC workflow and file structure of the tests.

Basically this:
```
npx nyc playwright test tests/parallell
npx nyc playwright test tests/sequential --workers=1
```

LavissaWoW:
@SchrodingersGat Aight, here we go. PUI tests are working.

I've done a few things:

1. I've made a new directory `src/frontend/tests/destructive`. 
2. The default `playwright.config.ts` file now ignores the above directory
3. I made a new `playwright-destructive.config.ts` that only runs the tests in the directory in pt. 1. These tests are only run with 1 worker.
4. I modified the QC workflow to run the default config first, then run the destructive config. 
5. We have a lot of flakyness in our tests, so I increased the retry count in the playwright configs to 2.

Note: Codecov seems to be referencing a whole lot of stuff not related to this PR

LavissaWoW:
With working tests, I'm considering this ready for a final review

SchrodingersGat:
@LavissaWoW thanks, I'll hope to assign some time to review this in the next couple of days

LavissaWoW:
@SchrodingersGat Never heard back from you here.
Seems like some parts need a rewrite now tho.

SchrodingersGat:
not stale

LavissaWoW:
Not stale, indeed. Author's just a bit busy at the moment. Will resolve this soon

LavissaWoW:
Not stale

PR #6992, opened 2024-04-10
[CI] Re-add playwright reporting

matmair:
This re-adds the playwright reporting now that I have figured out why it was flaks (report config section missing).

Discoverd during #6987


matmair:
REady for review and merge

matmair:
Yer likely, you can download the HTML report after the tests fail an see in which state the UI was - this can help

PR #6993, opened 2024-04-10
[PUI] Fix licensing dialog

matmair:
This helps if the licensing for either platform is not available or empty; currently, the modal just crashes.

Discovered during #6987

### DIFF
--- a/src/frontend/src/components/modals/LicenseModal.tsx
+++ b/src/frontend/src/components/modals/LicenseModal.tsx
@@ -40,7 +40,7 @@ export function LicenceView(entries: Readonly<any[]>) {
         </Accordion>
       ) : (
         <Text>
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/modals/LicenseModal.tsx
```diff
@@ -38,6 +38,10 @@ export function LicenceView(entries: Readonly<any[]>) {
             </Accordion.Item>
           ))}
         </Accordion>
+      ) : (
+        <Text>
+          <Trans>No Information provided - this likely a server issue</Trans>
```
SchrodingersGat, line 43: Please change `this likely` to `this is likely`

SchrodingersGat:
Thanks @matmair

PR #6996, opened 2024-04-10
Improve DB efficiency for user settings

SchrodingersGat:
- Prefetch related user
- Prevents 1+N issue

### Before

image

### After

image


matmair:
Very nice! In theory we/you should get alerts about these kinds of errors in sentry - do they show up there?

SchrodingersGat:
I haven't seen sentry reporting anything like this.

PR #6998, opened 2024-04-11
Admin shell

SchrodingersGat:
Adds optional integration with [django-debug-shell](https://github.com/djk2/django-admin-shell)

- Allows optional shell access in development mode, for users with superuser access
- Requires `DEBUG` mode to be enabled (disabled for production)
- Requires environment var / configuration parameter to be set
- Package is *not* installed as part of default requirements (but is as part of dev-requirements)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 # Dummy requirements file to trigger the package pipeline
-# The new requirements file is located in src/backend/requirements.txt
+# The backend requirements file is located in src/backend/requirements.txt
 #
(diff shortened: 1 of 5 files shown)


matmair:
@SchrodingersGat we might want to include a check if the package is installed in the env and write a warning if not - so that there is no confusion why it is not functioning.

SchrodingersGat:
@matmair in such a case there is a very obvious warning:

image

IMO this is very much a developer-only feature, and you have to go through a lot of steps to actively enable it - so, one would assume that in such a case the provided warning is sufficient?

SchrodingersGat:
Never mind, added it.

Issue #6886, opened 2024-03-28
Batch Code from Build Order not pre-populated in New Build Output Form


simonkuehling:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When creating a Build Order, I can set a Batch Code in the form - which I thought is intended to be used to pre-populate the form for creating new Incomplete Build Outputs (is it?).

The batch code is not carried into the build output form right now. Is this a bug, or do I misunderstand the "Batch Code" field for a Build Order?

### Steps to Reproduce

1) Add a `Batch Code` to a Build Order:
![24-03-28 15-15-35 0107](https://github.com/inventree/InvenTree/assets/296454/0b62f4fb-711d-4a42-8651-e5f019fed653)

2) Create a new Build Output in that Build Order:
![24-03-28 15-15-37 0108](https://github.com/inventree/InvenTree/assets/296454/196ab6fa-7bdb-4db6-9130-e56ec6962998)

### Expected behaviour

I would expect that the batch code would be populated in the build output form.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.15.0 dev
Django Version: 4.2.10
Commit Hash: c827133
Commit Date: 2024-03-18

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.1.0-rpi7-rpi-v8-aarch64-with
Installer: DOC

[...]

Issue #6364, opened 2024-01-30
Generate batch code does not generate from PO receive line items


Elitearcanist:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When on a PO, receive line items (which generates new stock items) does not automatically generate batch codes.

This may be more of a missing feature but validation plugin function generate_batch_code feels incomplete without functionality in POs.

### Steps to Reproduce

Using a Inventree server with plugins enabled
1. enable a verification plugin, the sample verification plugin works
2. Go to a PO with unreceived line items
3. Click to receive all line items or receive an individual item
4. Click to add a batch code
5. A code is not auto-generated

### Expected behaviour

When a batch code is added to a line item in the receive menu it should autofill the entry with a code generated from the generate_batch_code function.

### Deployment Method

- [X] Docker
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.13.0 dev
Django Version: 3.2.20
Commit Hash: 01f2aa5
Commit Date: 2023-07-12
Commit Branch: master
Database: postgresql
Debug-Mode: True
Deployed using Docker: True
Platform: Linux-4.18.0-513.9.1.el8_9.x86_64-x86_64-with
Installer: DOC

Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.0.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}, {'name': 'EventPlugin', 'slug': 'sampleevent', 'version': None}, {'name': 'CustomValidator', 'slug': 'validator', 'version': '0.3.0'}, {'name': 'AVL Demo', 'slug': 'companyavldemo', 'version': '0.1.0'}]

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/inventree) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/inventree/InvenTree/issues/6364">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/inventree/InvenTree/issues/6364/pledge.svg?darkmode=1">
[...]

Issue #6098, opened 2023-12-15
Extend batch code generation


SchrodingersGat:
Add an API endpoint returning the "next" batch code:

- Allow batch code to be generated based on provided `POST` fields (e.g. 'part', 'location', etc)
- Expose to `ValidationPlugin` mixin (allowing custom implementation)
- Generate when receiving against purchase order
- Generate when creating against build order

Further notes:

- API Endpoint for generating a "serial number" for a new part, based on provided fields
- Also exposed to the plugin interface
- Auto updates in form fields based on selected part

#### Linked Issues

- https://github.com/inventree/InvenTree/issues/6364

SchrodingersGat:
This is closed in https://github.com/inventree/InvenTree/pull/7000

Note: Only implemented in the *new* interface (React / PUI)

PR #7000, opened 2024-04-11
Batch code generation

SchrodingersGat:
This PR implements a solution for https://github.com/inventree/InvenTree/issues/6098, namely allowing batch codes to be generated via a `POST` request to an API endpoint.

Any data provided to the request is made available to plugins, so a custom plugin could theoretically use this information for the purpose of batch code generation.

This also provides a solution for https://github.com/inventree/InvenTree/issues/6364 (at least in the new react interface), by adding a new `useGenerator` hook function that makes use of this new API endpoint.

### Related Issues

- Closes https://github.com/inventree/InvenTree/issues/6364
- Closes https://github.com/inventree/InvenTree/issues/6098
- Closes https://github.com/inventree/InvenTree/issues/6886

### Tasks

- [x] Maybe make the endpoint "generic" - allowing for future expansion on a single endpoint?
- [x] Extending form functionality to the "build order" forms
- [x] Auto generating batch codes wherever they are needed in the forms interface
- [x] Add a similar endpoint for generating serial numbers

### Further Work

Further work could extend on this by:

- [ ] Add a similar endpoint for generating internal part numbers
- [ ] Add a similar endpoint for generating order codes (purchase / sales / build / return)

generators for serial numbers and internal part numbers would hook into existing plugins, allowing these plugins to seamlessly inject data into the front-end forms dynamically


matmair (approved):
LGTM; it seems like the API schema changed though - could you bump the API version?

src/frontend/src/hooks/UseGenerator.tsx
```diff
@@ -0,0 +1,70 @@
...
+        setQuery({
+          ...query,
+          ...params
+        });
+      }
+
+      queryGenerator.refetch();
+    },
+    [query]
+  );
```
wolflu05, line 45: You could make the ref to the update function static if you would do something like this. This would have the advantage, that the useMemo from the form fields doesn't need to trigger all the time the query state updates.
```suggestion
  const update = useCallback(
    (params: Record<string, any>, overwrite?: boolean) => {
      if (overwrite) {
        setQuery(params);
      } else {
        setQuery((query) => ({
          ...query,
          ...params
        }));
      }

      queryGenerator.refetch();
    },
    []
  );
```

wolflu05 (approved):
LGTM

SchrodingersGat:
Bumping this to the next release cycle, still some work to do here.

PR #7002, opened 2024-04-11
[CI] Add nicer OpenAPI diffs

matmair:
Adds a diff for OpenApi to the schema check step to the step summary. This can be used to quickly evaluate the effective schema diffs.


matmair:
Sample: https://github.com/matmair/InvenTree/actions/runs/9860962099/attempts/1#summary-27228345488

matmair:
@SchrodingersGat @wolflu05 should the diff also be added as a comment?

SchrodingersGat:
As a comment in the PR, or where?

matmair:
As a PR comment. Not sure if it would be too noisy

wolflu05:
Do you mean this with the diff?
image

In what scenarios would that be helpful? Maybe if the user forgets to bump the api version? Or is that a different story?

SchrodingersGat:
I don't think we need it as a comment.

matmair:
@wolflu05 it helps to see the real effect on the schema without making the diff yourself. This example shows that only the info field changed so a new API revision is unneeded.
I find it hard to guestimate the real-world effects of changes in serializers and this diff helps to visualize the impact quickly.

wolflu05:
Would it be possible to please also provide an example for how this would look like if there was a bigger change in the api schema, e.g. an endpoint was added?

matmair:
@wolflu05 I have merged this with #7667 for a sample - here is the output https://github.com/matmair/InvenTree/actions/runs/9964062291/attempts/1#summary-27531527133

wolflu05:
ok, thank you very much for that example. That's quite a large amount of data for such a small change (adding 2 notes fields to two models), because this shows the changes very detailed. I don't think we need that as a comment too, as if the API schema is failing and we want to check the changes, we can easily jump into the Action output summary.

matmair:
The small code change effects the response on 21 Endpoints, most of the them both on POST and GET. So not really a small change if you look at the schema - that is exactly the value of this change, it allows for easier discovery of effected endpoints.

SchrodingersGat:
@matmair LMK when you are happy for me to merge this in then :)

matmair:
@SchrodingersGat yes, ready to merge

Issue #7005, opened 2024-04-12
[PUI] Focus on form fields


SchrodingersGat:
When opening a new form modal, I would like to be able to specify the field to focus on by default. If no field is specified, focus on the first visible field.

It looks like this should be possible with [react-form-hook setFocus](https://react-hook-form.com/docs/useform/setfocus) but I cannot get it working.

@wolflu05 do you have any insights here?

wolflu05:
We need to directly pass the ref to the input instead of destructuring it first (we have to do that for all form field types ignorer for it to work correctly):

```diff
          <TextInput
             {...reducedDefinition}
-            ref={ref}
+            ref={field.ref}
             id={fieldId}
             type={definition.field_type}
```

Then we can utilize the `setFocus` method to set focus after the initial render.

PR #7008, opened 2024-04-12
Form focus

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/7005
- Adds ability to focus on a particular field

### DIFF
--- a/src/frontend/src/components/forms/ApiForm.tsx
+++ b/src/frontend/src/components/forms/ApiForm.tsx
@@ -66,6 +66,7 @@ export interface ApiFormProps {
   pathParams?: PathParams;
   method?: 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE';
   fields?: ApiFormFieldSet;
+  focus?: string;
   initialData?: FieldValues;
   submitText?: string;
   submitColor?: string;
@@ -278,6 +279,10 @@ export function ApiForm({
 
   // Fetch initial data on form load
   useEffect(() => {
+    if (props.focus) {
+      form.setFocus(props.focus);
+    }
+
     // Fetch initial data if the fetchInitialData property is set
     if (props.fetchInitialData) {
       queryClient.removeQueries({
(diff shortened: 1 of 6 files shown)


SchrodingersGat:
@wolflu05 does not seem to be working, but perhaps I am calling `form.setFocus` in the wrong place?

wolflu05:
Have you tried initially loading the form and then just make a slight change to the code (e.g. add a console log) and then hit cmd s to trigger fast refresh? Then try to focus the browser window (don't click inside otherwise the focus is lost again), and see if the field is focused now?

I suspect that when this useEffect normally runs the field is not already registered. You could find that out by also logging form.getValues() before the setFocus call.

SchrodingersGat:
Is there a hook which runs only after the form has been opened / initially rendered?

wolflu05:
Have you tried adding the setFocus method to the `useEffects` dependencies, and also use a sperate effect?

```tsx
useEffect(() = {
  console.log("Here", form.getValues())
  form.setFocus(focus);
}, [form, focus]); // and also try form.setFocus directly
```

If nothing helps, the only idea I have is to schedule a timer, but that's not a great solution:

```tsx
useEffect(() = {
  const id = setTimeout(() => {
    console.log("Here", form.getValues())
    form.setFocus(focus);
  }, 1); // normally 1ms should work here as this then gets executed on the next tick after the first render, but make sure to also test some significantly longer values just to make sure (e.g. 1000). 
  return () => clearTimeout(id);
}, [form, focus]);
```

SchrodingersGat:
@wolflu05 thanks, seems to work pretty well now :)

wolflu05:
Im not sure if adding the fields to the useEffect dependencies has some side effects. E.g. in forms where a state is connected to a field'a value, the fields variable changes all the time and so entering a text in such a input field results in focusing the focus field after each keypress. (But I have not tested it)

SchrodingersGat:
The form focus should only happen once, I think. Once it is set it is not tested again. But let's keep an eye out for any issues

PR #7009, opened 2024-04-12
[PUI] Mobile ignore warning in dev

matmair:
This adds in dev enviroments the option to ignore the mobile warning and use PUI on a mobile screen anyway. Useful for debugging.

### DIFF
--- a/src/frontend/src/states/LocalState.tsx
+++ b/src/frontend/src/states/LocalState.tsx
@@ -33,6 +33,8 @@ interface LocalStateProps {
   addDetailDrawer: (value: number | false) => void;
   navigationOpen: boolean;
   setNavigationOpen: (value: boolean) => void;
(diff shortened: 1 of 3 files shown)


wolflu05:
Thank you very much for this change, this makes testing especially form the mobile phone (when one being away from the computer) much easier. I would also really appreciate it if there would be a dedicated option to also enable this on prod instances e.g. through the `window.INVENTREE_SETTINGS.ignore_mobile_check`.

matmair:
That is explicitly not a goal. PUI is not optimized in any way for mobile and breaks a lot (navigation can not be closed, drawers are hard to recognise, scrollbars are everywhere, modals break).
I am hard stop against enabling PUI on mobile till there is a covering and passing test suite for mobile formfactors and all CUI functions work in PUI.

PR #7022, opened 2024-04-14
[PUI] Add quick login via url

matmair:
This PR:
- Adds the url based login method (user + pwd) that we use for the demo to PUI
- Switches most tests (all that do not explicitly test user login UI) to use this method as it is easier to compute

### DIFF
--- a/src/frontend/src/pages/Auth/Login.tsx
+++ b/src/frontend/src/pages/Auth/Login.tsx
@@ -52,7 +52,6 @@ export default function Login() {
 
     // check if we got login params (login and password)
     if (searchParams.has('login') && searchParams.has('password')) {
-      console.log('got login params');
       doBasicLogin(
         searchParams.get('login') ?? '',
         searchParams.get('password') ?? ''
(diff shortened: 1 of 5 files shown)


matmair:
ready for review

src/frontend/src/pages/Auth/Login.tsx
```diff
@@ -48,6 +49,17 @@ export default function Login() {
     }
 
     checkLoginState(navigate, location?.state?.redirectFrom, true);
+
+    // check if we got login params (login and password)
+    if (searchParams.has('login') && searchParams.has('password')) {
+      console.log('got login params');
```
SchrodingersGat, line 55: Please remove this debug msg :)

matmair: I thought this might be helpful when debugging the demo instance but I have removed it now

matmair:
@SchrodingersGat this would be ready for review and handy as it cuts down on test run time

PR #7057, opened 2024-04-17
[PUI] Add playground / pricing tests

matmair:
This PR adds tests for the playground and pricing area


src/frontend/tests/defaults.ts
```diff
@@ -1,6 +1,6 @@
 export const classicUrl = 'http://127.0.0.1:8000';
 
-export const baseUrl = `${classicUrl}/platform`;
+export const baseUrl = `./platform`;
```
SchrodingersGat, line 3: @matmair I ran into issues using a relative URL like this - once you traverse down into lower URLs the resulting URL can become something like:

`/platform/parts/platform/stock`

matmair: using the classic endpoint means that there is no coverage so we will have to find a middle-way

matmair:
@SchrodingersGat I have witched back to a PUI base URL - there was no coverage when using PUI paths.

Issue #5882, opened 2023-11-08
Report Generation for new interface


SchrodingersGat:
- Support report generation in new interface (React)
- Enable label printing also
- Potentially refactor the existing label printing / report generation API endpoints (breaking)

SchrodingersGat:
Not stale

matmair:
Also see https://github.com/inventree/InvenTree/issues/5841

Issue #6417, opened 2024-02-05
Possible breaking changes for 1.0


matmair:
With somewhat soon (once #5212 is done) 1.0 approaching, we can start thinking about breaking changes we might want to implement to make long-term maintainability easier.

```[tasklist]
### Tasks
- [ ] https://github.com/inventree/InvenTree/issues/5015
- [ ] https://github.com/inventree/InvenTree/issues/5578
- [ ]  moving config files out of the code source
- [ ] do not require any envs on docker startup, assume local subfolder /data
- [ ] rename default python `env` to `.venv`
- [ ] respect AUTH_USER_MODEL
```

See also this discussion https://github.com/inventree/InvenTree/pull/6173#issuecomment-1928113213

matmair:
Thoughts @inventree/maintainer @inventree/triage

matmair:
Open to more ideas if others arise

matmair:
Another point of discussion: rename the latest branch from `master` to `main`. The latter becomes more and more the default and I think that it also is the more logical naming. Thoughts @SchrodingersGat?

SchrodingersGat:
> Another point of discussion: rename the latest branch from `master` to `main`. The latter becomes more and more the default and I think that it also is the more logical naming. Thoughts @SchrodingersGat?

Not opposed to that

SchrodingersGat:
@matmair what are your thoughts on this one? Do you want to complete any of the remaining points before we release `1.0.0`?

matmair:
I think it would be unwise to do any of them so shortly before release of a stable. We can always tackle them later.

PR #7074, opened 2024-04-21
Report printing refactor

SchrodingersGat:
This PR presents a major refactoring of how we handle report templates. In particular, a lot of model duplication and API endpoint duplication is removed. This is a precursor to supporting reports in the new interface, and potentially expanding the functionality of the reporting framework into the future.

Note that this is a *breaking* change as we are removing a number of models, and also deleting a number of existing API endpoints

### Related Issues

- Closes https://github.com/inventree/InvenTree/issues/5882
- Closes https://github.com/inventree/InvenTree/issues/5841
- Related to https://github.com/inventree/InvenTree/issues/6417

### Features / Updates

- Create new generic `ReportTemplate` database model
- Migrate all existing report template tables into this new one
- Reduce *many* report API endpoints to a single generic set

### TODO

- [x] Refactor label printing models / API in a similar fashion
- [x] Commonize the "report context" data across labels / reports
- [x] Allow template table(s) to be filtered by "model type"
- [x] Update unit testing
- [x] Add playwright testing for report and label printing via web interface
- [x] Update documentation
- [x] Pull "context" data directly into the documentation from the codebase
- [x] Test that all CUI report printing functionality continues to function correctly
- [x] Test that all CUI label printing functionality continues to function correctly
- [x] Implement report printing in new PUI interface
- [x] Implement label printing in new PUI interface
- [x] refactor template editor in PUI to use new API 
- [x] Update python bindings to support new API endpoints - https://github.com/inventree/inventree-python/pull/226
- [x] Update python bindings again now that we have changed the plugin API - https://github.com/inventree/inventree-python/pull/229
- [x] Update app to support new API endpoints - https://github.com/inventree/inventree-app/pull/489
- [x] Update app again now that we have changed the plugin API - https://github.com/inventree/inventree-app/pull/491
- [x] Refactor "include_installed" option for stock report
- [x] Refactor `generate_filename` method - ensure it works for reports and labels
- [x] Get label printing working again (label size is not observed)
- [x] Automatically add "default" label templates on launch
- [x] Automatically add "detault" report templates on launch
- [x] Ensure that both builtin "label printing" plugins still work
- [x] Reimplement / refactor API endpoints for "printing" labels and reports with proper serializers
- [x] Ensure backwards compatibility with existing label printer plugins
- [x] Perform full migration run to ensure old labels and templates are migrated correctly
- [x] Move templating db model mixin into `report.mixins`


matmair:
looks promising, I am still convinced that reports and labels should be the same thing but we have to start somewhere.

wolflu05:
> Looks promising! I am still convinced that reports and labels should be the same thing but we have to start somewhere.

I somehow have the same feeling. Especially if we encourage users to use report assets / snippets for labels.

And I have a few questions regarding this refactor:
1. what impact has this change on the PUI template editor?
2. are we still gonna save templates as files and not storing the content in the db?
3. is there an option to render a template without saving it first, but provide the template data in the request body?

SchrodingersGat:
> > Looks promising! I am still convinced that reports and labels should be the same thing but we have to start somewhere.
> 
> I somehow have the same feeling. Especially if we encourage users to use report assets / snippets for labels.
> 
> And I have a few questions regarding this refactor:
> 
> 1. what impact has this change on the PUI template editor?
> 2. are we still gonna save templates as files and not storing the content in the db?
> 3. is there an option to render a template without saving it first, but provide the template data in the request body?

1. The PUI template editor will still function - I'll just have to tweak the URLs a bit
2. No, at least not yet. We will have to consider the implementation and security implications, and I do not believe the library that @matmair linked is ready for django 4.2 yet?
3. No, but I guess we could make an API endpoint for previewing a template? Now that all templates are "the same" this would be a much simpler activity. I'd leave that for a separate PR though.

matmair:
I think we will have to create our own db-based template renderer; I can not find something that fits our thread model

wolflu05:
Just a thought I already have since a while: maybe we can build the new api in a way that we can also expose the current available context and their sub-/sub-/... attributes via an api endpoint which we can later use in the PUI template editor to provide auto completion with some description to the properties.

wolflu05:
Not sure of what the scope of this PR is, but maybe we can somehow get the request object out of the template rendering functions when doing this refactor. This would simplify things with workers much, as Pickeling the request object does not work. And IMHO having the request while rendering a template is not really necessary. The only thing I can think of is having the user that triggered the printing job, but we should rather pass that as a separate context argument. For more info see #6772

src/backend/InvenTree/templates/js/translated/label.js
```diff
@@ -56,145 +58,103 @@ function printLabels(options) {
...
-            contentType: "application/json",
-            dataType: "json",
-            accepts: { json: "application/json" },
-            async: false,
-            success: (res) => { printingOptionsRes = res },
-            error: (xhr) => showApiError(xhr, formOptions.url)
-        });
+    const baseFields = {
+        template: {},
+        plugin: {},
```
wolflu05, line 111: Is that right, that plugin is now expected to be a numeric id and not the plugin slug anymore? Is there still an option then to print with only the slug, or do I need two requests then? Because the slug is guessable by 3rd party extensions, the id not, that can be different between instances.

SchrodingersGat: The plugin needs to be specified by *id* now - this is so that we can use a foreignkeyrelatedfield on the serializer - and reference to the *pluginconfig* object - and expose the serialized fields to the documented API.

I think that the downside (have to use ID values rather than "slug" values) is acceptable because it simplifies the implementation and increases consistency...

wolflu05: Ok, so printing with the machine plugin or the pdf printer plugin from a third party tool, now requires a second request to first ask for the corresponding id if that plugin?

How have you handled this now in the template editor to preview the file? Is there now a shared endpoint with the reports to preview a label directly rather than needing to request a pdf from a printing plugin first and then making a second request to grab the file?

wolflu05:
May I ask you, if you could please add a quick explanation of how the new system works now compared to the old? Like how are the templates stored now, what endpoints are available and how the templates interfere with plugins?

wolflu05:
Thank you very much for your explanation to get a overview of this change. It really helped to understand the fundamentals.

> And you can filter here by model type:

1. Is that normal that in your screenshot the names are sometimes lower and sometimes uppercased?

2. I'm not sure if it's just me, but this filtering looks like bad UX to me, as:
  - a lot of clicks are required to filter for a specific template
  - by default all templates are shown
  - the filters are not persisted in the url query params, so sharing a link with a particular template selected is not possible
  Don't get me wrong, I don't say the previous method of filtering was the best, but it was really easy to navigate between the different template types.

---

Do I understand this right:
1. an API endpoint to post template code against for previewing is up for further work to enable a more "live" preview feeling?
2. labels can also be rendered as a pdf via the /label/print api endpoint now without using a plugin. And to that point labels are now completely similar to reports. The only difference between reports and labels is that reports don't have the plugin print integration?

SchrodingersGat:
> 1. Is that normal that in your screenshot the names are sometimes lower and sometimes uppercased?

That's due to the internal "verbose name" values of the database tables - I'll look into that

> 2. I'm not sure if it's just me, but this filtering looks like bad UX to me, as:

This is how table filtering works in PUI. I agree that it could be improved. Happy to review a PR :)

> an API endpoint to post template code against for previewing is up for further work to enable a more "live" preview feeling?

Yep, could now quite easily support a "preview" endpoint. Not in the scope of this refactor though.

> labels can also be rendered as a pdf via the /label/print api endpoint now without using a plugin. And to that point labels are now completely similar to reports. The only difference between reports and labels is that reports don't have the plugin print integration?

Correct.

matmair:
So is there something we can help with to get this over the finish line?

SchrodingersGat:
@matmair getting the unit tests updated to address the new API is the big blocker currently. Most of the other things are done!

matmair:
@SchrodingersGat schould we split that up? One does inventree-python, one inventree for example?

SchrodingersGat:
> @SchrodingersGat schould we split that up? One does inventree-python, one inventree for example?

I'd appreciate that! I've started work on the inventree-python tests - https://github.com/inventree/inventree-python/pull/226 - perhaps you want to tackle the outstanding printing unit tests here? :)

matmair:
@SchrodingersGat on it!

wolflu05:
One small question, what are revisions in context of these template models? How do they affect business logic, and how do they get bumped? For every typo fix in a template? Or is there a way to say when it should be a new revision?

SchrodingersGat:
@wolflu05 good question. As originally implemented the "revision" number increments on every save. However this is perhaps not the best approach? Open to suggestions, now would be the time to change if we wanted to.

SchrodingersGat:
Yes this sounds like it would be good for a future PR.

src/backend/InvenTree/report/api.py
```diff
@@ -1,164 +1,314 @@
...
+        if self.request:
+            plugin_key = self.request.query_params.get('plugin', None)
+            plugin = self.get_plugin_class(plugin_key)
+            plugin_serializer = self.get_plugin_serializer(plugin)
+
+            if plugin_serializer:
+                kwargs['plugin_serializer'] = plugin_serializer
+
+        serializer = super().get_serializer(*args, **kwargs)
+        return serializer
```
matmair, line 200: @SchrodingersGat are you sure this is working? I can not get PrintingOptionsSerializer  running in OPTIONS calls.

src/backend/InvenTree/report/api.py
```diff
@@ -1,164 +1,314 @@
...
+
+        return None
+
+    def get_serializer(self, *args, **kwargs):
+        """Return serializer information for the label print endpoint."""
+        plugin = None
+
+        # Plugin information provided?
+        if self.request:
+            plugin_key = self.request.query_params.get('plugin', None)
```
matmair, line 192: At least this needs to change but I am not sure the `get_plugin_class` fnc is really working as intended
```suggestion
            plugin_key = self.request.data.get('plugin', None)
```

src/backend/InvenTree/report/api.py
SchrodingersGat: I'll check again when I am back at my computer but yeah it has been working

matmair: The OPTIONS calls too? I have a very hard time getting the tests running consistently in this regard

matmair: I have found the issue, you can not reload/activate/deactivate plugins between determining the plugin pk and doing any call using it as it seems to change

src/backend/InvenTree/report/api.py
matmair: I will provide a better tested fix for this in the testing PR

src/backend/InvenTree/templates/js/translated/label.js
SchrodingersGat: Ok, so the plugin should be selected by slug, not by ID?

SchrodingersGat:
@matmair @wolflu05 regarding the requests to retain plugin lookup by "slug" (rather than "pk"): I am happy to do this, but I would suggest that we move all the plugin lookup in the API to use slug rather than PK.

e.g.  `/api/plugin/inventreebarcode/` - rather than `/api/plugin/17/`

The reason for this is to allow the printing endpoint to use the plugin slug for the foreign key field. For consistency I'd just roll that out across all the plugin endpoints.

If you are both OK with that, I'd submit that as a separate PR, and wait for that to be merged before continuing with this.

SchrodingersGat:
I am bumping this to the next release cycle, as there is still significant work to do here.

src/backend/InvenTree/report/models.py
```diff
@@ -120,531 +143,361 @@ def save(self, *args, **kwargs):
...
-        # TODO: Support custom filename generation!
-        # filename = kwargs.get('filename', 'report.pdf')
+        context = self.get_context(instance, request)
 
         # Render HTML template to PDF
-        wp = WeasyprintReportMixin(
+        wp = WeasyprintReport(
             request,
             self.template_name,
             base_url=request.build_absolute_uri('/'),
```
wolflu05, line 193: Does this really work if request is None? And do we really need the request object at all in here?

SchrodingersGat: @wolflu05 I'll have to look into this further but it does look like we can remove the request object entirely.

src/frontend/src/pages/Index/Settings/AdminCenter/Index.tsx
```diff
@@ -61,10 +65,6 @@ const CurrencyTable = Loadable(
   lazy(() => import('../../../../tables/settings/CurrencyTable'))
 );
 
-const TemplateManagementPanel = Loadable(
-  lazy(() => import('./TemplateManagementPanel'))
-);
```
wolflu05, line 66: Why have you removed the lazy loading here? This means that now when someone opens the admin center the complete huge chunk for the code editor is loaded too.

SchrodingersGat: This is because we now use two instances of the "template table" in the panel - and (as far as I can tell) you cannot use lazy importer for components you are passing props to? At least it does not seem to work that way...

wolflu05: I would suggest moving them both into one separate file and have this index route just for linking all sub-pages of the admin center together instead of defining some page logic in the Index.tsx of the admin center.

But why do you mean lazy should not work with props? That should work fine, we would just need to use generics [here](https://github.com/inventree/InvenTree/blob/afc66d4cc9a868e3e080c968fda525f383fe3b63/src/frontend/src/functions/loading.tsx#L19-L24) so that the props will be properly introspected for TS. But if you ignore the TS error, that lazy loading should work also with props.

src/backend/InvenTree/report/api.py
```diff
@@ -261,242 +407,39 @@ def print(self, request, items_to_print):
...
-                pdf_content = ContentFile(pdf, 'test_report.pdf')
-            except TemplateDoesNotExist:
-                return
-
-            StockItemAttachment.objects.create(
-                attachment=pdf_content,
-                stock_item=item,
-                user=request.user,
-                comment=_('Test report'),
-            )
```
matmair, line 320: @SchrodingersGat was this removal intentional? I can not find an obvious replacement in the new code

src/backend/InvenTree/plugin/builtin/labels/inventree_machine.py
```diff
@@ -112,8 +112,9 @@ def __init__(self, *args, **kwargs):
             super().__init__(*args, **kwargs)
 
             view = kwargs['context']['view']
-            template = view.get_object()
-            items_to_print = view.get_items()
+            # TODO @matmair Re-enable this when the need is clear
+            template = None  # view.get_object()
+            items_to_print = None  # view.get_items()
```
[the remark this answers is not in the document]
wolflu05: The idea is to give printing drivers the ability to decide wherever a printer would be able to print this job. E.g. 
- A printer which has only 10x10mm labels cannot print large 30x70mm labels, therefore it wouldn't make sense to show that printer to the user for selection 
*(plugins would need to add a settings field where one could enter the currently available label sizes in the printer, and then override the [`get_printers`](https://docs.inventree.org/en/stable/extend/machines/label_printer/#machine.machine_types.LabelPrinterBaseDriver.get_printers) function to then filter for printers that could print that size of template)*. 
- Same for the actual items, maybe it's not that likely to print with a printer in a different factory when printing a location label. But since it could happen, I have decided to let this one open for future implementation.

src/backend/InvenTree/report/api.py
matmair: @SchrodingersGat any insights on this?

SchrodingersGat:
@wolflu05 @matmair last remaining hurdle is the failing playwright test. The test runs successfully but gets stuck on "teardown". Any ideas? Really struggling on this one.

SchrodingersGat:
@matmair @wolflu05 thanks for your help on this one. Merging in now, we have some time before `0.16.0` to work out any issues which may rise from such a significant diff.

src/backend/InvenTree/plugin/builtin/labels/inventree_machine.py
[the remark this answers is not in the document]
wolflu05: @SchrodingersGat @matmair What happend with this?

matmair:
Should we maybe open a follow up issue with all the future ideas for this area that were voiced here?

SchrodingersGat:
> Should we maybe open a follow up issue with all the future ideas for this area that were voiced here?

Yes I think that's a great idea

src/backend/InvenTree/plugin/builtin/labels/inventree_machine.py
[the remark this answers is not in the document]
matmair: I do not think this was adressed

src/backend/InvenTree/report/api.py
matmair: @SchrodingersGat was this intentional? I will add it to the open issues if not

tristanle22:
Hello, any chance of this update being included in the near future releases, along with #9532?

SchrodingersGat:
@tristanle22 we are working hard on the final issues before our next release - not many to go now: https://github.com/inventree/InvenTree/milestone/17

All merged features will be available as part of that release.

tristanle22:
Great!

PR #7081, opened 2024-04-22
Pin hashes in requirements 

matmair:
This PR introduces requirement hash pinning. This secures against supply chain attacks in which dependencies are replaced with malicious versions.
It also switches workflows over to use a requirement files.


SchrodingersGat:
Why does each package get multiple hashes in the requirements file?

matmair:
@SchrodingersGat it dependes on how many files/distributions are offered, this can be seen on Pypi itself too:https://pypi.org/project/certifi/#copy-hash-modal-205cb7d5-a79a-4f73-aeb7-f5c1c778db8c

matmair:
Most packages nowaday release 2 files: a source tarball and a wheel (that is a pre-build distribution format)

SchrodingersGat:
Ok! Please fix the conflict and I'll merge straight away

matmair:
@SchrodingersGat mergefix is in

PR #7111, opened 2024-04-23
[CI] Ignore migrations in backend reporting

matmair:
Removes information on migrations from coverage - leaving them to the dedicated migration tests.


matmair:
Adressing this later

PR #7113, opened 2024-04-23
[PUI] Test modals

matmair:
Adds tests for general PUI modals (AboutInstance, License Info, About org, QR modal) and fixes a key error in the license modal.


matmair:
@SchrodingersGat this adds a bit better coverage on modals

PR #7114, opened 2024-04-23
[PUI] Test for errors in console

matmair:
This PR adds a fixture to the base test to ensures that not errors are raised on the console.

It also fixes a few common errors:
- missing keys
- duplicate keys
- unstable keys (re-renders)
- wrong nesting of objects (a td directly within a table)

A few errors are ignored for reporting:
- errors that start with `ERR:` - those are mainly fetching actions, which are taken at the start of a user sessions and might still be unauthenticated
- chrome internal errors
- API calls to the barcode API - which is called with failing commands in the tests


matmair:
@SchrodingersGat this is ready for review; the non-covered code from the patch would require a lengthy new test suite so I would like to cover this at a later point

SchrodingersGat:
Thanks @matmair this is a great addition to the testing framework

matmair:
@SchrodingersGat trying to get PUI testing baseline build out as hard as I can ;-)

PR #7119, opened 2024-04-25
[FR] Increase Length of URL limit on import

### LINKED ISSUE #7118: [FR] Increase Length of URL limit on import

### Problem statement

See [Discussion 7078](https://github.com/inventree/InvenTree/discussions/7078)

### Suggested solution

Possible options:
- Increase length of URL limit to something greater than 200 (Maybe 300?).
- Make the maximum length adjustable in the Admin settings somewhere.
- Remove the limit and replace it with content validation that validates and then accepts the link regardless of length if it is valid.

Not much I can do here... If the supplier's or manufacturer's website requires a link longer than 200 characters to link to the part on their site, then the limit isn't high enough.  The only other option would be to run our own URL shortening service and manually shorten any link that's too long on a one-off basis.

### Examples of other systems

_No response_

- [ ] I want to develop this.

matmair:
This increaes the max_length for links to 200 chars.
Why 2k? It seems reasonably long enough, the longest I could find on alibaba (not that I would use them as a supplier) is about 800ish - double it to be safe and this should be long enough.

Why not DB controllable? That would introduce a new surface to test for and make assumptions about the data model hard, which for me outweighs the possible gains.
Why not no limit? Opens us up to Denial of Service type of problems, we had that in the past and got according feedback from security researchers (see our Security Advisories).

Fixes #7118

### DIFF
--- a/src/backend/InvenTree/company/models.py
+++ b/src/backend/InvenTree/company/models.py
@@ -513,6 +513,7 @@ class ManufacturerPart(
         null=True,
         verbose_name=_('Link'),
         help_text=_('URL for external manufacturer part link'),
+        max_length=2000,
     )
 
     description = models.CharField(
@@ -829,6 +830,7 @@ class SupplierPart(
         null=True,
         verbose_name=_('Link'),
         help_text=_('URL for external supplier part link'),
+        max_length=2000,
     )
 
     description = models.CharField(


SchrodingersGat:
Should we expand this to *all* URL fields?

matmair:
We can but not sure it is required as there have been no complaints regarding the other fields

SchrodingersGat:
No harm in adding this in now, I suppose.

PR #7121, opened 2024-04-26
Fix CI filters

matmair:
I seem to have overlooked this in the refactor


SchrodingersGat:
Not sure how we missed this one! Thanks for the fix

PR #7112, opened 2024-04-23
Fixes to make compatible with Python 3.12

matmair:
This PR:
- Changes the code base to work with python 3.12
- Adds a matrix to the coverage tests to make sure upper and lower bounds are covered

Closes https://github.com/inventree/InvenTree/issues/6424
Contains #5777

PR #7126, opened 2024-04-28
CI: Add pre-commit hook to use modern testing asserts

matmair:
During https://github.com/inventree/InvenTree/pull/7112 is stumbled on the usage of a few depreceated testing assertations syntaxes.

This PR adds a pre-commit hook to ensure modern syntax is used. This change offers 2 benefits:
- make upgrades to newer versions easier. unittest is part of the standardlib and interfaces might change with major upgrades - this hook ensures we should not be blindsighted by them
- using more specific equals enables error messages to be more specific. For example changing
```python
        self.assertTrue(response.status_code in [401, 403])
```
to
```python
        self.assertIn(response.status_code, [401, 403])
```
Will throw an error containing the error_code value that was not found in the list instead of just False.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,3 +84,7 @@ repos:
 #    rev: 3.0.0
 #    hooks:
 #    -   id: shellcheck
+-   repo: https://github.com/isidentical/teyit
+    rev: 0.4.3
+    hooks:
+    -   id: teyit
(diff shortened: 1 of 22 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@matmair much appreciated 😸

Issue #7123, opened 2024-04-26
[FR] Ukrainian language support


ybilopolov:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

Would be great to have Ukrainian translation for the app.

### Suggested solution

Author would willingly commit into the process of translation, however Ukrainian is not even listed in the list of languages, available for translation in crowdin.

### Describe alternatives you've considered

There is no clear alternative for potential non-English-speaking Ukrainian users.

### Do you want to develop this?

- [X] I want to develop this.

matmair:
We do not support ukrainian currently as there are no translators for it. I have enabled support on [our translation](https://crowdin.com/project/inventree/uk) platform, please head over there and submit your translations.

PR #7129, opened 2024-04-28
Add ukrainian as a language option

matmair:
Closes https://github.com/inventree/InvenTree/issues/7123

This adds ukrainain as a language option, although the translations are obviously missing.


SchrodingersGat:
@matmair I believe you also need to add a `.d.ts` file in the frontend locale directory

matmair:
@SchrodingersGat file is added

SchrodingersGat:
Thanks @matmair

PR #7147, opened 2024-05-01
Add field description to the currency_exchange_retrieve API call

martonmiklos:
While poking around with the OpenAPI codegen I ended up generating blank response from the currency/exchange endpoint. 

I fixed it with a custom serializer with the knowledge recently learned from wolflu05.

kép

### DIFF
--- a/src/backend/InvenTree/common/api.py
+++ b/src/backend/InvenTree/common/api.py
@@ -127,6 +127,7 @@ class CurrencyExchangeView(APIView):
     permission_classes = [permissions.IsAuthenticated]
     serializer_class = None
 
+    @extend_schema(responses={200: common.serializers.CurrencyExchangeSerializer()})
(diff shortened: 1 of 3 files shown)


src/backend/InvenTree/common/api.py
```diff
@@ -127,6 +127,7 @@ class CurrencyExchangeView(APIView):
     permission_classes = [permissions.IsAuthenticated]
     serializer_class = None
 
+    @extend_schema(responses={200: common.serializers.CurrencyExchangeSerializer()})
```
matmair, line 130: Question: Are you sure we need ti intialize the serializer here?

matmair (approved):
LGTM

SchrodingersGat:
@martonmiklos thanks for the fix :)

Issue #7149, opened 2024-05-02
Stock Level (from Remove Stock option) Displays Stock #


imager2019:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

From the Stock menu option and selecting the item(s) and Remove Stock option. The Stock level displays a # sign
However, some other products (noticed today with a new recently added item), displays the correct stock level.
![stock-level-issue](https://github.com/inventree/InvenTree/assets/49574724/2031b25f-54b2-4346-a782-d66293c0b349)

### Steps to Reproduce

From Stock menu option
Select item(s)
Select Remove Stock option
Display will show items along with current stock levels, or # if it cannot display

### Expected behaviour

Expected to have stock level displayed.

Some products (seems to be newer added items) have the correct stock level displayed.

### Deployment Method

- [ ] Docker
- [X] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

# Version Information:
InvenTree-Version: 0.14.5
Django Version: 4.2.11
Commit Hash: b32f081a
Commit Date: None
Commit Branch: stable
Database: sqlite3
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-5.15.0-105-generic-x86_64-with-glibc2.31
Installer: PKG
Target: ubuntu:20.04
[...]

PR #7150, opened 2024-05-03
Stock transfer dialog fix

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/7149

### DIFF
--- a/src/frontend/src/forms/StockForms.tsx
+++ b/src/frontend/src/forms/StockForms.tsx
@@ -284,10 +284,10 @@ function StockOperationsRow({
   };
 
   const stockString: string = useMemo(() => {
-    if (!!record.serial) {
-      return `#${record.serial}`;
-    } else {
+    if (!record.serial) {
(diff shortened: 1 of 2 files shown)


src/frontend/src/forms/StockForms.tsx
```diff
@@ -283,6 +283,14 @@ function StockOperationsRow({
     input.removeFn(input.idx);
   };
 
+  const stockString: string = useMemo(() => {
+    if (!record.serial) {
+      return `${record.quantity}`;
+    } else {
+      return `#${record.serial}`;
+    }
+  }, record);
```
wolflu05, line 292: @SchrodingersGat This should be:
```suggestion
  }, [record]);
```

Doesn't ts detect that issue?

SchrodingersGat: @wolflu05 nice catch as always - fix incoming in https://github.com/inventree/InvenTree/pull/7151

PR #7154, opened 2024-05-03
Performance improvement for notifications API endpoint

SchrodingersGat:
- URL: http://localhost:8000/api/notifications/?format=json

### Before

image

### After

image


SchrodingersGat:
Ah, this has potential but needs some more work. The actual serialized content varies a lot based on the "target" and "source" content types so the prefetch does not always work...

Might be a good idea to add a "link" field to the notification model so we can cache this data, and not have to do (expensive) lookup.

PR #7158, opened 2024-05-04
Refactor login state management

SchrodingersGat:
- Previously relied only on presence of cookie
- Cookie may not actually be *valid*
- Inspect actual login state by looking at userState values
- Ensures better sequencing of global state API requests
- Login state is now correctly preserved across browsers

### DIFF
--- a/src/frontend/src/components/items/MenuLinks.tsx
+++ b/src/frontend/src/components/items/MenuLinks.tsx
@@ -51,11 +51,11 @@ export function MenuLinks({
   highlighted?: boolean;
 }) {
   const { classes } = InvenTreeStyle();
-  highlighted = highlighted ?? false;
 
   const filteredLinks = links.filter(
     (item) => !highlighted || item.highlight === true
   );
+
   return (
     <SimpleGrid cols={2} spacing={0}>
       {filteredLinks.map((item) => (
--- a/src/frontend/src/components/items/MenuLinks.tsx
+++ b/src/frontend/src/components/items/MenuLinks.tsx
@@ -45,13 +45,13 @@ function ConditionalDocTooltip({
 
 export function MenuLinks({
   links,
-  highlighted
+  highlighted = false
 }: {
   links: MenuLinkItem[];
   highlighted?: boolean;
 }) {
   const { classes } = InvenTreeStyle();
-  highlighted = highlighted || false;
+  highlighted = highlighted ?? false;
 
   const filteredLinks = links.filter(
     (item) => !highlighted || item.highlight === true
--- a/src/frontend/src/functions/auth.tsx
+++ b/src/frontend/src/functions/auth.tsx
@@ -169,20 +169,7 @@ export const checkLoginState = async (
 
   // Not yet logged in, but we might have a valid session cookie
   // Attempt to login
-  let result: boolean = false;
-
-  await api
-    .get(apiUrl(ApiEndpoints.user_me))
-    .then((response: any) => {
-      if (response.status == 200) {
-        result = true;
-      }
-    })
-    .catch(() => {});
-
-  if (result) {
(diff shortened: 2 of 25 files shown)


src/frontend/src/functions/auth.tsx
```diff
@@ -150,24 +161,35 @@ export function checkLoginState(
...
+    .then((response: any) => {
+      if (response.status == 200) {
+        result = true;
+      }
+    })
+    .catch(() => {});
+
+  if (result) {
+    await fetchUserState();
+  }
```
matmair, line 185: Question: Can we just await fetchUserState? Seems like we will have 2 requests to user_me this way

src/frontend/src/views/DesktopAppView.tsx
```diff
@@ -17,7 +16,7 @@ import { useUserState } from '../states/UserState';
 
 export default function DesktopAppView() {
   const [hostList] = useLocalState((state) => [state.hostList]);
-  const [fetchUserState] = useUserState((state) => [state.fetchUserState]);
+  const { fetchUserState, isLoggedIn } = useUserState();
```
matmair, line 19:
```suggestion
  const [fetchUserState, isLoggedIn] = useUserState((state) => [state.fetchUserState, state.isLoggedIn]);
```

matmair:
It seems not all of the new login flow is covered by tests, I think that should be a goal so we do not inadvertently break anything.

src/frontend/src/functions/auth.tsx
SchrodingersGat: @matmair good call, I think you are correct. I'll make the change.

matmair (approved):

matmair (approved):

SchrodingersGat:
Ok, can confirm that this now works under the following arrangements:

- Hosted locally, vite dev server
- Netlify preview against demo server (cross site)
- Netlify preview against localhost (cross site)

SchrodingersGat:
@matmair any insight into the playwright tests here? They are all passing locally for me.

src/frontend/tests/baseFixtures.ts
```diff
@@ -64,7 +64,8 @@ export const test = baseTest.extend({
         url != 'http://localhost:8000/api/barcode/' &&
         url != 'http://localhost:8000/api/news/?search=&offset=0&limit=25' &&
         url != 'https://docs.inventree.org/en/versions.json' &&
-        !url.startsWith('chrome://')
+        !url.startsWith('chrome://') &&
+        !url.startsWith('http://localhost:5173/node_modules/.vite/')
```
matmair, line 68: That is not a good idea IMO. We will miss a lot of real errors that way. Those are also raised to the user

SchrodingersGat: @matmair fair call. It didn't even fix the errors I am seeing...

src/frontend/src/components/items/MenuLinks.tsx
```diff
@@ -45,13 +45,13 @@ function ConditionalDocTooltip({
...
   links,
-  highlighted
+  highlighted = false
 }: {
   links: MenuLinkItem[];
   highlighted?: boolean;
 }) {
   const { classes } = InvenTreeStyle();
-  highlighted = highlighted || false;
+  highlighted = highlighted ?? false;
```
wolflu05, line 54: Mutating props is a really bad practice, props are immutable in react. This should be a second variable, and if used inside a dependency array it needs to be wrapped in a useMemo. Basically this could also be removed, as the default value is already assigned at the function definition.

SchrodingersGat: @wolflu05 I have simplified this now, thanks for the insight

PR #7164, opened 2024-05-05
Added test statistics

### LINKED ISSUE #5995: [FR] Test statistics

martonmiklos:
Fixed #5995

This PR adds test statistics views to the part and build pages which provides overview about the accumulated count of the passed/failed tests.


SchrodingersGat:
@martonmiklos can you please merge in the latest changes from master, and then I can review :)

martonmiklos:
> @martonmiklos can you please merge in the latest changes from master, and then I can review :)

Done. There are two things what might need to be looked into: as far as I saw filtering for datetime fields is not implemented in the PUI yet, so the date filtering is only available in the old UI.

I have not managed to hide the pagination (it does not makes any sense as we will always have 3 rows). 
I saw the enablePagination property, but it does not hide the pagination bar. 
Maybe I will report it to upstream if there is no trick to do that.

SchrodingersGat:
@martonmiklos where are you at with this one? :)

martonmiklos:
> @martonmiklos where are you at with this one? :)

@SchrodingersGat  could you please help how to properly specify links with two parameters here:
https://github.com/martonmiklos/InvenTree/blob/a7cafc792e46c0e09441431126b4d3200328c983/src/backend/InvenTree/build/templates/build/build_base.html#L304
It works, but tests are failing, and I have not found any examples in the existing code.

SchrodingersGat:
I would suggest that you make a "generic" test statistics endpoint, and rather than pinning it against specific database instances, pass those as query parameters you can filter by.

Then you can have a single API endpoint and you can filter by:

- Stock Item
- Part
- Build Order
- etc

matmair:
Hi there @martonmiklos!
1) Is this done / ready for final review?
2) Please stop force pushing/rebasing, this makes reviewing useless

martonmiklos:
> Hi there @martonmiklos!
> 
>     1. Is this done / ready for final review?
> 
>     2. Please stop force pushing/rebasing, this makes reviewing useless

Hi @matmair 
I tried to get the CI passing, but now I have no idea why (or is it related to my changes) the SQlite failed.

SchrodingersGat:
@martonmiklos the sqlite test is currently broken for python 3.12, we are still struggling to understand this one. Are you happy with the PR otherwise?

SchrodingersGat:
Please fix the conflicts here and if you're satisfied with this I'm happy to merge it in

SchrodingersGat:
@martonmiklos apologies but it looks like there are still some conflicts. If you can address this in the next couple of days, I'll hold off merging anything else so that we can get this one done :)

martonmiklos:
> @martonmiklos apologies but it looks like there are still some conflicts. If you can address this in the next couple of days, I'll hold off merging anything else so that we can get this one done :)

Done!

SchrodingersGat:
@martonmiklos a number of the frontend tests (playwright) are failing here - are they OK locally for you?

martonmiklos:
> @martonmiklos a number of the frontend tests (playwright) are failing here - are they OK locally for you?

I managed to get the PUI locally and discovered (together with an another issue in the old UI) that I messed up the API urls for PUI. Now it works locally.

SchrodingersGat:
Ok, something is going on with the PUI tests, other PRs are failing too, when they should not be. I will have to look into this tomorrow

SchrodingersGat:
@martonmiklos thanks for the hard work here :) I Have fixed the other PUI tests so this is good to go now. Thanks for the ongoing contributions!

SchrodingersGat:
@martonmiklos when testing this on the "build order" page (for a part with only a small number of tests / results), there are a large number of queries running:

image

image

I'll open a new issue to address this, I would appreciate it if you could look into optimizing the API / database queries.

PR #7184, opened 2024-05-08
Fix for details links

SchrodingersGat:
- Do not open as a new link
- Instead, use internal 'navigate'
- Otherwise, triggers a login sequence again
- Major improvement in workflow


matmair (approved):
LGTM

Issue #7193, opened 2024-05-09
About box broken link and missing email settings point to a broken link in docs


Petrox:
### Body of the issue

I know these are independent issues, but I don't want to spam the issuetracker:

1.
The https://docs.inventree.org/en/0.14.5/app/app/ link in the about dialog is a 404:
![image](https://github.com/inventree/InvenTree/assets/3036509/835b7719-fbe0-4f39-b1c4-1a74d9ca8e81)

Should probably be: https://docs.inventree.org/en/stable/app/app/   (the mobile app is quite independent from the server side so 99.9% of users will be using the latest stable mobile app but 80% of them won't use the latest server version anyways)

2.

Shall someone install inventree and then not set up the email settings causes inventree to display a notification in the system info dialog:

![image](https://github.com/inventree/InvenTree/assets/3036509/2cc8524e-2301-47e0-b2f4-6624f54c4472)

The "Email settings not configured" link points to a 404:

![image](https://github.com/inventree/InvenTree/assets/3036509/86a17b51-999b-47c1-a424-2d54daef82e5)

Should be https://docs.inventree.org/en/stable/start/config/#email-settings

Petrox:
I have found "latest/app/app" at multiple places in code and docs, so I'll go with that.

Petrox:
https://docs.inventree.org/en/latest/settings/email/ is there, but https://docs.inventree.org/en/0.14.5/settings/email/ is 404. 
I have checked and the files in docs/settings/email are one year old, so it's not the fault of the app, but somehow the generated documentation on the website seems to be holey.

PR #7195, opened 2024-05-10
Fixed the app/app link to point to the latest app version. #7193 

Petrox:
This is a oneliner HREF change

Closes https://github.com/inventree/InvenTree/issues/7193

### DIFF
--- a/src/backend/InvenTree/InvenTree/version.py
+++ b/src/backend/InvenTree/InvenTree/version.py
@@ -104,7 +104,7 @@ def inventreeDocUrl():
 
 def inventreeAppUrl():
     """Return URL for InvenTree app site."""
-    return f'{inventreeDocUrl()}/app/app/'
+    return f'https://docs.inventree.org/en/latest/app/app/'
 
 
 def inventreeCreditsUrl():


SchrodingersGat:
@Petrox thanks for the fix.

FYI the underlying issue was that the `0.14.5` docs build was not live - I have now fixed that.

However I agree that the "app" docs link should always point to stable as that is not pinned to a particular server version.

SchrodingersGat (approved):

Petrox:
> However I agree that the "app" docs link should always point to stable as that is not pinned to a particular server version.
Note: since there were multiple links pointing to **LATEST/app/app** I used that instead of the stable I mentioned first.

matmair:
@SchrodingersGat you sure this makes sense? Linking to `latest` means any change to the docs structure in the future will break any number of deployed versions of InvenTree. Docs and core code are in the same repo so the potential for them to drift apart is far less likely when being referenced stably compared to pointing to the very latest version of docs.

SchrodingersGat:
@matmair on second thoughts I think you are right - if (for whatever reason) we change the docs structure then this will break things. 

@Petrox note that I have addressed the upstream issue (in the documentation host), so I will revert this fix.

Petrox:
No problems.

Btw linking to stable of the app is technically more accurate than linking
to the specific version. Since the mobile app is up to date or one version
behind the stable latest published version, while the inventree instances
are usually not updated every weekend. (I would guess)

And just fyi the codebase has links to the latest/app/app already (and
still).

May I help clean those up?

SchrodingersGat:
> And just fyi the codebase has links to the latest/app/app already (and
still).

Can you link me to where those are?

I have been investigating another solution - adding static redirects on the documentation server for topics which are not linked to a particular "version" of InvenTree backend:

e.g.

- https://docs.inventree.org/app
- https://docs.inventree.org/python

These redirect to the "stable" docs at the appropriate page - and can be adjusted at any time if we move the page locations.

This way the docs links in the code can be kept static (for these items). Thoughts?

PR #7202, opened 2024-05-11
Add location type to location overview

SergeoLacruz:


wolflu05:
This is a nice addition, totally forgot the details pages when adding the location type feature.

Maybe it's a good idea to also add this addition to PUI.

SchrodingersGat:
@SergeoLacruz a good suggestion by @wolflu05 - can you add this to the new interface also?

SergeoLacruz:
Good Idea but...
It looks simple but changes to LocationDetail.tsx do not result in any change. I have seen that there are invoke tasks like frontend-compile, frontend-build, ... but neither does anything except showing errors like No such file or directory: 'install', yarn: error: no such option: --emptyOutDir ...
It took me quite some time to understand how the actual UI works. Actually I do not have the time and the need to dig into a complete new system. There are more important tasks on the list in the area of supplier management, connection to CAD, ...
It will take some time before I can come back to this.

matmair:
No worries, it is a completely different system

wolflu05 (approved):
I have added the suggested addition now in #7238, so this PR LGTM now.

SchrodingersGat:
@SergeoLacruz thanks for the contribution :)

PR #7207, opened 2024-05-13
Nginx example

SchrodingersGat:
Re-introduce sample file for running behind nginx proxy.

### DIFF
--- a/contrib/container/Caddyfile
+++ b/contrib/container/Caddyfile
@@ -1,7 +1,10 @@
-# Caddyfile for Inventree
+# Example Caddyfile for Inventree
 # The following environment variables may be used:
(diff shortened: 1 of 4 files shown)


SchrodingersGat:
@matmair happy to do that for sure. Where is the nginx file for the installer?

matmair:
@SchrodingersGat can we just link to https://github.com/matmair/InvenTree/blob/05c2fe93ec99b4e846b00266aa00dead7ce054e1/contrib/container/docker-compose.yml#L135-L136 instead of serving the same file 2 in the repo?

SchrodingersGat:
@matmair sorry I don't follow your example here - where is the second copy of the nginx conf? Do you mean `/contrib/packager.io/nginx.prod.conf/`?

PR #7224, opened 2024-05-14
Plugin API lookup key

SchrodingersGat:
Changes the plugin API endpoints to use the 'key' ('slug') rather than primary key for lookup.

This allows plugin configs to be accessed via repeatable, known values rather than pk values.

- Ref: https://github.com/inventree/InvenTree/pull/7074
- Ref: https://github.com/inventree/inventree-python/pull/227


src/backend/InvenTree/InvenTree/api.py
```diff
@@ -546,10 +546,18 @@ class MetadataView(RetrieveUpdateAPI):
 
     MODEL_REF = 'model'
 
+    def __init__(self, *args, **kwargs):
+        """Initialize the MetadataView."""
+        super().__init__(*args, **kwargs)
+
```
matmair, line 552: Question: Is this required? Seems to do nothing on a skim

src/backend/InvenTree/plugin/api.py
```diff
@@ -352,21 +355,21 @@ def get_object(self):
...
         Both the 'slug' and 'key' must be valid, else a 404 error is raised
         """
-        key = self.kwargs['key']
+        setting_key = self.kwargs['setting']
 
         # Look up plugin
-        plugin = check_plugin(
-            plugin_slug=self.kwargs.get('plugin'), plugin_pk=self.kwargs.get('pk')
-        )
+        plugin = check_plugin(self.kwargs.pop('key', None), None)
```
matmair, line 361: Question: Do we need to check for existence of `key` and raise a ValidationError?

src/backend/InvenTree/plugin/api.py
```diff
@@ -397,7 +400,8 @@ def get(self, request):
                     })
 
         result = PluginSerializers.PluginRegistryStatusSerializer({
-            'registry_errors': error_list
+            'registry_errors': error_list,
+            'active_plugins': PluginConfig.objects.filter(active=True).count(),
```
matmair, line 404: Comment: Nice add

matmair (approved):
Very clean diff, nice!

matmair:
Seems like a few tests need to be adapted to use key instead of pk to resolve urls

src/backend/InvenTree/plugin/api.py
wolflu05: Definitely, this was my plan on the longer run to add more status info here about plugins. And that's also why I added no dedicated errors endpoint, because I thought we can later extend this.

src/backend/InvenTree/InvenTree/api.py
SchrodingersGat: Thanks, was a stub from an experiment that is no longer needed - I will fix it.

src/backend/InvenTree/plugin/api.py
SchrodingersGat: The `check_plugin` method already does this:

image

matmair: ok I somehow got it to not trigger with my tests, will investigate that further; Only 2 tests two fix left: https://github.com/matmair/InvenTree/actions/runs/9086781305/job/24973262465

Issue #7227, opened 2024-05-14
Updating to 0.15.0 unable to execute invoke update (docker)


LastEdit:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [X] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

I tried updating from 0.14.5 to 0.15.0 in container using the standard commands. Unable to complete

### Steps to Reproduce

1. docker compose pull
2. docker compose run --rm inventree-server invoke update
3. Error

### Relevant log output

```bash
[+] Creating 1/0
 ✔ Container inventree-db  Running                                         0.0s 
Loading config file : /home/inventree/data/config.yaml
Installing required python packages from 'src/backend/requirements.txt'
Requirement already satisfied: pip in /usr/local/lib/python3.11/site-packages (24.0)
Requirement already satisfied: setuptools in /root/.local/lib/python3.11/site-packages (69.5.1)
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'src/backend/requirements.txt'


docker compose run inventree-server sh

[+] Creating 1/0
 ✔ Container inventree-db  Running                                         0.0s 
Loading config file : /home/inventree/data/config.yaml
/home/inventree # ls
base_requirements.txt      install_build_packages.sh
data                       requirements.txt
gunicorn.conf.py           src
init.sh                    tasks.py
/home/inventree # cd src
/home/inventree/src # ls
backend
/home/inventree/src # cd backend
/home/inventree/src/backend # ls
InvenTree
/home/inventree/src/backend # cd InvenTree/
/home/inventree/src/backend/InvenTree # ls
InvenTree             label                 report
_testfolder           locale                script
build                 machine               stock
common                manage.py             templates
company               order                 users
config_template.yaml  part                  web
generic               plugin
gunicorn.conf.py      plugins


```

SchrodingersGat:
@LastEdit the new docker image is building and should be available in the next few hours

PR #7228, opened 2024-05-14
Docker fix

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/7227
- Fixes missing file in docker image

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -85,9 +85,9 @@ jobs:
           docker run --rm inventree-test invoke --list
           docker run --rm inventree-test gunicorn --version
           docker run --rm inventree-test pg_dump --version
-          docker run --rm inventree-test test -f /home/inventere/init.sh
-          docker run --rm inventree-test test -f /home/inventere/tasks.py
-          docker run --rm inventree-test test -f /home/inventere/gunicorn.conf.py
+          docker run --rm inventree-test test -f /home/inventree/init.sh
+          docker run --rm inventree-test test -f /home/inventree/tasks.py
+          docker run --rm inventree-test test -f /home/inventree/gunicorn.conf.py
           docker run --rm inventree-test test -f /home/inventree/src/backend/requirements.txt
           docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/manage.py
       - name: Build Docker Image
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -95,6 +95,7 @@ jobs:
         run: docker compose --project-directory . -f contrib/container/dev-docker-compose.yml build --no-cache
       - name: Update Docker Image
         run: |
+          docker compose --project-directory . -f contrib/container/dev-docker-compose.yml run inventree-dev-server invoke install
           docker compose --project-directory . -f contrib/container/dev-docker-compose.yml run inventree-dev-server invoke update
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yaml
```diff
@@ -85,12 +85,17 @@ jobs:
           docker run --rm inventree-test invoke --list
           docker run --rm inventree-test gunicorn --version
           docker run --rm inventree-test pg_dump --version
+          docker run --rm inventree-test test -f /home/inventere/init.sh
+          docker run --rm inventree-test test -f /home/inventere/tasks.py
+          docker run --rm inventree-test test -f /home/inventere/gunicorn.conf.py
```
matmair, line 90: possible typos
```suggestion
          docker run --rm inventree-test test -f /home/inventree/init.sh
          docker run --rm inventree-test test -f /home/inventree/tasks.py
          docker run --rm inventree-test test -f /home/inventree/gunicorn.conf.py
```

.github/workflows/docker.yaml
```diff
@@ -85,12 +85,17 @@ jobs:
...
+          docker run --rm inventree-test test -f /home/inventree/src/backend/requirements.txt
           docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/manage.py
       - name: Build Docker Image
         # Build the development docker image (using docker-compose.yml)
         run: docker compose --project-directory . -f contrib/container/dev-docker-compose.yml build --no-cache
       - name: Update Docker Image
         run: |
+          docker compose --project-directory . -f contrib/container/dev-docker-compose.yml run inventree-dev-server invoke install
           docker compose --project-directory . -f contrib/container/dev-docker-compose.yml run inventree-dev-server invoke update
```
matmair, line 99: Comment: I think this is redundant as `inv update` already calls `inv install`

contrib/container/Dockerfile
```diff
@@ -128,6 +128,7 @@ COPY --from=prebuild /root/.local /root/.local
 
 # Copy source code
 COPY src/backend/InvenTree ${INVENTREE_BACKEND_DIR}/InvenTree
+COPY src/backend/requirements.txt ${INVENTREE_BACKEND_DIR}/requirements.txt
```
matmair, line 131: Comment: Ok that is an embarrassing mistake, that is on me. Interesting that the CI did not fail

matmair (approved):

.github/workflows/docker.yaml
SchrodingersGat: Intentional to check that the checks actually work :)

.github/workflows/docker.yaml
SchrodingersGat: I agree, but 'inv update' was previously *not* failing in these tests..

contrib/container/Dockerfile
SchrodingersGat: Yes, it has been passing CI for quite some time now! I cannot work out how the failure during `invoke install` is ignored.

.github/workflows/docker.yaml
matmair: Ah, that makes sense. I was really confused about how you would typo the project name

contrib/container/Dockerfile
matmair: That is interesting - is the raise not working?

SchrodingersGat: I only just added the raise

PR #7255, opened 2024-05-17
Bump pre commit versions

matmair:
This is mainly a maintenance chore with one important change: uv now supports including the `# via packagename` lines making dependabot diffsets much easier to read. To make the diff even smaller I have moved the config for uv to the pyproject.yaml file.


SchrodingersGat:
@matmair looks like this is causing a CI failure on python `3.12`

SchrodingersGat:
@dependabot rebase

matmair:
I cant seem to get regex running - it seems like it could be https://github.com/mrabarnett/mrab-regex/issues/413 but not sure.

matmair:
Interesting observation: something in the pipeline is not Python version agnostic although we pass versions through every I can think of

SchrodingersGat:
@matmair I have a new PR here to hopefully fix the playwright issue - https://github.com/inventree/InvenTree/pull/7511

PR #7260, opened 2024-05-19
Prevent deletion of part which is used in an assembly

SchrodingersGat:
Adds protection against deleting a part which is a component in an assembly.

@matmair @wolflu05 thoughts here? Should this behaviour be behind a configurable setting?

### DIFF
--- a/src/backend/InvenTree/InvenTree/models.py
+++ b/src/backend/InvenTree/InvenTree/models.py
@@ -122,6 +122,20 @@ class PluginValidationMixin(DiffMixin):
         self.run_plugin_validation()
         super().save(*args, **kwargs)
 
+    def delete(self):
+        """Run plugin validation on model delete.
+
+        Allows plugins to prevent model instances from being deleted.
+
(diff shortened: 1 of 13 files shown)


wolflu05:
Sorry that I cannot add some production feedback here, I have not really used assemblies yet. But just from thinking about it,  if we delete an part, the part for the assembly is missing, that would mean, we either delete the assembly too, or remove the part from the assembly or deactivate the part somehow that its not shown in search anymore.

SchrodingersGat:
> Sorry that I cannot add some production feedback here, I have not really used assemblies yet. But just from thinking about it, if we delete an part, the part for the assembly is missing, that would mean, we either delete the assembly too, or remove the part from the assembly or deactivate the part somehow that its not shown in search anymore.

I think that this is not quite right. The change I am implementing here prevents the *sub_part* from a BOM being deleted - which would in practice mean that a BOM item disappears (without warning) from the BOM of an existing assembly. This change prevents the part from being deleted while it is still in the BOM for a different assembly.

PR #7261, opened 2024-05-19
fix: SELinux labels for Caddyfile

p-fruck:


SchrodingersGat:
@p-fruck thanks for the fix, sorry for the oversight!

PR #7263, opened 2024-05-20
Docs: Update plugins and compose

p-fruck:
Seems like the codebase has changed quite a bit since my last working dev setup :) Had some trouble setting everything up, I hope I got the stuff with the new plugin path and compose command correctly?


SchrodingersGat (approved):

SchrodingersGat:
@p-fruck thanks for the updates to the docs :)

PR #7280, opened 2024-05-21
Add documentation on 'validate_model_deletion'

SchrodingersGat:


matmair (approved):
LGTM; I like the direct linking to souce code!

matmair:
Any idea why it seems that inventree-python CI tests are failing on all jobs lately?

SchrodingersGat:
> Any idea why it seems that inventree-python CI tests are failing on all jobs lately?

Yep I will fix that tonight.

Issue #6306, opened 2024-01-21
[React] New markdown editor


SchrodingersGat:
For fields that support markdown (like "notes") we currenly use easymde.

However it looks like mdx-editor would be a better choice - https://github.com/mdx-editor/editor

It supports mode advanced editing features (including image resizing) and a better, more "modern" look and feel. Better react support too.

matmair:
Something more modular (enableing mentions and inline links linke GitHub has) like tiptap might also be intresting

wolflu05:
I like that there is a WYSIWYG and diff editor, but still a source editor. But we need to consider that change well, because there are some special markdown flavors involved that maybe people will use and break later if we migrate to something else.

PR #7284, opened 2024-05-21
[PUI] Notes editor

SchrodingersGat:
- Implement new "markdown editor" for PUI.
- Uses [MDXEditor](https://mdxeditor.dev/)
- Closes https://github.com/inventree/InvenTree/issues/6306


wolflu05:
> @SchrodingersGat what is the difference between MDXEditor and the editor we are currently using?

I would be really interested on the answer as well.

SchrodingersGat:
This new editor has a few additional features (in particular the ability to resize images) which I found appealing, and it much more in line with the visual look and feel of the new interface. 

However diving into "image resize" a bit more, that requires custom `` tags (currently we disallow any html tags) and so would require a back-end change anyhow

SchrodingersGat:
As a separate issue I would like to explore removing the "notes" from the API endpoints where they currently reside, and put them into separate endpoints (like "metadata"). The "notes" fields can potentially house a lot of data and I do not see the benefit of serializing each notes field for every item when we are generating a list of items...

PR #7305, opened 2024-05-23
Fix sales order shipment / completed buttons

SchrodingersGat:
- Template logic needed tweaks
- Ref: https://github.com/inventree/InvenTree/pull/6449/


SchrodingersGat:
@matmair weird that all the tests are failing. Might be an issue with uv?

Looks like a new version just dropped: https://pypi.org/project/uv/

PR #7309, opened 2024-05-23
Reintroduce uv into CI actions

SchrodingersGat:
- Removed in https://github.com/inventree/InvenTree/pull/7305
- New version of `uv` has broken CI pipeline

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -49,8 +49,7 @@ runs:
         shell: bash
         run: |
           python3 -m pip install -U pip
-          pip3 install -U invoke wheel
-          pip3 install uv==0.1.45
+          pip3 install -U invoke wheel uv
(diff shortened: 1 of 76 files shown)


SchrodingersGat:
@matmair I have raised this so we don't neglect to add `uv` back into the pipeline. But it looks like the recent updates to `uv` have ... broken things

matmair:
Ok this seems somewhat intentional / not considered a bug https://github.com/astral-sh/uv/issues/3765#issuecomment-2125648857

matmair:
And https://github.com/astral-sh/uv/pull/3767

matmair:
At this point, I think we should either pin the latest working version or just go back to pip. There are 200 more important issues than fixing the CI every few uv updates

SchrodingersGat:
I'd support just going back to pip. `uv` has caused a *lot* of headaches and I agree it is not worth our time chasing this down

PR #7320, opened 2024-05-24
Merge pull request from GHSA-2crp-q9pc-457j

SchrodingersGat:
* ensure API login only works if mfa is not required

* add migration to log out users

* add migration to clear users


wolflu05:
*I just comment here, as I cannot comment anymore on the published advisory.* 

Maybe we should document the added logout migration to the advisory description, so that people know about this *restriction*.

PR #7322, opened 2024-05-24
Handle exception on migration

SchrodingersGat:
Fixes bug in migration file which was introduced as part of https://github.com/inventree/InvenTree/security/advisories/GHSA-2crp-q9pc-457j


matmair:
Looks fine but seems to still fail. IMO we can also remove this migration, it is added security

PR #7329, opened 2024-05-24
Label printing bug fixes

SchrodingersGat:
Addresses some bugs in label printing framework

### DIFF
--- a/src/backend/InvenTree/plugin/base/label/mixins.py
+++ b/src/backend/InvenTree/plugin/base/label/mixins.py
@@ -177,9 +177,9 @@ class LabelPrintingMixin:
                 self.print_label(**print_args)
             else:
                 # Offload the print task to the background worker
-                # Exclude the 'pdf_file' object - cannot be pickled
 
-                kwargs.pop('pdf_file', None)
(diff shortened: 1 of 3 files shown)


wolflu05:
That Pickering error is exactly #6772. and that's why I think the request object should not be part of the rendering methods. Because rendering should be done in the background. But that only works if the request (which obviously cannot be pickeld) is not in the arguments passed to the worker.

wolflu05:
@SchrodingersGat @matmair what do you think on that: https://github.com/inventree/InvenTree/pull/7329#issuecomment-2129673252 ?

TL;DR I think rendering a label/report template should be done on a background worker for performance and architecture reasons. But that requires we completely strip out the request object from the args/kwargs as the request object cannot be pickelt. (That would also solve part of #6772)

SchrodingersGat:
@wolflu05 I am completely on board to remove the request object entirely, and allow label printing to be done entirely in the background worker thread. To achieve this we would have to change what we pass through - in fact all we need to pass through is:

- The `LabelTempltae` ID
- The `slug` of the plugin
- The `pk` values of the model instances to be printed
- The `pk` of the user

Issue #7321, opened 2024-05-24
"Category" field of "Create Part" form is having problems


SchrodingersGat:
### Discussed in https://github.com/inventree/InvenTree/discussions/7318

<div type='discussions-op-text'>

<sup>Originally posted by **Century1026** May 24, 2024</sup>
I first discovered the problem when the title "Category" was not able to be translated.

Then I made changes on title names and help text.
![b2fb0ca92d783834732c7606abf6267](https://github.com/inventree/InvenTree/assets/106154819/27db7f5f-b47e-4fc4-9c7f-e33cd2eed69a)
![e88d4cfdac1030c87d97bfab821fb57](https://github.com/inventree/InvenTree/assets/106154819/e6e1f5ee-be87-485f-97ae-27b322a4bb5a)
However, it turned out only the "Category" field is having no changes at all.
![2d4c2348a8c4ca8caafb891273b64df](https://github.com/inventree/InvenTree/assets/106154819/31226236-b496-403c-956f-1ca873475255)
 Is the title of "Category" field defined somewhere else?</div>

PR #7343, opened 2024-05-26
Updates for metadata information

SchrodingersGat:
- Override 'label' values with 'verbose_name' values
- Only if 'label' is *not* translated, but 'verbose_name' is
- Allows the translated model fields name to be pushed through to the metadata framework
- Closes https://github.com/inventree/InvenTree/issues/7321

----------------------

This PR fixes a subtle bug which existed in our "metadata" implementation, which extracted database field information and provided it to the DRF API.

Any custom serializer fields which do not have an explicitly defined label would default to their attribute name, which was an untranslated string.

So, something like this:

image

Would result in a `label` value of "Category" (which is simply extracted from the name of the serializer field) even if it pointed to an underlying database model field. The "verbose_name" of the field (in this case `Part Category`) was not obtained.

Worse, the auto-generated labels were not exposed to the translation framework at any point. So, any custom DRF fields were never getting translated.

------------------------

This patch adds some extra introspection steps, and will override a raw stirng value with a translated string value (if it exists). It also adds explicit mapping from `verbose_name` (on the model field) to `label` (on the serializer field).

*Note: If a translated string exists on the serializer field it will not be overridden by the model field*

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -13,6 +13,10 @@ site/
 # Generated API schema files
 docs/api/schema/*.yml
 
+# Temporary cache files
+url_cache.txt
+invoke-commands.txt
+
 # Temp files
 releases.json
 versions.json
--- a/docs/docs/develop/contributing.md
+++ b/docs/docs/develop/contributing.md
@@ -96,7 +96,7 @@ The HEAD of the "stable" branch represents the latest stable release code.
 
 ## API versioning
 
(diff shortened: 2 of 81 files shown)


matmair (approved):
LGTM; Should we add tests for this (translation)?

SchrodingersGat:
> LGTM; Should we add tests for this (translation)?

Yeah, I'll add some unit tests so that we don't get any regressions

PR #7357, opened 2024-05-27
[PUI] Mantine tree

SchrodingersGat:
- Replaces existing navigation trees with new native mantine component


wolflu05:
Have you done some performance testing with this change on an instance with large amount of locations/categories?

matmair (approved):
LGTM

SchrodingersGat:
> Have you done some performance testing with this change on an instance with large amount of locations/categories? Would be really interested in, how this performs.

As far as I have tested it seems just as performant as the current solution.

PR #7367, opened 2024-05-28
PO shipped complete

SchrodingersGat:
Adds some configuration and more unit testing for "shipped" status for sales orders


matmair:
@SchrodingersGat is there a specific reason all the status codes are moved around? This makes things a bit more difficult / less obvious for #4289

SchrodingersGat:
> @SchrodingersGat is there a specific reason all the status codes are moved around? This makes things a bit more difficult / less obvious for #4289

This was just for separation of the apps - stock status values seemed to belong in `./stock/`, etc.

Issue #7385, opened 2024-06-01
Docker Install - Ensure that REDIS is not enabled in .env during DB Initialization


yodaphone:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When installing via Docker Compose, if the REDIS variables are enabled in `.env` file during the Initial DB Setup, the setup errors out & the fails. The documentation must make it clear NOT to enable REDIS cache before the DB Initialization is completed

When i moved my docker compose, .env & Caddyfile to another server, it failed to build & thats how I found out. I rebuilt the VM and when I disabled it in the `.env` file

```
Traceback (most recent call last):
  File "/root/.local/lib/python3.11/site-packages/django_redis/cache.py", line 29, in _decorator
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.local/lib/python3.11/site-packages/django_redis/cache.py", line 99, in _get
    return self.client.get(key, default=default, version=version, client=client)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/.local/lib/python3.11/site-packages/django_redis/client/default.py", line 260, in get
    raise ConnectionInterrupted(connection=client) from e
django_redis.exceptions.ConnectionInterrupted: Redis ConnectionError: Error -3 connecting to inventree-cache:6379. Try again.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/inventree/src/backend/InvenTree/manage.py", line 24, in <module>
    main()
  File "/home/inventree/src/backend/InvenTree/manage.py", line 20, in main
    execute_from_command_line(sys.argv)
  File "/root/.local/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/root/.local/lib/python3.11/site-packages/django/core/management/__init__.py", line 416, in execute
    django.setup()
  File "/root/.local/lib/python3.11/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/root/.local/lib/python3.11/site-packages/django/apps/registry.py", line 124, in populate
    app_config.ready()
  File "/home/inventree/src/backend/InvenTree/InvenTree/apps.py", line 59, in ready
[...]

PR #7388, opened 2024-06-02
Cache updates

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/7385
- Improvements to global cache support
- Override global caching during certain operations
- Update documentation
- Enable redis cache for devcontainer
- Finer control over caching settings

### DIFF
--- a/src/backend/InvenTree/InvenTree/cache.py
+++ b/src/backend/InvenTree/InvenTree/cache.py
@@ -32,8 +32,11 @@ def is_global_cache_enabled():
     - Test if the user has enabled and configured global cache
     - Test if it is appropriate to enable global cache based on the current operation.
     """
+    host = cache_host()
+
     # Test if cache is enabled
-    if not cache_setting('enabled', False, typecast=bool):
+    # If the cache host is set, then the "default" action is to enable the cache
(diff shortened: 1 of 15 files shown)


src/backend/InvenTree/InvenTree/cache.py
```diff
@@ -0,0 +1,105 @@
...
+        logger.warning('Global cache is enabled, but no cache host is configured!')
+        return False
+
+    # The cache should not be used during certain operations
+    if any((
+        InvenTree.ready.isRunningBackup(),
+        InvenTree.ready.isRunningMigrations(),
+        InvenTree.ready.isRebuildingData(),
+        InvenTree.ready.isImportingData(),
+        InvenTree.ready.isInTestMode(),
```
matmair, line 53: Q: How should caching be tested if it is always disabled in testing?

matmair (approved):
LGTM; at some point we should probably test for this

src/backend/InvenTree/InvenTree/cache.py
SchrodingersGat: We could override this like we do for other test cases with a special environment variable.

Issue #7383, opened 2024-06-01
Invalid SITE_URL value:  InvenTree server cannot start.


decryptedchaos:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [x] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

i have my own docker <-- Proxy  (Traefik) disabling it is not an option,  and i don't really think its the cause,  but as such,  i h ave disabled the use of caddy for inventree because my traefik is configured to auto deploy  containers to a DNS wildcard  containername.docker.example.com    this setup should work just as Caddy does as it just proxys the dns to the container on its service port.  so i don't think its a problem.. 

But i can't get the inventree-server to stop flapping (restarting)   it just contantly loops with this error 

Invalid SITE_URL value: '"http://inventree.docker.example.com"'. InvenTree server cannot start.

Yet, the variable SITE_URL  is configured properly and works.   the server just never starts. 

### Steps to Reproduce

Described above.

SchrodingersGat:
please provide full log output for the server. And I assume the url you have provided is not your actual URL?

decryptedchaos:
Yeah clearly,  i just don't want my url posted publicly for obvious reasons.  

There isn't really any log  it just repeats what i said over and over with standard info lines above and below 

Python version 3.11.9 - /usr/local/bin/python
Invalid SITE_URL value: '"http://inventree.docker.*******.com"'. InvenTree server cannot start.
Loading config file : /home/inventree/data/config.yaml

SchrodingersGat:
@decryptedchaos I tried to reproduce this fault, by using the literal URL `http://inventree.docker.example.com` - the server did not fail in this case. Is there perhaps something particular about the specific URL you are using?

Here is the validation being performed (in https://github.com/inventree/InvenTree/blob/cdac7465b22e0eaaf1f2d2a469f2ef32502d03db/src/backend/InvenTree/InvenTree/settings.py#L987)

![image](https://github.com/inventree/InvenTree/assets/10080325/bf3870ac-98b3-4f46-b7c3-b982c286da9d)

decryptedchaos:
No nothing special,  the only thing that's even remotely different on my setup is that i don't use your supplied Caddy proxy like i say in my OP, because i already have a configured traefik proxy serving multiple containers that i can't simply change for one image.  

[...]

PR #7392, opened 2024-06-03
Config var fix

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/7383
- Automatically remove extraneous quotes around config variables


SchrodingersGat:
This approach is too simple and causes a number of breakages for little benefit. closing it out.

PR #7404, opened 2024-06-04
Setting refactor

SchrodingersGat:
Adds helper functions for "get" and "set" for server-side settings.

- Code reduction
- Improves "cyclical" imports


matmair (approved):
LGTM; should we add a pre-commit hook that ensures the new helpers are used / the model is not accessed directly?

SchrodingersGat:
@wolflu05 @matmair any insights into why CI might be failing here, on SQLITE / python 3.9, but works on 3.12?

matmair:
Not really sure - looks like the DB is not cleaned

SchrodingersGat:
I cannot reproduce locally on a 3.9/sqlite setup...

SchrodingersGat:
@matmair found the issue, a new setting was being created in the validator for a different setting.

matmair:
That is a tricky one to find!

PR #7405, opened 2024-06-04
Panel placeholders

SchrodingersGat:
- Force panel content to be specified
- Allows us to see which panels still need to be implemented


matmair (approved):
LGTM; This makes it easier to find all the places we need to fill

PR #7425, opened 2024-06-10
[plugins] allow static files for plugins

SchrodingersGat:
- Adds support for static files for plugins
- Anything in the "static" subdir of each plugin is copied across
- Runs automatically when a plugin is activated
- Runs as part of the `invoke static` task
- Runs as part of the `invoke plugins` task

This PR is part of preparation for bring UI plugins into the new react interface. Ideally, the bulk of any "frontend" plugin code can be served by the static file server, which is much more efficient than rendering everything on the backend on each request. It is left up to the plugin developer to determine which files should be served statically.

Plugin static files are copied into the global static directory, to `./plugin/<slug>/` - where `slug` is the key of the plugin instance.

Files are copied only for "active" plugins

### DIFF
--- a/src/backend/InvenTree/InvenTree/management/commands/collectplugins.py
+++ b/src/backend/InvenTree/InvenTree/management/commands/collectplugins.py
@@ -1,0 +1,13 @@
+"""Management command to collect plugin static files."""
+
+from django.core.management import BaseCommand
+
+
+class Command(BaseCommand):
+    """Collect static files for all installed plugins."""
+
+    def handle(self, *args, **kwargs):
+        """Run the management command."""
+        from plugin.staticfiles import collect_plugins_static_files
+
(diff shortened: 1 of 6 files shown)


wolflu05:
@SchrodingersGat This has broken plugins that were already using the static files, as now the static files are served by a different path. Ref: wolflu05/inventree-bulk-plugin#80 This should definitely be marked as breaking.

Additionally the automatic static files collection has not worked for me. I tried to install my plugin in a fresh docker production environment myself, but it had not collected the plugin static files itself, I needed to run `invoke update`. Same for the migrations.

Also I discovered a very weird bug, with the plugin registry. I was just hitting reload a couple of times in this recording:

https://github.com/user-attachments/assets/d462c35e-23b1-474e-b585-acb7cb751b94

PR #7435, opened 2024-06-12
Remove the New Stock Item button from Stock view in case the user has no Stock write mermission

SergeoLacruz:
Strange...
If a user has only view permission to Stocktake, stock location and stock item, the button is still visible. If the user pushes it
an error appears, not allowed. 
If I grant the user VACD permissions to Build and Purchase order via groups, the user can create a stock item using
this button. Might be a bug. 
This pull request just removes the button.


SchrodingersGat:
@SergeoLacruz thanks, this is a nice addition :)

PR #7420, opened 2024-06-08
Single table for file attachments

SchrodingersGat:
Refactoring how "attachments" are stored, moving to a single table approach.

### Tasks

- [x] Create new table
- [x] Data migration
- [x] Remove old tables
- [x] Expose new model to API
- [x] Add a mixin class for models which will have attachments associated with them
- [x] Mixin class must provide method for accessing attachments
- [x] Expose a list of models we wish to associate attachments with
- [x] Remove old API endpoints
- [x] Update attachment tables in PUI
- [x] Update attachment tables in CUI
- [x] Update unit tests
- [x] Extract user information on upload
- [x] Unit test for data migration
- [x] Unit tests for `InvenTreeAttachmentMixin` class mixin
- [x] Work out how to determine user permissions in backend (i.e on file upload)
- [x] Work out how to determine user permissions in user interface
- [x] Update python bindings - https://github.com/inventree/inventree-python/pull/234
- [x] Update mobile app - https://github.com/inventree/inventree-app/pull/505

### Related Issues

- Closes https://github.com/inventree/InvenTree/issues/5429
- Closes https://github.com/inventree/InvenTree/issues/5703

PR #7447, opened 2024-06-15
 Remove add/edit/delete access to stock item attachments when user is not owner

SergeoLacruz:
I think it is better if the use cannot add or delete attachments if the are not owner.


SchrodingersGat:
@SergeoLacruz this change will be overridden by the refactor in this PR - https://github.com/inventree/InvenTree/pull/7420

Essentially we are moving all attachements into a single database table.

So, we will need to consider how attachment permissions are handled in this case.

Let's wait until #7420 is merged and then revisit this.

Issue #7429, opened 2024-06-11
Loading large tables of data via python API bloats Inventree server memory usage


zoopsel:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When running a python program that requests data via the inventree-python api from a large table, the inventree server's memory usage bloats in size to multiple gigabytes.
In can happen that during this time all the memory of the system is used up and linux kills the server.

We have quite a lot of StockItemTracking entries in our Inventree database (9.5 MB), and I would like to fetch them all ( `StockItemTracking.list(api)` ) in order to do some quick in memory processing on the data.
First, we had to increase the timeout value of the connection, as this is set to 10 seconds by default.
Then we are able to fetch all these elements from the server but it takes quite lot of time (>10mins) and the memory used by the Inventree process goes from 260MB to 8.2GB.
When the fetching process eventually finishes, the memory used by Inventree remains at around 4GB and does not decrease back to 260MB.

This happens when using the Inventree develop server (`invoke server`) and when using Inventree with Gunicorn.

Is that normal behaviour or might this be some caching bug?

### Steps to Reproduce

1. Have a table with lots of entries in the Inventree database such that the database table has a sixe in the order of MBs.
2. Create a python script and fetch this table via the api. (you probably will have to increase the timeout value)
3. Use a process monitor and look at memory usage.

### Expected behaviour

No memory bloat.

### Deployment Method

- [ ] Docker
- [ ] Package
- [X] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

0.15.3 (but also on 0.12.9)

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

### Relevant log output

```shell
Neither the server console nor the console of the python script fetching the data produce any output.
```

SchrodingersGat:
Is this specific to fetching via the python bindings? Or any such request to the API?

In any case this is a serious issue, the memory should definitely not be bloating out like that.

- How many items are in the table you are calling?
- Does this happen with multiple tables or is limited to a single table?
- Any other testing you have performed which might inform this ?

zoopsel:
I encountered the bug when using the python bindings, never tried other ways.
Other tables work fine. For example the StockItems i can fetch via python api bindings without any problems.
In our database there are 29'074 stock items and the whole table has a size of 4.5 MB.
In the stock item tracking table there are 103'669 items and the size is 9.5MB.

[...]

PR #7451, opened 2024-06-16
Improve stock item tracking API query

SchrodingersGat:
- Cache related model lookups into single DB queries
- Significant improvements to query speed
- Ref: https://github.com/inventree/InvenTree/issues/7429

------------------

As part of testing for https://github.com/inventree/InvenTree/issues/7429 I discovered that the "stock item tracking" API endpoint is very inefficient! This fix provides some caching so that a significantly reduced number of database queries are actually applied.

For an example dataset with ~1000 rows in the DB table:

| - | N Queries | Time | 
| --- | --- | --- |
| **Before** | 1985 | 3.4s |
| **After** | 47 | 0.44s |

### DIFF
--- a/src/backend/InvenTree/stock/api.py
+++ b/src/backend/InvenTree/stock/api.py
@@ -1464,7 +1464,9 @@ class StockTrackingList(ListAPI):
                 deltas = item['deltas']
 
                 if key in deltas:
-                    item['deltas'][f'{key}_detail'] = related_data[deltas[key]]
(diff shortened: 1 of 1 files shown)


matmair:
@SchrodingersGat how did the CI test for database query count not trigger this? Do we need to adapt that or change the tests?

SchrodingersGat:
Looking into it, we don't seem to have any API tests for this endpoint

PR #6293, opened 2024-01-19
Remove django-allauth-2fa, dj-rest-auth and django-user-sessions    

matmair:
This PR remove django-allauth-2fa and replaces it with the inline functions of django-allauth. This reduces the dependency footprint and eliminates a bunch of custom creations.
It also adapts the schema generation process to include all allauth endpoints (they do not use DRF so normal generation is not working).

Update: Most of the planned things we would be interested in are now implemented and the scope shifted a bit - it makes sense to just ripp all auth extensions out and use only allauths built-in things. We have a lot of grown auth-adjacent code that can probably be removed/rewritten.

### Related Issues

- Closes #6281
- Closes #8708
- Closes #8949

### TODO 
- [x] refactor authentication forms && detection of enabled features (see 8912)
- [x] implement custom link target lookup for email from django-allauth
- [x] show user ID in MFA flow
- [x] implement email verification flow
- [x] re-implement registration flow
- [x] re-implement password reset flow
- [x] re-implement password change
- [x] keep auth context somewhere for the local session to refer back to
- [x] move now removed system status info into main status (registration, mfa, sso available)
- [x] switch back to browser flow for auth APIs
- [x] rename all new auth URLs to fix to the current schema
- [x] include schema for auth with central API
- [x] write changelog

[Coverage](https://app.codecov.io/gh/inventree/InvenTree/pull/6293?dropdown=coverage)

This relies on several upstream PRs:
- https://codeberg.org/allauth/django-allauth/pulls/4236 (custom frontend URLs)
- https://codeberg.org/allauth/django-allauth/pulls/4235 (allowing email registration even if MFA is enabled)

PR #7469, opened 2024-06-18
[PUI] Add 2FA login

### LINKED ISSUE #6200: [PUI] Enable 2FA for new interface
Need to enable 2FA functionality via the API, and implement for:

- React interface
- Mobile app
- etc

In addition to implementing 2FA support via API endpoints, need to setup the user interface elements for the new platform UI

matmair:
This PR adds the option to use the CUI 2FA login method for PUI. The user is automatically send to the appropriate screen, can enter their default 2FA method (TOTP code, backup code, HOTP) and is then redirected to the PUI home screen.

https://github.com/inventree/InvenTree/assets/66015116/d723d5ef-a6e4-4c5a-b33a-6352bfe83415

This method respects CRSF and is stateless - which should make it pretty hard to circumvent.

Closes #6200


SchrodingersGat:
@matmair seems like a very good approach to work around this issue

wolflu05:
Awesome that you implemented this, this was stopping me from using 2FA on my home instance. The question is just, what are we gonna do if we deprecate CUI? Will we still keep that view for entering the TOPT code? Or is there a possibility, we can directly implement that into PUI?

matmair:
@wolflu05 my plan is to leave this in when we remove CUI, it is only a very small part - you can check out what this removal would look like here: https://github.com/matmair/InvenTree/pull/174

I do not think we should implement the different 2fa factors ourself (we support 3 rn) as that would create a lot of surfaces to make costly security issues.

matmair:
This is ready for review/merge - testing the frontend is proving to be very hard because we can't add MFA methods via API (yet - that is coming with https://github.com/inventree/InvenTree/pull/6293)

wolflu05:
> In the long run I really want to switch to allauths new included methods

But this would enable use to build our own 2FA input directly in PUI and really get rid of all CUI code. In your linked allauth roadmap they mention: 
> Add first-class API support. As part of the delivering the API, an OpenAPI specification and example React example application will be delivered, showcasing full SPA access to all of the django-allauth flows.

If they support 2FA via API, its not "our responsibility" with custom API code to built it secure anymore, because as I understand, they want to implement the critical backend API layer we could just use, and build our own frontend around it?

matmair:
Yes, the switch will enable that. It will however require a complete review and possibly rewrite of our auth processes/integrations. This might take a while as it has grown over the years.

SchrodingersGat:
@matmair nice work, thanks for the fix

PR #7474, opened 2024-06-19
Added details on stock ownership control

SergeoLacruz:
Add background info and an example


SchrodingersGat:
@SergeoLacruz thanks for this addition :)

PR #4488, opened 2023-03-14
[Feature] Add RMA support

SchrodingersGat:
This PR adds support for "RMA" or "Return Order" functionality.

Initial implementation may be somewhat "bare bones" and need fleshing out in greater detail in subsequent PRs.

- Closes https://github.com/inventree/InvenTree/issues/3298

### TODO

- [x] Add setting to enable / disable `Return Order` functionality (default = off)
- [x] Implement "detail" screen for individual Return Order
- [x] Implement API endpoints for attachments
- [ ] ~~Implement "calendar view" for Return Order List~~ - future work
- [x] Enable search results in quick-search bar
- [x] Enable search results in search page
- [x] Add "line item" model
- [x] Add "extra line item" model
- [x] Add functionality for returning goods
- [x] Clean and add database migrations
- [x] Add models for return reports (and implement functionality for this)
- [x] Generate a "nice" basic report for return orders
- [x] Unit testing for ReturnOrder API endpoints
- [x] Unit testing for Contact API endpoints
- [x] Unit testing for returnorder report generation
- [x] Unit testing for page permissions
- [x] Split `order.js`
- [x] Reinvigorate `contact` model
- [x] Add "point of contact" to purchase order and sales order modesl
- [x] Add dataset for new "Contact" model - https://github.com/inventree/demo-dataset/pull/40
- [x] Add dataset for ReturnOrder - https://github.com/inventree/demo-dataset/pull/41
- [x] Add app issue for supporting "point of contact" - https://github.com/inventree/inventree-app/issues/285
- [x] Add app issue for supporting return orders - https://github.com/inventree/inventree-app/issues/286
- [x] Add documentation for "contacts" - https://github.com/inventree/inventree-docs/issues/455
- [x] Add documentation for "Return Order" - https://github.com/inventree/inventree-docs/issues/458
- [x] Update documentation for "Purchase Order" - https://github.com/inventree/inventree-docs/issues/455
- [x] Update documentation for "Sales Order" - https://github.com/inventree/inventree-docs/issues/455
- [x] Add python integration for new return order functionality - https://github.com/inventree/inventree-python/issues/177

PR #7479, opened 2024-06-19
More tests for `users` app

matmair:
This PR adds more tests for the APIs in the `users` app - which is partly responsible for auth.
It also removes some dead and duplicate code and adds docs for the `/api/user/roles` endpoint (during cleanup  discovered that this was missing).


src/backend/InvenTree/users/models.py
matmair, line 1: @SchrodingersGat you added this in https://github.com/inventree/InvenTree/pull/4488 but I can not find any usage in the code base - any idea if this is still required?

SchrodingersGat:
@matmair thanks for these enhancements :)

PR #7488, opened 2024-06-21
Fix non null field 'description'

mp-strachan:
Fix typo in migration '0026_auto_20240422_1301.py' to resolve error: `TypeError: LabelTemplate() got unexpected keyword arguments: 'decription'`


SchrodingersGat:
@mp-strachan good catch, thanks for the PR

Issue #5028, opened 2023-06-12
[Roadmap] Implement proper revision system


SchrodingersGat:
Currently the "revision" field is a simple text field, does not have any functional meaning.

Future development should focus on implementing proper "revision" for Part:

- A "new" revision overrides old one
- Revisions are linked together
- When viewing a part detail page, show available revisions

```[tasklist]
### Tasks
- [x] Prevent deletion of part which is used in an assembly -https://github.com/inventree/InvenTree/pull/7260
- [x] Add "locked" attribute to Part model - https://github.com/inventree/InvenTree/pull/7527
- [x] Prevent update of BOM items / parameters / etc for locked parts - https://github.com/inventree/InvenTree/pull/7527
- [ ] Add "revision_of" attribute to Part model - https://github.com/inventree/InvenTree/pull/7585
- [ ] Implement validation rules for variants / revisions - https://github.com/inventree/InvenTree/pull/7585
- [ ] Add revision selection to PUI - https://github.com/inventree/InvenTree/pull/7585
```

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/inventree) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/inventree/InvenTree/issues/5028">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/inventree/InvenTree/issues/5028/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/inventree/InvenTree/issues/5028/pledge.svg">
</picture>
</a>

wolflu05:
What about returning the active data (I think we discussed that there should be a active/latest marked revision) for the current requests (an new rev parameter is obmitted) so it's downwards compatible. And if there is an rev parameter we search for the particular revision. Can this be implemented with an Mixin? Wouldn't that remove so much work for the future if we do it properly?

martonmiklos:
Uhm I do not want to do feature creeping, but PLM like change managment features would be very useful  in some small business use cases. 
I am thinking mainly about enforced approval processes of the revision changes. For e.g. if someone wants to make a new revision active then an approval process needs to be followed in order to make the change. This could contain approvals from various parties involved in the products lifecycle (for e.g. test engineer, contract manufacturer, etc..) If all approval done the new revision is made active automatically.

[...]

PR #7527, opened 2024-06-28
[Feature] Part lock

SchrodingersGat:
- Adds ability to "lock" parts:
- Ref: https://github.com/inventree/InvenTree/issues/5028

### Features

- Cannot delete locked parts
- Cannot edit or delete BOM items for locked assemblies
- Cannot edit or delete parameters for locked parts

### TODO 

- [x] Unit testing for editing / deleting locked BOM items
- [x] Unit testing for editing / deleting locked parameters
- [x] Unit testing for deleting locked parts
- [x] Update documentation

### Ideas For Future Work

- Add global setting to control "locked" action for BOM items
- Add global setting to control "locked" action for part parameters
- Control who can lock / unlock a part
- Determine which "part" fields are restricted from being edited when the part is locked


matmair (approved):
This looks very nice! Do you think this makes sense to abstract some of the logic into a general mixin? I can see locking being useful for orders and attachments too.

PR #7532, opened 2024-06-28
Plugin static files fix

SchrodingersGat:
Fixes for copying plugin static files


SchrodingersGat:
@matmair did you recently run into some CI issues with a regex import? Any idea what was causing this? I do not see any changes in this PR which would cause an issue

matmair:
Yes @SchrodingersGat, they just magically went away. I have looked at the changes and no idea what is causing this

PR #7547, opened 2024-07-02
Detect python version for installer

matmair:
This PR introduces autodetection of available Python versions on a system during installation - if no Python environment was set up before.
Should reduce the need for manual intervention during package install.


SchrodingersGat:
@matmair very nice indeed, thanks for taking the time to look into this :)

matmair:
@SchrodingersGat I am working on a few other items in this label area, expect more package / installer PRs in the following days

PR #7553, opened 2024-07-03
Display barcode text near qrcode

matmair:
This PR adds an optional display of the encoded barcode data underneath QR codes. Currently only in CUI - barcode modals are not implanted in PUI.
grafik


matmair:
@SchrodingersGat this is now ready for review. this regex error in 3.12 is very distracting - i can not reproduce it locally so no idea how we will fix it

SchrodingersGat:
@matmair thanks, great work. Agreed that the regex issue is very frustrating. Will need to allocate some time to look into it.

Issue #5755, opened 2023-10-19
[PUI] Check user permission when viewing pages


SchrodingersGat:
User might not have "view" permission for a given page.

If not, redirect away, or show error message?

matmair:
I would run a permission check and if it is a missing permission show an error, else log out (user might be unauthenticated).

SchrodingersGat:
Not stale

matmair:
@SchrodingersGat we now have permission tests in PUI - is this still valid?

SchrodingersGat:
Yes this is still valid - we do not currently check if a user has the required permissions to view any particular page in PUI.

PR #7554, opened 2024-07-03
[PUI] Error pages

SchrodingersGat:
- Adds framework for displaying various error pages to the user based on API response status
- Prevents non-staff users from accessing system settings
- Prevents non-staff users from accessing admin page
- Handle 404 errors (and other errors) for detail view pages
- Closes https://github.com/inventree/InvenTree/issues/5755


src/frontend/src/pages/InstanceDetail.tsx
matmair, line 1: Comment: Shouldn't this be in layout / components?

src/frontend/src/pages/InstanceDetail.tsx
```diff
@@ -0,0 +1,28 @@
...
+    return <LoadingOverlay />;
+  }
+
+  if (status >= 500) {
+    return <ServerError status={status} />;
+  }
+
+  if (status >= 400) {
+    return <ClientError status={status} />;
+  }
```
matmair, line 25: Question: What is the error state if the user has viewing perms but can not do a change action? Are they immediately thrown off the page?

matmair (approved):
LGTM;
Site Note: It seems like the pattern of a navigationtree and details occurs a bunch of time - this could be a good candidate for a generic component in the future

src/frontend/src/pages/InstanceDetail.tsx
SchrodingersGat: I do not think that this update will have any issues here - the permissions checks should happen separately. Certainly will keep an eye out for it

Issue #6891, opened 2024-03-29
Single-line installer is not looking for already exsisting installs


matmair:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

The installer is just installing sources and the package - not checking if there is already a config (that might be missing mandatory new keys)

### Steps to Reproduce

1. Have an old instal (0.9.0 or similar)
2. Install a current version (0.14.0 or similar)
3. Have an install missing BACKUP_DIR and SITE_URL config keys, making the install more or less unusable
4. Get Errors during the install because of the already existing env

### Expected behaviour

Just working

### Deployment Method

- [ ] Docker
- [ ] Bare metal

### Version Information

0.14.0

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

PR #7555, opened 2024-07-04
Package is not checking for already existing installs

matmair:
This PR adds logic to detect what version was installed before and initially ensures that updates from versions below 0.9.0 get SITE_URL and BACKUP_DIR set correctly.
If something is missing it is logged and the setup is exited with status_code 1

Fixes https://github.com/inventree/InvenTree/issues/6891

### DIFF
--- a/contrib/packager.io/functions.sh
+++ b/contrib/packager.io/functions.sh
@@ -344,7 +344,7 @@ function final_message() {
 
 function update_checks() {
   echo "# Running upgrade"
-  old_version = $2
+  old_version = $1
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
Thanks for your continued efforts on maintaining and improving the installer :)

matmair:
I strongly intend to close all ([2](https://github.com/inventree/InvenTree/labels/installer)) remaining installer issues before `0.16.0` release so I can bring back the DO image in good faith.

PR #7556, opened 2024-07-04
Fix SSO theme selection

### LINKED ISSUE #7257: Theme selection not showing when using SSO

### Describe the bug*

When using a SSO provider (in this case OIDC) it seems that selecting the theme in the user settings does not change the displayed theme.

When selecting the theme in settings, the request to the API is made and the response is 302 (as expected) and the setting is updated in the `common_colortheme` table (see snippet below), however the theme never changes for the user. It works as expected for a non SSO user but for SSO it seems to fail.

```
inventree=# select * from common_colortheme;
 id |    name     |      user       
----+-------------+-----------------
  1 | dark-reader | admin
 38 | darker      | Matthew Nickson
```

### Steps to Reproduce

1. Setup SSO provider (openid_connect)
2. Login
3. Change default theme to dark-reader
4. Note no change

### Expected behaviour

Theme updates

### Deployment Method

- [ ] Docker
- [ ] Package
- [X] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.15.1
Django Version: 4.2.12
Commit Hash: be3b22c
Commit Date: 2024-05-15

Database: postgresql
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-6.8.4-3-pve-x86_64-with-glibc2.35
Installer: GIT

Active plugins: False

The demo site doesn't use SSO so this doesn't apply

### Relevant log output

One slightly interesting thing from the server.err.log is this for the SSO user (there are multiple for each role so I have just included one below):
```
/opt/inventree/env/lib/python3.10/site-packages/django/core/cache/backends/base.py:117: CacheKeyWarning: Cache key contains characters that will cause errors if used with memcached: ':1:role_Matthew Nickson_return_order_delete'
  warnings.warn(warning, CacheKeyWarning)
```

matmair:
This PR switches the `ColorTheme` model to use a strong ForeignKey instead of a string of the username. This should solve problems with association.

Note: User themes only work once the auth cycle is fully finished (ie. once you are on the Home page)

Closes #7257


matmair:
@SchrodingersGat thoughts? I am unable to find a technical reason to use why a string was used for the ref

SchrodingersGat:
@matmair LGTM :)

SchrodingersGat:
@matmair thanks for the fix

PR #7562, opened 2024-07-05
Ensure schema generation is more stable

### LINKED ISSUE #6882: API schema generation is not stable between dev enviroments
With the recent problems, I discovered that the API schema generation is highly unstable between dev and CI environments.

Main problems:
- plugin URLs are included by default
- the local currency config influences descriptions/enums
- server endpoints changes

I think at least 1 and 2 should be changed by default to make comparisons simpler

_Originally posted by @matmair in https://github.com/inventree/InvenTree/issues/6836#issuecomment-2023123850_

matmair:
This PR adds changes to the `schema` command to make the schemas generated on different systems more comparable. For this, a few settings are enforced by default:
- The server URL is fixed
- Plugins are turned off
- Currencies are set to AUD,CNY,EUR,USD for generation only

To enable this 2 auxiliary changes are made:
- `get_global_setting` can now be supplied with an environment variable that overwrite the database values
- the schema task has now a no_default switch to disable the defaults

Closes #6882


SchrodingersGat:
@matmair very clean implementation, I like it!

PR #7563, opened 2024-07-05
Add substitute details to BOM export

### LINKED ISSUE #5179: [FR] Add substitute details to BOM export

### Problem statement

When exporting a BOM (.xlsx) with substitute parts it only includes the substitute part name and description. As these are not necessarily unique fields, it would be helpful to also include the IPN.

Additionally, when sharing the Bill of Materials with an external manufacturer. They need to know the parts manufactuer and supplier details. So, adding these additional columns would also help.

### Suggested solution

Add the following columns to the BOM export:

- Substitute IPN
- Substitute Manufacturer
- Substitute MPN
- Substitute Supplier
- Substitute SKU

As there may be multiple manufacturers and suppliers, they should follow the same structure as Inventree uses for the standard parts. So, for example:
- Substitute_Part1_Manufacturer_0
- Substitute_Part1_Manufacturer_1
- Substitute_Part2_Manufacturer_0
- etc.

Instead of using Substitute parts, I can list multiple manufactures under each standard part. But this causes other issues for example, if the standard part is used in multiple products then that enforces that the substitutes are allowed in all products. This defeats the purpose of having BOM specific substitutes. 

### Examples of other systems

_No response_

- [ ] I want to develop this.

mredpath2:
The following changes to BOM Export as per #5179 :

1. Added substitute part IPN to BOM export when "Include Alternative Parts" is selected
2. Add Part Revision to BOM export as requested by @hugozijlmans

I also thought about adding substitute manufacturer and supplier details. But I wasn't sure if this would clog up the BOM export with too many columns if there were multiple alternates and they each had multiple SPN and MPN? So, I haven't included this change. 

Closes #5179


SchrodingersGat:
@mredpath2 thanks for the additions :)

SchrodingersGat:
You will need to address the style issues so that the rest of the tests can run

mredpath2:
> You will need to address the style issues so that the rest of the tests can run

Should be resolved now :)

SchrodingersGat:
@mredpath2 you will also need to update the unit test which checks the columns in the generated BOM file:

image

matmair (approved):
Looking great!

hugozijlmans:
Thanks @mredpath2. I'm not sure if many people were missing this feature however I'm very glad it's implemented as usually we use partnumber as BOM line item but for custom items (in our case PCB or PCBA) it's very helpfull to use the revision to show it's still the same part but a new version.

SchrodingersGat:
@mredpath2 thanks for the contribution :)

PR #7580, opened 2024-07-07
Add variable to stop creation of admin account

matmair:
Adds option to disable admin account creation with the installer


SchrodingersGat:
Is this new option discoverable, either in the docs on in the installer help?

matmair:
No, I need it mostly for repeatable e2e testing

PR #7581, opened 2024-07-08
Add cleanup task

### LINKED ISSUE #7559: [FR] Add invoke task to remove compiled files

### Problem statement

Some parts of InvenTree create custom compiled files (translations, static, python byte code) that means that static, reproducible deployment is hard with board tools.

### Suggested solution

Add an invoke task to remove all dynamic files created by `inv update`

Write custom bash scripts (this is how I currently do it) and keep them up to date in a separate place. I can do that with the deep insights into the inner workings of our deployment methods but it is not accessible to the wider user base.

### Examples of other systems

_No response_

- [X] I want to develop this.

matmair:
This PR introduces a task to clean up compiled files. This mainly is useful for the Debian package as it enables us to clean generated files before installing a new version. This solves a few potential file conflicts and removes the warnings you get on updates.

Closes #7559


tasks.py
```diff
@@ -1350,3 +1350,17 @@ def docs_server(c, address='localhost:8080', compile_schema=False):
         c.run('python docs/extract_schema.py docs/schema.yml')
 
     c.run(f'mkdocs serve -a {address} -f docs/mkdocs.yml')
+
+
+@task
+def clear_generated(c):
+    """Clear generated files from `inv update`."""
+    # pyc/pyo files
+    c.run('find . -name "*.pyc" -exec rm -f {} +')
```
SchrodingersGat, line 1359: Can we ensure that `c.run` is executed from the local dir, no matter where the `invoke` command is called from? Otherwise this may present an issue

matmair: I have added a general helper to ensure this

SchrodingersGat:
@matmair please fix conflicts and I'm happy to merge :)

SchrodingersGat:
Thanks for the improvements!

PR #7585, opened 2024-07-08
Revision Improvements

SchrodingersGat:
This PR enhances support for part "revisions" - now in addition to the existing "revision" field, the "Part" model now has the "revision_of" field, which points to another part. This allows a particular part to be "deprecated" but kept in the database.

User interface elements have been added (in the PUI in particular) to easily switch between part revisions, and make them obvious when available

- Closes https://github.com/inventree/InvenTree/issues/5028
- Ref: https://github.com/inventree/demo-dataset/pull/70
- Adds `revision_of` field to the `Part` model
- Implements rules for managing revisions
- Adds (simple) support for revisions in legacy UI
- Adds full support for revisions in new UI

### TODO

- [x] Update / improve revision documentation
- [x] Add playwright tests

### Screenshots

The "revision selector" dropdown appears for parts where there are multiple revisions available:

image

image

Clicking on a particular revision takes you to that revision.

--------------------------------------

looking for feedback at this stage, before progressing much further.

@matmair @wolflu05 @bberg115


src/frontend/src/pages/part/PartDetail.tsx
```diff
@@ -878,6 +984,40 @@ export default function PartDetail() {
               setTreeOpen(true);
             }}
             actions={partActions}
+            detail={
+              enableRevisionSelection ? (
+                <Stack gap="xs">
+                  <Text>{t`Select Part Revision`}</Text>
+                  <Select
```
wolflu05, line 991: Is there a reason why you do all that manually? Wouldn't a StandaloneField in combination with the RelatedModelField work?

wolflu05 (commented):
The base implementation looks really promising for larger instances that have the need of revisioned parts.

src/backend/InvenTree/build/serializers.py
```diff
@@ -205,7 +207,7 @@ class Meta:
     quantity = serializers.DecimalField(
         max_digits=15,
         decimal_places=5,
-        min_value=0,
+        min_value=Decimal(0),
```
matmair, line 210: Q (non blocking): Is this essential to revisions?

src/backend/InvenTree/company/serializers.py
```diff
@@ -166,6 +166,8 @@ def annotate_queryset(queryset):
 
     image = InvenTreeImageSerializerField(required=False, allow_null=True)
 
+    email = serializers.EmailField(required=False, default='', allow_blank=True)
```
matmair, line 169: Q (Non blocking): Is this related to revisions?

src/backend/InvenTree/part/models.py
```diff
@@ -661,6 +661,49 @@ def validate_ipn(self, raise_error=True):
...
+                    'revision_of': _('Part cannot be a revision of itself')
+                })
+
+            # Part cannot be a revision of a part which is itself a revision
+            if self.revision_of.revision_of:
+                raise ValidationError({
+                    'revision_of': _(
+                        'Cannot make a revision of a part which is already a revision'
+                    )
+                })
```
matmair, line 679: I think having trees of revisions would be helpful

src/backend/InvenTree/part/models.py
```diff
@@ -841,15 +884,13 @@ def validate_unique(self, exclude=None):
...
                 })
 
-        # Ensure unique across (Name, revision, IPN) (as specified)
-        if (
-            Part.objects.exclude(pk=self.pk)
-            .filter(name=self.name, revision=self.revision, IPN=self.IPN)
-            .exists()
-        ):
-            raise ValidationError(
-                _('Part with this Name, IPN and Revision already exists.')
```
matmair, line 851: C (blocking): This change in business rules is breaking and I would like to discuss if this should be a setting

src/frontend/src/pages/part/PartDetail.tsx
```diff
@@ -341,6 +352,7 @@ export default function PartDetail() {
         label: t`Price Range`,
         value_formatter: () => {
           const { data } = useSuspenseQuery({
+            refetchOnWindowFocus: false,
```
matmair, line 355: C (non blocking): This appears multiple times, should this maybe be a default passed to the queryclient?

src/frontend/src/tables/bom/BomTable.tsx
```diff
@@ -347,6 +349,26 @@ export function BomTable({
     table: table
   });
 
+  const validateBom = useApiFormModal({
```
matmair, line 352: Q (non blocking): Is this related to revisions?

matmair (approved):
Looks good in general.
Regarding the backend I would prefer if we would work with a bit more flexible model to enable adoption to different:
- a mixin for the `Part` model (and future ones like Orders, Stock, ...) that adds a `revision_of` and a `revision` FK field
- a generic `Revision` model that has metadata enabled and is used to keep information about the creation of a revision (user, date, note, delta)

As for creating new revisions: it would be great if we could implement the `StateTransitionMixin` and `handle_transition` to add the possibility for plugins to influence the creation of new revisions. I could provide that as a separate PR too.

src/frontend/src/pages/part/PartDetail.tsx
SchrodingersGat: Because we construct the list of options locally, (rather than fetch directly from the field) and then pass to the select, it won't quite work (without some refactoring of the `RelatedModelField`

src/backend/InvenTree/build/serializers.py
SchrodingersGat: No, looks like these have bled through from my attempts to support the new django-rest-framework requirements. Not essential here, but doesn't hurt either

src/backend/InvenTree/company/serializers.py
SchrodingersGat: No, looks like these have bled through from my attempts to support the new django-rest-framework requirements. Not essential here, but doesn't hurt either

src/backend/InvenTree/part/models.py
SchrodingersGat: I thought about this, but having a potentially recursive relationship here makes it very hard to determine the total number of revisions for a part. This is why we utilize `MPTT` for the part variant - but we cannot have a second tree structure attached to the part model.

The limitation here is not too restrictive ( I think ) - as long as you point the "revision_of" to the same base part, you can have as many revisions as you like.

Feel free to suggest another approach!

src/backend/InvenTree/part/models.py
SchrodingersGat: Good catch this was not an intentional removal. Have added it back in.

src/frontend/src/tables/bom/BomTable.tsx
SchrodingersGat: I added it in as I was creating a number of revision parts (with BOMs) for the demo dataset - https://github.com/inventree/demo-dataset/pull/70

src/frontend/src/pages/part/PartDetail.tsx
SchrodingersGat: Good suggestion, I have added a default value to the query manager

src/frontend/src/pages/part/PartDetail.tsx
wolflu05: Ok, makes sense. Was just thinking from a Quick Look that the endpoint you fetch is a simple related model endpoint which the related model field could handle.

src/backend/InvenTree/part/models.py
matmair: I think from a user perspective it is limiting as you need to manually re-structure multiple parts to achieve the same result. But I can see this taking a lot of effort to include in the PR.

matmair (approved):

bberg115:
Sorry for the delayed response - I think the screen shots and the conversation above all seems good! 

It's not something we've discussed, as far as I know, but how will this change effect the build order process? Will you be able to choose the revision of a BOM or be forced to latest?

SchrodingersGat:
> @SchrodingersGat what is your opinion on my statement about the model structure? I can see this easily developing into something similar to reports/attachments where we duplicated code to multiple places for a generic function till we needed a refactor to simplify it back down.

It is a fair point. However at this stage, I can't see a clear delineation of which parts would be "generic" and which would be split into the separate model. Currently, I have created no new models, and this is a very simple change to "revert" if we need to. I would like to suggest that we wait until such time as we do want to add "revisions" to other models, and at that point, we make the effort to genericise the implementation. Until then, I'm not sure how we would implement it and we might have to refactor that anyway.

SchrodingersGat:
> It's not something we've discussed, as far as I know, but how will this change effect the build order process? Will you be able to choose the revision of a BOM or be forced to latest?

@bberg115 not in this initial stage. Happy to discuss that as part of a future PR.

SchrodingersGat:
@bberg115 please feel free to raise any issues that you see with this new implementation :)

PR #7599, opened 2024-07-09
[Bug] Project code fix

SchrodingersGat:
Fixes bug introduced in data importer refactor PR which broke the "project code" field in the order serializer.

Reverts the bug and also introduces some new unit tests to prevent regression.


matmair (approved):
LGTM

PR #7606, opened 2024-07-09
Bump development deps

matmair:
This bumps the various pre-commit dependencies, specifically extending codespells detections and re-adding direct attribution for dependencies in requirements.txt's.


matmair:
This is ready for review/merge

PR #7610, opened 2024-07-10
Fix "Scan barcode" modal input

### LINKED ISSUE #7529: "Scan barcode" modal does not accept any special characters

### Describe the bug*

When scanning a 2D supplier barcode using a USB barcode scanner into the "Scan barcode" modal, the Record and Group separator symbols do not get sent along to the server. I think this happens because these control characters get filtered out by the browser before they reach the input field.

Additionally, when the barcode scanner enters these characters using something like `ctrl`+ `]` or `ctrl` + `6`, these combinations still get sent to the browser itself, changing tabs or doing other stuff. As far as I can see this is because the input field in the modal only stops the propagation of `keyup` events, not the `keydown` events that actually trigger these shortcuts.

### Steps to Reproduce

1. Open the "Scan barcode modal"
2. Scan a 2D barcode with record and group separators

### Expected behaviour

The scanned characters get added to the input field and sent to the server.

### Deployment Method

- [X] Docker
- [ ] Package
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.15.4
Django Version: 4.2.12
Commit Hash: fa1a9da
Commit Date: 2024-06-16

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.8.6-100.fc38.x86_64-x86_64-with
Installer: DOC

Active plugins: False

### Relevant log output

_No response_

tomvaneyck:
Fixes #7529.
I've taken another look, and although `event.preventDefault()` stops some shortcuts from being executed by the browser, things like `ctrl`+`T` and `ctrl`+`Tab` are still captured and executed[^1]. However, as far as I know no barcodes contain characters that collide with these shortcuts. It of course also depends on the barcode reader.

I've also addressed the issue of receiving the "alt key codes" by pressing `alt`+`<some numbers>`. Our barcode scanner can be set to only emit these codes which avoids the shortcut problem from above, but it takes way longer to input a barcode this way.

[^1]: https://stackoverflow.com/questions/59952382/using-preventdefault-to-override-opening-new-tab

### DIFF
--- a/src/backend/InvenTree/templates/js/translated/barcode.js
+++ b/src/backend/InvenTree/templates/js/translated/barcode.js
@@ -315,8 +315,32 @@ function barcodeDialog(title, options={}) {
 
         var barcode = $(modal + ' #barcode');
 
+        // The control characters that are used in barcodes.
+        const controlKeys = {
+            "BracketRight": String.fromCharCode(29), // Group separator
+            "Digit6": String.fromCharCode(30), // Record separator
+        }
+        let altValue = 0;
+
         barcode.keydown(function(event) {
-            event.preventDefault();
+            if (event.ctrlKey) {
+                // Prevent most of the keyboard shortcuts.
+                // Not all of them will be blocked, browser don't allow this:
+                // https://stackoverflow.com/questions/59952382/using-preventdefault-to-override-opening-new-tab
+                event.preventDefault();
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@tomvaneyck thanks for submitting this fix. I don't have a way of checking this manually but it looks reasonable.

PR #7614, opened 2024-07-10
Add debug option to the sheet printer plugin

martonmiklos:
I spent some time debugging an issue with my labels printed through the sheet printer (looks to be a wheasyprint issue).
At some point I went ahead and added a debug settings to being able to see the raw HTML output. It might be useful to others in the future.

### DIFF
--- a/src/backend/InvenTree/plugin/builtin/labels/label_sheet.py
+++ b/src/backend/InvenTree/plugin/builtin/labels/label_sheet.py
@@ -12,6 +12,7 @@ import weasyprint
 from rest_framework import serializers
 
 import report.helpers
+from InvenTree.helpers import str2bool
 from plugin import InvenTreePlugin
 from plugin.mixins import LabelPrintingMixin, SettingsMixin
 from report.models import LabelOutput, LabelTemplate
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@martonmiklos great addition, thanks!

src/backend/InvenTree/plugin/builtin/labels/label_sheet.py
```diff
@@ -64,9 +65,22 @@ class InvenTreeLabelSheetPlugin(LabelPrintingMixin, SettingsMixin, InvenTreePlug
...
+        }
+    }
 
     PrintingOptionsSerializer = LabelPrintingOptionsSerializer
+    debug = None
+
+    def in_debug_mode(self):
+        """Check if the plugin is printing in debug mode."""
+        if self.debug is None:
+            self.debug = str2bool(self.get_setting('DEBUG'))
```
SchrodingersGat, line 83: Oops, this function should return a value, right?

SchrodingersGat:
@martonmiklos I think I merged this too hastily - please see comment above. The `in_debug_mode` method should return a value. Please submit a new PR to address this :)

src/backend/InvenTree/plugin/builtin/labels/label_sheet.py
wolflu05: Also, what's the purpose of setting self.debug if it's not defined? The plugin settings are cached. And secondly this means that we cannot disable debug mode at runtime anymore.

src/backend/InvenTree/plugin/builtin/labels/label_sheet.py
```diff
@@ -135,11 +149,15 @@ def print_labels(
...
-        html = weasyprint.HTML(string=html_data)
-        document = html.render().write_pdf()
+        if self.in_debug_mode():
+            # Render HTML to PDF
+            html = weasyprint.HTML(string=html_data)
+            document = html.render().write_pdf()
+
+            output.output = ContentFile(document, 'labels.pdf')
+        else:
+            output.output = ContentFile(html_data, 'labels.html')
```
wolflu05, line 159: Isn't that just swapped? I mean if in debug mode it should return the html, but you return the pdf.

SchrodingersGat: @wolflu05 correct, I was way too hasty on this. @martonmiklos please review

martonmiklos: Ah, copy paste engineering tricked me, let me fix it.

PR #7622, opened 2024-07-11
Fix issues made in #7614

martonmiklos:


src/backend/InvenTree/plugin/builtin/labels/label_sheet.py
```diff
@@ -82,6 +82,8 @@ def in_debug_mode(self):
         if self.debug is None:
             self.debug = str2bool(self.get_setting('DEBUG'))
 
+        return self.debug
```
SchrodingersGat, line 85: What is the reasoning behind first checking if self.debug is None? Surely would be simpler to just return the setting check

src/backend/InvenTree/plugin/builtin/labels/label_sheet.py
```diff
@@ -149,7 +151,7 @@ def print_labels(
         # Render to a single HTML document
         html_data = self.wrap_pages(pages, **document_data)
 
-        if self.in_debug_mode():
```
SchrodingersGat, line 152: I think this would read cleaner if you flipped the cases rather than negating the logic check

src/backend/InvenTree/plugin/builtin/labels/label_sheet.py
martonmiklos: I have to admit that I have actually I have copy pasted this from here:
https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/plugin/builtin/labels/inventree_label.py#L46
I think the original author's intention was to cache the settings.

SchrodingersGat: Ok, this caching behaviour does not work as expected. Once you first print a label, if you then change the "debug" setting, the original value of the setting is still used (until the server is restarted).

It is also a bug in the implementation you copied from.

Can you please remove the caching behaviour? I have tested it locally and it is not correct.

If possible it would be great if you could also remove the caching behaviour from the other plugin where you copied it from.

martonmiklos: I removed caching, tested it works on the fly.
I am wondering what was the original motivation of this the caching.

 PR coming soon for the other plugin.

SchrodingersGat:
Thanks for the fix @martonmiklos

PR #7625, opened 2024-07-11
Correct tooltips

martonmiklos:
In the test templates list there are two actions where the tooltip says test result, however it relates to test template: 
kép


SchrodingersGat:
Cheers @martonmiklos

Issue #7571, opened 2024-07-06
[Importer] Support Default Values


SchrodingersGat:
- Follow up to https://github.com/inventree/InvenTree/pull/6911
- Ref: https://github.com/inventree/InvenTree/issues/3101

### Tasks

- [ ]  Support "default values" when importing dataset
- [ ] Enable per-column specification of "default" values in the admin portal
- [ ] Show "original row data" for each row when importing data

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/inventree) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/inventree/InvenTree/issues/7571">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/inventree/InvenTree/issues/7571/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/inventree/InvenTree/issues/7571/pledge.svg">
</picture>
</a>

PR #7635, opened 2024-07-12
[PUI] Add BOM import tool

SchrodingersGat:
Adds a specific BOM import tool for the new interface:

### New Features

- Uses the new generic API-driven importer
- Adds "override" values which can be defined per import session
- In the context of a part BOM, the "part" field is fixed, but other data can be imported
- Allows "default" values to be defined per column for an import session
- Default values are used if no value found in the file (for that column / row combination)

### Related Issues

- Closes https://github.com/inventree/InvenTree/issues/7571

### Example Screenshots

image

image

image


wolflu05:
Nice addition, was just about thinking to suggest implementing a `Literal` value field instead of mapping a column, to manually supply a value instead of adding e.g. a category id to a list of part names to import.

PR #7640, opened 2024-07-12
Remove caching of the DEBUG settings from the label printer plugin

martonmiklos:
As we discussed in #7622


SchrodingersGat:
Thanks @martonmiklos

PR #7645, opened 2024-07-13
Switch to zoneinfo from pytz

matmair:
We target 3.9 which has zoneinfo for timezone info instead of pytz.


SchrodingersGat:
Nice. Should we also remove pytz from the requirements?

matmair:
@SchrodingersGat sadly not, some of our dependencies still depend on it

matmair:
I need to figure out why CI works on my machine but not on GH

matmair:
This is more complex than initially thought. https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html
I am amazed we have not had problems with pytz before.

SchrodingersGat:
Wow, what a deep dive!

SchrodingersGat:
@matmair it would be great to get this one closed out if you have some time to assign to it :)

matmair:
@SchrodingersGat this is more a skill not a time issue ;-) I can look into it after sso and selection lists

SchrodingersGat:
Ok, LMK if you are banging your head against the wall on it!

matmair:
@SchrodingersGat can you take a look at this? Especially the change in the test assertations. Maybe I am missing something but the previous assertations seem strangely off by a few minutes.

1337joe:
It looks like this is a somewhat known problem with pytz:
https://stackoverflow.com/questions/27531718/datetime-timezone-conversion-using-pytz
https://stackoverflow.com/questions/60996205/python-datetime-timezone-conversion-off-by-4-minutes

Long story short, using `replace(tzinfo=...` (what InvenTree was using) can make invalid assumptions, where `localize(time)` would try to correct for that.

For what it's worth, your updated test values all look correct to me.

matmair:
@SchrodingersGat ready for review and merge; I have marked it as breaking as a precaution, I do not think it should break anything

SchrodingersGat:
@matmair seems like a minimal change overall, in the end. Thank you for looking into this one.

Issue #7600, opened 2024-07-09
Wrong item purchase price when using Enter


dieselburner:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When entering purchase order item I use tab and enter keys, and as a result I believe the Auto Pricing is not automatically unchecked when I hit enter after updating item price. As a result, added item loses manually entered price and shows calculated pricing as per supplier price breaks.

Add item:

![Screenshot from 2024-07-09 12-23-43](https://github.com/inventree/InvenTree/assets/1078344/dad55ad2-2bf2-4215-8a11-4ce79639ff3a)

Result:

![Screenshot from 2024-07-09 12-19-17](https://github.com/inventree/InvenTree/assets/1078344/f4f31b10-df2f-4cfd-8238-d92a33fe0dff)

### Steps to Reproduce

1. Create purchase order
2. Hit _Add Line Item_
3. Select _Part_
4. Use Tab key to navigate to _Puchase Price_
5. Enter required price
6. Hit Enter key

### Expected behaviour

Manually entered purchase price is shown in entered item.

### Deployment Method

- [X] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

# Version Information:
InvenTree-Version: 0.15.4
Django Version: 4.2.12
Commit Hash: fa1a9da
Commit Date: 2024-06-16

Database: sqlite3
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-3.10.108-x86_64-with
Installer: DOC

### Please verify if you can reproduce this bug on the demo site.

- [X] I can reproduce this bug on the demo site.

SchrodingersGat:
@dieselburner thanks for reporting.

rocheparadox:
@SchrodingersGat , I will try to fix this.

SchrodingersGat:
Thanks @rocheparadox

rocheparadox:
This issue arises due to the underlying model of the events of forms as defined in forms.js

https://github.com/inventree/InvenTree/blob/76f8a2ee9e42c383383c846b892d1874720ac089/src/backend/InvenTree/templates/js/translated/forms.js#L1634

The jquery documentation for the *change* event goes as follows

> For select boxes, checkboxes, and radio buttons, the event is fired immediately when the user makes a selection with the mouse, but for the other element types the event is deferred until the element loses focus.

[...]

PR #7660, opened 2024-07-15
Added onInput event for fields in forms - fix for issue #7600

rocheparadox:
An onInput event is added for fields in forms that gets triggered everytime an input is detected in the field. This fixes the purchase price not being updated when the enter key is hit.

- Closes https://github.com/inventree/InvenTree/issues/7600


SchrodingersGat:
@rocheparadox thanks for the fix :)

PR #7667, opened 2024-07-16
Add notes to mp and sp

wolflu05:
This PR adds the notes field to the manufacturer and supplier part. It adds a frontend implementation for both, CUI and PUI.


matmair (approved):
LGTM, very clean

SchrodingersGat (approved):

SchrodingersGat:
Great work @wolflu05

wolflu05:
Thanks for that feedback, but without all of your awesome refactoring into mixins this would not have been possible to implement so easy. I was really impressed after starting looking into it, it took me only like 20 minutes to figure out how the mixins should be used and implement.

PR #7678, opened 2024-07-18
Fix loading sequence for PUI form fields

SchrodingersGat:
- Ensure that "existing" values are honored
- Change hook execution ordering
- Delay field display until all fields are loaded

### DIFF
--- a/src/frontend/src/components/forms/ApiForm.tsx
+++ b/src/frontend/src/components/forms/ApiForm.tsx
@@ -139,6 +139,7 @@ export function OptionsApiForm({
       if (!props.ignorePermissionCheck) {
         fields = extractAvailableFields(response, props.method);
       }
+
       return fields;
     },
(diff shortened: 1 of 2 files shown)


wolflu05:
This PR has broken the forms if the server responds with an error:

1. Open the netlify preview for this PR and one for another previous PR
2. Goto a part and open the edit modal
3. Change the "Units" field to a non existing physical unit. E.g. `AA`
4. Click "Submit"
5. In the preview from this PR, the form is just reset to the default values, and no error is shown. In the other preview the validation error is shown.

SchrodingersGat:
@wolflu05 thanks for the report. I'll look into this (and add some additional playwright tests to ensure no further regressions)

SchrodingersGat:
@wolflu05 fix incoming in https://github.com/inventree/InvenTree/pull/7689

PR #7655, opened 2024-07-15
Optimize PUI package delivery for package installs 

matmair:
This PR:
- Adds git commit identifiers to the frontend CI-builds
- Includes the frontend resources with tag-builds of the package
- Adds a check to the frontend-download to ensure the frontend is only downloaded when necessary
- Adds better readable package build logs

It does not fix #7338 in all cases as the frontend can only by downloaded automatically for tagged releases

PR #7679, opened 2024-07-18
Backport Optimize PUI package delivery for package installs

matmair:
Backport for https://github.com/inventree/InvenTree/pull/7655

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -554,6 +554,8 @@ jobs:
         run: cd src/frontend && yarn install
       - name: Build frontend
         run: cd src/frontend && yarn run compile && yarn run build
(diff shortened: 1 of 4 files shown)


matmair:
I have no idea why but htis PR is not on stable or 0.15.x anymore. @SchrodingersGat did I do something obviously wrong with this backport?
https://github.com/inventree/InvenTree/blob/stable/contrib/packager.io/before.sh

SchrodingersGat:
Looks like this got merged into "stable" and not "0.15.x" . The "stable" branch gets overwrite on release, so this should be pushed to the 0.15.x branch. I guess I also did not notice this when I merged it. If you want to re submit against the correct branch I'll issue another point release.

matmair:
I will re-submit

PR #7684, opened 2024-07-18
Migrate Icons to Tabler icons and integrate into PUI

### LINKED ISSUE #7237: [FR] PUI & Category / Location / ... icons

### Problem statement

PUI doesn't show any defined icons, neither on e.g. stock location edit icon field nor the icon on the details pages, breadcrumbs, ...

### Suggested solution

Because we are now using tabler icons in PUI, the old fontawesome icons are not that easily integratabtle. In this FR I would start to suggest, that we should build our own, per instance, icon library into inventree where:
1. all icons can be get via a single url in json format
2. maybe the server can generate a custom sprite image with all icons available on a single picture where the icon position is sent in the json endpoint referenced in point 1
3. per default we should make the tabler icons available? But maybe add the font awesome icons as opt-in function?
4. admins can upload custom icons (not all icons users want are available on the tabler/FA icons package)
5. clients can the use this information to provide consistent icons across systems (PUI, App, reports, ...) with no differences between icon package dependency versions 

Of course we would need some migration from front awesome icons to the new icon system, but that is TBD.

Currently icons are integrated in:
- location icon
- location type icon
- category icon
But that could be later extended, if we build this generic, and maybe add an icon field type to the PUI form framework.

Just add the font awesome icons to PUI too, but I think that's not a very clean solution, and users also are not able to define custom icons. Because that's also a big point for me, not all icons are available in the font awesome icons package.

### Examples of other systems

_No response_

- [x] I want to develop this.

wolflu05:
This PR is a proposal which migrates the category/location icons to the tabler icons package and validates the icon string on the backend. To do so, the backend needs a list of available icons, and that's why I built the whole system from the beginning a bit more modular without much more effort.

This approach has a couple of advantages compared to if we just use the PUI integrated tabler icons package:
- Server can validate the icon provided by the user and raise an Validation error if its not a valid icon
- No version mismatches between the tabler icons package and different clients (pui, cui, app, reports, ...)
- The frontend build doesn't get polluted with \~2MB of additional icon data (~+100%)
- Its independent from the client and stack, it only needs to understand a woff, woff2 or ttf font file provided by the server
- Allows icons to be used in report/label templates rendered by the server
- Users can provide their own icon packages (with specific icons for different sizes of drawers, bags, ESD bags, ... or electrical components for part categories, which are not available in the tabler icons set.) as a plugin (and also for others to use).

## TODO

- [x] Add `/api/icons` endpoint
- [x] Add tabler icons download/update script 
- [x] Validate saved icons
- [x] Bump API version
- [x] return default icon if no icon is defined from API
- [x] PUI Icon picker
- [x] PUI Render icons along the UI
- [x] CUI explain icon format and change link to the tabler icons package
- [x] CUI Render icons along the UI
- [x] Printing integration
- [x] Plugin Integration
- [x] Add migration script
- [x] Documentation
  - [x] how the icon format is structured
  - [x] how to use icons in reports
  - [x] how to create an icon pack plugin
  - [x] how to migrate icons
- [x] Tests

I suggest to add something like the following text to the release notes on the top to communicate this change:

---

## ⚠️ Migrate Icons to Tabler icons (#7684)

Starting with InvenTree 0.16, the font awesome icons for part categories and stock locations are no longer available. They are replaced by the [tabler icons](https://tabler.io/icons) package and user data is now validated to check that the icons even exist. This refactoring allowed us to:
- Build an icon picker for PUI which speeds up the process of picking the right icon.
   image
- Icons can be used natively inside of label/report templates.
- Users can provide their own, custom crafted icon packs through plugins.

After updating to this version, all existing icons need to be migrated to the tabler icons package. To simplify this process, a migration command is available.

1. Generate an icon map of all currently used icons. (Optionally, the `--include-items` flag can be used to add the referenced inventree items to the exported icons for easier identification):
    ```bash
    python src/backend/InvenTree/manage.py migrate_icons --output-file /tmp/icon-map.json
    ```
2. Edit the generated file at `/tmp/icon-map.json`. For each item in the dict, assign a `new_icon`. E.g. `ti:package:outline` (Read more about the icon format at [the docs](https://docs.inventree.org/en/stable/stock/stock/#icons))
3. Check if every new icon is valid: 
    ```bash
    python src/backend/InvenTree/manage.py migrate_icons --input-file /tmp/icon-map.json
    ```
4. If everything is valid, import the new icons now:
    ```bash
    python src/backend/InvenTree/manage.py migrate_icons --input-file /tmp/icon-map.json --import-now
    ```
5. Adjust any default icon (`PART_CATEGORY_DEFAULT_ICON` and `STOCK_LOCATION_DEFAULT_ICON`)
6. The icon migration process is done, and InvenTree can be used as before.

---

Closes #7237


wolflu05:
This PR is ready for review now. I added some detailed description to the initial PR comment. (The failed, flaky PUI test seems unrelated.)

SchrodingersGat:
@wolflu05 fantastic work here, this is a huge enhancement :) I particularly like that we can use custom icons in the report templates, and (with additional work) in the app too!

One thing I note is that when editing a "location type" (in the "admin section of PUI) - the edit modal takes significant time to open (at least one second). Perhaps there is some optimizing on the backend that could take place here?

wolflu05:
> One thing I note is that when editing a "location type" (in the "admin section of PUI) - the edit modal takes significant time to open (at least one second). Perhaps there is some optimizing on the backend that could take place here?

@SchrodingersGat The issue is not the backend, it is the frontend. I have improved the performance of the icon picker with my last commit significantly now. Please test again, it should be by far better now.

SchrodingersGat:
@wolflu05 very nice, the updates make it a lot more responsive! I'm happy to merge this in, thanks for the great work here :)

matmair:
@wolflu05 it seems like migration tests failed on this after merged: https://github.com/inventree/InvenTree/actions/runs/10069548588/job/27836706327
https://github.com/inventree/InvenTree/actions/runs/10069548597/job/27838227324 - any idea if this is related?

SchrodingersGat:
I noticed this too, trying to get to the bottom of it. Strangely the migration testing library has a new version out today, with seems related, but we are not using that yet. Very weird.

I'm away from my computer for a couple of days so unable to dig into this right now

wolflu05:
Very strange. The only thing I touched related to migrations was this migration:

https://github.com/inventree/InvenTree/blob/96abd0898c043d0cdbab846647d15bc189709bdf/src/backend/InvenTree/part/migrations/0127_remove_partcategory_icon_partcategory__icon.py#L7-L29

Could this `migrations.SeparateDatabaseAndState(...)` have caused this? Not sure how to solve this otherwise, this was the only solution I found online. I just wanted to rename the models `icon` property to `_icon`, but keep the db column at `icon` to add a getter and setter for `.icon`. Without that separation of database and state operations, it would first generate SQL to rename the column to `_icon` and then a second query to rename it back to `icon`. Any ideas how to solve this somehow else?

matmair:
I am not sure, this all is very strange. They should have run due to changes in migrations.

I am not sure this PR was the root cause but it seems to be the first one where this appeared in the last few days.

wolflu05:
Let's continue discussion on #7730

PR #7689, opened 2024-07-19
[PUI] form error fix

SchrodingersGat:
- Fixes bug introduced in https://github.com/inventree/InvenTree/pull/7678
- Reduce duplicate API queries when loading a form
- Add unit testing via playwright

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -52,7 +52,7 @@
         "dayjs": "^1.11.10",
         "embla-carousel-react": "^8.1.6",
         "html5-qrcode": "^2.3.8",
-        "mantine-datatable": "^7.11.1",
+        "mantine-datatable": "^7.11.2",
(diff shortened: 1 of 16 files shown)


SchrodingersGat:
@wolffam please LMK if you spot any issues now :)

matmair:
@SchrodingersGat did you mean to tag Lukas? Because that is not him

wolflu05 (commented):
Thanks for that quick fix, that indeed fixes the error with the not showing errors, but I now found out, that that your initial changes have also broken the edit forms (see video), which is not fixed by this fix.

https://github.com/user-attachments/assets/3560d221-adc2-4fb1-adaf-91bf1df3067b

SchrodingersGat:
@wolflu05 thanks, I think I have fixed this now and added some more end-to-end tests to cover it.

PR #7691, opened 2024-07-19
[PUI] Tab permissions

SchrodingersGat:
Hide top-level navigation tabs based on the users permissions.

e.g. if the user does not have the "purchase order" role, the tab will be hidden:

image

Ref: https://github.com/inventree/InvenTree/discussions/7654


SchrodingersGat:
Oops, branched from wrong commit...

PR #7699, opened 2024-07-21
Bump docker image alpine base from 3.18 to 3.20

matmair:
Our base image uses a year-old alpine (https://endoflife.date/alpine) that has some os-level vulnerabilities. I would like to propose a move to 3.2o - which does not have those vulnerabilities and seems stable.

PR #7717, opened 2024-07-23
Revert postgres version to 13

SchrodingersGat:
Bump from version 13 to 14 causes data incompatibility and needs to be very carefully considered.

Ref: https://github.com/inventree/InvenTree/pull/7699


matmair:
As mentioned in https://github.com/inventree/InvenTree/pull/7699#issuecomment-2244855413 I think the quickest / cleanst solution to get a "safe" image out to latest is to just revert #7699 completly

PR #7730, opened 2024-07-24
try to fix tests

wolflu05:
This PR fixes the failing migrations tests first discovered after the merge of #7684 in: https://github.com/inventree/InvenTree/actions/runs/10069548588/job/27836706327
https://github.com/inventree/InvenTree/actions/runs/10069548597/job/27838227324 

Somehow this seems to not be directly related to that PR, but was first caught after that merge. It may be related to https://github.com/wemake-services/django-test-migrations/pull/294 ? Because when upgrading to 1.4.0 from 1.3.0 the issue is gone.

### DIFF
--- a/src/backend/InvenTree/part/test_migrations.py
+++ b/src/backend/InvenTree/part/test_migrations.py
@@ -110,7 +110,6 @@ class TestParameterMigrations(MigratorTestCase):
         t2 = PartParameterTemlate.objects.create(name='Template 2', units='AMPERE')
 
         # Create some parameter values
-        print('===>', repr(Part), repr(a), repr(PartParameter), repr(t1))
         PartParameter.objects.create(part=a, template=t1, data='1.0')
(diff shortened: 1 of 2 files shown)


wolflu05:
Im trying to debug this issue. What I already found out:

1. when reverting back to the commit prior to my icons merge commit, the migration tests pass
2. when removing the separate database and state operations and place the unnecessary migrations directly in the operations list, it's still not working
3. somehow this seems to not be the the correct instance of part, was trying to debug that with my last run

matmair:
Maybe the issue is in the testing code itself?

SchrodingersGat:
I also found that updating to the latest version of `django-test-migrations` "fixed" the tests too....

wolflu05:
You're right, updating "fixed" the issues. (See my last run)

matmair (approved):
LGTM; great that we figured that out

SchrodingersGat:
Nice work, thanks @wolflu05

PR #7739, opened 2024-07-25
Backport Optimize PUI package delivery for package installs

matmair:
Fix for wrong merge target in https://github.com/inventree/InvenTree/pull/7679


SchrodingersGat:
Thanks @matmair

PR #7747, opened 2024-07-27
Add option to disable MFA

SchrodingersGat:
Adds installation level option to disable MFA entirely.

By default, MFA is enabled, and requires manual intervention (at the sysadmin level) to turn it off.

### DIFF
--- a/docs/docs/start/config.md
+++ b/docs/docs/start/config.md
@@ -297,6 +297,10 @@ Alternatively this location can be specified with the `INVENTREE_BACKUP_DIR` env
 
 InvenTree provides allowance for additional sign-in options. The following options are not enabled by default, and care must be taken by the system administrator when configuring these settings.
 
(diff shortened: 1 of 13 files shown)


matmair:
@SchrodingersGat would it not be easier and less error prone to stop addition of new factors? That would be a much smaller diff and less logic branches that could cause issues

wolflu05:
I thought about also suggesting that. That would be much easier.

SchrodingersGat:
@matmair sure, if you can suggest a cleaner alternative I'd be happier with that. This did turn out a bit clunky.

matmair:
@SchrodingersGat https://github.com/matmair/InvenTree/tree/disable-mfa-ref would be a minial implementation without frontend changes

SchrodingersGat:
@matmair that looks cleaner. Can you submit a PR?

PR #7757, opened 2024-07-29
Add option to disable MFA

matmair:
This PR adds a installation level option to disable MFA entirely. MFA is kept enabled by default but can be turned off via `config.yaml` or environment variable.

Replaces https://github.com/inventree/InvenTree/pull/7747
See https://github.com/inventree/InvenTree/discussions/7746


matmair:
@SchrodingersGat do we want to reflect the disabling of MFA in the UI? I see this as a very fringe use case - mostly for the demo site

SchrodingersGat (approved):

SchrodingersGat:
Nice one @matmair thanks for implementing

Issue #7709, opened 2024-07-22
Automatic plugin static files collection broken after install 


wolflu05:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

I have broken down the issues with plugin static files into multiple smaller ones:

### ~~1. Static files path~~ (fixed with #7763)

<details><summary>Details</summary>
<p>

If an `AppMixin` plugin before had a `static` folder and the django app was loaded, `invoke static` had already collected the static files in the root of the static folder so a file in `my-plugin/static/panel.js` was served by `/static/panel.js`.
*(to remove clashes between files with the same name from multiple plugins, I placed my static files in `my-plugin/static/my-plugin/panel.js`, so it was served by `/static/my-plugin/panel.js`)*

Now after the mentioned PR, a file in `my-plugin/static/panel.js` is served by `/static/plugins/my-plugin/panel.js`. I do not think this change is bad, it just needs to be documented as breaking. One good think would be to provide a `{% plugin_static "panel.js" %}` template tag which resolves to the particular plugin static url (`/static/plugins/my-plugin/panel.js`) and we back port that to 0.15 which resolves to `/static/panel.js`

</p>
</details> 

### 2. Automatic static file copying and migrations are not run after activating a plugin

I can reproduce this in a fresh docker setup (but not in my dev setup running locally). But no idea why only in a production docker setup. (Sometimes only very few times it actually succeeds, but then the migrations are run two times, the first time without any error, and the second they error out, because obviously the table already exists, that the migrations should create). Maybe that's some type of race condition.

0. Goto AdminCenter > Plugins > Settings and Activate plugins that uses apps
1. Goto AdminCenter > Plugins > Install plugin and use `inventree-bulk-plugin` (or `inventree-kicad`, but that only tests with migrations)
2. Activate that plugin
3. See this error in the docker logs: `inventree-worker  | Plugin registry has no record of plugin 'inventree-bulk-plugin'`
   - See that the folder `static/plugins/inventree-bulk-plugin` does not exist (folder should exist)
   - See that the migrantions have not run (should have run)
5. Run `invoke update` the files now exists and the migrations have run

### ~~3. When deactivating a plugin, the the static **files** don't get removed~~

When deactivating a plugin, the static files get don't removed automatically, a `invoke static` is needed. And even then, nested folders remain.

### ~~4. AppMixin Plugins with static files causes an error~~ (fixed with #7763)

[...]

PR #7763, opened 2024-07-30
Plugin static files

wolflu05:
This PR fixes issue 1 and 4 from https://github.com/inventree/InvenTree/issues/7709#issuecomment-2244844925

This PR is marked as breaking to appear at the top of the changelog. Something like the following text should be added to the changelog for users to explain the change:

---

### ⚠️ Plugin static files (#7425) (#7763)

Previously if a plugin had the `AppMixin` and contained a `static` folder, static files where collected as part of `invoke static` like a normal django app directly in the `static` folder (e.g. `my-plugin/static/test.js` was accessible by `/static/test.js`). Now plugin static files are collected from every plugin that contains a `my-plugin/static` folder (No need for the `AppMixin` anymore) in `/static/plugins/<plugin-slug>/` meaning the `test.js` in the static folder from the previous example is now accessible from `/static/plugins/<plugin-slug>/test.js`.

The `{% plugin_static "test.js" %}` template tag from `{% load plugin_extras %}` can be used to get the correct path for a static file provided by a plugin. It is also backported to 0.15.x (#7764) and acts as a proxy to the normal `static` tag there. 

---

### DIFF
--- a/src/backend/InvenTree/InvenTree/ready.py
+++ b/src/backend/InvenTree/InvenTree/ready.py
@@ -115,6 +115,7 @@ def canAppAccessDatabase(
         'makemessages',
         'compilemessages',
         'spectactular',
+        'collectstatic',
     ]
 
     if not allow_shell:
@@ -125,7 +126,7 @@ def canAppAccessDatabase(
         excluded_commands.append('test')
 
     if not allow_plugins:
-        excluded_commands.extend(['collectstatic', 'collectplugins'])
+        excluded_commands.extend(['collectplugins'])
 
     for cmd in excluded_commands:
         if cmd in sys.argv:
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@wolflu05 so with these changes are you happy that existing plugin static files are honoured, and also the new approach?

wolflu05:
Sorry, but I don't understand your comment, could you please explain?

With this PR, the old approach using the AppMixin has no effect anymore, as the plugins get not loaded for the traditional Django static file collection. They only get copied with your added collectplugins script. That way there are no duplicated static files at two paths anymore.

Additionally I have fixed the issue with the recursive deletion.

SchrodingersGat:
@wolflu05 sorry, was not very good description.

I mean, that the new approach allows for static file collection whether or not the plugin has "app" support. So, are we happy that this is sufficient and covers all cases?

wolflu05:
Thanks for clarifying, Yes, I'm happy with it.

SchrodingersGat:
Ok, well thanks for the additional fixes and insights :)

PR #7752, opened 2024-07-29
Add how to install custom packages to the docker image section

wolflu05:
This PR is a proposal for the documentation to add some documenatation around how to install custom software in the production inventree image:

This would document a solution for #3562 and also #7737 without cluttering the official inventree docker image with unnecessary packages.

PR #7768, opened 2024-07-30
Fix small style issue

matmair:
Fixes a small style issue introduced in https://github.com/inventree/InvenTree/pull/7752


wolflu05:
Thank you, I completely overlooked that

PR #7773, opened 2024-07-30
Adjust docker labels to modern OCI schema

matmair:
This PR:
- switches the labels to the new opencontainers label schema
- makes the CI for docker a bit more robust


matmair:
This is probably better to be merged after the 0.16.0 release - not needed immediately

matmair:
This is ready for review

contrib/container/Dockerfile
```diff
@@ -11,6 +11,7 @@
 
 ARG base_image=python:3.11-alpine3.18
 FROM ${base_image} AS inventree_base
+ARG base_image
```
SchrodingersGat, line 14: Why hvae you added a blank `base_image` arg here? Won't this null out the default value?

matmair: No, it is a redefinition to carry the value over to this step from the previous one as ARGs are scoped per step. So the value as was definied in the previous step will be used.
https://docs.docker.com/reference/dockerfile/#scope

PR #7785, opened 2024-08-01
Make versions_check more reliable in forks

matmair:
This PR:
- Ensures version checks only run on GH actions (the required envs are only available there)
- Adapts the checks to also work on forks


.github/scripts/version_check.py
```diff
@@ -18,6 +18,9 @@
 
 import requests
 
+REPO = os.getenv('GITHUB_REPOSITORY', 'inventree/inventree')
+GITHUB_API_URL = os.getenv('GITHUB_REPOSITORY', 'https://api.github.com')
```
SchrodingersGat, line 22: These are both accessing the same environmental variable - is this intentional?

matmair: No, that is a typo. Good catch!

SchrodingersGat:
@matmair unfortunately this seems to have broken our docker builds:

image

matmair:
I see the problem, will submit a fix shortly

PR #7786, opened 2024-08-01
Make Tree Path searchable

wolflu05:
This PR adds:

- the `pathstring` property to the tree search fields to allow searching by pahtstring e.g. `Electronics/Passives/Resistors` will show the Resistors category that is in the `Passives` category, which itself is in the `Electronics` cagtegory.
- fixes the color scheme for the related model field in dark mode which was broken since the Mantine v7 bump.
- Icons are now displayed after the added margin left of tree elements in the related model field

image


SchrodingersGat (approved):

SchrodingersGat:
@wolflu05 thanks!

Issue #7603, opened 2024-07-09
[FR] Assembly pricing calculation without consumables


dieselburner:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

I love assembly pricing calculation feature as it precisely shows absolute minimum and maximum of overall assembly cost, depending on component purchase quantities and their price breaks.

However, when looking at assembly price range, it is calculated with consumables. Now, if I add to assembly, let's say, PCB stencil or solder paste, which are consumables, these will be not tracked in build order, which is also a great feature, but the assembly itself will not reflect the right cost anymore. 500g of solder paste I use cost around 150$. And if before a tiny board assembly was 5$ at a minimum and 20$ at a maximum, if I add this solder paste to BOM as consumable, board pricing will show 155$ at a minimum and 170$ at a maximum.

### Suggested solution

I would love to see an option to turn on assembly pricing calculation without consumables.

### Describe alternatives you've considered

Do not place consumables into BOM?

### Do you want to develop this?

- [ ] I want to develop this.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/inventree) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/inventree/InvenTree/issues/7603">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/inventree/InvenTree/issues/7603/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/inventree/InvenTree/issues/7603/pledge.svg">
</picture>
</a>

SchrodingersGat:
We could add a global option to exclude consumable items from BOM pricing calculation, I think this would provide the flexibility you are looking for?

dieselburner:
> We could add a global option to exclude consumable items from BOM pricing calculation, I think this would provide the flexibility you are looking for?

Yes! Thanks!

rocheparadox:
@SchrodingersGat shall I take this up?

SchrodingersGat:
@rocheparadox happy for you to tackle it! An extra setting, allowing "consumable" items to be excluded when performing BOM pricing calculation :)

rocheparadox:
[...]

PR #7787, opened 2024-08-01
Feature to exclude/include consumable price in BOM price range - FR #7603 

rocheparadox:
- A checkbox is added in the BOM panel in parts section to decide if the consumables' price have to be added to the BOM pricing
- A listener is attached to the checkbox to refresh the BOM table data when a change in the checkbox is detected.


rocheparadox:
@SchrodingersGat , Is there something else to be done from my side to get these changes merged? Thanks.

SchrodingersGat:
Hi @rocheparadox sorry I have not had a chance to review this yet. 

This approach is only going to impact the front-end representation, not the actual calculation (and storing) of BOM pricing. I think that you should:

**1. Add new global setting**

https://github.com/inventree/InvenTree/blob/e0a878467bb0fe6766ea410a89c9cff8d052faec/src/backend/InvenTree/common/models.py#L1636

Add a new setting to determine if consumable items should be included in pricing

**2. Update Calculation**

Update the `get_bom_price_range` method, to optionally include or exclude consumable parts:

https://github.com/inventree/InvenTree/blob/e0a878467bb0fe6766ea410a89c9cff8d052faec/src/backend/InvenTree/part/models.py#L2021

rocheparadox:
@SchrodingersGat , I will look into it. Thanks for the detailed explanation.

wolflu05:
maybe it would be also good to get this feature to PUI, because when CUI is dropped, this will be no longer available.

rocheparadox:
@SchrodingersGat, Would it not be a better, from a design standpoint, to provide this option as a **part** specific local decision? Doesn't adding this as a global setting restrict the users  throughout all BOM calculations? What if the users need to include the consumbale as part of their BOM pricing for one part and want to exclude in another?

SchrodingersGat:
@rocheparadox maybe - but at the moment our pricing management system does have provision for a "per part" option like this. 

We could certainly extend the pricing system, but, it will quickly become complex and everyone will want a slightly different approach...

This is where I think handing such decisions off to a custom pricing plugin system would make more sense.

If you have any ideas or input on how this should go, I'd be interested to hear them!

rocheparadox:
@SchrodingersGat , I believe that you are mentioning **Part.models.PartPricing** as the existing pricing management system for the **Part**. 

https://github.com/inventree/InvenTree/blob/70a52c938500f41c817bec8052e35e224b7eea16/src/backend/InvenTree/part/models.py#L2508

Nevertheless, I was thinking about adding a field called *include_consumbales* to that class and factoring that field in the BOM pricing calculations. However, as you mentioned, pricing itself would vary according to different users and it might be better left as a plugin development option. I am open to implementing either the global setting or this. I think, you know the user requirement better.

PR #7800, opened 2024-08-03
Fix docker repo names

matmair:
Fixes error introduced in https://github.com/inventree/InvenTree/pull/7785


SchrodingersGat:
THanks @matmair let's hope it now works fine :)

PR #7862, opened 2024-08-12
User defined states

matmair:
This PR adds the option to use user-defined states to models that use generic states.

For this, the model must be adapted to use the new `InvenTreeCustomStatusModelField` and the model's serializers must inherit `InvenTreeCustomStatusSerializerMixin`. The default behaviour of the serializer field is not changed but a new field is made available: `{field_name}_custom_key`. Custom states must always map to a `logical` key that points to the business logic. Custom states require unique keys that can not shadow logical keys. In theory, a model could make use of multiple fields with multiple custom states. This is not unit-tested with this PR but the code is developed with it in mind.

This PR does not change business logic but contains various frontend changes to render the custom field value instead of the (business) logic value. Both CUI and PUI were adapted to support the states. Colours for states now support enums (I would encourage usage of the enum)

Most API calls to serializers that use custom states experience 1-2 more DB calls due to the lookup of custom states. There is no caching in this PR implemented as consistency checks between different workers presented to be difficult.

```[tasklist]
### ToDo
- [x] Review of the rewrite of AllStatusViews by @wolflu05 - not sure why that was done that way and there was no testing coverage
- [x] Admin Center integration for state managment
- [x] Add docs
``` 

Included changes that are not strictly custom states:
- Adds filtering for contettypes API
- Adds contenttype model to PUI generic instance/form rendering
- Moves generic state API URL definitions to src/backend/InvenTree/generic/states

Closes https://github.com/inventree/InvenTree/issues/4289


src/backend/InvenTree/generic/states/api.py
```diff
@@ -73,18 +74,8 @@ class AllStatusViews(StatusView):
...
-            """Recursively discover status classes."""
-            for status_class in parent_status_class.__subclasses__():
-                name = '__'.join([*(prefix or []), status_class.__name__])
-                data[name] = {
-                    'class': status_class.__name__,
-                    'values': status_class.dict(),
-                }
-                discover_status_codes(status_class, [name])
-
-        discover_status_codes(StatusCode)
```
matmair, line 88: @wolflu05 do you have insight into why this was done this way? The git history shows that you authored it mainly. There was no coverage of the previous category and it seems everything works in PUI after the change.

wolflu05: Yes, this was changed so that nested inherotors of the status code class are discovered too. This is used for machine status.

There is an abstract `MachineStatus` class available here, which every machine type has to inherit and define its custom set of status codes:
https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/machine/machine_type.py#L25

Then all types are exposed unter `MachineStatus__xxx` where xxx is the name of the class inheriting the machine status class. The logic you removed is used to recursively detect status code inheritors and prepend the parent class name. You should see that the machine status codes were not shown properly now when navigating to the admin center > machines.

I actually thought I had added some tests for this, but it seems that I forgot. I would really appreciate if you could add one for me, as I personally won't have any time the next weeks to code something for inventree.

wolflu05 (commented):
> Most API calls to serializers that use custom states experience 1-2 more DB calls due to the lookup of custom states. There is no caching in this PR implemented as consistency checks between different workers presented to be difficult.

Why can't we just put this into the Redis/in-memory cache?

> Colours for states now support enums (I would encourage usage of the enum)

Great addition to validate the color value by providing an enum. Maybe it would be a good idea to also adjust the documentation here: https://docs.inventree.org/en/stable/extend/machines/overview there is the old way, using just a string, mentioned quite few times.

matmair:
A cache is an optional dependency currently and in-memory is per instance - which is not working with docker-compose in my testing. I do not want to add additional setup complexity that we would need to debug for users.

I can try to update that part of the docs.

SchrodingersGat:
@matmair if I am reading this correctly:

- We have a set number of "business logic" states (X) which we define on the backend
- Users can create new "custom display states" (Y) where each Y state must map to an X state
- The new Y states do not impact business logic, and are only for "display" purpose
- When changing from state Y1 to state Y2, we check the corresponding transition between X1 and X2 states

Am I missing anything?

matmair:
@SchrodingersGat from a business logic standpoint that is it. Y must have a unique key (for that X + Y + state and state + model + Y combination) - which enables filtering in plugins and possibly even with filters in the future.

src/backend/InvenTree/generic/states/api.py
matmair: I have amended the function and tests (API side) to cover this use case.

matmair:
This is technically done, I will try to finish the docs by the end of the week.

matmair:
@SchrodingersGat this is ready for review and merge. I don't know if we should backport it though as I am unsure how the API revision should be ported to 0.16.x.

It would be cool if we could get it into 0.16.x but probably more hassle than it is worth.

SchrodingersGat:
@matmair I think that given the amount of significant change here, we should let it have a round of testing in the `0.17.x` cycle before pushing it out to stable. Happy with that?

matmair:
@SchrodingersGat i have changed the target

matmair:
I have adjusted the api_version so that this can be merged after https://github.com/inventree/InvenTree/pull/7520

docs/docs/concepts/custom_states.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Custom States
+---
+
+## Custom States
```
SchrodingersGat, line 5: @matmair it would be really helpful in the "status" documentation for each model (purchase order etc) we can link through to here. Currently this new page is not linked from anywhere else...

matmair: Due to the newness of the feature, I do not think it should be highly advertised but I will add links

SchrodingersGat:
Thanks for the hard work on this one @matmair - due to the large number of changes here we will need to monitor closely until the 0.17.x release

PR #7868, opened 2024-08-14
Fix usePartFields hook

SchrodingersGat:


src/frontend/src/forms/PartForms.tsx
```diff
@@ -13,6 +13,8 @@ export function usePartFields({
 }: {
   create?: boolean;
 }): ApiFormFieldSet {
+  const settings = useGlobalSettingsState.getState();
```
wolflu05, line 16: Are you sure, that this works? 

What was your initial issue, that you tried to solve with this? That the useMemo was not retriggered when the global state updated?

Isn't the Zustand syntax something like this in a reactive way then:
```suggestion
  const settings = useGlobalSettingsState();
```

SchrodingersGat: The initial issue was that the call to settings was buried inside the lower `useMemo` call which was only updated when `create` changed. So, AFAIK, `settings` value was never being re-evaluated

wolflu05: No, it should be reevaluated, on the initial render of the component and if create would have changed.

Using .getState is used to get the state outside of the component. E.g. in a hook. Ref: https://github.com/pmndrs/zustand?tab=readme-ov-file#readingwriting-state-and-reacting-to-changes-outside-of-components

If you want to subscribe to the state and bind it to a react state, just call the store, the store is a hook. But I'm not sure if this whole change is really necessary.

PR #7880, opened 2024-08-14
Fix for usePartFields

SchrodingersGat:
- Ref: https://github.com/inventree/InvenTree/pull/7868#discussion_r1716423685

@wolflu05 is this what you mean?


wolflu05 (approved):
Yes, this should trigger an update of the useMemo now if the global settings state changes somehow. Not sure if your change before has worked (I'm away from the computer so I haven't tested this, just thought this might not have worked as expected before, so I commented).

Issue #7737, opened 2024-07-25
Chinese font support is missing in Docker containers.


zly20129:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When printing custom labels, Chinese cannot be output, you must manually add Chinese font support to successfully print Chinese labels.
Please see [Bug Solution Record](https://github.com/inventree/InvenTree/issues/7735).

### Steps to Reproduce

1. Creat a Chinese headline part；
2. Printer part label to pdf with Simple part label template；
3. You can see no zh_cn text will be printed.

### Expected behaviour

Possible problem to be solved: add Chinese font support in the officially given Docker container image.
Reason for needing to add font support to official mirrors: The update process given in the official help documentation requires the mirrors to be pulled and downloaded from the server again, and I'm not sure if this overrides the operation of adding font support manually, which would result in the user needing to once again re-add Chinese font support manually.

### Deployment Method

- [X] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

```
# Version Information:
InvenTree-Version: 0.15.6
Django Version: 4.2.14
Commit Hash: 23a9485
Commit Date: 2024-07-23

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-5.4.0-189-generic-x86_64-with
Installer: DOC

[...]

PR #7881, opened 2024-08-14
Update default fonts for docker image

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/7737
- Adds more comprehensive font support for back-end operations (e.g. printing)

### DIFF
--- a/contrib/container/Dockerfile
+++ b/contrib/container/Dockerfile
@@ -75,7 +75,7 @@ RUN apk add --no-cache \
     mariadb-client mariadb-connector-c \
     && \
     # font support
-    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-terminus font-inconsolata font-dejavu font-noto font-noto-cjk font-noto-extra \
+    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-terminus font-noto font-noto-cjk font-noto-extra \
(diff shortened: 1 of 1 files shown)


contrib/container/Dockerfile
```diff
@@ -74,8 +74,9 @@ RUN apk add --no-cache \
     # MySQL / MariaDB client
     mariadb-client mariadb-connector-c \
     && \
-    # fonts
-    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-noto terminus-font && fc-cache -f
+    # font support
+    apk --update --upgrade --no-cache add fontconfig ttf-freefont font-terminus font-inconsolata font-dejavu font-noto font-noto-cjk font-noto-extra \
```
matmair, line 78: What purpose servers `font-inconsolata font-dejavu`? `font-noto-cjk` should be enough for Chinese / Korean / japanese

matmair (changes requested):

contrib/container/Dockerfile
SchrodingersGat: Agreed, I'll update the PR and remove those. I was initially just following the recommendations from the alpine font support page.

matmair (approved):

PR #7886, opened 2024-08-14
[PUI] Lazy loading of global modals

SchrodingersGat:
- Slight improvement to initial loading time by lazy-loading some global modals


wolflu05 (approved):
Good idea, LGTM

matmair (approved):
LGTM; @wolflu05 do you think we could detect things like this in CI and highlight it?

wolflu05:
Not sure how this would work. Maybe there is an eslint plugin, or the netlify analytics could detect that.

SchrodingersGat:
I'm going to close this out for now - the gains are marginal and it is leading to a number of weird issues that are hard to debug. Not worth the effort currently.

PR #7900, opened 2024-08-16
Improve "issued by" filter for Build list API

SchrodingersGat:
- Allow filtering by groups
- Improve filter labelling
- Refactor "assigned_to" filter


matmair:
@SchrodingersGat is this changing default behaviour?

SchrodingersGat:
@matmair *slightly* - the "issued by" filter now accepts owners (user or group) rather than simply user. Not a significant change really, but I'll mark is as breaking so it gets visibility in the release.

PR #7917, opened 2024-08-19
[PUI] Fix for BOM pricing donut chart

SchrodingersGat:
- Mantine charts tooltip can't handle '.' character.


matmair (approved):
LGTM;
Is this bug reported upstream? Does not seem like something that is intended

PR #7918, opened 2024-08-19
[PUI] Order Currency

SchrodingersGat:
Correctly fill out currency when adding a new line item to an order


src/frontend/src/pages/sales/SalesOrderDetail.tsx
```diff
@@ -77,6 +80,14 @@ export default function SalesOrderDetail() {
     }
   });
 
+  const orderCurrency = useMemo(() => {
```
matmair, line 83: This seems to be needed multiple times; could this be generalised?

matmair (approved):
LGTM

src/frontend/src/pages/sales/SalesOrderDetail.tsx
SchrodingersGat: The internals of the memo are slightly different. I think that generalizing this would end up with more code...

PR #7941, opened 2024-08-21
[Build] Create child builds

SchrodingersGat:
- Adds option (when creating a build order) to automatically generate child build orders
- Default = False (requires user intervention)

### TODO

- [x] Add unit tests
- [x] Add note to documentation


matmair (approved):
Intresting addition. There might be room to do a wizard for creating subassemblies.

PR #7939, opened 2024-08-21
[PUI] Add Link/Unlink Barcode action

matmair:
This PR adds link and unlink actions to the various interfaces.

Todo
- [x] Refactor BarcodeActionDropdown to remove duplications
- [x] Add functional tests

~~-[ ]Re-Enable camera-based scanning on link~~ -> will be new PR

Fixes #7920

PR #7960, opened 2024-08-22
[PUI] Add image based QR code assigment

matmair:
Follow up to https://github.com/inventree/InvenTree/pull/7939

This adds a button to scan codes for barcode linking.
I am not sure about the styling and would be thankful for feedback.

New look for barcode linking
grafik

Current look for general QR scans:
grafik


wolflu05:
This already looks pretty good. I like that there is a select to use a different camera (this is pretty useful in safari to use the iPhone camera as a scanner).

I have to suggestions to improve the design.
1. make the stop button as high as the input field (I think there is an extra size type for the action button in mantine for this
2. make the input at the bottom full width and add the icon to the inputs left section.

Maybe this component can be also used later for the related model field to just input barcode data there.

matmair:
I have adapted it to this: 
grafik
Should we use this setup for all scanning purposes?

matmair:
Ready for review and merge if @wolflu05 is happy

SchrodingersGat:
@matmair looks good to me!

wolflu05 (approved):
Thanks for the design adjustments. It looks great now.

PR #7929, opened 2024-08-19
[CI] Switch to using Crowdin GitHub Action

matmair:
Fixes #7910

PR #7975, opened 2024-08-23
fix: translation action wrong location

matmair:
Addresses issues after https://github.com/inventree/InvenTree/pull/7929 merge:
- Compiled static in translation PR
- path to source is lost


matmair:
Note: Translations seem to still work with the shallow paths - so there might be the option to transfer all existing translations and use shallow paths going forward. There is no technical need to mirror the path structure in Crowdin as long as it is respected in the transform back.

Issue #7964, opened 2024-08-22
[BUG] User Defined States Issue


SchrodingersGat:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

Newly introduced feature "custom user states" breaks existing installations.

- Ref: https://github.com/inventree/InvenTree/pull/7862

When editing the "status code" for a stock item, a 500 error is raised:

![image](https://github.com/user-attachments/assets/841a1ce9-afb4-42b0-91d7-e45fa9fdf738)

![image](https://github.com/user-attachments/assets/2168b366-f292-442f-97d9-ff750afcf6c4)

```

Traceback (most recent call last):

File "/opt/inventree/env/lib/python3.10/site-packages/rest_framework/views.py", line 506, in dispatch
response = handler(request, *args, **kwargs)

File "/opt/inventree/env/lib/python3.10/site-packages/rest_framework/generics.py", line 288, in patch
return self.partial_update(request, *args, **kwargs)

File "/opt/inventree/env/lib/python3.10/site-packages/rest_framework/mixins.py", line 82, in partial_update
return self.update(request, *args, **kwargs)

File "/opt/inventree/src/backend/InvenTree/InvenTree/mixins.py", line 36, in update
self.perform_update(serializer)

File "/opt/inventree/env/lib/python3.10/site-packages/rest_framework/mixins.py", line 78, in perform_update
serializer.save()

File "/opt/inventree/src/backend/InvenTree/InvenTree/serializers.py", line 295, in save
super().save(**kwargs)

File "/opt/inventree/env/lib/python3.10/site-packages/rest_framework/serializers.py", line 207, in save
self.instance = self.update(self.instance, validated_data)

File "/opt/inventree/src/backend/InvenTree/stock/serializers.py", line 495, in update
return super().update(instance, validated_data)

File "/opt/inventree/src/backend/InvenTree/generic/states/fields.py", line 161, in update
reference = get_logical_value(

File "/opt/inventree/src/backend/InvenTree/generic/states/custom.py", line 43, in get_logical_value
return InvenTreeCustomUserStateModel.objects.get(key=value, model__model=model)

File "/opt/inventree/env/lib/python3.10/site-packages/django/db/models/manager.py", line 87, in manager_method
return getattr(self.get_queryset(), name)(*args, **kwargs)

File "/opt/inventree/env/lib/python3.10/site-packages/django/db/models/query.py", line 637, in get
[...]

PR #7976, opened 2024-08-23
Fix state changes on stock items

matmair:
Fixes https://github.com/inventree/InvenTree/issues/7964


matmair:
needs https://github.com/inventree/InvenTree/issues/7977

SchrodingersGat:
@matmair thanks for giving this some attention :)

PR #7979, opened 2024-08-24
Fix path

matmair:
Fixes another small bug introduced in https://github.com/inventree/InvenTree/pull/7929 /  https://github.com/inventree/InvenTree/pull/7975


SchrodingersGat:
@matmair this has fixed the specific pathing issue, but now running into a git identity problem:

image

Issue #7973, opened 2024-08-23
[PUI] Markdown Horizontal Dividers in Notes not parsed correctly in platform UI


simonkuehling:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

We used a lot of horizontal dividers (three dashes like ```---```) in the notes field throughout inventree. In the new platform UI those seem to not be escaped correctly as the rendered notes text field cut off everything after the first of these horizontal dividers. 

### Steps to Reproduce

1. in the old UI add a line with three dashes like ```---``` to a notes field with more text before and after this line
2. view the note in the new UI
3. only text from before the divider is rendered (both in ```view``` and in ```edit``` mode)

### Expected behaviour

properly parse/escape three dashes to account for legacy content in the database from before the PUI upgrade

### Deployment Method

- [X] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

# Version Information:
InvenTree-Version: 0.17.0 dev
Django Version: 4.2.15
Commit Hash: 5515e07
Commit Date: 2024-08-20

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.1.0-rpi7-rpi-v8-aarch64-with
Installer: DOC

[...]

PR #7980, opened 2024-08-24
Notes editor

SchrodingersGat:
This PR reverts the implementation of the markdown editor, from using [MDXEditor](https://mdxeditor.dev/) to [react-simple-mde](https://www.npmjs.com/package/react-simplemde-editor).

This brings the new user interface back in line with the old one. It also cuts down significantly on the required front-end packages.

The initial switch to MDXEditor was made due to some problems which (at the time) I could not overcome. However these have now been addressed, and overall `react-simple-mde` provides a cleaner user experience.

- Closes https://github.com/inventree/InvenTree/issues/7973


matmair:
Is this a fix for the notes editor errors?

PR #7985, opened 2024-08-25
Update crowdin.yml

SchrodingersGat:
- Add documentation strings
- Attempt to "fix" dest parameters
- Part of ongoing efforts to get crowdin working again...

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,12 +1,14 @@
+# Configuration file for Crowdin project integration
+# See: https://crowdin.com/project/inventree
+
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@matmair ok the upload / download of translations seems to be working quite well now, and the paths match up on crowdin now.

However we still have "duplicate" translations which means that the "new" files are showing up as only having a small number of unique translations:

image

So I think we still need to:

1. Download the existing translations
2. Remove the old GitHub integration
3. Re-upload the translations against the directory structure
4. Re-approve all existing translations

Do you agree that this is the correct approach?

PR #8005, opened 2024-08-26
[PUI] Make actions recognisable

matmair:
Action Dropdowns are not very recognisable right now. Therefore, I propose adding indicators similar to the screenshots below to buttons with more than 1 action. I am not sure how to style that, below solution is not the final goal - maybe a single indicator could be enough for example.

Current
grafik

Proposed
grafik

Fixes https://github.com/invenhost/InvenTree/issues/98


SchrodingersGat:
Maybe a "down caret" indicator would be clearer here? It seems a bit more "traditional". But I agree with the concept in general

wolflu05:
I like the suggestion from @SchrodingersGat

matmair:
What do we think about this: 
grafik

SchrodingersGat:
That's pretty clean IMO

wolflu05:
I think no its really hard to tell, which "down caret" belongs to what action, maybe an outline would help to add some separation, but not sure how that will look like.

matmair:
It looks atrocious with an outline
grafik

wolflu05:
That's true. Maybe its just me having problems to see the separation.

SchrodingersGat:
What about a mantine indicator with the caret attached directly to the button

https://mantine.dev/core/indicator/

matmair:
I have tried another thing: `light` 
grafik

matmair:
I find vertical dots also match the flow better:
grafik

matmair:
And the same thing in dark mode with dots instead of chevrons
grafik

SchrodingersGat:
Light with chevrons is my pick :)

matmair:
Ok then this is ready for merge - light with chevron is the current state of the PR.

> And maybe we add the chevron to the vertical dot menu too?

I would not add chevrons to the dot menu, I think that 3 dots are universally understood as "more" - that is why I suggested usage of them in the first place.

SchrodingersGat:
> I would not add chevrons to the dot menu, I think that 3 dots are universally understood as "more" - that is why I suggested usage of them in the first place.

In this case, should we add a special dropdown menu just for the "more" actions - I see that you have added a lot of instances of `noindicator` - this could be refactored into a special dropdown with the *dots* icon and no indicator?

matmair:
Sure, I can look into that.

matmair:
@SchrodingersGat I have added a special dropdown to adress this

matmair:

SchrodingersGat:
Nice work, thanks for this!

PR #8010, opened 2024-08-27
Update LICENSE copyright year to 2024

dekhu48:


matmair:
Updating the copyright year is completely meaningless in a legal sense. Because of that we stopped upping it

SchrodingersGat:
@matmair would it be best to remove the dates entirely then?

matmair:
@SchrodingersGat at the starting year can be interesting for disputes - I would just leave it as is

dekhu48:
Closing PR

PR #8013, opened 2024-08-27
Add row context (right click) additionally to actions in tables

matmair:
This draft PR adds right-click context menus to tables. The content can be determined by a custom callback or computed from the row actions.
I am interested in feedback - you should be able to play with it using the preview deployment.
grafik
grafik

Closes https://github.com/invenhost/InvenTree/issues/128


SchrodingersGat:
I think we need to have a larger conversation about how we interact with tables - and make the "user expectation" consistent.

There are:

- Row actions (via buttons)
- Row actions (via context menu)
- Some rows take you somewhere else when you click them
- Some rows open up a context drawer when clicked on

Is there a way we can visually indicate what happens when you left-click or right-click on a row, do you think?

matmair:
I am not aware that other apps provide this, have not seen this pattern somewhere else consciously. Maybe linking somewhere else should just be a highlighted link somewhere at the beginning of the table to make it clear.

matmair:
@SchrodingersGat how should we proceed here?

SchrodingersGat:
@matmair I am happy with this in principle. Looking at the current state, there are some visual styling issues that need to be addressed:

Here is a regular row-context menu (via the drop-down in the "actions" column):

image

And the same menu accessed via right-click context menu:

image

So, the colors need fixing, and some padding around the menu item

matmair:
I will look into reusing the renderer from the actions dropdown

matmair:
@SchrodingersGat I have addressed the styling; ready for review & merge

SchrodingersGat (approved):

SchrodingersGat:
Fantastic, looks great now. This is a really nice addition.

Issue #7997, opened 2024-08-26
`render_currency` in templates show incorrect amount of decimals


GoryMoon:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

When using the `render_currency` helper in templates it will render too many decimals until you have both `decimal_places` and `max_decimal_places` set.

The docs say that if `decimal_places` is specified it will ignore `max_decimal_places`.
https://docs.inventree.org/en/stable/report/helpers/#currency-formatting

From looking at the code this line will always pick the max amount of decimals without respecting the other setting.
https://github.com/inventree/InvenTree/blob/d1c835485982650dfa4a9c6b51142ab9660584d4/src/backend/InvenTree/InvenTree/helpers_model.py#L238

### Steps to Reproduce

This is observable on the demo site without any changes.

The `InvenTree Sales Order` uses `max_decimal_places=2` on all prices, looking at the output it shows 4 decimals.
Adding `decimal_places=2` will show the correct amount of decimals.

### Expected behaviour

To have `render_currency` only show the configured amount of decimals like the Docs say.

### Deployment Method

- [X] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

# Version Information:
InvenTree-Version: 0.16.0
Django Version: 4.2.15
Commit Hash: 3f2ac5a
Commit Date: 2024-08-25

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.1.21-v8+-aarch64-with
Installer: DOC

[...]

PR #8017, opened 2024-08-28
Adjust "render_currency" behaviour

SchrodingersGat:
- Fixes https://github.com/inventree/InvenTree/issues/7997


GoryMoon (approved):
Looks good, it will do what I want.
Either the docs need to be updated regarding that min/max is always used or the code need change to ignore those if a `decimal_places` is supplied.
The docs have this not on the min/max values: `(ignored if decimal_places is specified)`

https://docs.inventree.org/en/stable/report/helpers/#currency-formatting

PR #8018, opened 2024-08-28
Remove references to old setting

SchrodingersGat:
After the [report generation refactor](https://github.com/inventree/InvenTree/pull/7074) we no longer separate report generation concepts based on the "target" of the report.

This PR removes two settings which are no longer relevant.


src/backend/InvenTree/common/models.py
```diff
@@ -1704,20 +1704,6 @@ def save(self, *args, **kwargs):
...
-            'validator': bool,
-        },
-        'REPORT_ATTACH_TEST_REPORT': {
-            'name': _('Attach Test Reports'),
-            'description': _(
-                'When printing a Test Report, attach a copy of the Test Report to the associated Stock Item'
-            ),
-            'default': False,
-            'validator': bool,
-        },
```
matmair, line 1720: Do we not want to reimplement this function with the new templating?

SchrodingersGat: As I see it we have the following options:

1. Add a new setting to dictate if reports (not just "test reports") are automatically saved as attachments against the linked model
2. Add a "run time" option (when printing report templates) where the user can select if reports are saved as attachments
3. Implement a plugin which does this (or a subset of these things)

However as the concept of a "test report" no longer exists, the current setting is invalid.

matmair: Then we should maybe mention somewhere that we have completely removed this functionality. It would be problematic if someone depended on the reports being attached and they just were not anymore in silence.

matmair (approved):

src/backend/InvenTree/common/models.py
SchrodingersGat: Do you have a preference for 1/2/3 above?

matmair: No, I am not using this feature right now

wolflu05: Im not using this feature, but I like 1). Maybe it would be good to have the ability to specify this per repot type, maybe also per template, if it should be attached or not.

SchrodingersGat: @wolflu05 that's a good idea, I like the idea of specifying it per template

PR #8023, opened 2024-08-28
Check plugin registry hash when registering events

SchrodingersGat:
- Before registering events, check plugin registry for reload
- Ensures that background worker plugin registry is up-to-date


SchrodingersGat:
Closing, I missed that `check_reload` is called a few lines later. My problem was actually in my plugin code...

PR #8027, opened 2024-08-28
[PUI] Make breadcrumbs adjustable

matmair:
Makes usage of breadcrumbs more uniform and adds a user setting to optionally show the current item in the breadcrumbs too.


matmair:
see https://github.com/inventree/InvenTree/discussions/7962

src/frontend/src/components/nav/PageDetail.tsx
```diff
@@ -50,12 +54,21 @@ export function PageDetail({
     ]
   ]);
 
+  // breadcrumb caching
+  const computedBreadcrumbs = useMemo(() => {
```
SchrodingersGat, line 58: @matmair I would suggest that a simpler solution, rather than adding another `last_crumb` parameter, would be to simply remove the last one in the list, if the setting is not enabled?

matmair: Not all pages have breadcrumbs where it makes sense to remove the last one. I find an explicit interface better than one that assumes things and needs to be negated.

SchrodingersGat: Ok, I'm happy with that then

src/frontend/src/pages/build/BuildDetail.tsx
```diff
@@ -542,8 +542,8 @@ export default function BuildDetail() {
             editAction={editBuild.open}
             editEnabled={user.hasChangePermission(ModelType.part)}
             imageUrl={build.part_detail?.image ?? build.part_detail?.thumbnail}
-            breadcrumbs={[
-              { name: t`Manufacturing`, url: '/manufacturing' },
+            breadcrumbs={[{ name: t`Manufacturing`, url: '/manufacturing' }]}
+            last_crumb={[
```
SchrodingersGat, line 546: It is a small request but can we make this `lastCrumb` instead of `last_crumb` - the convention here is for camelCase

SchrodingersGat:
@matmair looks very good, just a small change requested regarding variable naming

src/frontend/src/pages/build/BuildDetail.tsx
matmair: done

SchrodingersGat:
Thanks @matmair :)

PR #8037, opened 2024-08-29
Fix query return type for part detail page

SchrodingersGat:
- Sub-variables of 'null' value cannot be accessesd


SchrodingersGat:
Closing, branched from wrong commit.

PR #8043, opened 2024-08-30
Surpress error if no changes exsist

matmair:
If no changes in the translations happen, we get errors currently: https://github.com/inventree/InvenTree/actions/runs/10624771577/job/29453665879


SchrodingersGat:
Thanks for this, that should reduce the inbox noise

PR #8053, opened 2024-09-02
A bit more testing

matmair:
This PR is a follow up to https://github.com/inventree/InvenTree/pull/8048 and adds more coverage


src/backend/InvenTree/stock/migrations/0065_auto_20210701_0509.py
```diff
@@ -4,22 +4,6 @@
...
-    for app in apps.get_app_configs():
-
-        label = app.label
-
-        migrations = MigrationRecorder.Migration.objects.filter(app=app).order_by('-applied')[:5]
-
-        print(f"{label} migrations:")
-        for m in migrations:
-            print(f" - {m.name}")
-
```
matmair, line 22: @SchrodingersGat I would remove this as I do not see a business value in it. Do you think it needs to be kept?

SchrodingersGat: @matmair I added this to assist with user debugging - there have been a few cases where I have needed to see the state of a particular app, and having `invoke.showmigrations part` is far simpler to accomplish. I think it has its use

matmair: But was has that to do with running the command on every instance through migrations (one time)

SchrodingersGat: @matmair oh, sorry I totally misread the context of this code. Yeah, that looks like an old debug hook. In fact if you look down further in the file, that code is not even getting called!

SchrodingersGat: That can 100% be removed

matmair:
@SchrodingersGat this is ready for review; stock is still not 100% but there is a lot of surface to cover and small PRs are less likely to break stuff

SchrodingersGat (approved):

SchrodingersGat:
Thanks for the additional tests :)

PR #8061, opened 2024-09-03
Add option ot mark the build item as complete if all tests passed

### LINKED ISSUE #6006: [FR]Add option to being able to mark build outputs completed if all required tests are passed

### Problem statement

When thinking on the #5037 the following came into my mind: in some scenarios it would be useful if the test results could mark the given build output completed. 
For e.g. in the case if the final test passed (and the build output is ready for completion see #5037) then the output could be marked completed.

### Suggested solution

Add a "Finish build output if this test passed" parameter to the test templates.
Maybe if we implement the #5037 we need to reword this a bit to something "Finish build output if this (and all other required) tests are passed" 

-

### Examples of other systems

-

- [X] I want to develop this.

martonmiklos:
Fixes #6006


SchrodingersGat:
@martonmiklos thanks for looking at this, and apologies for the delayed response.

Looking at the new model field implementation here, I think that it should be an option against the `BuildOrder` instance, rather than the `Part` instance. 

@matmair @wolflu05 any thoughts or preferences here?

martonmiklos:
> @martonmiklos thanks for looking at this, and apologies for the delayed response.

No worries. 

> 
> Looking at the new model field implementation here, I think that it should be an option against the `BuildOrder` instance, rather than the `Part` instance.
> 
> @matmair @wolflu05 any thoughts or preferences here?

I see your point, but I would say both options could make sense. In my use case most of the products got packaged as the part of the last teststep so closing the build order when all units passed makes sense. Enabling the "autoclose" feature for each build order created would not makes too much sense in this case.

SchrodingersGat:
@martonmiklos I was envisioning a much "simpler" approach, where you can set the "default value" for the "auto close" option as a global setting, but then adjust on a per-build level as needed.

Having this as a "per part" option seems a bit messy (for lack of a better word) especially as most parts cannot be "built" (they are components, e.g. ) and so the option does not even make sense for them. At a bare minimum the option should be hidden for parts which are not "assemblies".

The issue here is that we do not have any other "dependent attributes" - the attributes such as "purcahsesable", "trackable" , "salable" (etc) are all independent of each other.

SchrodingersGat:
@martonmiklos thinking about this further, I believe that this would be much better as a custom plugin. This way we do not need to add new internal logic to multiple points in the code to implement a feature which most people would disable anyway.

We have a comprehensive set of event triggers  which could be hooked into by a plugin to "auto-complete" a build when the tests are completed. And, the plugin could expand with additional complexity as required, without needing to change core behaviour.

SchrodingersGat:
@martonmiklos I'm going to close this out, but if you want to discuss further how to implement this as a plugin I'd be happy to assist

PR #8084, opened 2024-09-05
Packaging switch to an empty buildpack

matmair:
This PR switches the package building process to an empty build pack. We are not using the functions the build pack provides and this
- makes the process faster
- removes the need to keep versions of build packs in sync and working
- makes it possible to package for more platforms (CentOS, newer versions of Debian, ubuntu)


matmair:
This has problems that seem complex to debug; Will re-open later

PR #8054, opened 2024-09-02
Add SelectionList concept

matmair:
This PR adds the concept of SelectionLists - a way to define reusable choices for parameters (optionally controllable by plugins).

This is helpful to load and sync a list of choices into multiple instances. The usage of SelectionLists could be extended in the future.

<details><summary>Screenshots</summary>
<p>

### CUI

grafik

grafik

grafik

### PUI

grafik

grafik

grafik

grafik

</p>
</details> 

ToDo:
- [x] PUI unit testing
- [x] locking SelectionList
- [x] list deactivation logic

PR #8093, opened 2024-09-09
[PUI] Fix translation of model type strings

SchrodingersGat:
This PR fixes a long-standing issue in PUI in which labels for model types are not correctly translated. This is because they were defined in a `const` datatype which did not change, but the lingui locale rendering is dynamic.

So, labels are now exposed via a function which is computed at runtime, allowing dynamic translation.

This was mainly noticeable in the "Search Results" drawer:

### Before

image

### After

image


matmair:
I just encountered this in https://github.com/inventree/InvenTree/pull/8054 and this new mechanism feels very unintuitive as information is now spread in multiple places. Is there another way we can handle this? I would prefer if adding a new model was intuitive and not 3 CI runs to get right.

SchrodingersGat:
@matmair fair point, we could clean this up a bit, i'm happy to take suggestions :)

PR #8104, opened 2024-09-10
[PUI] Refactor keeping modelinfo together

matmair:
Follow up to https://github.com/inventree/InvenTree/pull/8093. This PR reverts the organisation of model information to the old model with 1 important change: labels and lables_multiple are now functions to enable live translations.


SchrodingersGat:
Nice, this is much cleaner :)

Issue #8103, opened 2024-09-10
Docker Compose - inventree-server crashing due to missing django_q_schedule table?


victorhooi:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [X] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

I am trying to deploy InvenTree locally using the provided Docker compose file at:

https://github.com/inventree/InvenTree/blob/master/contrib/container/docker-compose.yml

I have placed the following three files in a single directory:
* https://github.com/inventree/InvenTree/blob/master/contrib/container/docker-compose.yml
* https://github.com/inventree/InvenTree/blob/master/contrib/container/.env
* https://github.com/inventree/InvenTree/blob/master/contrib/container/Caddyfile

I then run:
```
docker compose pull
docker compose up -d
```
(At first, I got hit with a gotcha around the Caddyfile - it wasn't mentioned this was needed - and when Docker compose tried to spin everything up, it hit this error:

```
❯ docker compose up --detach
[+] Running 4/5
 ✔ Container inventree-cache   Running                                                                                                                                                                                                                                    0.0s 
 ✔ Container inventree-db      Running                                                                                                                                                                                                                                    0.0s 
 ✔ Container inventree-worker  Running                                                                                                                                                                                                                                    0.0s 
 ✔ Container inventree-server  Started                                                                                                                                                                                                                                    0.0s 
 ⠙ Container inventree-proxy   Starting                                                                                                                                                                                                                                   0.1s 
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/Users/bytedance/inventree-docker/Caddyfile" to rootfs at "/etc/caddy/Caddyfile": mount /Users/bytedance/inventree-docker/Caddyfile:/etc/caddy/Caddyfile (via /proc/self/fd/6), flags: 0x5000: not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
```
(It seems at least one other person hit this - https://github.com/inventree/InvenTree/discussions/7401 - it might be worth documenting it - or figuring out why the Docker compose file doesn't fail gracefully here?)

Anyway, clearing that error - the next issue is that `inventree-server` and `inventree-worker` are both crashing.

[...]

PR #8126, opened 2024-09-13
Revert int.worker to worker

SchrodingersGat:
- Prevent existing docker compose installs from breaking
- Ref: https://github.com/inventree/InvenTree/issues/8103

This PR reverts the "worker" task to simply being called "worker" (not "int.worker"). Otherwise, when we push the new changes to stable, anyone running on a docker setup will find that their "worker" task will not start any more. Their (existing) docker compose file will reference the "worker" task

### DIFF
--- a/docs/docs/start/docker_install.md
+++ b/docs/docs/start/docker_install.md
@@ -247,7 +247,7 @@ index 8adee63..dc3993c 100644
 -        image: inventree/inventree:${INVENTREE_TAG:-stable}
 +        image: inventree/inventree:${INVENTREE_TAG:-stable}-custom
 +        pull_policy: never
-         command: invoke int.worker
+         command: invoke worker
          depends_on:
              - inventree-server
 ```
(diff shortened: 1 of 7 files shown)


docs/docs/start/docker_install.md
```diff
@@ -247,7 +247,7 @@ index 8adee63..dc3993c 100644
 -        image: inventree/inventree:${INVENTREE_TAG:-stable}
 +        image: inventree/inventree:${INVENTREE_TAG:-stable}-custom
```
matmair, line 248: Q: Is this intended to be included?

matmair (commented):
LGTM for now; I would prefer if we could change this with 1.0

matmair (approved):

docs/docs/start/docker_install.md
SchrodingersGat: No, nice catch..

SchrodingersGat: I think that the  github diff preview is being tricked, as we have a "diff" in the documentation...

PR #8137, opened 2024-09-17
UI plugins custom features

wolflu05:
This PR enhances the UI mixin (which currently only can provide panels) with the capability to also integrate in lots of different other parts of the UI via a generic system. E.g. table actions, row actions, custom navigation entries, ... Currently only custom template editors and previews can be provided.

I'm currently working on this label template drag and drop editor plugin: https://github.com/wolflu05/inventree-template-editor-plugin and have released the first version now. Feel free to test it.

## TODO

- [x] refactor user interface related API and serializer code into separate files
- [x] basic plugin ui feature implementation
- [x] integrate template editors into plugin system
- [x] integrate template previews into plugin system
- [x] provide exportable types
- [x] add code editor sample to ui sample plugin to use in tests
- [x] refactor `ref` out of the getFeature functions as parameter and instead provided them if necessary via the renderContext, and also have the ability to specify the return type to make the system more generic for other usecases (e.g. only return a list of navigation entries, ...)
- [x] allow to specify the function name via the source file string divided by a colon
- [x] bump API version
- [x] add tests
- [x] documentation


matmair:
Very interesting, I like the approach @wolflu05!

wolflu05:
This is ready for review now.

src/frontend/src/components/plugins/PluginContext.tsx
```diff
@@ -1,18 +1,26 @@
...
+import { useLocalState } from '../../states/LocalState';
+import {
+  SettingsStateProps,
+  useGlobalSettingsState,
+  useUserSettingsState
+} from '../../states/SettingsState';
+import { UserStateProps, useUserState } from '../../states/UserState';
 
-/*
+/**
```
matmair, line 20: Q: Why is this second star needed?

src/frontend/src/components/plugins/PluginContext.tsx
```diff
@@ -34,3 +39,37 @@ export type PluginContext = {
   theme: MantineTheme;
   colorScheme: MantineColorScheme;
 };
+
+export const useInvenTreeContext = () => {
+  const host = useLocalState.getState().host;
```
matmair, line 44: Q: Why is the state invocated this way? I think a change in `host` would not be propagated with this kind of invocation

src/frontend/tests/pui_printing.spec.ts
```diff
@@ -84,6 +84,15 @@ test('PUI - Report Printing', async ({ page }) => {
 test('PUI - Report Editing', async ({ page }) => {
   await doQuickLogin(page, 'admin', 'inventree');
 
+  // activate the sample plugin for this test
+  await page.request.patch(`${classicUrl}/api/plugins/sampleui/activate/`, {
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Basic YWRtaW46aW52ZW50cmVl`
```
matmair, line 91: Q: Could this be calculated/shared with the credentials in line 85 to ensure this does not break if we change default credentials?

src/frontend/tests/pui_printing.spec.ts
```diff
@@ -104,5 +113,42 @@ test('PUI - Report Editing', async ({ page }) => {
...
+    .click();
+  const msg = (await consoleLogPromise).args();
+  expect(await msg[0].jsonValue()).toBe('updatePreview');
+  expect((await msg[1].jsonValue())[0]).toBe(newTextareaValue);
+
+  // deactivate the sample plugin again after the test
+  await page.request.patch(`${classicUrl}/api/plugins/sampleui/activate/`, {
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Basic YWRtaW46aW52ZW50cmVl`
```
matmair, line 148: Same as on line 94

matmair (approved):
LGTM; There seems to be no coverage on `usePluginUIFeature` and `getPluginTemplateEditor`, `getPluginTemplatePreview` and most changes on `TemplateTable.tsx)` reported on codecov - is that an issue with reporting or are these changes not tested?

src/frontend/src/components/plugins/PluginContext.tsx
wolflu05: That's what the jsdoc spec says. (Ref: [jsdoc.app/about-getting-started](https://jsdoc.app/about-getting-started#:~:text=Each%20comment%20must%20start%20with%20a%20/**%20sequence%20in%20order%20to%20be%20recognized%20by%20the%20JSDoc%20parser.%20Comments%20beginning%20with%20/*%2C%20/***%2C%20or%20more%20than%203%20stars%20will%20be%20ignored.)) You can also see the difference in vscode, if there are two `**` vscode parses and highlights the jsdoc correctly.

src/frontend/src/components/plugins/PluginContext.tsx
wolflu05: You're right, normally that should be different. But I have copied it from the initial implementation from @SchrodingersGat , maybe he had something in mind with it? I have now changed it.

src/frontend/tests/pui_printing.spec.ts
wolflu05: I have changed this now although the `admin:inventree` credentials are referenced already in a lot of different places.

wolflu05:
> There seems to be no coverage on usePluginUIFeature, getPluginTemplateEditor, getPluginTemplatePreview and most changes on TemplateTable.tsx reported on codecov - is that an issue with reporting or are these changes not tested?

@matmair I was wondering about that too, but all of that code should normally be covered by my added playwright test. That's why I actually made the effort to activate the sample plugin and ensure the sample template editor is shown when doing template editing. And also make sure that passing the code back and forth between the editors works. But I have no idea why codecov does not display that. Maybe you can help me here?

src/frontend/tests/pui_printing.spec.ts
wolflu05: You're right, that makes sense.

src/frontend/src/components/plugins/PluginContext.tsx
matmair: I was not aware of that, thank you for the explanation

matmair (approved):

wolflu05:
Yeah, the template editors are not rendered on the initial component render, as first the data from the server has to be loaded, but that shouldn't cause any problems, as that's how we do it everywhere? But what's interesting though is if you look at the history of the codecov comment, the previous comment actually hasn't contained the missing tsx files, only the python file.

wolflu05:
@matmair Maybe that's the same problem like https://github.com/mxschmitt/playwright-test-coverage/issues/22, but unfortunately there is no response on that issue . I guess you have used that template to build the coverage system? Maybe we need to try the official coverage solution: https://playwright.dev/docs/api/class-coverage

wolflu05:
The code that is marked as uncovered is definitely being run:

Code marked as not covered:
image

I have added a `throw new Error(...)` here:
image

And in the output when running the tests, you can see that this error is raised.

So this is definitely a problem with the coverage reporting.

matmair:
This is indeed very curious; do you think it is realistic that we could fix it ourselves or do we have to just ignore it for now?
I would really like to enforce code coverage rules but that will be hard with this not working

wolflu05:
Not sure, I have no idea what's going on there, I even cannot get coverage to work locally.

wolflu05:
@matmair can we merge this pr now, as the lines are actually be covered by the tests? Then you also have something to test coverage with your new approach.

matmair:
I would prefer if we figure this out before, coverage of PUI, in general, is very slacking. We can go ahead and merge but than I would expect an issue to keep track of this.

wolflu05:
As said before, I have absolutely no clue on this, why it's not counted. I have never done frontend coverage in any other project, so I'm no help here. @SchrodingersGat do you have any idea here?

wolflu05:
@SchrodingersGat would you mind taking a review at this, so we can merge this? You can activate the sample ui plugin or install the inventree-template-editor-plugin for  testing.

wolflu05:
@SchrodingersGat @matmair the coverage issue has been resolved. It was caused by `await page.context().close();` as the last line of the playwright test. I have no idea why this was added. I have just extended the test and didn't keep care what this is doing. I have removed it now, and coverage seems to be reported fine now. So there is nothing in the way anymore.

SchrodingersGat:
@wolflu05 thanks for your work here, it is looking very promising! Hopefully we can have some fully integrated plugins working soon to test it out with

PR #8147, opened 2024-09-19
[PUI] Panels plugin fix

SchrodingersGat:
This is a follow up to https://github.com/inventree/InvenTree/pull/7470

Realized after the fact that the CUI and PUI plugin mixins use the same method name for delivering custom panels. This means that a plugin cannot bridge the gap and support both CUI and PUI.

To address this, the method name in the `UserInterfaceMixin` class has been changed.

### DIFF
--- a/docs/docs/extend/plugins/ui.md
+++ b/docs/docs/extend/plugins/ui.md
@@ -18,9 +18,9 @@ When rendering certain content in the user interface, the rendering functions ar
 
 ## Custom Panels
 
-Many of the pages in the InvenTree web interface are built using a series of "panels" which are displayed on the page. Custom panels can be added to these pages, by implementing the `get_custom_panels` method:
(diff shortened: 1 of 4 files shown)


SchrodingersGat:
@wolflu05 this will impact on your outstanding PR, so a review would be appreciated :)

wolflu05:
Interesing, haven't thought about that yet. But I suggest naming it `get_ui_panels` to stay consistent? Thoughts @matmair ?

matmair (approved):
LGTM in general; @wolflu05 suggestion re name would keep it more consistent - I would change it if possible

SchrodingersGat:
Done! @wolflu05 if you can have similar naming for the other plugin features e.g. `get_ui_xxx` that would be great :)

wolflu05:
> Done! @wolflu05 if you can have similar naming for the other plugin features e.g. `get_ui_xxx` that would be great :)

That's why I suggested it, I already use `get_ui_features`.

PR #8148, opened 2024-09-19
WIP: PUI testing coverage use bable plugin

matmair:
THIS IS FOR TESTING PURPOSE - DO NOT MERGE


wolflu05:
I guess you have removed the eslint from precommit, because it didn't download. I had this issue too a few days ago, and found out you can also disable it by running SKIP=eslint git commit .... without commiting changed in the precommit config.

PR #8171, opened 2024-09-24
[PUI] Part charts fix

SchrodingersGat:
Bug fixes and refactoring for part charts / tables

### DIFF
--- a/src/frontend/src/hooks/UseTable.tsx
+++ b/src/frontend/src/hooks/UseTable.tsx
@@ -54,6 +54,8 @@ export type TableState = {
 export function useTable(tableName: string): TableState {
   // Function to generate a new ID (to refresh the table)
   function generateTableName() {
+    // Ensure we strip out any hyphen characters from the original name
+    tableName = tableName.replaceAll('-', '');
     return `${tableName}-${randomId()}`;
   }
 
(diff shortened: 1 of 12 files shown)


src/frontend/src/pages/part/PartStocktakeDetail.tsx
```diff
@@ -23,6 +30,36 @@ import { TableColumn } from '../../tables/Column';
...
+/*
+ * Render a tooltip for the chart, with correct date information
+ */
+function ChartTooltip({ label, payload }: ChartTooltipProps) {
+  if (!payload) {
+    return null;
+  }
+
+  if (label && typeof label == 'number') {
+    label = formatDate(new Date(label).toISOString());
```
wolflu05, line 42: Reassigning props is an anti pattern. That's why some code quality tools suggest to wrap the props with readonly. And for such "more" resource intensiv functions it would be better to wrap them in a useMemo


```suggestion
    const formattedLabel = useMemo(() => formatDate(new Date(label).toISOString()), [label]);
```

src/frontend/src/hooks/UseTable.tsx
```diff
@@ -54,6 +54,8 @@ export type TableState = {
 export function useTable(tableName: string): TableState {
   // Function to generate a new ID (to refresh the table)
   function generateTableName() {
+    // Ensure we strip out any hyphen characters from the original name
+    tableName = tableName.replaceAll('-', '');
```
wolflu05, line 58: Same here, props should not be reassigned, they are readonly.

src/frontend/src/pages/part/PartStocktakeDetail.tsx
SchrodingersGat: Yeah, great point. I have addressed this now, thanks.

PR #8191, opened 2024-09-26
Build order consume

SchrodingersGat:
This is a large(ish) feature branch which allows stock to be consumed by a build order *before* the build order is complete.

This allows more complex (and useful) workflows, where the build might take some significant time, but the stock is consumed at the start. e.g. a batch of PCBs are loaded, but then require a lot of testing, etc.

Currently, the allocated stock must "languish" in the build order, and this makes tracking of real stock quantities difficult.

Given that this feature might take a while to integrate and test, I am targetting it (for now at least) against 1.0.0. This will *not* be implemented in the legacy user interface.

### Tasks

- [x] Update documenation around new workflows
- [x] User interface updates
- [x] Playwright testing
- [x] Updated unit testing
- [x] Ensure partial allocation vs partial consumption works well
- [x] Add dummy data to the demo dataset - https://github.com/inventree/demo-dataset/pull/95
- [x] Ensure that the "REQUIRED" counts still make sense (take "consumed" into account)
- [x] Check that the "required" counts are correct in the API (unit testing)
- [x] Ensure table filters work correctly for the build-line table


matmair (approved):
Nice addition!

matmair (approved):

Issue #5232, opened 2023-07-12
Packing List support


mabroens:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

Hi team, as i am progressing with my report templates for invoices (will upload this later once 100% sattified) i thought it was time to look again at creating a Packing List or pick list including serialnumbers for stock items.

i had a look at https://github.com/inventree/InvenTree/issues/62  where matmair provided me a link to [packing lists](https://docs.inventree.org/en/stable/report/pack/).  but unfortunatly this results in: 
![image](https://github.com/inventree/InvenTree/assets/130482028/8aaaa6dc-cf71-4760-827b-8ba6e4750857)

for version 0.11.0 it shows:
![image](https://github.com/inventree/InvenTree/assets/130482028/589af031-de87-45e1-96a1-74d61df92c70)

and in https://github.com/inventree/InvenTree/pull/748 
i noticed:
![image](https://github.com/inventree/InvenTree/assets/130482028/a79714f8-a26b-410e-89d2-24b4ab57d3f0)

is there anything left from previous work (not finisched) that i can take as starting point. then later when i am done i can upldoad it and you can post the template again.

I also had a look at: https://docs.inventree.org/en/0.6.0/report/context_variables/#stockitem which let me to a inventree_test_report_base.html (https://github.com/inventree/InvenTree/blob/master/InvenTree/report/templates/report/inventree_test_report_base.html)
which is the only thing i could find referencing serial numbers.

so tried following:
```
 {% for line in lines.all %}
<tr>
<td>
	<div class='thumb-container'>
		<img src='{% part_image line.part %}' class='part-thumb'>
	</div>
</td>
<td>
	<div class='part-text'>
		{% if line.part.revision %} 
				<strong>{{ line.part.IPN }} {{line.part.revision}}-Sample</strong>
		{% else %} 
				<strong>{{ line.part.name }}</strong>
		{% endif %}
		<br><br>
		{{ line.part.description }}
		</div>
</td>
{% if stock_item.is_serialized %}
	<td>{% trans "Serial Number" %}: {{ stock_item.serial }}</td>
{% else %}
	<td>{% trans "Quantity" %}: {% decimal stock_item.quantity %}</td>
{% endif %}

<td></td>
<td></td>
</tr>
    {% endfor %}

```
but that did not work for me.

Br Martijn

### Suggested solution

provide me some hint or older doc that could be reused

[...]

Issue #62, opened 2018-04-17
PackingList / BuildList


SchrodingersGat:
When shipping or building items, it can be helpful to print out a list of all the items, with a picture, description, quantity.

Then shipping peoples can easily check off each item as they go.

SchrodingersGat:
This packing list can be generated from:

* Purchase Order
* Sales Order / Customer Order
* Internal Stock Movement

mabroens:
Hi @SchrodingersGat ,

i have started using Inventree 2 weeks agoo and LOVE it. its great i have been able to upload my parts to the db using the inventree-digikey python app. works :-) 
but now that i have all my stock in and i have cretaed a build order for an external assembly house (I design PCB's) i also need to ship certain parts from my lab to them. therfore i would like to create a packinglist for the items that i'm shipping (or hand carry) to the assembly house. how can i create this packinglist?

you mentioned tha this is possible from: This packing list can be generated from:

> Purchase Order
> Sales Order / Customer Order
> Internal Stock Movement

but i cant figure out how to.
please advice.
Br Martijn

matmair:
You need to create a report template. Read more in the docs about [reports](https://docs.inventree.org/en/stable/report/report/) and [packing lists](https://docs.inventree.org/en/stable/report/pack/). This is an advanced topic.
Please do not reuse old, closed issues.

SchrodingersGat:
@mabroens the report documentation as linked by matmair above should contain all you need. Otherwise please open a new issue or discussion

PR #8194, opened 2024-09-26
Shipment Reports

fuzeman:
This pull request implements sales order shipment reports which will enable greater flexibility for the generation of packing lists.

I found some previous discussions in #5232 and #62 around packing lists, but the solutions discussed appeared to be only suitable for generating a packing list of an entire sales order. I usually have long-running sales orders with weekly shipments requiring individual packing slips and it will be great to have this integrated inside InvenTree.

-----

**Sample Report (included in pull request):**

SalesOrderShipment


matmair (approved):
Looks good to me, thank you for the PR. Let's see if all style checks pass - this is good to merge from my POV

SchrodingersGat:
@fuzeman nice addition, thanks!!

fuzeman:
Thanks.

I've been using InvenTree almost daily for build orders and stock management over the last 3 years and finally had the push to make my first contribution. Hopefully more PR's to come when time allows.

SchrodingersGat:
Contributions are always welcome 🤗 And it is wonderful to hear such feedback

fuzeman:
Just noticed that the "Attach to Model on Print" functionality probably won't work as `SalesOrderShipment` doesn't support attachments. It would be simple enough to enable attachments on `SalesOrderShipment`, but I'm not sure how we would view them. Maybe they should be aggregated and displayed under the `SalesOrder` attachment list?

I'll take a look, but let me know if you have any thoughts.

fuzeman:
After taking a closer look at how attachments are implemented, the simplest option would probably be a custom `create_attachment` implementation on `SalesOrderShipment` which stores the attachments on the parent `SalesOrder` instance. The only remaining concern would be if `SalesOrderShipment` should have `InvenTreeAttachmentMixin` which would create a redundant `SalesOrderShipment` model type on the `Attachment` model.

SchrodingersGat:
@fuzeman I think your suggested approach is reasonable, but could we have a situation where a user could manually create an attachment against a `SalesOrderShipment`? Probably not, or not easily, I'd imagine. In which case the redundant model type is likely fine. There is nowhere where the user has to manually select which model to upload an attachment against.

fuzeman:
@SchrodingersGat Thanks. See PR #8231.

PR #8228, opened 2024-10-01
[PUI] Plugin settings UI

SchrodingersGat:
Adds functionality to render custom "configuration" UI elements for a plugin, in the new interface.

Expands existing functionality similar to how custom panels are implemented.

The plugin can dynamically render content to this window:

image

### TODO

- [x] Playwright / UI tests


SchrodingersGat:
And another example, from the [wireviz plugin](https://github.com/inventree/inventree-wireviz/pull/44):

image

src/frontend/src/pages/Index/Settings/UserSettings.tsx
```diff
@@ -148,16 +148,14 @@ export default function UserSettings() {
...
         title={t`Account Settings`}
         subtitle={
           user?.first_name && user?.last_name
             ? `${user?.first_name} ${user?.last_name}`
             : null
         }
         shorthand={user?.username || ''}
-        switch_link="/settings/system"
-        switch_text={<Trans>Switch to System Setting</Trans>}
-        switch_condition={user?.is_staff || false}
```
matmair, line 160: Q: Why is this removed?

src/frontend/src/pages/Index/Settings/SystemSettings.tsx
```diff
@@ -306,10 +306,9 @@ export default function SystemSettings() {
       {user.isStaff() ? (
         <Stack gap="xs">
           <SettingsHeader
+            label="system"
             title={t`System Settings`}
             subtitle={server.instance || ''}
-            switch_link="/settings/user"
-            switch_text={<Trans>Switch to User Setting</Trans>}
```
matmair, line 312: Q: Why is this removed?

matmair (approved):
LGTM in general, there seems to be features removed without a callout in the PR message.

Could we add frontend tests to ensure the integration actually works?

wolflu05 (commented):
Nice that you have implemented this, this is a great way forward.

However I'm not sure if we really should build lots of similar functionality. Instead, I would prefer if we could build one generic method for integrating custom frontend portions into the ui, so that we don't have to build the same thing (an api endpoint, a plugin function to get the content, a method for rendering a template and providing context from the plugin to the js function, ...) over and over again with different plugin apis (that potentially also confused plugin devs and provides bad dx.

That's why I started to implement the custom ui features as a very generic and abstract interface with custom ts types that can handle various contexts provided by different sources. (With the inventree npm package, this gets even better from a dx perspective)

Maybe for this integration that's a bit overkill to use the custom ui features, but for other ui integrations I would really like to see the generic way improved rather than than implementing the same thing over and over again?

What do you think @matmair ?

SchrodingersGat:
@wolflu05 I'm all for refactoring / making this more generic. Maybe we can do that on next feature integration, or as a separate PR?

SchrodingersGat:
> Q: Why is this removed?

@matmair I have slightly tweaked the mechanism for switching between "user settings" / "system settings" / "admin section" - the functionality is still there but visually cleaned up a bit

src/frontend/src/pages/Index/Settings/SystemSettings.tsx
SchrodingersGat: See comments below

SchrodingersGat:
Playwright testing has now been added too

matmair (approved):

matmair (approved):

PR #8235, opened 2024-10-03
Fix build item over-allocation checks

fuzeman:
This pull request fixes an issue with `BuildItem` over-allocation checks which prevent allocation changes to over-allocated items. Looking a bit closer at related code it appears `SalesOrderAllocation` already handled this correctly, and I've aligned these changes with that implementation.

---------------

Not sure if this is a concern but I noticed the quantity checks between `BuildItem` and `SalesOrderAllocation` don't appear to be aligned with one casting to `decimal.Decimal` and the other not. Not sure if this is breaking anything, but it doesn't look quite right to me.

https://github.com/inventree/InvenTree/blob/f8c7635a8d42f8860dc39198d7962f5286bcd853/src/backend/InvenTree/build/models.py#L1609-L1617
https://github.com/inventree/InvenTree/blob/f8c7635a8d42f8860dc39198d7962f5286bcd853/src/backend/InvenTree/order/models.py#L2014-L2025

### DIFF
--- a/src/backend/InvenTree/build/models.py
+++ b/src/backend/InvenTree/build/models.py
@@ -1606,12 +1606,19 @@ class BuildItem(InvenTree.models.InvenTreeMetadataModel):
                     'quantity': _(f'Allocated quantity ({q}) must not exceed available stock quantity ({a})')
                 })
 
-            # Allocated quantity cannot cause the stock item to be over-allocated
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@fuzeman thanks for the fix, looks like a pretty subtle one to work out!

PR #7904, opened 2024-08-18
Add namespaces to tasks

matmair:
This PR adds namespaces to tasks (namely dev and int) and moves most tasks to these new namespaces. I decided against hiding the moved tasks from the listing command as that might be confusing to users.

```ash
Available tasks:

  backup                            Backup the database and media files.
  export-records                    Export all database records to a file.
  frontend-download                 Download a pre-build frontend from GitHub if you dont want to
                                    install nodejs on your machine.
  import-records                    Import database records from a file.
  install                           Installs required python packages.
  migrate                           Performs database migrations.
  plugins                           Installs all plugins as specified in 'plugins.txt'.
  remove-mfa                        Remove MFA for a user.
  restore                           Restore the database and media files.
  static                            Copies required static files to the STATIC_ROOT directory, as
                                    per Django requirements.
  superuser                         Create a superuser/admin account for the database.
  update                            Update InvenTree installation.
  version                           Show the current version of InvenTree.
  wait                              Wait until the database connection is ready.
  dev.delete-data                   Delete all database records!
  dev.docs-server                   Start a local mkdocs server to view the documentation.
  dev.frontend-dev                  Start frontend development server.
  dev.gunicorn                      Launch a gunicorn webserver.
  dev.import-fixtures               Import fixture data into the database.
  dev.schema                        Export current API schema.
  dev.server                        Launch a (development) server using Django's in-built
                                    webserver.
  dev.setup-dev                     Sets up everything needed for the dev environment.
  dev.setup-test                    Setup a testing environment.
  dev.test                          Run unit-tests for InvenTree codebase.
  dev.test-translations             Add a fictional language to test if each component is ready for
                                    translations.
  dev.translate                     Rebuild translation source files. Advanced use only!
  int.clean-settings                Clean the setting tables of old settings.
  int.clear-generated               Clear generated files from `invoke update`.
  int.export-settings-definitions   Export settings definition to a JSON file.
  int.frontend-build                Build frontend.
  int.frontend-check                Check if frontend is available.
  int.frontend-compile              Generate react frontend.
  int.frontend-install              Install frontend requirements.
  int.frontend-trans                Compile frontend translations.
  int.rebuild-models                Rebuild database models with MPTT structures.
  int.rebuild-thumbnails            Rebuild missing image thumbnails.
  int.render-js-files               Render templated javascript files (used for static testing).
  int.showmigrations                Show the migration status of the database.
  int.translate-stats               Collect translation stats.
[...]

PR #8239, opened 2024-10-05
Add missing namespaces to contributing page

1337joe:
I was rebuilding my dev environment and the instructions in the docs didn't work as listed.

1. Without django installed in my environment `invoke` wouldn't run.
2. The `dev` namespace was missing from the invoke arguments.

It looks like this page was largely missed in PR #7904. I didn't find any other pages that didn't have the namespace specified where necessary.


SchrodingersGat (approved):

SchrodingersGat:
@1337joe thanks for the fix!

Issue #8234, opened 2024-10-03
Issue with login when pushing Sample Data


Timieh:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

Issues with login. 
When a new Inventree setup is finished. 
Login is fine! 
When you push the demo sample data into it. 
the login wouldn't work anymore. 

Temp. fix is run again:
`docker compose run inventree-server invoke superuser`

### Steps to Reproduce

Fresh install Debian 11
Fresh Install docker
Fresh install production
(Everything fine until this point). 
Push of sample data (according to guide). 
Try to re-login into inventree instance. 
Issue, user/password wrong / not recognized. 

(Temp. fix is run again:
`docker compose run inventree-server invoke superuser`)

### Expected behaviour

Expected behaviour is to push the demo sample data. 
And still keep the existing users. 

### Deployment Method

- [x] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

Docker (latest possible ~2 Okt 2024)

Inv 0.16.4
Api 232
python 3.11.0
django 4.2.15

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

matmair:
The sample data contains users and groups so your login credentials are probably overwritten. There is also the option to clear the database when loading sample data, that might also be the issue

Timieh:
Issue occors when executing:
`docker compose run --rm inventree-server invoke setup-test -i`

(So didn't do the empty db command)

SchrodingersGat:
Loading the demo data completely overrides the database. There is no way to "combine" the data and it is not intended that way

Timieh:
Understood, Maybe nice to have a note in the 'production server'. manual ?
 
https://docs.inventree.org/en/stable/start/docker_install/#ssl-configuration

SchrodingersGat:
@Timieh fair point! Are you able to open a PR with an update to the docs?

Timieh:
Happy to assist @SchrodingersGat , but sorry i don't now how to handle that. 
If I can get some small guide about it (new to this). I'm happy to help about this and future stuff ;)

SchrodingersGat:
For some useful information on how to create a pull request on GitHub - https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request

[...]

PR #8243, opened 2024-10-06
Add additional information on demo dataset

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/8234

@Timieh a small-target PR to add some additional docs as requested


Timieh:
Great!

PR #8246, opened 2024-10-07
[Refactor] Serial generation

SchrodingersGat:
This PR refactors the creation of large numbers of serialized stock items.

Previously each item was created individually and thus was very inefficient. 

| Operation | Old | New |
| --- | --- | --- |
| Create 500 new stock items, serialized | > 9,000 queries, 14 seconds | 28 queries, 0.9 seconds |
| Serialize existing stock item, quantity 500 | > 9,000 queries, 23 seconds  | 34 queries, 0.9 seconds |
| Create build order outputs, quantity 500 |  6,500 queries, 8 seconds | 2,000 queries in 3.2 seconds |


src/backend/InvenTree/stock/models.py
```diff
@@ -458,44 +459,98 @@ def report_context(self):
         & Q(expiry_date__lt=InvenTree.helpers.current_date())
     )
 
-    def update_serial_number(self):
-        """Update the 'serial_int' field, to be an integer representation of the serial number.
+    @classmethod
+    def create_serial_numbers(cls, serials: list, **kwargs) -> QuerySet:
```
matmair, line 463: C: I think this should be an internal method (_ prefix) as it should not be used by plugins etc.

src/backend/InvenTree/stock/models.py
```diff
@@ -458,44 +459,98 @@ def report_context(self):
...
             serial_int = plugin.convert_serial_to_int(serial)
 
             if serial_int is not None:
-                # Save the first returned result
-                # Ensure that it is clipped within a range allowed in the database schema
-                clip = 0x7FFFFFFF
-
-                serial_int = abs(serial_int)
-
-                serial_int = min(serial_int, clip)
```
matmair, line 485: We had an issue with DoS attacks with too large serial numbers - that was fixed by this check; have you implemented an equivalent in the new code?

matmair (approved):
LGTM; I have a small sec concerns, see detailed comments

src/backend/InvenTree/stock/models.py
SchrodingersGat: Fair enough, I will do so.

src/backend/InvenTree/stock/models.py
SchrodingersGat: Nice catch, I've re-implemented this now.

matmair (approved):

SchrodingersGat:
Waiting for https://github.com/inventree/InvenTree/pull/8269 to be merged

Issue #8251, opened 2024-10-07
Prevent HTML Encoding in QR Code Payload for Part Data


skydiablo:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

Currently, when generating a QR code for a part, the resulting payload is HTML-encoded, which affects the output format. For example, instead of generating a plain text JSON payload like this:

```
{"part": 1}
```

the payload is encoded as:

```
{&quot;part&quot;: 1}
```

This encoding causes unnecessary transformations, especially when parsing the data from the QR code. It would be more useful if the QR code generator produced the payload in plain text format without HTML-encoding special characters such as " and &.

### Steps to Reproduce

Go to a part in InvenTree.
Generate a QR code for that part.
Scan the QR code and check the payload. It will be HTML-encoded (e.g., {&quot;part&quot;: 1}).

### Expected behaviour

The payload should not be HTML-encoded and should instead output raw JSON data, like:

```
{"part": 1}
```

### Deployment Method

- [ ] Docker
- [ ] Package
- [x] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

0.16.4

### Please verify if you can reproduce this bug on the demo site.

- [x] I can reproduce this bug on the demo site.

skydiablo:
so, i have checked this, i think the default escaping is the problem. this fix it, just patch this [line](https://github.com/inventree/InvenTree/blob/198a39a33cf304c8682c41cea958c39d6bc1420c/src/backend/InvenTree/part/templates/part/part_base.html#L461) from

```
            '{{ part.barcode }}',
```
to
```
            '{{ part.barcode|safe }}',
```

regards, volker.

SchrodingersGat:
@skydiablo so, it is only in the displayed text that the error is occurring, not the actual generated barcode itself?

skydiablo:
no, the payload of the QR-code itself is escaped, see above, there is a PR to fix this issue for the payload

PR #8255, opened 2024-10-08
fix part qr lable

skydiablo:
in ref to this issue: https://github.com/inventree/InvenTree/issues/8251

Closes https://github.com/inventree/InvenTree/issues/8251

### DIFF
--- a/src/backend/InvenTree/part/templates/part/part_base.html
+++ b/src/backend/InvenTree/part/templates/part/part_base.html
@@ -458,7 +458,7 @@
     $("#show-qr-code").click(function() {
         showQRDialog(
             '{% trans "Part QR Code" escape %}',
-            '{{ part.barcode }}',
+            '{{ part.barcode|safe }}',
         );
     });
 


SchrodingersGat:
@skydiablo thanks for the fix  :)

wolflu05:
Aren't other pages also affected by this now?


And isn't that introducing an xss, because we cannot trust the output of the barcode generator, since this can be influenced by how the barcode plugin is configured, or doesn't this matter since only admins can configure the barcode plugin?

SchrodingersGat:
> Aren't other pages also affected by this now?

@wolflu05 you are correct, I misread the intent of the fix here and thought it was applied generically to all pages.

@skydiablo would you be able to submit another PR which addresses all instances of this? Perhaps even roll it into a custom template tag and make it generic.

> And isn't that introducing an xss, because we cannot trust the output of the barcode generator, since this can be influenced by how the barcode plugin is configured, or doesn't this matter since only admins can configure the barcode plugin?

As the data is encoded into a string, we just need to escape any `'` characters, to ensure it does not run as javascript, right?

skydiablo:
hey, i would realy like to do a generic fix, but my python skills are equal zero ;) i think the problem is easy to addresse:

- find all occurrences that could be possible
- build a generic qr-code widget
  - define output to 'safe' but escape ''' characters

right?

SchrodingersGat:
@skydiablo please refer to this PR - https://github.com/inventree/InvenTree/pull/8286

This has basically implemented what you suggested above, and runs the barcode data through bleach to ensure that it is "safe" to render into the templates

Issue #6643, opened 2024-03-05
[PUI] Improve home page style


30350n:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

With PUI now being promoted in the 0.14.0 release blog post, I think it might be a good idea to clean up the home page UI a bit, as this is the first thing everyone sees. (Not content, but style wise, as currently parts of it seems quite cluttered).

![image](https://github.com/inventree/InvenTree/assets/34186858/78a2fd91-7a9e-43a9-ac3e-22232b4b6699)

Issues from the screenshots:
1. The left- and rightmost tabs (Home and Sales) have their right/left margins respectively, set to 0, which causes the underline not to be centered anymore.
2. The styling of the light gray boxes clashes with the inner darker ones.
  2.1 Corners should be rounded
  2.2 Vertical/Horizontal margins should be equal (or atleast not *this* dissimilar)
3. The left/right navigation buttons overlap with the content below
  3.1 The buttons also look a little pixelated

![image](https://github.com/inventree/InvenTree/assets/34186858/3619337f-a361-4328-93f6-21e2f851d55f)

4. The settings menu pops up on top of the main nav bar, I think it'd look better if it's top right corner would be at the bottom right corner of the nav bar.

### Suggested solution

I'd be happy to look into trying to improve the UI design here and I'd probably be capable of doing so if this was vanilla html/css, but from what I can tell this seems to be built on top of "mantine" so I don't really know where I'd even start. If someone could throw me a pointer here on where and how to start, this would be very helpful ^^

### Describe alternatives you've considered

-

### Do you want to develop this?

- [X] I want to develop this.

matmair:
2 was a stylistic choice
2.1 this was chosen to contrast the interactive carousel with the static background of the widget
2.2 widgets follow a fixed grid width/height while content does not so the content would need to be adjusted. Keep in mind that the content is a touch-enabled carousel


You can submit a PR for sure; I have not bothered to finish the design as core functions are still missing

30350n:
Any pointers regarding where to start? Where does the styling happen in the first place?

wolflu05:
Your point one already bothered me too and I actually thought I would have committed that already in one of my PRs, but turns out I forgot. You just need to remove those lines: 
https://github.com/inventree/InvenTree/blob/master/src/frontend/src/globalStyle.tsx#L128-134

[...]

Issue #3062, opened 2022-05-24
[FR] Improved Dashboard System / Index Page Improvements


SchrodingersGat:
The current "home page" needs a major overhaul and should show information pertaining to the logged-in user.

For example a technician should be able to see all the build  orders assigned to them, basically their "assigned tasks".

Instead of *boring* tables, some "dashboardy" widgets which are information dense.

The dashboard will also be exposed to the plugin interface, so custom dashboard "widgets" can be added.

*Open for ideas...*

SchrodingersGat:
@eeintech @matmair @maksimstojkovic any ideas here? What would you like to see as a "dashboard" when you log into InvenTree?

matmair:
@SchrodingersGat the option to load a lot less data at the beginning would be nice. So everything loads faster and that there are fewer API requests.

maksimstojkovic:
@SchrodingersGat possibly a similar bar to the page-actions menu which allows sales/purchase/build orders to be created directly, before reloading into the page of the said order. It could be debated whether the inclusion of the bar would consume too much screen real estate, but in the future it could be modified to include other frequently used actions (ideally this is a data-driven decision).

eeintech:
I would vote for a "recent activity" feed:
- part added, by which user and when
- part supplier update, by which user and when
- BoM recently validated, by which user and when
- stock updates, by which user / PO / SO and when it was updated
- BO status updates, by which user and when it was updated
- etc.

feurer98:
I think you can never please everyone.
That's why I would give the user a possibility to design it himself. 
As a reference like MyPage from Openproject: [Docs](https://www.openproject.org/docs/getting-started/my-page/)

Maybe just keep this option in mind and then introduce it on a testable frontend.

matmair:
> I would vote for a "recent activity" feed:
> - part added, by which user and when
> - part supplier update, by which user and when
> - BoM recently validated, by which user and when
> - stock updates, by which user / PO / SO and when it was updated
> - BO status updates, by which user and when it was updated
> - etc.

Nice idea. Could be a pretty expensive database operation without caching thought.

maksimstojkovic:
Thoughts on a single-page scrollable dashboard with tiles/widgets that can be enabled/disabled?
I find that a fair bit of the information available on screen in the existing dashboard is not used when deciding to navigate to a specific order.

[...]

PR #8278, opened 2024-10-12
[PUI] Dashboard refactor

SchrodingersGat:
This PR presents a major refactor of the "dashboard" feature in PUI, bringing it closer towards a release state.

We were already using [react-grid-layout](https://github.com/react-grid-layout/react-grid-layout) for the new dashboard - this PR moves the implementation forward, adding new features, refactoring, etc

### Related Issues

- Closes https://github.com/inventree/InvenTree/issues/3062
- Closes https://github.com/inventree/InvenTree/issues/6643

### Main Goals

- [x] Remove duplicate "home" / "dashboard" pages
- [x] Improve usability of dashboard
- [x] Add method of "adding" new widgets to the dashboard
- [x] Add method of "removing" widgets from the dashboard
- [x] Replicate all previous "home page" functions as "dashboard" items
- [x] Provide a library of "builtin" dashboard items
- [x] Save and restore separate layout "per user"
- [x] Allow plugins to register dashboard items
- [x] Cleanup global navigation menu (left-hand drawer)
- [x] Playwright tests
- [x] Consolidate UI plugin API settings
- [x] Single entrypoint for accessing UI features on backend
- [x] Add "news" widget as a dashboard item
- [x] Review existing (legacy CUI) frontpage items
- [x] Add unit tests for "dashboard" API endpoints
- [x] Update documentation

### Screenshots

| Description | Image |
| --- | --- |
| Sample dashboard with builtin items, and items loaded via plugins | image |
| Add new item | image | 
| Remove Items | image |
| Adjust layout | image |


src/backend/InvenTree/plugin/base/ui/api.py
```diff
@@ -54,6 +54,44 @@ def get(self, request):
         )
 
 
+class PluginDashboardList(APIView):
```
wolflu05, line 57: Adding new api endpoints and implementing the same thing over and over again with different interfaces was what I was trying to avoid when I built the ui features implementation using a generic system in a very typesafe way. But that's up to you. Or @matmair what do you mean?

src/frontend/src/components/dashboard/DashboardLayout.tsx
```diff
@@ -0,0 +1,290 @@
...
+const ReactGridLayout = WidthProvider(Responsive);
+
+/**
+ * Save the dashboard layout to local storage
+ */
+function saveDashboardLayout(layouts: any, userId: number | undefined): void {
+  const data = JSON.stringify(layouts);
+
+  if (userId) {
+    localStorage?.setItem(`dashboard-layout-${userId}`, data);
```
wolflu05, line 21: Wouldn't it make sense to store the dashboard layout on the server in a user setting?

wolflu05 (commented):
Very nice that you have combined the dashboard and home pages now. We are one step closer to a final release now.

src/backend/InvenTree/plugin/base/ui/api.py
SchrodingersGat: Ah, ok I see. So a single endpoint for fetching *all* types of UI integrations? I can work on that.

src/frontend/src/components/dashboard/DashboardLayout.tsx
SchrodingersGat: Great idea! I'll see how that would work

SchrodingersGat: I have this working, however the "layout" data does take up a lot of space. The `UserSetting.value` field currently has a restricted length of 2,000 characters. This equates to (roughly) 30 widgets - assuming a separate setting per "layout breakpoint" (i.e. "sm" / "md" / "lg" / "xl" which depend on the screen size).

This seems pretty limited IMO. We could bump up the length restriction on the "value" field, and also have a single setting for each layout "breakpoint".

Alternatively, we could live with the dashboard being cached to the browser (at least for now)?

SchrodingersGat: I'm going to leave it as is for this PR at least. Perhaps we can work on this as a future PR.

src/backend/InvenTree/plugin/base/ui/api.py
SchrodingersGat: @wolflu05 please check this out again now - there is only a single API endpoint to access these features now

src/frontend/src/components/plugins/PluginUIFeatureTypes.ts
```diff
@@ -35,11 +36,7 @@ export type TemplateEditorUIFeature = {
     template_type: ModelType.labeltemplate | ModelType.reporttemplate;
     template_model: ModelType;
   };
-  responseOptions: {
-    key: string;
-    title: string;
-    icon: InvenTreeIconType;
-  };
+  responseOptions: PluginUIFeature;
```
wolflu05, line 39: @SchrodingersGat sorry, I only had a chance now to review this *properly*.

The idea here was, that each different feature can respond with some different options (which are defined in the "options" dict key provided by the `get_ui_features` function on the plugin class). That's why I build a TS generic type. Where each feature gets its base things defined (`feature_type`, `requestContext`, `responseOptions`, ...). I already defined them for `TemplateEditorUIFeature` (and `TemplatePreviewUIFeature`) here. (Your dashboard things are missing) Then some generic type helpers can be used where needed to type things. This becomes pretty handy, when we publish the inventree-ui npm package. UI feature developers then can just do something like this:

```ts
import { PluginUIGetFeatureType, TemplateEditorUIFeature } from '@inventree/ui';

export const getFeature: PluginUIGetFeatureType<TemplateEditorUIFeature> = (params) => {
  //
};
```  

You now only changed the responseOptions for the template editor feature to some "generic" options (Same in the plugin definition. It's not really clear now what are custom options specific to that feature and what are mandatory things like source. From my two examples this may be not really clear, because they both contained the same options, but I was planning, that other features may need to have different options. E.g. not all features need a description or icon. How do you want to move forward with this now, especially with inventree 1.0?

SchrodingersGat: @wolflu05 i"m happy to look at expanding the typing system as part our efforts towards 1.0 :)

PR #8307, opened 2024-10-17
Refactoring / Enhancements

SchrodingersGat:
~~WIP PR to provide historical salse information (broken down by month for e.g.) for a particular part.~~

*Note: The original intent of this PR is probably better served as a plugin. But there are a number of fixes / improvements which are still good additions*.


src/frontend/src/tables/company/CompanyTable.tsx
```diff
@@ -157,16 +157,11 @@ export function CompanyTable({
...
           tableFilters: tableFilters,
           tableActions: tableActions,
           enableDownload: true,
-          rowActions: rowActions,
-          onRowClick: (row: any) => {
-            if (row.pk) {
-              let base = path ?? 'company';
-              navigate(`/${base}/${row.pk}`);
-            }
-          }
```
matmair, line 169: Q: Why is the direct link removed?

matmair (commented):
Lots of small things, seems reasonable

src/frontend/src/tables/company/CompanyTable.tsx
SchrodingersGat: Ah good catch. The "open link in new tab by holding control" wasn't working for this table, but I forgot that the link can be passed in. I'll fix this.

PR #8314, opened 2024-10-18
Add example for INVENTREE_CUSTOMIZE environment variable

SergeoLacruz:
Our admin struggled a bit with this :-)


docs/docs/start/config.md
```diff
@@ -369,6 +369,15 @@ The logo and custom messages can be changed/set:
 | INVENTREE_CUSTOMIZE | customize.navbar_message | Custom message for navbar | *Not specified* |
 | INVENTREE_CUSTOMIZE | customize.hide_pui_banner | Disable PUI banner | False |
 
+The INVENTREE_CUSTOMIZE environment variable must contain a json object with the keys from the table above and
+the wanted values. Setting the variable is a bit tricky in bash because of json syntax. Example:
```
matmair, line 373: Why would you set them in bash? That is the dumbest of all config ways to do this

SchrodingersGat: @matmair what would you suggest here for a better example?

matmair: I would suggest to not recommend bash for configuring inventree outright. Once the server is restarted the settings are gone if not set correctly (which we do not explain or recommend anywhere). We provide and document either the yaml or environment variables everywhere else, this inclusion will just lead to confusion.

SergeoLacruz: Good morning,
I do not care is you use bash, csh, tcsh, ksh,sh or whatever kind of system. I just wanted to add something to a position where we stumbled. Because from the table alone, it is not obvious what the content of the variable should look like. Three times the same variable name in the table looks strange without further description. 
Michael

SergeoLacruz:
I removed the bash example and just mentioned the content of the variable. OK now?

SchrodingersGat:
Yes this reduced sample is fine, thanks

PR #8317, opened 2024-10-19
[PUI] Switch linting to biome

matmair:
Currently eslint and prettier are setup up (probably incorrectly as they do not seem to work) in pre-commt for PUI linting/formatting. This PR replaces both with biome, which is faster, does basically the same things, and seems to work.

It also adds auto fixes for various stuff like consistent quote styles or removing unused imports.


matmair:
@SchrodingersGat @wolflu05 thoughts on this? No idea why eslint is not working but this fixes it.

SchrodingersGat:
Nice this looks great @matmair

SchrodingersGat (approved):

SchrodingersGat:
@wolflu05 any thoughts on this one?

SchrodingersGat:
@matmair are you happy for me to merge this?

wolflu05:
I have never used biome before, but it seems like a cool tool after a quick look. (it's also rust based, which promises speed). So sure we can try it out.

I would suggest adding some config for vscode to use biome as a formatter, so it works smoothly together with format on save.

SchrodingersGat:
> I would suggest adding some config for vscode to use biome as a formatter, so it works smoothly together with format on save.

How would you suggest doing that?

wolflu05:
Seems like they have a guide about it: https://biomejs.dev/reference/vscode/

matmair:
I have added a extension accommodation and base settings

wolflu05:
Thanks 👍

matmair:
Thank you for your input @wolflu05  - it is valued as this whole frontend tech stuff is not really my forte.
@SchrodingersGat this is ready for review and merge from my side

SchrodingersGat:
Awesome, great work as always @matmair

PR #8331, opened 2024-10-22
[API] Sales order filters

SchrodingersGat:
- Fixes for SalesOrderLineItem API endpoints
- Improvements to PUI tables
- Updated unit testing


matmair (approved):
LGTM;
Random though: Do you think we could dynamically load the filters (including translations) from the backend? Would reduce duplication.

SchrodingersGat:
@matmair some of the translated labels for the filters are already introspected from the API - any filter which directly matches an available field in the API endpoint can use the translated string from that field

PR #8332, opened 2024-10-22
Task group

SchrodingersGat:
- Adds "group" attribute to background tasks for improved filtering
- Improvements to "notify low stock" calls


matmair (approved):
Looks promising; could be used for statics / reporting in a future step

PR #8366, opened 2024-10-25
Add level indicators and path to stock location and part category dropdowns

### LINKED ISSUE #8350: Full hierarchy in 'Default Location' dropdown not presented to user

### Describe the bug*

When choosing a default stock location for a part or category, the hierarchy of stock locations is not visible in Platform UI whereas it was in Classic UI.

Based on the hierarchical way in which stock locations can be set up in InvenTree, I set up all my stock locations with the location itself _not_ including details of its parent locations.

For example, I have two stock locations with the name "B01", but one is a child of Drawer Unit 1 and one is a child of Drawer Unit 2. This has always worked well in Classic UI due to the way that the full location hierarchy is always visible.

Image

In Platform UI, the hierarchy is hidden, and I am presented with two apparently identical options of "B01" or "B01", with no additional information to help me decide which one I need to select.

Image

I have tried to illustrate the differences in behaviour in this short clip. First, the expected behaviour in Classic UI, followed by the new behaviour in Platform UI.

https://github.com/user-attachments/assets/5945129b-43d2-4292-9bfb-72f971afa480

### Steps to Reproduce

1. Add a new part category
2. Select the **Default Location** dropdown box
3. Begin to type the name of a location
4. In Platform UI, filtered results of search may be indistinguishable from one another.

### Expected behaviour

The hierarchy of stock locations is presented to the user as it is in Classic UI.

### Deployment Method

- [x] Docker
- [ ] Package
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.16.5
Django Version: 4.2.15
Commit Hash: 6e37f0c
Commit Date: 2024-10-07

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-5.10.0-27-amd64-x86_64-with
Installer: DOC

Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.1.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}, {'name': 'InvenTreeCurrencyExchange', 'slug': 'inventreecurrencyexchange', 'version': '1.0.0'}, {'name': 'InvenTreeLabel', 'slug': 'inventreelabel', 'version': '1.1.0'}, {'name': 'InvenTreeLabelMachine', 'slug': 'inventreelabelmachine', 'version': '1.0.0'}, {'name': 'InvenTreeLabelSheet', 'slug': 'inventreelabelsheet', 'version': '1.0.0'}, {'name': 'DigiKeyPlugin', 'slug': 'digikeyplugin', 'version': '1.0.0'}, {'name': 'LCSCPlugin', 'slug': 'lcscplugin', 'version': '1.0.0'}, {'name': 'MouserPlugin', 'slug': 'mouserplugin', 'version': '1.0.0'}, {'name': 'TMEPlugin', 'slug': 'tmeplugin', 'version': '1.0.0'}, {'name': 'Brother Labels', 'slug': 'brother', 'version': '0.9.0'}]

### Relevant log output

_No response_

chris-thorn:
This fixes #8350 and makes the stock location dropdown in the new interface closely match the dropdown in the classic interface.

Adds hyphens to indicate hierarchy level, and shows stock location path instead of simply stock location name.


SchrodingersGat:
@chris-thorn the implementation looks good here. Can you please implement the same thing for "part category" too?

chris-thorn:
I hope I've correctly updated this PR with the requested changes!

SchrodingersGat:
It looks like some of the tests are failing now due to the adjusted naming scheme.

You will need to adjust the following unit tests (in `./src/frontend/tests`):

### Forms - Stock Item Validation
image

### Pages - Index - Scan (StockLocation)

image

In both cases you can see that the expected text has changed, causing the test(s) to fail...

wolflu05:
But do we really want to have the full path name as name on each level? I mean if the tests expect something different,we change that behavior?

SchrodingersGat:
@wolflu05 AFAIK this is only intended to change the rendering in form drop downs - as is the current behaviour in CUI

wolflu05:
Im not sure what I would prefere, I guess none of both. In the end my goal is to implement a tree picker like I did in CUI for all tree models, and then show all parent levels, when a search query matches only a child. But ok then change this for now.

chris-thorn:
> But do we really want to have the full path name as name on each level? I mean if the tests expect something different,we change that behavior?

@wolflu05 I agree that neither is ideal, and that a tree picker would be perfect for this.

Just to clarify my reasoning for this PR: Right now in PUI, if stock location or part category names don't include information about the hierarchy when you search for a location or category, you potentially see multiple identical names but with no indication of which is the correct one. It's only when you hover over it that you can see the hierarchy.

```
Bay 1
├── Shelf 1
│   └── Bin 1
└── Shelf 2
    └── Bin 1
```

Searching for `Bin 1` will produce two results but will not display information about their parent locations, so you don't know whether you're picking the bin on `Shelf 1` or `Shelf 2`.

Until the tree picker is ready for implementation, it's my feeling that the dropdown should behave like the one in CUI.

@SchrodingersGat Ah, sorry about that - I will update the PR today.

chris-thorn:
I've changed [pages/pui_scan.spec.ts](https://github.com/inventree/InvenTree/blob/c062d5e48df72df8c19db962dadbc5bf868e3fc8/src/frontend/tests/pages/pui_scan.spec.ts#L70) but I couldn't work out what needed changing in [pui_forms.spec.ts](https://github.com/inventree/InvenTree/blob/c062d5e48df72df8c19db962dadbc5bf868e3fc8/src/frontend/tests/pui_forms.spec.ts#L42). `Electronics Lab` is a top-level stock location so shouldn't require any additional path detail... Unless I've misunderstood?

SchrodingersGat:
@chris-thorn are you using the interactive window in playwright to determine what specific locators you should be looking for?

SchrodingersGat:
> ERROR: cannot install on alpine distribution - only Ubuntu and Debian are supported'

Ah, this might be a real hole in our process. I don't use the devcontainer for frontend testing, and so have not run into this issue.

@matmair @wolflu05 do you have any experience running playwright inside our devcontainer?

This might be another good reason to move to a debian based docker image...

@chris-thorn are you running on windows? My dev setup is under WSL (ubuntu) and might be easier to get setup there?

chris-thorn:
Right - that makes sense. Yeah, I'm running the devcontainer in Windows.

I will have a play with WSL, or just try running directly on an Ubuntu server.

SchrodingersGat:
For the purpose of this PR, change line `/src/frontend/tests/pui_forms.spec.ts:43` from:

```javascript
await page.getByRole('option', { name: /Electronics Lab/ }).click();
```

to

```javascript
await page.getByText('Electronics production facility').click();
```

matmair:
I have not used playwright within devcontainers

SchrodingersGat:
@chris-thorn are you able to make the small fix as outlined above?

chris-thorn:
> @chris-thorn are you able to make the small fix as outlined above?

Sorry, was away over the weekend. I've now push the change to this branch.

SchrodingersGat:
@chris-thorn nice work, thanks for the contribution :) I hope that you can continue to implement good features like this!

Issue #5754, opened 2023-10-19
[PUI] Remove custom themes


SchrodingersGat:
Current support for "custom themes" will be removed when we make the switch to the new react UI

matmair:
@SchrodingersGat what is the action item here? Remove the backend function?

SchrodingersGat:
Yes, basically all the current theming support, once we remove the existing UI code

wolflu05:
But we could still sync the theme for one user, I mean there is still a dark and light mode?

SchrodingersGat:
@wolflu05 yes that is still there, this is just a reminder to cull the old code once we switch

matmair:
Will be needed after PUI is GA

matmair:
This is addressed in #8384

SchrodingersGat:
Duplicate of https://github.com/inventree/InvenTree/pull/8384

PR #8384, opened 2024-10-28
Remove CUI

matmair:
This PR removes CUI and adjacent helpers that are no longer required.

Part of the preparation for 1.0 as [listed in the blog](https://inventree.org/blog/2024/09/23/ui-roadmap).

closes https://github.com/inventree/InvenTree/issues/5754


SchrodingersGat:
That's a *lot* of code removed :)

matmair:
@SchrodingersGat looking through coverage several questions have popped up:
- do we still need `BomUploadTemplate` and `BomDownload`
https://github.com/matmair/InvenTree/blob/11a5f309b51186f64f808548ecebfbec2da6c09d/src/backend/InvenTree/part/views.py#L12-L25
- we have a lot of properties that are only used in the HTML templates, not the API - should we keep or remove them?
- there a bunch of django-import-export stuff still in here - should I remove that too?

SchrodingersGat:
> do we still need BomUploadTemplate and BomDownload

For now - all the features of BomDownload have not been ported across to the new framework yet.

> we have a lot of model properties that are only used in the HTML templates, not the API - should we keep or remove them?

Happy to remove if they are only ever called from the HTML templates. I'd like that as a separate PR so we can review each removed function closely

> there is a bunch of django-import-export stuff still in here - should I remove that too?

Let's do that as a separate PR - I want to ensure that all the import/export functionality has been replicated in the new system

matmair:
> Happy to remove if they are only ever called from the HTML templates. I'd like that as a separate PR so we can review each removed function closely

Should this be one PR per model or all models in one?



> Let's do that as a separate PR - I want to ensure that all the import/export functionality has been replicated in the new system

I have something rough for this in the pipeline, will keep it rough till 0.17.0 is through

matmair:
@SchrodingersGat this is ready for review and merge once 0.17.0 is done; there will be some smaller follow-ups probably (removing all ajax, removing some frontend dependencies in backend). Keeping this in this state to leave it review-able.

SchrodingersGat:
@matmair please fix conflicts here :)

matmair:
@SchrodingersGat fixed

SchrodingersGat:
@matmair should we remove the frontend URL prefix "platform" now? Or would you like to do that as a separate PR

matmair:
I would like to keep changes to PUI separate from this whole CUI removal

matmair:
+ We need to consider the migration story with old CUI links

SchrodingersGat:
What a milestone! Thanks for all the amazing effort on this one @matmair 

Obviously as this is a *huge* change, we will have a lot of follow up and potential triage here. Let's keep an eye on it .

PR #8386, opened 2024-10-28
fix link to MIT license

matmair:
Fixes the opensource.org link and ignores opensource.org as it seems to not wokr on CI


matmair:
@SchrodingersGat @wolflu05 this seems to work locally but not on CI - any idea why?

SchrodingersGat:
The website may be rate limiting or blocking requests coming from github?

matmair:
Seems like it, my tests in https://github.com/inventree/InvenTree/pull/8384 are failing because of this. Should we stop linting for this URL or wait a few days?

SchrodingersGat:
Maybe we just stop linting this one?

matmair:
@SchrodingersGat added an ignore

PR #8395, opened 2024-10-30
[PUI] Table refactor

SchrodingersGat:
- Split table header component into different file, for better maintainability
- Add option to remove header entirely


matmair (approved):
Looks good; it seems that the header less feature was not used - is that planned in this PR?

PR #8396, opened 2024-10-30
[PUI] stock table

SchrodingersGat:
- Add "SKU" column to StockItem table
- Add "MPN" column to StockItem table
- Add extra fields to data export

### DIFF
--- a/src/backend/InvenTree/stock/serializers.py
+++ b/src/backend/InvenTree/stock/serializers.py
@@ -340,7 +340,7 @@ class StockItemSerializer(
     - Includes serialization for the item location
     """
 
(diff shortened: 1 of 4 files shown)


src/backend/InvenTree/stock/serializers.py
```diff
@@ -340,7 +340,7 @@ class StockItemSerializer(
     - Includes serialization for the item location
     """
 
-    export_exclude_fields = ['tracking_items']
+    export_exclude_fields = ['notes', 'tags', 'tracking_items']
```
matmair, line 343: Q: Why are notes removed?

matmair (approved):

Issue #8372, opened 2024-10-26
[FR] Return order for non-serializable (trackable) items


baltuonis:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

We are using InvenTree to manage our small non-profit book warehouse. So far it served us a lot and we love how nifty it is.

We are trying to register Return Orders for books that get returned from our sellers/book fairs.
Unfortunately Return Orders only allow serialized (trackable) parts to be returned.
We ship dozens of books weekly so it is not practical to serialize (track) them.

### Suggested solution

Is there a way to return non-serialized (non trackable) parts?

### Describe alternatives you've considered

Modify the code ourselves?
Is there a simple/non-destructive way to modify the code for behavior this to work?
We have invested in this system and it would be tedious to migrate.

### Do you want to develop this?

- [x] I want to develop this.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/inventree) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/inventree/InvenTree/issues/8372">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/inventree/InvenTree/issues/8372/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/inventree/InvenTree/issues/8372/pledge.svg">
</picture>
</a>

SchrodingersGat:
The return order was focused on "serialized" items to allow full traceability of items. If you have a method of discriminating between individual stock items (other than the serial number) then I have no *idealistic* objection to removing this limitation

I would probably suggest controlling this via a configurable setting, so that the behaviour could be configured.

Would you be willing to contribute some funding to this issue? It should not take a significant amount of time to implement this.

baltuonis:
Hey @SchrodingersGat , Thanks for quick response!


1. I'm not very familiar with InvenTree internals, but as a KISS suggestion - could it just return the item(s) to the stock without location set?
2. Ideally for our use-case it: would allow setting returned amount and destination location.
3. I agree on configurable setting. I see how a lot of users won't be comfortable with this loose returns tracking.
4. I'm helping a volunteer organisation to modernize their processes so this is entirely on me. How much $$$ would make a worthy contribution? And how long could it take to implement? (weeks? months?)

[...]

PR #8400, opened 2024-10-31
[FR] Support non-serializable (part) item Return Orders

baltuonis:
Adding support for non-serializable (part) item Return Orders

Feature request: #8372


baltuonis:
@SchrodingersGat I was not sure how to add a parameter to disable/enable this. If you could point me to the right location, I could try adding it.

SchrodingersGat:
@baltuonis thanks for submitting the PR! This approach is not exactly what I had in mind, what I envisioned was that you would simply remove the restriction that the returned item had to be serialized:

https://github.com/inventree/InvenTree/blob/feefa60a8f11971cadd43cc58e33aa43ce68f325/src/backend/InvenTree/order/models.py#L2408

What you have implemented here is a lot more complex, and I am supposing that what you actually want to be able to achieve is to create a "return order" for a stock item which does not "exist" in the system? It that correct?

For example, if a non-serialized item has been shipped to a customer, that item should be able to be returned via the existing processes, if we allow "non serialized" items to be added to a ReturnOrder via the `ReturnOrderLineItem` model. I do not think that these extra models you are adding here are necessary.

Perhaps you can expand a bit more on what exactly you are trying to achieve, based on the context above?

baltuonis:
@SchrodingersGat right, so this was my first "hold my beer, I'm gonna make a PR in 15 mins" approach.
But then I've bumped into an issue where I can't return more than 1 item in the same Return Order (RO), because of *unique(item,order)* constraint. 

If I want to return them, I have to create a bunch of stock items for the same product. Now imagine, we have return orders with 17 units of item1, 5 units of item2, etc. That where I had to add `quantity` field.

Another problem (as you've suggested) is that sometimes there are items that are not present in the main warehouse and we have to issue an RO that includes them... That's where I added `part` field. 

So then I've realized that adding those fields + all the surrounding logic inside a `ReturnOrderLineItem` would result in a weird polymorphic object that will complicate everything long term.

That's why in the end I've added a separate section with a new model. I think that could be turned off/on with a flag in admin settings.

SchrodingersGat:
@baltuonis can I suggest a simpler approach, which I think should work for you here.

Let's say you have sent a quantity of 100x books to a customer. This appears as a single StockItem (with quantity 100x) at the customer.

The customer wants to return 15 of these books. Rather than adding another layer of complication with a new model that maps a "quantity" of the item to be returned, I'd suggest the following process:

1. "Split" the single stock item into two, 1 with quantity 85 and 1 with quantity 15.
2. Issue a return order for the quantity 15
3. Repeat the process for returning other smaller quantities

This utilizes processes which already exist, and does not require this additional alternative process.

To make this happen you would need to make at least the following adjustments to the codebase:

a) Allow stock items which are "at" a customer to be "split" - currently this is prevented, I think
b) Enable non-serialized stock items to be part of return orders

Does this make sense? I strongly feel it is a cleaner alternative which will work better for you anyway

SchrodingersGat:
@baltuonis I have a PR here which will implement this - https://github.com/inventree/InvenTree/pull/8590

I think this new approach is a bit cleaner, and avoids multiple different edge case conditions that might otherwise need to be accounted for.

If you can provide a small amount of funding against https://github.com/inventree/InvenTree/issues/8372 (via the "Fund This Issue" button) to support the project that would be appreciated :)

SchrodingersGat:
Closed by https://github.com/inventree/InvenTree/pull/8590

PR #8401, opened 2024-10-31
Prevent multiple background workers without global cache

SchrodingersGat:
This PR enforces a single background worker if global caching (e.g. via redis) is not enabled.

This is because without a global cache, there is no way to properly  lock the individual workers to ensure that a scheduled task is only run once.

If global cache is enabled (currently we support redis) then multiple workers can be run in parallel, as the global cache allows tasks to be scheduled correctly.

The underlying issue was discovered as part of https://github.com/inventree/InvenTree/pull/8399 - some very hard-to-observe race conditions!


matmair (approved):
LGTM; should probably be marked as breaking and mentioned in the release notes as this has the potential to slow down systems a lot as a lot fewer queues are available for background tasks

SchrodingersGat:
@matmair good idea about the breaking change. I'll mark it as such and add a note to the docs too

PR #8403, opened 2024-10-31
Purchase Order Destination

SchrodingersGat:
- Adds "destination" field to the PurchaseOrder model
- Specify (at the order level) where items should be received into
- Can be overridden by "destination" field on the PurchaseOrderLineItem model

### TODO

- [x] Update documentation
- [x] Playwright testing


matmair (approved):
LGTM

PR #8411, opened 2024-11-02
Bug fix for plugin table

SchrodingersGat:
- Allow "deactivation" of missing plugin
- Required before it can be deleted


wolflu05:
Is that the bug that a plugin could sometimes not directly be activated after installing it because something weird is in the browser cache?

SchrodingersGat:
No, that must be a different thing?

wolflu05:
ok, thats a really strange one, because I cannot reproduce it every time. But maybe that's already fixed. Ill see if I see it sometime again.

PR #8431, opened 2024-11-05
Improve optional chaining checks

SchrodingersGat:
- Fixes bug where part_detail is potentially undefined


matmair (approved):
Lgtm; solves a few warnings from the sast too

PR #8434, opened 2024-11-05
[PUI] Updates for AdminButton

SchrodingersGat:
### Tasklist

- [x] Hide if django admin interface is disabled
- [x] Use correct admin URL
- [x] Add playwright tests for admin buttons on various pages

### DIFF
--- a/src/backend/InvenTree/InvenTree/api.py
+++ b/src/backend/InvenTree/InvenTree/api.py
@@ -230,13 +230,22 @@ class InfoView(AjaxView):
             'debug_mode': settings.DEBUG,
             'docker_mode': settings.DOCKER,
             'default_locale': settings.LANGUAGE_CODE,
-            # Following fields are only available to staff users
-            'system_health': check_system_health() if is_staff else None,
-            'database': InvenTree.version.inventreeDatabase() if is_staff else None,
-            'platform': InvenTree.version.inventreePlatform() if is_staff else None,
-            'installer': InvenTree.version.inventreeInstaller() if is_staff else None,
-            'target': InvenTree.version.inventreeTarget() if is_staff else None,
         }
+
+        if is_staff:
+            # Following fields are only available to staff users
+            data.update({
+                'system_health': check_system_health() if is_staff else None,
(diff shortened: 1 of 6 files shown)


src/backend/InvenTree/InvenTree/api.py
```diff
@@ -230,14 +230,23 @@ def get(self, request, *args, **kwargs):
...
-            'database': InvenTree.version.inventreeDatabase() if is_staff else None,
-            'platform': InvenTree.version.inventreePlatform() if is_staff else None,
-            'installer': InvenTree.version.inventreeInstaller() if is_staff else None,
-            'target': InvenTree.version.inventreeTarget() if is_staff else None,
         }
 
+        if is_staff:
+            # Following fields are only available to staff users
+            data.update({
+                'system_health': check_system_health() if is_staff else None,
```
matmair, line 238: These checks are redundant with the check introduced above

src/backend/InvenTree/InvenTree/api.py
```diff
@@ -230,14 +230,23 @@ def get(self, request, *args, **kwargs):
...
             'default_locale': settings.LANGUAGE_CODE,
-            # Following fields are only available to staff users
-            'system_health': check_system_health() if is_staff else None,
-            'database': InvenTree.version.inventreeDatabase() if is_staff else None,
-            'platform': InvenTree.version.inventreePlatform() if is_staff else None,
-            'installer': InvenTree.version.inventreeInstaller() if is_staff else None,
-            'target': InvenTree.version.inventreeTarget() if is_staff else None,
         }
 
+        if is_staff:
```
matmair, line 235: This means that the API response is not uniform and depends on the status of the current user. This makes integrations and probing checks like in https://tools.invenhost.com/check/ more of a hassle; is there a particular reason for this change?

matmair:
Grteat to see the ToDos adressed but I am not sure about the API response change

SchrodingersGat:
@matmair I'll revert the changes to the API endpoint as suggested.

Issue #7990, opened 2024-08-25
Custom Stock Status cannot be set from Stock Actions


gunstr:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

A configured custom stock status such as "TestState" is not visible in the drop down list in the StockAction modal.

I can set the TestState on an item from the Edit Stock Item modal but I noticed that if then TestState is selected as a filter to the Stock Item list no itmes are visble. I do not know if this is related or maybe it should be a separate issue?

### Steps to Reproduce

This test is done on the Demo site only, I have not installed PUI locally yet.

1. Create a Custom Stock status
2. Open the Stock Items list, click Stock Actions and select Change Stock Status

![image](https://github.com/user-attachments/assets/04c3d97e-0fc8-4e65-9168-3d747bb78b67)

3. In the Status dropdown the TestState option is not visible

![image](https://github.com/user-attachments/assets/8812dfc8-9063-48a2-aa27-25842f2cc9dc)

### Expected behaviour

It should be possible to set custom stock status also from the Stock Action modal

### Deployment Method

- [ ] Docker
- [ ] Package
- [ ] Bare metal
- [X] Other - added info in Steps to Reproduce

### Version Information

InvenTree-Version: 0.17.0 dev
Django Version: 4.2.15
Commit Hash: 6951748
Commit Date: 2024-08-25
Commit Branch: null
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-5.15.0-100-generic-x86_64-with
Installer: DOC
nullActive plugins: false

### Please verify if you can reproduce this bug on the demo site.

- [X] I can reproduce this bug on the demo site.

SchrodingersGat:
@gunstr thanks for reporting, we will have to look into this

matmair:
@SchrodingersGat can we move this to 1.0?

SchrodingersGat:
@matmair sure, it requires a bit more attention and I would like to get the new release out.

gunstr:
@SchrodingersGat @matmair is your recommendation not to use the cusom states at all until you have worket this trough a bit more?

SchrodingersGat:
@gunstr I think that for this particular issue, you can manually adjust the stock status per stock item, rather than as a group?

gunstr:
[...]

PR #8438, opened 2024-11-06
[Refactor] Custom states

SchrodingersGat:
This PR is an update to the "custom states" functionality, to improve usability and extend user interface integration

- Closes https://github.com/inventree/InvenTree/issues/7990
- Closes https://github.com/inventree/InvenTree/issues/8612#

### Tasks

- [x] Improved back-end validation
- [x] Improved form management for creating / editing a custom state
- [x] Display proper custom state in PUI for individual detail views
- [x] Display proper custom state in PUI for tables
- [x] Ensure matching logical state is set when transitioning to a custom state
- [x] Check that "custom" state can only be set to appropriate values
- [x] Add custom states to the demo dataset - https://github.com/inventree/demo-dataset/pull/86
- [x] Extend existing unit testing to ensure that all new validations checks are tested
- [x] Add playwright tests for custom states (after demo dataset is updated)
- [x] Ensure table filtering works with custom state values - https://github.com/inventree/InvenTree/issues/8612#

### Future Work / Extensions

- Revisit removing "model" field from CustomState table (or at least making it optional)
- Update documentation on custom states


SchrodingersGat:
@matmair I have been looking at the implementation of custom states and there are a few areas which I think can be cleaned up a bit.

One question I have is: what is the value of the `model` (ContentType) field on the `InvenTreeCustomUserStateModel` model? We already map to a status group, which implies a given model type. I don't see the value of having a separate field (which the user has to know how to fill in) - which can conflict with the specified statuscode.

I would suggest that we remove the `model` field and rely on the `reference_status` field to perform this functionality

matmair:
@SchrodingersGat there is only an implicit link between status groups and models. When you start shipping custom models in plugins and using custom states there you need some way to prefilter possible selections.

matmair (approved):

SchrodingersGat:
I don't understand the use case here - in what circumstance would you want to make a custom state which overrides a `BuildStatus` code, but links to a `SalesOrderShipment` model? (for example)

matmair:
The use case is enabling order when you have 40+ state groups and need to filter to manage them. If you see no value you can remove this and I will patch my custom fork to support, it seems anyway like I will never be able to use upstream

wolflu05:
*(Just a question, no opinion here, I just want to understand your design choice)*

Do you have any example scenario where custom states and models are in a relationship other than n:1? (E.g. the same custom state is assigned to multiple models. If not, doesn't that say that each custom state has only one model, which implicitly means there is a reliable link between a state and the model.

matmair:
Implicit is the key here; there is no reliable way to map a custom state to a model, thus automations that need to make this link will not be able to function anymore if it is removed. I do not rely on implicit mapping in my business critical lob apps.

SchrodingersGat:
Looking into this further, I think that if you want to link a "status group" to a particular database model, that should be done at the class level, rather than linking for each custom status field. Otherwise, you could end up with a (supposedly valid) situation as below:

| Status Reference | Linked State | Custom State | Linked Model |
| --- | --- | --- | --- |
| PurchaseOrderStatus | PENDING | AWAITING_APPROVAL | PurchaseOrder |
| PurchaseOrderStatus | PENDING | OTHER_STATUS_A | PurchaseOrder |
| PurchaseOrderStatus | PENDING | OTHER_STATUS_B | User |

There is no validation that custom status values must point to the *same* model, or that the linked model must be in any way related to the parent "status reference". To me, this does not make much sense. I would suggest that we a `get_model_class()` method to the `StatusCode` class - which by deafult returns `None`. Then, any subclass can reimplement this to point to a given db model class (if required).

It removes the requirement that the user *knows* what model a custom status needs to point to (currently this must be picked as it is a required field).

matmair:
I really would have liked to discuss these things before the original PR was merged. I built a ton of plugins and integrations around the merged (and thus I thought agreed upon) state and at this point, I am just frustrated when I open up GitHub with the instability of interfaces.

SchrodingersGat:
@matmair I probably should have reviewed this more closely the first time around. As part of trying to fix https://github.com/inventree/InvenTree/issues/7990 I have identified a number of bugs / shortcomings of the current implementation. I am trying my best to rectify these without too many changes to the existing interface.

One of the outstanding issues is the "model allocation". I would like to understand more about how you are using it, and whether a strict in-code link between the particular StatusCode sub-class and a DB model would be sufficient

src/backend/InvenTree/InvenTree/models.py
```diff
@@ -78,6 +78,81 @@ def has_field_changed(self, field_name):
         return field_name in self.get_field_deltas()
 
 
+class StatusCodeMixin:
```
matmair, line 81: I think this would better fit into the generic / states part of the code base to ensure we provide proper decoupling in tests and code

Issue #8155, opened 2024-09-20
Problem after installing 0.16.2


tnemrap:
### Deployment Method

- [x] Installer
- [ ] Docker Development
- [ ] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

After update to 0.16.2 i get 
OSError: No translation files found for default language en-us.

### Steps to Reproduce

I just run inventree run cli update

### Relevant log output

Installing plugin packages from '/etc/inventree/plugins.txt'
Python version 3.11.2 - /opt/inventree/env/bin/python3
Traceback (most recent call last):
  File "/opt/inventree/src/backend/InvenTree/manage.py", line 24, in <module>
    main()
  File "/opt/inventree/src/backend/InvenTree/manage.py", line 20, in main
    execute_from_command_line(sys.argv)
  File "/opt/inventree/env/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/opt/inventree/env/lib/python3.11/site-packages/django/core/management/__init__.py", line 416, in execute
    django.setup()
  File "/opt/inventree/env/lib/python3.11/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/opt/inventree/env/lib/python3.11/site-packages/django/apps/registry.py", line 116, in populate
    app_config.import_models()
  File "/opt/inventree/env/lib/python3.11/site-packages/django/apps/config.py", line 269, in import_models
    self.models_module = import_module(models_module_name)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1206, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1149, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
[...]

PR #8441, opened 2024-11-06
fix: Narrow scope of cleanup

matmair:
This fixes the `clear-generated` task. It was previously too widely scoped and had the potential to brick users envs if they were under /opt/inventree (which is the default).
This fix narrows the scope to the "src" directory - which is always guaranteed to be InvenTree internal stuff.

Fixes https://github.com/inventree/InvenTree/issues/8252 https://github.com/inventree/InvenTree/issues/8155

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -1443,14 +1443,14 @@ def docs_server(c, address='localhost:8080', compile_schema=False):
 def clear_generated(c):
     """Clear generated files from `invoke update`."""
     # pyc/pyo files
-    run(c, 'find . -name "*.pyc" -exec rm -f {} +')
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
What happens if the invoke task is run from outside the source directory? Should we also confirm that the command is being run from the correct directory first?

matmair:
that should not be a problem, the command uses the `run` wrapper that ensures that all commands are run from the root of the install

matmair:
See https://github.com/inventree/InvenTree/blob/c062d5e48df72df8c19db962dadbc5bf868e3fc8/tasks.py#L139-L150

PR #8445, opened 2024-11-07
Sales order improvements

SchrodingersGat:
This PR provides some general UI improvements for managing sales orders.

Additionally, it allows stock items to be allocated against a sales order, but without specifying a shipment. This is a useful addition as it allows users to allocate stock items against the sales order first, and decide on shipping details later on.


matmair (approved):

src/frontend/src/tables/sales/SalesOrderLineItemTable.tsx
```diff
@@ -163,13 +187,15 @@ export default function SalesOrderLineItemTable({
...
-        )
+        render: (record: any) => {
+          return (
+            <ProgressBar
+              progressLabel={true}
+              value={record.allocated}
+              maximum={record.quantity}
+            />
+          );
+        }
```
matmair, line 198: Question out of curiosity - is there a functional difference between these 2 implementations? I can not see one - eager to improve my js knowledge

matmair (approved):

src/frontend/src/tables/sales/SalesOrderLineItemTable.tsx
SchrodingersGat: No, these are functionally equivalent, one returns a component, and the other a function which generates a component. I wonder if one is more "efficient" than the other? In any case, I think that this might have been a change I started to make and then nothing really came of it. I'll revert this diff, there's no need for it.

matmair: thank you for the explanation - I thought I missed something

PR #8450, opened 2024-11-07
UI consolidation

SchrodingersGat:
- Combine "suppliers" and "manufactuers" onto same tab

### DIFF
--- a/src/frontend/src/components/items/StylishText.tsx
+++ b/src/frontend/src/components/items/StylishText.tsx
@@ -1,4 +1,4 @@
-import { Accordion, Text } from '@mantine/core';
+import { Text } from '@mantine/core';
 
 import * as classes from '../../main.css';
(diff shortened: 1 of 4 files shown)


src/frontend/src/components/items/StylishText.tsx
```diff
@@ -1,4 +1,4 @@
-import { Text } from '@mantine/core';
+import { Accordion, Text } from '@mantine/core';
```
matmair, line 1: Q: Is this import really needed?

matmair (approved):

src/frontend/src/components/items/StylishText.tsx
SchrodingersGat: No, that looks like it has crept in by mistake, I'll fix that

PR #8454, opened 2024-11-08
Invoke error handling

SchrodingersGat:
Adds an error handler to the invoke calls in `tasks.py`. This should make installation errors clearer to the user, that an *ERROR* has occurred during the installation process (rather than failing somewhat silently):

image


matmair:
Good improvement! If you need feedback on a PR I will need a few days at least to get to it for the foreseeable future.

Issue #7869, opened 2024-08-14
[FR] Improve page titles for PUI


SchrodingersGat:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

Currently the page title only shows the instance name, and does not reflect the content of the current view.

This means that the "history" tab in the browser is not very useful:

![image](https://github.com/user-attachments/assets/7cc229f3-4893-4dee-8f9b-947781c0b5b8)

### Suggested solution

- Display information on the current "route" in the page title

### Describe alternatives you've considered

N/A

### Examples of other systems

N/A

### Do you want to develop this?

- [X] I want to develop this.

ATLAS2246:
I'd like to try this as a learning task, is that ok?

SchrodingersGat:
@ATLAS2246 absolutely :) your contribution would be warmly welcomed

SchrodingersGat:
@ATLAS2246 did you ever get anywhere with this?

PR #8467, opened 2024-11-12
[PUI] Page titles

SchrodingersGat:
- Update PUI page titles based on current route
- Closes https://github.com/inventree/InvenTree/issues/7869


src/frontend/src/tables/InvenTreeTable.tsx
```diff
@@ -498,6 +498,7 @@ export function InvenTreeTable<T extends Record<string, any>>({
       tableState.activeFilters,
       tableState.searchTerm
     ],
+    retry: 3,
```
matmair, line 501: Q: Is this intended to be in this PR?

SchrodingersGat: No, thanks for the catch

PR #8478, opened 2024-11-12
fix refreshInstance type and remove return

matmair:
Fixes a bunch of type errors ie. https://sonarcloud.io/project/issues?open=AZLptXluyTR5NnVS1-LJ&id=inventree_InvenTree

### DIFF
--- a/src/frontend/src/hooks/UseInstance.tsx
+++ b/src/frontend/src/hooks/UseInstance.tsx
@@ -1,4 +1,4 @@
-import { type QueryObserverResult, useQuery } from '@tanstack/react-query';
+import { useQuery } from '@tanstack/react-query';
 import { useCallback, useMemo, useState } from 'react';
 
 import { api } from '../App';
@@ -8,7 +8,7 @@ import { type PathParams, apiUrl } from '../states/ApiState';
 export interface UseInstanceResult {
   instance: any;
   setInstance: (instance: any) => void;
-  refreshInstance: () => Promise<QueryObserverResult<any, any>>;
+  refreshInstance: () => void;
   instanceQuery: any;
   requestStatus: number;
   isLoaded: boolean;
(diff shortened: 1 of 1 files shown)


src/frontend/src/hooks/UseInstance.tsx
```diff
@@ -8,7 +8,8 @@ import { type PathParams, apiUrl } from '../states/ApiState';
 export interface UseInstanceResult {
   instance: any;
   setInstance: (instance: any) => void;
-  refreshInstance: () => Promise<QueryObserverResult<any, any>>;
+  refreshInstance: () => void;
+  refreshInstancePromise: () => Promise<QueryObserverResult<any, any>>;
```
wolflu05, line 12: What is the reason for this change? You don't need to await the Promise if you don't want to where you use `refreshInstance`?

wolflu05: Dosen something like:
```diff
 return {
    sessionData,
    setSessionData,
    sessionId,
-    refreshSession
+    refreshSession: () => { refreshSession() },
    sessionQuery,
    status,
```

would fix the error to, but without cluttering the useInstance hook with unnecessary functions.

matmair: I do not see how this would fix the issue

SchrodingersGat: The additional wrapper promise function does seem clunky.. I'm sure there should be a better way to suppress the warning

matmair: I do note really care how this is solved but there are 28 issues with severity middle caused by the underlying change - those should be addresses https://sonarcloud.io/project/issues?rules=typescript%3AS6544&issueStatuses=OPEN%2CCONFIRMED&id=inventree_InvenTree

SchrodingersGat:
@matmair @wolflu05 I'm happy enough with this to merge :)

PR #8501, opened 2024-11-16
Enable redis by default for docker setup

SchrodingersGat:


wolflu05:
Can you add Resis as a dependency for the inventree server container? So that when Running commands like docker compose run ... invoke update  the cache boots up too.

PR #8502, opened 2024-11-16
Docker venv

SchrodingersGat:
This PR changes behavior under docker, using the python virtual environment even in "production" mode.

The purpose for this is to ensure that any extra python packages that get installed (i.e. plugins) persist between container runs, and across the webserver and worker containers.

### The Problem

Ref: https://github.com/inventree/InvenTree/issues/7709

The webserver and background worker processes run on separate container instances - meaning that they have different python package installation environments.

Thus, if a plugin is installed via the webserver (in the admin interface), it gets *installed* into the webserver container instance, but the background worker instance does not have a copy of the newly installed plugin code. When it tries to find it, it fails, and can result in some different subtle bugs (as well as the obvious "plugin does not work" issue).

As an added complication, if the background worker process is restarted, it *may* successfully result in the plugin being correctly installed inside that container, due to the auto-install behaviour of `plugins.txt` - however, this is kinda flaky and the static files don't get copied across reliably in this scenario.

### The Solution

In the `init.sh` entrypoint script for both containers, active a python virtual environment which points to a subdirectory on the externally mounted volume. This means that:

A) Installed plugin packages persist on the host machine (via mounted volume)
B) Installed plugin packages are shared between the webserver and background worker containers

### Testing

You can test that this works on an existing docker installation by adding the following line to the `.env` file:

```python
# Enable python virtual environment
INVENTREE_PY_ENV=/home/inventree/data/venv
```

Then, restart the docker containers.

Install a new plugin and verify that the plugin has been installed in the `<data>/venv/` subdirectory in the mounted volume

### Base Packages

Base python packages (that is, everything in `requirements.txt` needed for core InvenTree functionality) is installed *inside the container image* - when the container is built. Thus, only secondary packages (user installed plugins, etc) will persist in the external venv.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -40,7 +40,6 @@ services:
       INVENTREE_PLUGINS_ENABLED: True
       INVENTREE_SITE_URL: http://localhost:8000
       INVENTREE_CORS_ORIGIN_ALLOW_ALL: True
-      INVENTREE_PY_ENV: /home/inventree/dev/venv
 
     depends_on:
       - db
--- a/contrib/container/Dockerfile
+++ b/contrib/container/Dockerfile
@@ -37,7 +37,7 @@ ENV INVENTREE_PLUGIN_DIR="${INVENTREE_DATA_DIR}/plugins"
 # Location for python virtual environment
 # This will be created on the host system to persist *certain* packages
 # Note: *base* packages are installed into the docker image
-ENV INVENTREE_PY_ENV="${INVENTREE_DATA_DIR}/env"
+ENV INVENTREE_PY_ENV="${INVENTREE_DATA_DIR}/venv"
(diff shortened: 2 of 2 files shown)


wolflu05 (commented):
Thank you very much for investigating this. This could indeed fix the issues. I will have to take a closer look and test this myself the next days. This may also explain the <property xxx> strings shown instead of the plugin names, but not sure about that one.

SchrodingersGat:
It does not appear to by working correctly yet, as the server runs system python and invoke binnaries (`/usr/local/bin/python`) rather than the python in the "venv" directory (`/home/inventree/data/venv/bin/python`) - which means that the venv is bypassed anyway..

SchrodingersGat:
Closing this out in favour of https://github.com/inventree/InvenTree/pull/8503

Issue #4150, opened 2023-01-04
[FR] Add telemetry


matmair:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### Problem statement

We currently neither know which version users are running nor on which systems and if plugins or other 'new' features are used.

### Suggested solution

Implement user telemetry that reports back a few non-identifying instance metrics daily or weekly. Maybe also on plugin installs/activation. Must be Opt-In and all submitted data should be transparent - maybe a log of the last x events that were sent could be visible to superusers.

This should be discussed thoroughly before implementation - probably as a plugin. Has the potential to be controversial, especially with EU citizens.

### Describe alternatives you've considered

We could try to use other gauging mechanisms like switching the update checker to a custom server and using the user-agent to get passive data. That feels a bit sneaky though.

### Examples of other systems

I love the way octoprint does it: https://tracking.octoprint.org/. They expose the results to https://data.octoprint.org/ - which feels like a good idea in general. ELK stack + Grafana on a VPS might be a good solution.

### Do you want to develop this?

- [X] I want to develop this.

matmair:
@inventree/maintainer @inventree/triage @miggland @wolflu05 @SergeoLacruz @Bbillyben @martonmiklos @rkalman @Zontex thoughts? Please 👍 / 👎 if you think it is generally a good or bad idea.

Zontex:
@matmair I think it's necessary and developers like myself will support it. Should be built in, disabled by default and maybe after installation ask one time if to enable or not

Bbillyben:
Hi @matmair, 
This feature would be quite interesting, it's really frustrating to develop a tool and not having any clues of its use. 

I'm EU citizen, and a great part of my job is tied/linked to GPDR and personnal data, so i'm totally biased and quite paranoid about collecting datas.
It's freaking hard to be outside personnal data, and it tends to be harder and harder.
With enough good sources, probalistic matching could lead to good identification results, and the efficience increase with the number of collected datas.

And I'm always sceptical when a server is collecting data from me and stating it's only for internal use.

SergeoLacruz:
N'abbend,
Thanks for asking my opinion. 
As you probably know I am a German boy. So like BbillyBen I have my special opinion on things like that :-)
Anyhow I see the advantage for the developers. I think on huge projects with millions of users you cannot live without it because the usage scenarios are many and different. Very difficult to hunt bugs with little info. I know such systems from my company 
[...]

PR #8503, opened 2024-11-17
Plugins installation improvements

SchrodingersGat:
- Replaces https://github.com/inventree/InvenTree/pull/8502
- Closes https://github.com/inventree/InvenTree/issues/7709

### Summary

This PR improves plugin loading and install to ensure that the plugins are properly synced between different running containers, e.g. when running under docker. Ensures that the required system packages are installed before collecting plugins.

### Installation Process

- Install plugin via the web interface (frontend server)
- Plugin is installed and added to `plugins.txt`
- Activate plugin (via frontend server)
- Server offloads task (to background worker) to copy static files, etc
- Background worker checks `plugins.txt` - sees it has changed - and installs the new plugin
- Background worker then runs offline processing for the newly activated plugin


SchrodingersGat:
@wolflu05 I have refactored this (again) - a much simpler approach now which ensures that the plugins get installed into all running containers as appropriate

wolflu05:
Thanks, that looks promising, I'd like to take a closer look and test it myself, but that can take me some days...

SchrodingersGat:
Let me know when you've had a chance to test it. This is one of the final outstanding Issues for 0.17.0

SchrodingersGat:
@matmair if you have a chance to cast your eye over this I'd appreciate it :)

src/backend/InvenTree/InvenTree/tests.py
```diff
@@ -1184,18 +1184,8 @@ def test_initial_install(self):
...
-
-            # Set dynamic setting to True and rerun to launch install
-            InvenTreeSetting.set_setting('PLUGIN_ON_STARTUP', True, self.user)
-            registry.reload_plugins(full_reload=True)
-
-        # Check that there was another run
-        response = registry.install_plugin_file()
-        self.assertEqual(response, True)
+        registry.reload_plugins(full_reload=True, collect=True)
+        self.assertGreater(len(settings.PLUGIN_FILE_HASH), 0)
```
matmair, line 1188: Q: Does this test ensure that the installation actually runs with this setup?

src/backend/InvenTree/plugin/installer.py
```diff
@@ -63,39 +66,54 @@ def handle_pip_error(error, path: str) -> list:
...
+    """Return the file hash for the plugins file."""
+    import hashlib
+
+    pf = settings.PLUGIN_FILE
+
+    if not pf or not pf.exists():
+        return None
+
+    with pf.open('rb') as f:
+        return hashlib.md5(f.read()).hexdigest()
```
matmair, line 116: Q (non-blocking): Can we use this oportunity to move to a sha hashing function? md5 always comes up in security reviews and toolings.
```suggestion
        return hashlib.sha256(f.read()).hexdigest()
```

matmair, line 116: C: In the future, we could use this cool new(ish) function for efficient file digesting: https://docs.python.org/3/library/hashlib.html#hashlib.file_digest

src/backend/InvenTree/plugin/installer.py
```diff
@@ -108,8 +126,17 @@ def install_plugins_file():
         logger.warning('Plugin file %s does not exist', str(pf))
         return
 
+    cmd = [
+        'install',
+        '--disable-pip-version-check',
+        '--no-cache-dir',
```
matmair, line 132: Q: Are you sure we do not want to use caching?

src/backend/InvenTree/plugin/installer.py
```diff
@@ -203,15 +238,8 @@ def install_plugin(url=None, packagename=None, user=None, version=None):
...
-    # Check if we are running in a virtual environment
-    # For now, just log a warning
-    in_venv = sys.prefix != sys.base_prefix
-
-    if not in_venv:
-        logger.warning('InvenTree is not running in a virtual environment')
-
     # build up the command
-    install_name = ['install', '-U']
+    install_name = ['install', '-U', '--no-cache-dir', '--disable-pip-version-check']
```
matmair, line 242: Q: Same as above regarding caching

src/backend/InvenTree/plugin/plugin.py
```diff
@@ -239,9 +239,10 @@ def file(cls) -> Path:
         """File that contains plugin definition."""
         return Path(inspect.getfile(cls))
 
-    def path(self) -> Path:
+    @classmethod
+    def path(cls) -> Path:
         """Path to plugins base folder."""
-        return self.file().parent
+        return cls.file().parent
```
matmair, line 245: Q: Is this related or just nice to have?

matmair (approved):
Looks good in general; It would really be good to have https://github.com/inventree/InvenTree/issues/4150 available to know how prevalent instances without workers are

matmair:
@SchrodingersGat I might be blind but where exactly is the offloading happening? Maybe I am too tired but I seem to miss that change in the diff

src/backend/InvenTree/plugin/installer.py
SchrodingersGat: No probably not - I have reverted this now

src/backend/InvenTree/plugin/plugin.py
SchrodingersGat: This was needed for another approach I was trying - but in the end it was a nice addition

src/backend/InvenTree/plugin/installer.py
matmair: @SchrodingersGat this is only available in 3.11 - we target 3.9

SchrodingersGat:
> @SchrodingersGat I might be blind but where exactly is the offloading happening? Maybe I am too tired but I seem to miss that change in the diff

When you activate a plugin, the collection of static files is offloaded to the background worker:

image

For this to succeed, the background worker needs to have the same set of plugins installed!

src/backend/InvenTree/plugin/installer.py
SchrodingersGat: Ah, ok good point.

matmair:
Ah now I found the link, the check migrations function is doing the installing

wolflu05:
Sorry, I had no time yet, I was planning to test this on the weekend. But if anything pops up then, I'll just reply here.

PR #8519, opened 2024-11-19
Add empty context menu

SchrodingersGat:
- For tables without "row actions"
- Show an "empty" menu
- For UX consistency across all tables

### DIFF
--- a/docs/docs/develop/react-frontend.md
+++ b/docs/docs/develop/react-frontend.md
@@ -125,7 +125,7 @@ npx playwright show-report path/to/report
 
 ### No Tests Found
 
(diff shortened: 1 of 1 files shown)


wolflu05:
Maybe we can add a small text like _No actions_ or so in the context menu.

matmair (approved):

wolflu05:
> Maybe we can add a small text like _No actions_ or so in the context menu.

@SchrodingersGat thoughts?

SchrodingersGat:
@wolflu05 oops, didn't see your comment before merging. That would be a good addition I think

PR #8523, opened 2024-11-19
Improve plugin panels

SchrodingersGat:
- Do not prepend plugin name unless necessary
- Allows for "cleaner" navigation URLs

e.g. for the "wireviz" plugin:

image


matmair:
@SchrodingersGat does this ensure the order of panels between 2 plugins with the same panel name is stable?

SchrodingersGat:
> @SchrodingersGat does this ensure the order of panels between 2 plugins with the same panel name is stable?

There should be no change to this behaviour, the order of insertion is fixed - assuming that the API always delivers panels in the same "order"

PR #8524, opened 2024-11-19
Relation update

SchrodingersGat:
- Ref: https://github.com/inventree/InvenTree/discussions/8416
- Adds "note" field to PartRelated model
- Implement new table column in PUI


wolflu05:
We might could also overengineer this by adding a second table like relation types which can be selected. That way we can keep this field much clear/filter/… by common relations types. E.g. show all additional equipment relations, …  But that’s just a thought

matmair:
I think using a state would probably be user-friendly as custom states would make it easy to extend the selection without adding new interfaces that need management functionality and discovery

turist79:
> Using a state would probably be user-friendly as custom states would make it easy to extend the selection without adding new interfaces that need management functionality and discovery. That could be added as a separate PR to keep this one in the imminent 0.17.0

Do you mean the same "states" used for stock items or some other feature?

turist79:
> We might could also overengineer this by adding a second table like relation types which can be selected. That way we can keep this field much clear/filter/… by common relations types. E.g. show all additional equipment relations, … But that’s just a thought

Keeping both Note and Relation type as separate table (dictionary) is the best solution here. I have never seen an example when organizing things from the start wasn`t beneficial later. The Relation type thing colud contain: direction flag (A->B, or bi-directional), optional conversion rate. This approach would allow smooth implementation of a lot of usefull things actually.

Of course, just my 2c again :)

matmair:
@turist79 happy to review PRs

turist79:
> @turist79 happy to review PRs

sorry, but what does it mean?

matmair:
> sorry, but what does it mean?

I am happy to review functiong and tested code that you submit for inclusion implemeting your complicated suggestion

PR #8532, opened 2024-11-20
Add PO wildcard default setting

jacobfelknor:
This PR is definitely a work in progress, but I wanted to float this idea because it addresses one of the complaints I get about my InvenTree instance.

At my company, purchase orders may be prepended with a `P` or a `T` depending on the situation. For example, a reference may be `P-12345` or `T-12345`.  As a result, my PO reference pattern is `?-{ref:05d}`. 

However, a majority of POs get the `P` version of the reference, and the create PO form suggests the literal `?` from the reference, like `?-12345`. This means users must replace the `?` with a `P` in a majority of cases, which is annoying. If I change the reference to `P-{ref:05d}`, the `T` version is no longer accepted (which is correct and expected).

I propose adding a setting `PURCHASEORDER_REFERENCE_PATTERN_WILDCARD_DEFAULT` (and similar for other types of references) which suggests in the form a default value for the wildcard `?`, but continues to accept other characters in this position. If the setting is set to `None`, no replacement happens and the form continues to suggest `?`.

As I mentioned, this PR is incomplete in its current state and does not consider tests or other objects that use reference patterns. I also only implemented setting the value in the CUI because I have no familiarity with developing in the PUI yet (but will if this becomes real). This is more to gather feedback on the idea before I commit more time to it,


jacobfelknor:
> Just an idea/question for someone who is more familiar with the reference syntax. I guess if this is a custom implementation, maybe we could add some kind of `?{default:P}-{ref:05d}` syntax @SchrodingersGat ?

That does feel cleaner... especially because it probably means we can do it all in one place instead of a setting per object type

wolflu05:
I mean the syntax is still TBD. I'd really like to hear your ideas here, as I'm not so familiar with the ref syntax, never did something there.

jacobfelknor:
Yes, I will explore it

jacobfelknor:
@wolflu05 @SchrodingersGat 

This ended up being quite elegant in my opinion. Using the reference pattern `{?:P}-{ref:05d}` I'm able to accomplish what I wanted for POs, and this syntax works for other reference patterns as well

I still need to update tests/docs

jacobfelknor:
Ok, added to docs/tests. I wasn't sure exactly where to put the test, but I put it where it made most sense to me. Let me know if there's anything else I should do

SchrodingersGat (approved):

SchrodingersGat:
@jacobfelknor nice work, this is very clean indeed. The tests look appropriate, too.

Thanks for the great contribution!

PR #8534, opened 2024-11-20
Small SAST fixes

matmair:


src/frontend/src/components/importer/ImporterDrawer.tsx
```diff
@@ -81,6 +80,7 @@ export default function ImporterDrawer({
         return 3;
       case importSessionStatus.COMPLETE:
         return 4;
+      default:
```
SchrodingersGat, line 83: The 'default' case should return a value here still, otherwise you are changing functionality.

src/frontend/src/components/plugins/PluginUIFeature.tsx
```diff
@@ -84,7 +84,7 @@ export const getPluginTemplateEditor = (
     useEffect(() => {
       (async () => {
         try {
-          await func({
```
wolflu05, line 87: Isn't the await needed here so that asynchronous errors are catched too? But not sure about it.

src/frontend/src/components/importer/ImporterDrawer.tsx
matmair: Are you sure? `default` is empty in the current code base

src/frontend/src/components/plugins/PluginUIFeature.tsx
matmair: Shouldn't `func` be marked async for that?

wolflu05: This can be asynchronous if 

https://github.com/inventree/InvenTree/blob/master/src/frontend/src/components/plugins/PluginUIFeatureTypes.ts#L1is defined as a promise by the individual feature.

wolflu05: But you’re right, in this case it can’t be async, so its fine removing it. Good catch.

SchrodingersGat:
Thanks @matmair

src/frontend/src/components/plugins/PluginUIFeature.tsx
matmair: TBH all of those were caught by SonarCloud and were marked as middle to high importance so I mainly fixed them to get the quality gate back to good-ish before the release

Issue #6394, opened 2024-02-02
Strange Migration during CI


SchrodingersGat:
<img width="562" alt="image" src="https://github.com/inventree/InvenTree/assets/10080325/771600ff-cd34-4c19-883d-99407129ed89">

Ref: https://github.com/inventree/InvenTree/actions/runs/7754804489/job/21148965178?pr=6390

This migration is odd - where is it originating?

It *may* explain this unsolved issue - https://github.com/inventree/InvenTree/issues/5804

matmair:
This is still appearing - should be investigated. https://github.com/inventree/InvenTree/actions/runs/8156036945/job/22292838209#step:7:797

matmair:
This seems to stil appear - no idea why https://github.com/inventree/InvenTree/actions/runs/8870123046/job/24351576101#step:7:812

SchrodingersGat:
I can reproduce this locally (in devcontainer) by running `invoke test`. An empty migration file is created:

`./src/backend/InvenTree/InvenTree/migrations/0002_auto_20240521_0154.py`

```python
# Generated by Django 4.2.12 on 2024-05-21 01:54

from django.db import migrations


class Migration(migrations.Migration):

    dependencies = [
        ('InvenTree', '0001_initial'),
    ]

    operations = [
    ]

```

matmair:
@SchrodingersGat maybe we should just add that one?

SchrodingersGat:
Ok, finally found it - it's pretty silly.

https://github.com/inventree/InvenTree/blob/6250901c2ff851c0d93aaa27214602dc0040d406/src/backend/InvenTree/InvenTree/test_tasks.py#L148

We are intentionally creating an empty migration to ensure that it gets run by the tasks system

matmair:
Nice. So we can close this @SchrodingersGat?

PR #8537, opened 2024-11-21
Remove dummy migration file

SchrodingersGat:
- Remove dummy migration file created during unit tests
- Ref: https://github.com/inventree/InvenTree/issues/6394


matmair (approved):
lgtm

Issue #6920, opened 2024-04-02
[Docker] Custom docker container for Caddy proxy


SchrodingersGat:
Currently the docker setup requires a [custom caddy config file](https://docs.inventree.org/en/stable/start/docker_install/#required-files) which must be copied into the running [caddy container](https://hub.docker.com/_/caddy).

However we could simplify this by building our own proxy container which we then push to dockerhub / ghcr

This would reduce the number of required files

SchrodingersGat:
While a custom caddyfile container might make complex / custom deployments harder, users can always revert to their own custom container in such cases. In general this should make setup simpler. A good target for 1.0.0

wolflu05:
That will only work with named volumes, but then we need to instruct people to not change that named volume to a bind mount, but people will do that, ...

matmair:
Wouldn’t the server not work anymore at that point?

SchrodingersGat:
https://github.com/inventree/inventree-caddy

wolflu05:
> Wouldn’t the server not work anymore at that point?

I do not really understand what you mean by this.

SchrodingersGat:
Image is now live - https://hub.docker.com/r/inventree/inventree-caddy/tags

I am currently testing , and will push a simplified docker setup if it works well

SchrodingersGat:
https://github.com/inventree/InvenTree/pull/8538

SchrodingersGat:
Closing, see https://github.com/inventree/InvenTree/pull/8538

PR #7724, opened 2024-07-24
WIP: Websockets

matmair:
This PR is heavy WIP!! Do not run this in production

This PR adds ASGI support to the backend setup
- Adds daphne for ASGI support for `runserver`
- Adds uvicorn_worker to the gunicorn setup to enable handling both WSGI and ASGI
- Adds routing logic for ASGI
- Adds a first async component:  PageAttendanceDrawer. This shows other users that are on the same page right now and allows temporary discussions right there

### Tasks
- [ ] Add CI testing for ASGI
- [ ] Extend server-api to reflect if websockets are enabled
- [ ] Add setting to disable websockets
- [ ] Extend proxy settings
- [ ] Replace notification polling with websocket
- [ ] Add count of currently online users to page Attendance
- [ ] Only show page attendance if there other attendens

Closes https://github.com/inventree/InvenTree/issues/6262

PR #8538, opened 2024-11-22
Docker update

SchrodingersGat:
- Simplify docker setup procedure
- Now uses a pre-baked custom Caddy container
- Closes https://github.com/inventree/InvenTree/issues/6920
- Ref: https://github.com/inventree/inventree-caddy
- Ref: https://hub.docker.com/r/inventree/inventree-caddy

### DIFF
--- a/contrib/container/docker-compose.yml
+++ b/contrib/container/docker-compose.yml
@@ -98,7 +98,7 @@ services:
     # https://hub.docker.com/_/caddy
     inventree-proxy:
         container_name: inventree-proxy
-        image: caddy:alpine
+        image: inventree/inventree-caddy:latest
         restart: always
         depends_on:
             - inventree-server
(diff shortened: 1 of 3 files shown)


matmair:
@SchrodingersGat what is not really clear to me is the plan regarding keeping the image up to date with both upstream (caddy and inventree). Will there be a release for each caddy release and each inventree release? How does a user determine if they have the right container if we need to change the settings (for https://github.com/inventree/InvenTree/pull/7724 for example)?

wolflu05:
I personally find this setup much more complex than just downloading an extra Caddyfile, I mean I still have to download the extra .env file, why not also download an Caddyfile. We may could provide something like this to just copy to download all required files by one command:
```bash
curl --remote-name-all https://raw.githubusercontent.com/inventree/inventree/stable/contrib/container/{docker-compose.yml,.env,Caddyfile}
```

SchrodingersGat:
All reasonable points. On balance I don't think this is worth pursuing any further.

SchrodingersGat:
Another idea would be to generate a new `Caddyfile` into the shared volume (externally mounted) if one does not already exist (like we already do for `config.yaml` or `plugins.txt`), Then, the caddy image can make use of that when it launches

PR #8540, opened 2024-11-24
Remove custom context processors

SchrodingersGat:
**DO NOT MERGE UNTIL 0.17.0 IS RELEASED**

- Only merge after 0.17.0 release
- Remove code which injects custom context variables into CUI requests
- Not needed for new API-based PUI code
- Speeds up requests - remove unnecessary DB hits


matmair (approved):

matmair:
See also https://github.com/inventree/InvenTree/pull/8384

matmair:
@SchrodingersGat does this mean we can lower the general DB hit limit?

SchrodingersGat:
> @SchrodingersGat does this mean we can lower the general DB hit limit?

Maybe - i'm not exactly sure how many database hits we are saving here

PR #8543, opened 2024-11-24
Change picture for external stock location

SergeoLacruz:
I added a marking to the pic to better find where the external flag is visible.


SchrodingersGat:
Thanks @SergeoLacruz

PR #8544, opened 2024-11-25
API date filter updates

SchrodingersGat:
This PR adds a number of new filtering options to the API based on date values. Also adds associated table columns and filters to the PUI interface.


matmair (approved):
LGTM; getting filters into explicit functions will make introspection way easier down the line

PR #8547, opened 2024-11-25
Remove regex dependency

SchrodingersGat:
- Ref: https://github.com/inventree/InvenTree/issues/8495
- Remove external `regex` lib for backend
- Can provide similar functionality with `re` module


matmair (approved):

JamieSinn:
Looks like this will ideally also solve this issue: https://github.com/inventree/InvenTree/issues/8494

PR #8546, opened 2024-11-25
after_save_stock_item

skydiablo:
This pull request addresses the issue discussed in #8452

- Closes #8452

Issue #8452, opened 2024-11-08
Email Notification When Stock Level Drops Below Minimum?


skydiablo:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

I'm not receiving email notifications when a part's stock level drops below the set minimum, even though email functionality is working, and notifications are enabled in the user settings.

Could this issue (https://github.com/inventree/InvenTree/issues/7866) be addressing the same question?

### Steps to Reproduce

1. Set a minimum stock level for a part.
1. Ensure email notifications are enabled in the user settings.
1. Allow the stock level of that part to drop below the minimum.

### Expected behaviour

An email notification should be sent when the stock level falls below the minimum threshold.

### Deployment Method

- [ ] Docker
- [ ] Package
- [x] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

# Version Information:
InvenTree-Version: 0.16.8
Django Version: 4.2.15

Database: sqlite3
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-6.8.12-2-pve-x86_64-with-glibc2.31
Installer: None

Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.1.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}, {'name': 'InvenTreeCurrencyExchange', 'slug': 'inventreecurrencyexchange', 'version': '1.0.0'}, {'name': 'InvenTreeLabel', 'slug': 'inventreelabel', 'version': '1.1.0'}, {'name': 'InvenTreeLabelMachine', 'slug': 'inventreelabelmachine', 'version': '1.0.0'}, {'name': 'InvenTreeLabelSheet', 'slug': 'inventreelabelsheet', 'version': '1.0.0'}, {'name': 'DigiKeyPlugin', 'slug': 'digikeyplugin', 'version': '1.0.0'}, {'name': 'LCSCPlugin', 'slug': 'lcscplugin', 'version': '1.0.0'}, {'name': 'MouserPlugin', 'slug': 'mouserplugin', 'version': '1.0.0'}, {'name': 'TMEPlugin', 'slug': 'tmeplugin', 'version': '1.0.0'}, {'name': 'IPNGenerator', 'slug': 'ipngen', 'version': '0.1'}]

[...]

PR #8550, opened 2024-11-25
Add unit test

SchrodingersGat:
- Unit test to ensure event is fired when stock item is created OR updated
- Ref: https://github.com/inventree/InvenTree/pull/8546
- Ref: https://github.com/inventree/InvenTree/issues/8452


SchrodingersGat:
@skydiablo for your reference :)

PR #8572, opened 2024-11-27
[PUI] Fix stock item state rendering

matmair:
Fix rendering of stock states in dialogues - see #7990


SchrodingersGat:
Is rendering of `status_custom_key` sufficient? Does the value defer to the `status` value if no "custom" value is provided?

matmair:
yes, if `status_custom_key` is not set specifically it returns `status`

Issue #6022, opened 2023-12-02
[React] Implement existing Part functions


SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/5212

This issue serves as a list of functions / items / views which need to be reimplemented for the new react interface, specifically for the "part" views

### TODO

- [x] Implement part overview page
- [x] Implement part category page
- [ ] Implement "part detail" panel
- [ ] Implement "select part image" interface
- [ ] Implement "parametric part table"
- [ ] Implement "allocations" panel
- [ ] Implement "pricing" panel
- [ ] Implement "manufacturers" panel
- [ ] Implement "scheduling" panel
- [ ] Implement "stocktake" panel
- [ ] Implement "purchase orders" panel
- [ ] Implement "duplicate part" function
- [ ] Implement "order part" functionality
- [ ] Implement "label printing" functionality

matmair:
@SchrodingersGat do you have a screenshot of what from CUI exactly is missing?

SchrodingersGat:
I think that the only major missing component currently is the "order parts" wizard

PR #8602, opened 2024-11-30
[PUI] Order Parts Wizard

SchrodingersGat:
Implements an "order parts" wizard for PUI, reproducing (and improving upon) the existing CUI wizard.

### References

- Closes https://github.com/inventree/InvenTree/issues/6022

### TODO

- [x] Add "merge" option (like in CUI)
- [x] Generate new purchase orders as required
- [x] Playwright testing
- [x] Implement from "parts" table
- [x] Implement from "part detail" view
- [x] Implement from "sales order" view
- [x] Implement from "build order" view

### Future Work
- [ ] Provide initial "quantity" to order
- [ ] Adjust quantity based on supplier-part pack size
- [ ] Refactor "data import" wizard to use the new components and hooks
- [ ] Refactor "receive items" form to use bottom drawer
- [ ] Refactor other "process" forms to use drawer rather than modal

### Screenshots

image


wolflu05:
Thanks that you have implemented the order parts wizard. Just some things that I found while testing (I know that this is highly WIP):

- when trying to load more SPs, the table column size jumps. Maybe this needs to be fixed:
   
- The button to go to the next step is really hard to find/identify
- When I try to go to the next step it shows me an error:
      

I actually only tried it out yet to see if the following will be available too, but because of the third error from above, I cannot test it really, so I'll just throw that in for consideration:
Will there be an option/step to specify if you want to merge a line item or create a new line item if it already exists in the PO? With PUI we can even only show that if there is really a line item already existing.

PR #8611, opened 2024-12-01
[PUI] Make dashboard header look less like a widget

matmair:
This PR changed the header of the dashboard from
grafik

to
grafik

and moves the editing indicator
grafik


SchrodingersGat (approved):

SchrodingersGat:
Nice addition @matmair

PR #8616, opened 2024-12-02
Start documentation chapter for plugin tests

SergeoLacruz:
Lets start from here. I will continue as soon as I have something working...


SergeoLacruz:
Format looks ugly...
Will fix tomorrow
Good night :-)

SergeoLacruz:
Looks better now.

SergeoLacruz:
Thanks. I added your comments. I think we can publish it for the moment and add details as soon as we have them.

docs/docs/extend/plugins/test.md
```diff
@@ -0,0 +1,44 @@
+---
+Title: Unit Tests
+---
+
+## Unit Tests
+For complicated plugins it makes sense to add unit tests to your code. InvenTree
```
wolflu05, line 6: Maybe add something like "to ensure that plugins work correctly and are compatible with future versions too. You can run these tests as part of your ci against the current stable and latest tag to get notified when something breaks before it gets released as part of stable"???

wolflu05 (approved):

SchrodingersGat:
In the "do some work here" I think it would be a good idea to have some examples of what testing might actually be performed

SergeoLacruz:
Definitely...
But that's the next chapter :-)

PR #8623, opened 2024-12-02
[PUI] fix Table Context Menu Links

### LINKED ISSUE #8553: [PUI] Table Context Menu with Links
If there is a URL (link) in the table, it does not play nicely with the "context menu".

Image

This prevents the "right-click / save-as" option!

matmair:
These changes make it possible to disable context menus on right-clicks per column 
Closes #8553


SchrodingersGat:
@matmair seems like a reasonable fix :)

PR #8638, opened 2024-12-08
feat(backend): add reference sources model

matmair:
This PR adds a way to add arbitrary or restricted (text) references to any model.
Together with a minimal UI implementation, this enables linking external sources/IDs and internal models. This is a starting point that will allow the development of plugins that expose stable linked data import/exports or integrate more transparently into external systems.


SchrodingersGat:
@matmair are you happy for me to push this to the next release target?

matmair:
Let's move it, this will take a bit.

matmair:
Getting closer to the finish line: 

- [ ] missing some UI components
- [ ] docs
- [ ] an experimental flag for now

PR #8646, opened 2024-12-09
Refactor: Move settings definitions (and most validators) out of models.py

matmair:
The common models.py file is getting pretty huge and it is hard to easily reference where stuff belongs.
This PR refactors all the definitions and validators into the `common.setting` subdirectory - the core settings model logic is kept in models.py but all the static definitions are moved out.

I will try to add tests for all missing validators.


SchrodingersGat:
@matmair nice, this is a good refactor of what was a very large and complicated file

PR #8658, opened 2024-12-12
[Refactor] Barcode scanning

SchrodingersGat:
Refactoring implementation of custom supplier barcode scanning:

### Headline Items

- Code refactoring / simplification
- Allow greater introspection of data from generic "scan" endpoint
- Improved error handling for "check in items" handler
- Smoother integration with the app
- Allow client to explicitly choose *not* to auto-receive, but allow confirmation of extracted information

Ref: https://github.com/inventree/inventree-app/pull/577

One of the major improvements here is that a supplier barcode plugin will return partial data on a generic "scan" request. Previously it would *only* return information responding to a supplier part.

However, suppliers will send packing lists with other information (purchase order number is a good example). So now you can scan that code in the app, the supplier barcode will recognize and return *any* field which is provided, and allows the app to navigate to the purchase order detail page.


ius:
This looks pretty useful!

It (nearly) covers a use case I've been wanting to suggest: returning StockItems for 'minimized' EIGP 114.2018 barcodes containing `{supplier_order_number, supplier_part_no[, qty]}`. Would enable printing tiny datamatrix barcodes which resolve to a _StockItem_ without relying on _StockItem_ ids.

PR #8680, opened 2024-12-16
Docker improve build times

matmair:
Improve build times massively by using https://depot.dev - which works like magic for our arm64 builds. We might even be able to re-add raspy images.


wolflu05:
Amazing improvements...

>re-add raspy images

I do not think that is necessary to also target armv7l since every "modern" raspberry pi supports the 64bit raspberry pi OS which is aarch64 and therefore can run the arm64 images.

image

matmair:
Ok so we can skip that - thanks for the response

PR #8684, opened 2024-12-17
Remove "crispy forms" integration

SchrodingersGat:
- No longer doing any back-end form rendering


SchrodingersGat:
@matmair can we remove `./src/backend/InvenTree/InvenTree/forms.py` entirely now?

matmair (approved):

matmair:
@SchrodingersGat we have a bunch of auth logic in there, I will have to take a good look

SchrodingersGat:
@matmair ok I'll hold off on merging this one. It would be great if we can cleanly remove that whole file

SchrodingersGat:
Yeah, let's move them to a different file. Do you want to do that in a separate PR after this one?

matmair:
@SchrodingersGat I would appreciate it, not really a good feeling pushing to another persons branch

src/backend/InvenTree/InvenTree/settings.py
```diff
@@ -1336,11 +1331,12 @@
     'signup': 'InvenTree.forms.CustomSignupForm',
     'add_email': 'allauth.account.forms.AddEmailForm',
     'change_password': 'allauth.account.forms.ChangePasswordForm',
-    'set_password': 'allauth.account.forms.SetPasswordForm',
+    # 'set_password': 'allauth.account.forms.SetPasswordForm',
```
matmair, line 1334: Q (blocking): why is this changed?

matmair (changes requested):

matmair:
I have created a PR for the moving out, to be merged after this

src/backend/InvenTree/InvenTree/settings.py
SchrodingersGat: Right, I commented this out to get the code to launch - can we remove the custom 'set_password' form in `settings.py`?

matmair: I do not think so but I would have to try to be certain

SchrodingersGat:
@matmair fixed that line, can you please approve so that I can merge?

matmair (approved):

Issue #8692, opened 2024-12-17
[PUI] Use of external (USB) barcode scanner not possible in PUI


saschaludwig:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

Trying to use a external (USB) Barcodescanner will always result in "no results"

### Steps to Reproduce

Scan barcode inside a search field or open "scan qr code" modal doesn't find anything.
How do I scan codes with my barcode scanner within the new PUI? 

![Image](https://github.com/user-attachments/assets/208032d4-3f13-485e-9160-488a6ba04067)
![Image](https://github.com/user-attachments/assets/39b5649c-e719-48ea-8457-21ba410969f5)
![Image](https://github.com/user-attachments/assets/1af16605-6553-4c07-828b-ccdc107b4a73)

Although the demo site has another dialog, searching there also doesn't work:
![Image](https://github.com/user-attachments/assets/cd020b6f-cac5-4f9a-a38e-8ffc6f1cac68)

### Expected behaviour

Scan any qr code in any search field to find/view the corresponding item/location.

### Deployment Method

- [x] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

InvenTree-Version: 0.16.9
Django Version: 4.2.15
Commit Hash: e0756a7
Commit Date: 2024-11-30
Commit Branch: null
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.2.16-20-pve-x86_64-with
Installer: DOC
[...]

PR #8721, opened 2024-12-19
[PUI] Barcode scan

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/8692
- Fixes barcode "quick scan" dialog
- Implements using existing forms framework
- Error management

image

Notes: I have tested this with a USB barcode reader, and it works pretty well

This is not intended as a full featureset for this quick-scan dialog. Rather it is intended to get the scan dialog working and fixed for 0.17.x branch


SchrodingersGat:
Closing this, as I realise that this approach does not consider the "webcam" support at all.

PR #8734, opened 2024-12-21
Add separate dialog for 'Ship Order' button

1337joe:
When trying to use the PUI, I found it rather confusing that clicking the "Ship Order" button to move the order to the `SHIPPED` state opened the "Complete Order" dialog. The "Complete Order" dialog performed as I expected "Ship Order" to do, based on the CUI, but all the text indicated the wrong action was taking place.

I created a new dialog with the "Complete Order" functionality but the text from the CUI version of the "Ship Order" dialog.


SchrodingersGat:
@1337joe thanks, this is a nice addition :)

PR #8685, opened 2024-12-17
Remove django-import-export

SchrodingersGat:
This can now be replaced entirely by the new import / export framework which uses the DRF API.

### TODO 

- [x] Remove dead code
- [x] Update documentation for data import
- [x] Update documentation for data export
- [ ] ~~Reimplement "bom export" in new framework~~ - https://github.com/inventree/InvenTree/issues/8686

Issue #8686, opened 2024-12-17
[1.0.0] Reimplement old BOM exporter


SchrodingersGat:
"New" BOM export tool only exports flattened BOM with minimal information.

For 1.0.0, this should be reimplemented to support the "old" export style which allows cascading exporting and inclusion of additional fields.

This could be implemented via a plugin, but we would need to allow the exporter interface to hook into the plugin ecosystem.

### TODO

- [x] Allow export of "multi level" BOM data
- [ ] Include "pricing" information in BOM export
- [ ] Include "parameter" information in BOM export
- [ ] Include "stock" information in BOM export
- [ ] Include "supplier" information in BOM export
- [ ] Include "manufacturer" information in BOM export
- [x] Include "substitute" information in BOM export

### References

- https://github.com/inventree/InvenTree/pull/8685
- https://github.com/inventree/InvenTree/pull/9096

### Note

As part of this PR, remove the entirety of `./src/backend/InvenTree/part/bom.py`

matmair:
@SchrodingersGat I think plugins should be a second step, that introduces features that were not there before.

PR #8756, opened 2024-12-24
Cleanup

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/pull/8685#issuecomment-2560458321

Followup to https://github.com/inventree/InvenTree/pull/8685


matmair:
@SchrodingersGat this will require an API version bump as it removes a number of endpoints; maybe it makes more sense to include these in #8686

SchrodingersGat:
@matmair good point, I will roll the BOM exporter into this one.

I just have to devise a clean way to have a "custom" exporter within our new framework

SchrodingersGat:
@matmair @wolflu05 I have an idea here which requires some further fleshing out - feedback welcomed :)

### Current Situation

The new "data export" framework basically takes our existing DRF serializers, and pushes the data into a formatted file for download. It is very efficient, low code overhead, and integrates seamlessly with the PUI data tables. The intent here is that the table data (as presented and filtered) is basically "what you get" when you download the data via the table interface.

### Legacy BOM Exporter

The legacy BOM exporter contained a lot of extra information. Supplier parts, alternative part options, pricing data, stock information, etc. That's all fine, but it doesn't fit nicely within the concept of a simple "press this button to download the table data" approach.

### Custom Report

I see the "BOM exporter" (as it existed in CUI) working better as a custom report generator. There are custom operations in the background which do not take place in the regular "export serialized data" framework.

So, what I would like to propose is that we extend the "report generation" framework to allow arbitrary file generation. While currently it is limited to templated PDF files, there's no reason we can't have a "BOM export" plugin which generates an excel or .csv file for download.

However this would require a lot of ripping up / refactoring of existing code. Also, how do we handle custom reports that do not make use of the report "template"?

Before I go and start making changes, I'd love some feedback from you guys on how you think it should work - or potentially other ideas / approaches.

matmair:
I think the cleanest way would be to add a custom serializer that handles the BOM export and its options. We could add the option to use plugins to add serializers for custom exports.
Using the report framework does not seem intuitive - it is really oriented around templates right now and rewriting it would either break APIs for no apparent benefit or make testing really hard. IMO this most makes sense as a new tool and endpoint, not in the report framework.

SchrodingersGat:
Thanks for the feedback, that sounds pretty reasonable

SchrodingersGat:
Replaced by https://github.com/inventree/InvenTree/pull/9079

PR #8759, opened 2024-12-24
Fix missing permission for push schema action

matmair:
This PR fixes an error introduced by me when restricting the permissions of all flows. In this specific case we need to keep the credentials persistent.


matmair:
@SchrodingersGat this would be important to merge as this is stopping schema changes form being committed - see https://github.com/inventree/InvenTree/actions/runs/12475075825/job/34819238252#step:5:48

matmair:
Fixed the wrong schema in inventree/schema so QC runs through correctly again

PR #8779, opened 2024-12-27
feat(PUI): Make header tabs links to simpilfy new tab behaviour

matmair:
This PR enables opening header tab items per right-click to new tabs.


matmair:
@sur5r 👀

SchrodingersGat:
@matmair this is a nice addition. It would make sense to add it for the "Panel" nav tab links too?

matmair:
Propaply; should this be included in this patch?

SchrodingersGat:
@matmair yes I think so, should be only very minimal diff

matmair:
@SchrodingersGat added

src/frontend/src/components/buttons/AdminButton.tsx
```diff
@@ -66,7 +66,7 @@ export default function AdminButton(props: Readonly<AdminButtonProps>) {
         `${server.server.django_admin}${modelDef.admin_url}${props.id}/`
       );
 
-      if (event?.ctrlKey || event?.shiftKey) {
+      if (event?.ctrlKey || event?.shiftKey || event?.metaKey) {
```
wolflu05, line 69: As this is used over and over again, maybe it makes sense to extract this into a custom function?

SchrodingersGat:
@matmair I have noticed something here - the target link is different depending on how the click event is interpreted:

### Starting Point

From this starting point, I will click the "Manufacturing" tab:

image

### Left Mouse Button

Navigates to "Manufacturing" page in same tab

image

### Left Mouse Button + Control Key

Opens "Manufacturing" page in new tab

image

### Middle Mouse Button

Opens "Manufacturing" page (**via demo site**) in new tab:

image

Note that the URL in the address bar is different in this case!

The same behaviour hapens with the side panels too...

src/frontend/src/components/buttons/AdminButton.tsx
matmair: As those are seperate functions/interactions I would keep them seperate for now

matmair:
@SchrodingersGat I am not familiar with middle button. Which control key is this?

SchrodingersGat:
> I am not familiar with middle button. Which control key is this?

No control key, just click on the link by pressing the scroll wheel on the mouse. Typically this just opens a link in a new tab (similar to ctrl+click).

However, it bypasses the browser onClick event system somehow so we can't catch it like a normal click. Somehow, this type of click uses a different baseUrl when generating the link target

matmair:
Not really sure what to do about the middle click, I can not find where it is getting it's link in that case and why it is diefferent

sur5r:
Middle click doesn't work at all for me. It's not surprising given the main menu items are no links (a href) but buttons. I guess this is just a clash of worlds. This is no longer web browsing, so we can't expect it to behave as such.

matmair:
that is the thing - there should be/is an a element rendered - so it should work the same as "normal" html

SchrodingersGat:
@matmair it looks like most of the effects of this PR have been reverted, not sure if that is intentional? But the `<a>` elements are no longer in play

matmair:
I have re-introduced the rendering as <a/> elements; seems to work but I do not have a middle click for debug
would appreciate a review @sur5r  @SchrodingersGat

sur5r:
The links are there, but one doesn't end up on the correct view/subpage. It neither works with middle click nor with "copy link" and pasting the link into a new tab.

Example: Open a part and use the link from e.g. "Part pricing" to open an new tab. It opens the Part, but on "Part Details".

It seems to URLs are generated incorrectly, i.e. they point to `/part/301/details/pricing` instead of `/part/301/pricing`.

matmair:
Seems to work on my machine - are we talking about the same thing?

https://github.com/user-attachments/assets/cf33177d-bc33-44c2-b6df-1776a9d07fd7

sur5r:
No, I'm talking about these links:

image

matmair:
@sur5r it now should also work on the panels - can you please confirm?

SchrodingersGat:
Looks good, middle mouse button is working now! Thanks @matmair

PR #8790, opened 2024-12-30
[Setup] Support X-Forwarded-Proto header

SchrodingersGat:
Adds support for X-Forwarded-Proto header, which may be required for setups where InvenTree is run behind a secure proxy for example.

Additionally, this PR improves the documenation around setup / configuration of these options.


numanair:
This might not be relevant, but I (previously) noticed that I had to add `trusted_proxies` to the Caddyfile to allow displaying the correct client address in the settings interface. This is with another proxy in front of Caddy, but I thought I'd mention it here as it sounds like this is helping with that kind of setup.

SchrodingersGat:
> to allow displaying the correct client address in the settings interface

@numanair can you please expand on this, maybe a screenshot of what you are talking about?

numanair:
> > to allow displaying the correct client address in the settings interface
> 
> @numanair can you please expand on this, maybe a screenshot of what you are talking about?

So my setup is a Traefik proxy in front of the Caddy proxy/server. The front proxy handles SSL termination. Both are on the same Docker instance. I was using the example Caddy file. **In "Active Sessions" in the Inventree settings (CUI) it was showing the IP of the Traefik Docker container instead of real client IPs.** 

Caddy docs state:  
> If Caddy is not the first server being connected to by your clients (for example when a CDN is in front of Caddy), you may configure trusted_proxies with a list of IP ranges (CIDRs) from which incoming requests are trusted to have sent good values for these headers.

So my Caddyfile reverse_proxy section now looks like:

```
<snip>

	reverse_proxy {$INVENTREE_SERVER:"http://inventree-server:8000"} {
		trusted_proxies <fronted-proxy-ip>
	}
}
```

SchrodingersGat:
@numanair thanks, that's a good tip. I'll add some documentation around that in a separate PR

SchrodingersGat:
@numanair https://github.com/inventree/InvenTree/pull/8798 FYI

Issue #8742, opened 2024-12-23
Error printing label  cannot import name 'ImageDraw' from 'PIL' (/root/.local/lib/python3.11/site-packages/PIL/__init__.py)


Agilitytime:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

Hi,

its not possible to print a label.
After clicking on the Print button, the following error code plobs up.

Error printing label

cannot import name 'ImageDraw' from 'PIL' (/root/.local/lib/python3.11/site-packages/PIL/__init__.py)

it dusent matter, whitch Print method is used.

### Steps to Reproduce

1 go to Print Actions
2 Print Label

### Expected behaviour

Print a Label or Generate PDF

### Deployment Method

- [x] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

New Deployed Version, now Update etc.

77ae06b03ba5   inventree/inventree:stable   "/bin/ash ./init.sh …"   21 minutes ago   Up 21 minutes   8000/tcp                                                                                      inventree-worker
16124c1834c1   caddy:alpine                 "caddy run --config …"   21 minutes ago   Up 21 minutes   0.0.0.0:80->80/tcp, :::80->80/tcp, 0.0.0.0:443->443/tcp, :::443->443/tcp, 443/udp, 2019/tcp   inventree-proxy
db55c3ab38f9   inventree/inventree:stable   "/bin/ash ./init.sh …"   21 minutes ago   Up 21 minutes   8000/tcp                                                                                      inventree-server
6bbe316798dc   postgres:13                  "docker-entrypoint.s…"   21 minutes ago   Up 21 minutes   5432/tcp                                                                                      inventree-db
478056b8e58f   redis:7.0                    "docker-entrypoint.s…"   21 minutes ago   Up 21 minutes   6379/tcp

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

### Relevant log output

```shell

```

SchrodingersGat:
@Agilitytime please include the full error log - which should be available [via the admin interface](https://docs.inventree.org/en/stable/settings/logs/)

If you are not seeing any error messages there you may need to enable the [Log Report Errors](https://docs.inventree.org/en/stable/settings/global/#reporting) setting

gab696:
Hi,
I have the same problem, Did you found a solution ?

[...]

PR #8793, opened 2024-12-30
Fix package install in envs using python lower than 3.12 on debian like OSs

matmair:
This fixes issues where uv installs the wrong version of a package. It might be wise to use the uv lockfile in the future to ensure this does not happen again, that requires more work though

### DIFF
--- a/contrib/packager.io/functions.sh
+++ b/contrib/packager.io/functions.sh
@@ -292,6 +292,7 @@ function stop_inventree() {
 }
 
 function update_or_install() {
+  set -e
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
Is this in relation to any particular outstanding issues?

matmair:
as listed in the commits https://github.com/inventree/InvenTree/issues/8789 https://github.com/inventree/InvenTree/issues/8742 https://github.com/inventree/InvenTree/issues/8495 https://github.com/inventree/InvenTree/issues/8494

PR #8811, opened 2024-12-31
[Documentation] Remove package credits

SchrodingersGat:
Removes hard-coded credits.md from the documentation

- Already included in the UI and API
- Also includes some playwright testing fixes as uncovered


matmair:
I see a potential security/spam issue here. The urls in a python package can be set to whatever. Link them on our docs site might lead users to compromised or outdated adresses.

It should also discussed what value it brings to list all dependencies and sub-depndencies. The brevious behaviur was to only list the packages we directly used. Listing a few hundred frontend dependencies (that we probably do not even use as a lt are shaken out during build) servers not much value in my eyes.

SchrodingersGat:
@matmair I'd be happy (and would rather) remove this page from the docs entirely. We already have the up-to-date list of dependencies in the user interface and API. 

This PR was because I noticed that the credits page is woefully out of date.  But if you are happy for it to be removed, I'll do that.

matmair:
Than let's do that

matmair (approved):

SchrodingersGat:
@matmair confused why the Platform UI checks keep failing for this PR - nothing outside the "docs" directory has changed here...

matmair:
@SchrodingersGat maybe the change in the quick login? Can we try revert that?

Issue #8829, opened 2025-01-05
Docker stack is not publishing ports


Flintlocke89:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [x] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

Hello, relatively new to Docker and I can't seem to figure this out.
As far as I can tell I have followed the instructions (almost) exactly but when the compose spins up no ports are published, hence I have nothing to connect to.

### Steps to Reproduce

Followed [these instructions.](https://docs.inventree.org/en/latest/start/docker_install/#docker-installation)
Started with default docker-compose.yml, Caddyfile and .env

Commented out port 443 in the supplied docker-compose.yml as it conflicts with my existing reverse proxy
Added admin account details to .env file
Added `INVENTREE_WEB_PORT=7800` to .env file
Changed database name and passwords in .env file for security
Ran `docker compose run --rm inventree-server invoke update`
Ran `docker compose up -d`

### Relevant log output

Server:
```bash
Loading config file : /home/inventree/data/config.yaml

Python version 3.11.9 - /usr/local/bin/python

/root/.local/lib/python3.11/site-packages/allauth/exceptions.py:9: UserWarning: allauth.exceptions is deprecated, use allauth.core.exceptions

  warnings.warn("allauth.exceptions is deprecated, use allauth.core.exceptions")
```

Proxy:
```bash
{"level":"info","ts":1736073428.3572702,"msg":"using config from file","file":"/etc/caddy/Caddyfile"}

Error: adapting config using caddyfile: parsing caddyfile tokens for 'reverse_proxy': unrecognized subdirective {, at /etc/caddy/Caddyfile:74

{"level":"info","ts":1736073488.9328437,"msg":"using config from file","file":"/etc/caddy/Caddyfile"}

Error: adapting config using caddyfile: parsing caddyfile tokens for 'reverse_proxy': unrecognized subdirective {, at /etc/caddy/Caddyfile:74

{"level":"info","ts":1736073549.53401,"msg":"using config from file","file":"/etc/caddy/Caddyfile"}

Error: adapting config using caddyfile: parsing caddyfile tokens for 'reverse_proxy': unrecognized subdirective {, at /etc/caddy/Caddyfile:74
```

database: (although likely unrelated to the problem, has a lot of errors similar to the ones in this log snippet
[...]

PR #8830, opened 2025-01-05
Fix for buggy Caddyfile

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/8829
- Caddyfile had been committed without full testing


SchrodingersGat:
Merging as there are currently issues on master with the `structlog` import - but these changres are required for the `0.17.2` release

Issue #8827, opened 2025-01-05
Invoke Path Issues


SchrodingersGat:
### Overview

A "bug" has been introduced since the merging of https://github.com/inventree/InvenTree/pull/8747 - the `invoke` commands no longer work as expected.

### Description

When calling `invoke` (used for most system commands), the loaded virtual environment is not used as expected.

This occurs when `invoke` is installed as a system-level package which is available on `PATH` with higher priority than the version installed in the python virtual environment.

The outcome is that calling `invoke <command>` *does not use* the packages installed in the InvenTree virtual environment.

### Example - System Invoke

`invoke version` - invoke installed at `/usr/bin/invoke`

```
Warning: Dulwich module not found, git information will not be available.

InvenTree - inventree.org
The Open-Source Inventory Management System


Python paths:
Executable  /usr/bin/python3
Environment /usr

Installation paths:
Base        /home/inventree
Config      /home/inventree/dev/config.yaml
Media       /home/inventree/dev/media
Static      /home/inventree/dev/static

Versions:
Python      3.11.8
Django      4.2.16
InvenTree   0.18.0 dev
API         297
Node        v18.20.1
Yarn        1.22.19

Commit hash: None
Commit date: None
```

### Example - venv invoke

`python -m invoke version` - called from InvenTree venv

```
InvenTree - inventree.org
The Open-Source Inventory Management System


Python paths:
Executable  /home/inventree/dev/venv/bin/python
Environment /home/inventree/dev/venv

Installation paths:
Base        /home/inventree
Config      /home/inventree/dev/config.yaml
Media       /home/inventree/dev/media
Static      /home/inventree/dev/static

Versions:
Python      3.11.8
Django      4.2.17
InvenTree   0.18.0 dev
API         297
Node        v18.20.1
Yarn        1.22.19

Commit hash: ea9e3fb
Commit date: 2025-01-05
```

_Originally posted by @SchrodingersGat in https://github.com/inventree/InvenTree/issues/8747#issuecomment-2571472624_

SchrodingersGat:
This may well cause issues with the installer setup too

SchrodingersGat:
Not currently. I have tried a bunch of approaches but I cannot work out a way to cleanly determine if the "right" invoke binary is being used.

PR #8747, opened 2024-12-23
Refactor to use structlog

matmair:
Directly calling the logging module does bypass structlog, this fixes that

PR #8835, opened 2025-01-05
Docker fix

SchrodingersGat:
- Attemped fix for https://github.com/inventree/InvenTree/issues/8827
- Fixes "bug" introduced in https://github.com/inventree/InvenTree/pull/8747
- Ensure that `invoke` tool is correctly installed into venv


matmair (approved):

matmair:
@SchrodingersGat this looks good but as a real solution we should probably detect if we are running in the wrong env and call that out

PR #8836, opened 2025-01-05
fix: remove structlog from invoke internal paths

matmair:
This is not really a fix but a path that should re-enable running most tests and invoke commands. https://github.com/inventree/InvenTree/pull/8835 is still required.

Follow up to https://github.com/inventree/InvenTree/pull/8747


matmair:
@SchrodingersGat this seems to have fixed the issue

Issue #8594, opened 2024-11-29
[FR] Distribute front-end libraries for plugins


SchrodingersGat:
Distribute our frontend component library for use in custom plugins (via npm).

matmair:
From the non-exsistant response from NPM/GitHub I take it that we will need to use a name other than "inventree" for the namespace?

SchrodingersGat:
https://www.npmjs.com/org/inventreedb

wolflu05:
Just a collection of a few useful links that might help us here: 
- https://dev.to/receter/how-to-create-a-react-component-library-using-vites-library-mode-4lma
- https://vite.dev/guide/build#library-mode

SchrodingersGat:
@wolflu05 would you have any time over the next few weeks to assist with this? I'd really like to get this rolling, a number of plugin ideas which are stuck until we have a shared library option

wolflu05:
I’d be happy to help, but I’m a bit short on time currently. If you create a PR, I can jump in and assist with specific questions as I’m able. I’m definitely eager to get this sorted soon, since my plugin migrations depend on it too.

SchrodingersGat:
@wolflu05 that'd be great. I'll certainly tag you for a review!

Also check out https://github.com/inventree/InvenTree/pull/8851 - this is a precursor PR which will be required to allow us to distribute tightly-integrated core components (e.g. tables / forms)

matmair:
I think we will not get the inventree namespace - the tickets is getting 0 response from github https://support.github.com/ticket/personal/0/3005569

SchrodingersGat:
I have not received any response either. Let's proceed with `inventreedb` then!

SchrodingersGat:
@wolflu05 @matmair - WIP is now up here - https://github.com/inventree/InvenTree/pull/9323

Caveat: I have no idea what I am doing - so, input very very welcome :)

PR #8851, opened 2025-01-07
[UI] API Context

SchrodingersGat:
Adds global API context, allowing components (such as tables, forms) to be "reused" by plugins.

- Required for https://github.com/inventree/InvenTree/issues/8594
- Removes requirement for hard-coded API object
- Allows these components to be distributed as a "library"

### DIFF
--- a/src/backend/InvenTree/part/serializers.py
+++ b/src/backend/InvenTree/part/serializers.py
@@ -682,7 +682,7 @@ class PartSerializer(
     Used when displaying all details of a single component.
     """
 
-    import_exclude_fields = ['duplicate']
+    import_exclude_fields = ['duplicate', 'tags']
 
     class Meta:
         """Metaclass defining serializer fields."""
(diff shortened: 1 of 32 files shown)


src/backend/InvenTree/part/serializers.py
```diff
@@ -682,7 +682,7 @@ class PartSerializer(
     Used when displaying all details of a single component.
     """
 
-    import_exclude_fields = ['duplicate']
+    import_exclude_fields = ['duplicate', 'tags']
```
matmair, line 685: Q: Is this meant to be in this PR?

matmair (approved):
LGTM; does this work with the different login views?

SchrodingersGat:
> LGTM; does this work with the different login views?

I have intentionally *not* changed the login views. These do not need to be exposed to the plugin interface (not yet at least)

wolflu05 (commented):
LGTM

Issue #8852, opened 2025-01-07
[Reporting] Support generation of DataMatrix codes


SchrodingersGat:
Hi,
would it be possible to print Datamatrix codes ?

_Originally posted by @gab696 in https://github.com/inventree/InvenTree/discussions/8819_

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/inventree) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/inventree/InvenTree/issues/8852">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/inventree/InvenTree/issues/8852/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/inventree/InvenTree/issues/8852/pledge.svg">
</picture>
</a>

SchrodingersGat:
@gab696 for your reference

SchrodingersGat:
Fix will be in the next release cycle (1.0.0 at this stage)

PR #8853, opened 2025-01-07
Datamatrix

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/8852
- Adds support for generation of DataMatrix codes


wolflu05:
Does the current frontend, mobile app supports scanning data matrix codes?

SchrodingersGat:
> Does the current frontend, mobile app supports scanning data matrix codes?

Yes as long as the scanning tech supports it. The app 100% does I'll have to check the webcam scanner.

src/backend/InvenTree/report/templatetags/barcode.py
```diff
@@ -45,26 +53,31 @@ def clean_barcode(data):
...
 
     Returns:
-        base64 encoded image data
+        image (str): base64 encoded image data
 
     """
+    data = str(data).strip()
+
+    if not data:
+        return error_barcode()
```
matmair, line 79: Q: Would it not be better to raise at that point?

matmair (approved):
Very clean

src/backend/InvenTree/report/templatetags/barcode.py
SchrodingersGat: Good point, this is now implemented

Issue #6176, opened 2024-01-08
[FR] Stock Expiry when Receiving Line Item in PO


jacobfelknor:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

When receiving line items - thus generating stock - from the PO view, you are unable to set an expiry on the stock item being generated. Instead, we must then go to the received stock item created after the receive event and edit it to add the expiry. This isn't _that_ annoying, but there are options to add a batch code or location from this view.

### Suggested solution

Add another option for expire date for a stock item being received from a PO, similar to how you can add a batch code. In would also be helpful to add any stock item notes from this view

### Describe alternatives you've considered

As mentioned, we are just going to the detail view after the line item receive and adding the expire date after the fact. This is fine, but an extra step that could be forgotten or done incorrectly (accidently navigate to wrong stock item)

### Do you want to develop this?

- [ ] I want to develop this.

jacobfelknor:
If you have other suggestions for a better way of accomplishing this, let me know. I am still learning the system

SchrodingersGat:
I can see the usefulness of this. Would you be willing to look into contributing this feature?

jacobfelknor:
I could definitely take a look. It may be some time before I can get to it though, which is fine. I'm not particularly in a rush.

jacobfelknor:
For this to be considered complete, we probably also need to update inventree-python's `PurchaseOrderLineItem.receive` function

https://github.com/inventree/inventree-python/blob/master/inventree/purchase_order.py#L207-L220

SchrodingersGat:
Ok, assigned, feel free to ping me when you get a chance.

Note that we are currently pushing to refactor the entire front-end into react. So, any front-end / UI / UX features should be implemented *only* in the new React interface

jsherman256:
I would very much like this feature

SchrodingersGat:
@jacobfelknor @jsherman256 would either of you be willing to implement this feature, or alternatively provide funding towards its development?

jacobfelknor:
I will revisit

jacobfelknor:
I've just opened a PR #8679 

I still agree with my earlier statement:

> For this to be considered complete, we probably also need to update inventree-python's `PurchaseOrderLineItem.receive` function
> 
> https://github.com/inventree/inventree-python/blob/master/inventree/purchase_order.py#L207-L220

I can also look at that

jacobfelknor:
[...]

PR #8867, opened 2025-01-09
Add Expiry Date on Receive Line Item

jacobfelknor:
Addresses #6176

Add option in PUI receive line item form for adding an expiry. 
Update test `PurchaseOrderReceive.test_valid` to include an `expiry_date`

### DIFF
--- a/src/backend/InvenTree/order/models.py
+++ b/src/backend/InvenTree/order/models.py
@@ -800,8 +800,8 @@ class PurchaseOrder(TotalPriceMixin, Order):
         # Extract optional batch code for the new stock item
         batch_code = kwargs.get('batch_code', '')
 
-        # Extract optional expiry data for the new stock item
-        expiry_date = kwargs.get('expiry_date', '')
+        # Extract optional expiry date for the new stock item
+        expiry_date = kwargs.get('expiry_date')
 
         # Extract optional list of serial numbers
         serials = kwargs.get('serials')
--- a/src/backend/InvenTree/order/models.py
+++ b/src/backend/InvenTree/order/models.py
@@ -800,6 +800,9 @@ class PurchaseOrder(TotalPriceMixin, Order):
         # Extract optional batch code for the new stock item
         batch_code = kwargs.get('batch_code', '')
 
+        # Extract optional expiry data for the new stock item
+        expiry_date = kwargs.get('expiry_date', '')
+
         # Extract optional list of serial numbers
         serials = kwargs.get('serials')
 
@@ -863,6 +866,7 @@ class PurchaseOrder(TotalPriceMixin, Order):
                     purchase_order=self,
                     status=status,
                     batch=batch_code,
+                    expiry_date=expiry_date,
                     packaging=packaging,
                     serial=sn,
                     purchase_price=unit_purchase_price,
--- a/src/backend/InvenTree/order/serializers.py
+++ b/src/backend/InvenTree/order/serializers.py
@@ -717,6 +717,7 @@ class PurchaseOrderLineItemReceiveSerializer(serializers.Serializer):
             'quantity',
             'status',
             'batch_code',
+            'expiry_date',
             'serial_numbers',
             'packaging',
             'note',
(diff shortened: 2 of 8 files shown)


jacobfelknor:
weird - my pre-commit checks passed locally, but fail in actions. Any suggestions?

matmair:
Thank you for implementing this @jacobfelknor!
Loogs good in general. Due to the change in the API response you will need to bump the api version in `api_version.py` - and provide a short description in the text block (including a link to this PR). The changes in the API schema are visible in the CI - https://github.com/jacobfelknor/InvenTree/actions/runs/12701501246/job/35406119149#step:10:30

Regarding the pre-cmmit issue: not really sure what caused it - could also not get it to change anything on my setup. I simply reset that file for now to keep this PR going. Dependencies tend to be difficult in a lot of surprising ways.

SchrodingersGat:
@jacobfelknor nice work, implementation looks very clean!

For the UI side of things, please ensure that the UI elements are hidden if the `STOCK_ENABLE_EXPIRY` global setting is disabled. There are multiple examples in the existing frontend code showing how to use the `useGlobalSettingsState` hook to check this

src/backend/InvenTree/order/serializers.py
```diff
@@ -967,6 +975,7 @@ def save(self):
                         status=item['status'],
                         barcode=item.get('barcode', ''),
                         batch_code=item.get('batch_code', ''),
+                        expiry_date=item.get('expiry_date', ''),
```
SchrodingersGat, line 978: Change this to `expiry_date=item.get('expiry_date'),

src/backend/InvenTree/order/models.py
```diff
@@ -800,6 +800,9 @@ def receive_line_item(
         # Extract optional batch code for the new stock item
         batch_code = kwargs.get('batch_code', '')
 
+        # Extract optional expiry data for the new stock item
+        expiry_date = kwargs.get('expiry_date', '')
```
SchrodingersGat, line 804: Change this to `expiry_date = kwargs.get('expiry_date')`

jacobfelknor:
Sounds good, I will work on these changes now

src/backend/InvenTree/order/serializers.py
jacobfelknor: For this particular one, I ended up using `.get('expiry_date', None)` as it matched the pattern of those around it, even though the `None` is technically redundant.

src/frontend/src/forms/PurchaseOrderForms.tsx
```diff
@@ -298,6 +301,25 @@ function LineItemFormRow({
     }
   });
 
+  const [expiryDateOpen, expiryDateHandlers] = useDisclosure(false, {
+    onOpen: () => {
+      // check the default part expiry. Assume expiry is relative to today
+      const defaultExpiry = record.part_detail?.default_expiry;
```
jacobfelknor, line 307: I am assuming the number of days specified in `default_expiry` should be relative to the receive date.

I'm not well-versed in formatting javascript dates... but coming from Python this doesn't feel great. Open to suggestion here if there is a better way

SchrodingersGat: Could it be as simple as adding N days to the current date, right?

jacobfelknor: Yes... that is what I'm doing... but the formatting is what gets so verbose (lines 311-314). Does the form require me to format it in `"YYYY-MM-DD"`?

SchrodingersGat: We already have `dayjs` dependency in the frontend code, so you could do something like:

```javascript
// Generate a date 10 days in the future
const futureDate = dayjs().add(10, 'day').format('YYYY-MM-DD');
```

jacobfelknor: I like that much better. Thanks for the tip. Updated

SchrodingersGat:
@jacobfelknor thanks for the great work here :)

PR #8883, opened 2025-01-12
[UI] Reactive Details Columns

SchrodingersGat:
- Adjusts number of columns in "details" view based on screen width
- Ref: https://github.com/inventree/InvenTree/discussions/5328#discussioncomment-11805891


wolflu05:
@SchrodingersGat I think you're searching for https://mantine.dev/core/simple-grid/#responsive-props that would be a much cleaner way to handle this.

SchrodingersGat:
@wolflu05 I tried that initially but the width breakpoints are based on the screen width not the width of the element. e.g. it doesn't take the side panel width into account

PR #8893, opened 2025-01-14
fix(plugin): Ensure plugins are only installed if not deactivated

matmair:
the plugin reload mechanism did not consider if plugin installs were deactivated


matmair:
@SchrodingersGat there seems to be an error in the link validation logic

SchrodingersGat:
@matmair yeah looks like readthedocs have changed their build environment variables.

I have integrated a fix in https://github.com/inventree/InvenTree/pull/8895

PR #8905, opened 2025-01-16
chore(backend): bump deps

matmair:
Several backend updates


matmair:
Not really sure why Dependabot is not working... so doing this manually for now

PR #8912, opened 2025-01-19
feat: Do not change source on update/build

matmair:
This changes the invoke tasks to not extract translations on normal translation/compilation so that source code does not change during normal operations.


matmair:
@sur5r I think we talked about changing this to make updates cleaner. Let me know if there is anything else that we should change to make updates behave better

PR #8922, opened 2025-01-20
Plugin reload fix

SchrodingersGat:
This PR fixes another piece of the plugin puzzle.

When updating an existing plugin, which is already installed and has been loaded by the InvenTree server (or worker) process, the plugin module is not *reloaded* - as it is already loaded into `sys.modules`.

Thus, updating the plugin does not actually *reload* it correctly. This can mean that the "in memory" code does not match the "on disk" code. The problem is compounded further when it comes to synchronizing the plugin code between the `server` and `worker` processes - even if they point to the same code "on disk", the processes are not guaranteed to have the same code "in memory".

Thus, this PR forces any active third-party plugin modules to reload right at the point of mounting. This ensures that the "on disk" and "in memory" code is the same.

### Local Testing

I have tested this locally, by running the worker and server processes separately:

- `invoke dev.server`
- `invoke worker`

Then, installing a debug plugin with `-e` option (for editable install) and making live changes to the plugin code.

*Before this fix* the *server* process would reload the new code - but the *worker* process would attempt a reload (due to changed plugin hash) but just reload the *old* code from memory.

*After this fix* the *worker* process loads the correct (new) code when the reload is triggered.

### CI Testing

I am struggling to think of a way to test this in production / CI - but it would be a very good step.... @matmair @wolflu05 any thoughts here?

### DIFF
--- a/src/backend/InvenTree/plugin/registry.py
+++ b/src/backend/InvenTree/plugin/registry.py
@@ -453,6 +453,8 @@ class PluginsRegistry:
 
         Args:
             plugin: Plugin module
+            configs: Plugin configuration dictionary
+            force_reload (bool, optional): Force reload of plugin. Defaults to False.
         """
         from InvenTree import version
 
@@ -485,6 +487,7 @@ class PluginsRegistry:
 
         # Check if this is a 'builtin' plugin
         builtin = plugin.check_is_builtin()
+        sample = plugin.check_is_sample()
 
         package_name = None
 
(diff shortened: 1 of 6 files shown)


matmair:
@SchrodingersGat we could probably just include a plugin for testing that gets installed and edited. At that point I am however questioning how attackable we are making ourselves to xz type of attacks.
So testing data should probably not be distributed in any case.

src/backend/InvenTree/plugin/registry.py
```diff
@@ -510,6 +513,17 @@ def safe_reference(plugin, key: str, active: bool = True):
             # Initialize package - we can be sure that an admin has activated the plugin
             logger.debug('Loading plugin `%s`', plg_name)
 
+            # If this is a third-party plugin, reload the source module
+            # This is required to ensure that separate processes are using the same code
+            if not builtin and not sample:
+                if plugin_module := sys.modules.get(plugin.__module__):
+                    logger.debug('Reloading plugin `%s`', plg_name)
```
matmair, line 520: Is there a way to enhance this with information about the environment (ie if it is a worker)

matmair (approved):

PR #8932, opened 2025-01-20
refactor(backend): reduce tags

matmair:
Reduces unneeded frontend-related tags.


matmair:
@SchrodingersGat wdyt about these tags? Do you see any use for them after #6293 is merged?

SchrodingersGat:
These are all now provided via the API to the frontend code, right?

matmair:
Yes - but not sure if they are used/useful in reports.

wolflu05:
I can see a lot of these tags being used in reports. Maybe we can move them all into one file related to the report app?

matmair:
@wolflu05 can you be a bit more specific? Most of these tags were added for status or information modal or to control user specific rendering - which reports do not support IMO.

wolflu05:
I think the `render_date` tag would make sense to keep.

matmair:
That makes sense, I can re-add that one

SchrodingersGat (approved):

PR #8956, opened 2025-01-26
[Enhancement] Request cache

SchrodingersGat:
This PR adds an ephemeral cache which exists for the duration of a single web request.

The benefit of this is that external cache access (e.g. Redis), while being faster than a database hit, is still kinda slow. There are a number of cache checks which may be performed multiple times per single request - e.g. lookup for the `_PLUGIN_REGISTRY_HASH` key or the `PART_NAME_FORMAT` setting.

### Functionality 

- The first time a setting is accessed (within the context of a request), the setting object is cached. 
- Any subsequent access hits (within the same request) will return the cached value
- The cached values are cleared at the end of the request
- Cache is specific to its owning thread (i.e. does not cache across processes)
- This new cache is **not** automatically used - it has to be called manually
- Manually calling means we can be careful about where we apply it

### Performance

Some rough performance numbers, performing an API request for all items in the demo dataset:

| Model | Without Cache | With Cache | Improvement |
| --- | --- | --- | --- |
| Part | 0.891s | 0.725s | 19% |
| Stock Item | 2.332s | 2.0s | 14% |
| Build Order | 0.07s | 0.06s | 14% |
| BOM Item | 0.98s | 0.83s | 15% |

### Related Issues

- Discovered as part of profiling for https://github.com/inventree/InvenTree/pull/8950

### DIFF
--- a/src/backend/InvenTree/InvenTree/cache.py
+++ b/src/backend/InvenTree/InvenTree/cache.py
@@ -112,6 +112,11 @@ def get_cache_config(global_cache: bool) -> dict:
     return {'BACKEND': 'django.core.cache.backends.locmem.LocMemCache'}
 
 
+def clear_session_cache() -> None:
+    """Reset the session cache."""
+    thread_data.request_cache = {}
+
+
 def get_session_cache(key: str) -> any:
     """Return a cached value from the session cache."""
     request_cache = getattr(thread_data, 'request_cache', None)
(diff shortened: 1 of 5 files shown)


matmair (approved):

matmair:
Seems like there might be cache invalidation issues, judging from the test results

PR #8959, opened 2025-01-26
refactor(backend): remove contrib.messages

matmair:
We are not supporting the built-in message framework anyone. This PR removes the remnants.


matmair:
We need to keep the things registered as admin uses it - so only removal on the template

PR #8960, opened 2025-01-26
feat(backend): improve tag docs

### LINKED ISSUE #8874: No report tag docs
### Body of the issue

# Problem

The docs are missing a list of available tags and their functions, making requests regarding reporting repetitive to answer

# Suggested solution

 Add a generation step similar to the one for settings that exports all available tags.

matmair:
This PR:
- adds the admindocs app - enabling pretty good tags, filters, models and views auto docs in the admin interface
- adds invoke task and commands to export filter and tag definitions
- adds tags and filters to the schema repo
- adds list of all tags to docs
grafik
Closes  #8874


SchrodingersGat:
Nice, these will greatly improve discovery. A couple of small tweaks:

Something weird is going on with the navigation actions at the bottom - maybe the table isn't closing out properly?

image

The "tags" and "filters" headings should be one greater indent level:

image

wolflu05:
My end goal would still be to somehow directly implement this data into the template editor. The best solution would be to somehow write an LSP integration and swap out the editor for Monaco. But that will be a lot of work.

matmair:
@SchrodingersGat i will address those
@wolflu05 fully agree on that.

This is first and foremost a cheap way to achieve a good chunk of the intended functionality and to make sure that we have better insights in how the tags change over time.

matmair:
@SchrodingersGat rendering issues are addressed now

SchrodingersGat:
Thanks!

PR #8970, opened 2025-01-28
[Pricing] Add option to convert received items currency

SchrodingersGat:
Adds a new configurable option to automatically convert the "cost" of an incoming purchase order item to the base currency.

This option is useful when you want to ensure that the converted cost of a stock item is calculated at time of receipt, correctly taking into account the conversion rates *at that point in time*.

The purchase order line item retains information on the original order currency, while the stock item unit cost is converted into the "base" currency.

Additionally, this PR improves the documentation around currency support.

Closes https://github.com/inventree/InvenTree/issues/8980


matmair (approved):

SchrodingersGat:
@matmair @wolflu05 if you have some time, I'd appreciate a look at the failing playwright tests here. There is something peculiar about this PR which causes the unit tests supporting the report printing functionality to fail.

From reviewing the data it appears that the template files are all returning 404 errors - but, I cannot see any indication of what (in this PR in particular) may be causing this...

matmair:
No idea what this could be

PR #8979, opened 2025-01-28
fix(contrib): Invoke Path Issues

### LINKED ISSUE #8827: Invoke Path Issues

matmair:
Fixes #8827


SchrodingersGat:
This is a clever approach! Unfortunately it immediately fails on our devcontainer (which is running off the docker image).

This is because our docker image installs invoke as part of the base image, rather than into the venv.

image

SchrodingersGat:
We could adjust our devcontainer post-install step to manually run `pip install -U invoke --ignore-installed` - and do similar things in our CI

matmair:
@SchrodingersGat this check checks an environment variable that is set at the beginning of the base image - I think it is save to assume this should be available in any image build around our Dockerfile

matmair:
If I understand our images correctly it is normal that the invoke path is not correct in them

SchrodingersGat:
>  think it is save to assume this should be available in any image build around our Dockerfile

The exception here is that in the devcontainer setup, we install the python packages into a virtual environment (because we may need to change them during development). As such, the `invoke` version check then fails.

So I think a good addition here is that we specifically install another copy of invoke into this venv as part of the devcontainer post-install step.

Also, any devs who are using the devcontainer setup will need to run this command too. So we should add "breaking" tag to this PR

matmair:
@SchrodingersGat implemented

.devcontainer/postCreateCommand.sh
```diff
@@ -7,6 +7,9 @@ git config --global --add safe.directory /home/inventree
 python3 -m venv /home/inventree/dev/venv --system-site-packages --upgrade-deps
 . /home/inventree/dev/venv/bin/activate
 
+# Ensure the correct invoke is available
+pip3 install invoke
```
SchrodingersGat, line 11: You will need to add `--ignore-installed` to force it to install into the venv

matmair: Thank you, added

SchrodingersGat (approved):

SchrodingersGat:
Thanks, hopefully this puts this issue to bed :)

PR #8984, opened 2025-01-29
Sales Order Barcode Allocate

fuzeman:
I discovered that when allocating items to a sales order by scanning stock item barcodes, once you complete the first shipment and move to another shipment, no stock items can be allocated to the new shipment.

As can be seen in the commit, previously with the `delivery_date` filter completed shipments would need a delivery date added before items could be allocated to a new shipment. Changing this to `shipment_date` enables multiple shipments to be allocated and shipped together easily.

-----------

Looking at the app, there probably should be a way to allocate into specific shipments. I might be able to take a look when I have some time, happen to use Dart + Flutter myself in a couple of production apps.

### DIFF
--- a/src/backend/InvenTree/plugin/base/barcodes/api.py
+++ b/src/backend/InvenTree/plugin/base/barcodes/api.py
@@ -654,7 +654,7 @@ class BarcodeSOAllocate(BarcodeView):
             return shipment
 
         shipments = order.models.SalesOrderShipment.objects.filter(
-            order=sales_order, delivery_date=None
+            order=sales_order, shipment_date=None
         )
 
         if shipments.count() == 1:


SchrodingersGat:
@fuzeman nice catch, this seems like a sensible fix.

Regarding the app - adding multiple shipment support is a great idea! If you are able to investigate that it would be fantastic. I'd be happy to help you get to know your way around the codebase

SchrodingersGat:
> What are you thinking long-term with the app? I'm more familiar with Flutter, but switching to react-native could be a consideration to enable the sharing of code with the new react interface.

There are no plans to change to react native, although I have considered it (for code sharing purposes obviously).

It would be a good move from that perspective but a *LOT* of work without any support or potential payoff.

And we would need to ensure that all the current flutter libraries (which are already pretty fragile) have react native equivalents that work cross platform.

Mobile dev sucks

fuzeman:
Yeah, good points. It sounds natural on the surface to merge/share code between the two, but I'm sure there will be a lot of pain along the way to achieve that.

PR #8991, opened 2025-01-29
fix(backend): ensure only the intended files are pushed

matmair:
Follow up to https://github.com/inventree/InvenTree/pull/8989


SchrodingersGat:
Do you intend to leave in those debug messages?

matmair:
@SchrodingersGat yes, should make debugging prod issues easier.

PR #8992, opened 2025-01-29
Sales Order Shipment Search

fuzeman:
Fixes searching in the sales order shipment template editor preview and enables searching from the search drawer. I think this will be very helpful when you need to find shipments by invoice or tracking number.


matmair (approved):
Looking good, thanks @fuzeman

SchrodingersGat:
This looks nice! I would suggest that you add a playwright test for this, which would act as an end-to-end test for the new API features and UI integration.

fuzeman:
I've added tests for search by tracking number and invoice number to `Sales Orders - Shipments`, let me know if that's what you were thinking.

---------------

I finally got tests running, but discovered the first `page.goto` was kicking back to the login screen and causing the tests to stall and timeout. For example, the following line must be removed for the test to complete:
https://github.com/inventree/InvenTree/blob/95874d9097eb7629e5484ba5d3320ea6f88cbc9c/src/frontend/tests/pages/pui_sales_order.spec.ts#L103

I haven't included any changes in regards to this issue, guessing something is still partially broken in my development environment.

SchrodingersGat:
@fuzeman please merge in master once https://github.com/inventree/InvenTree/pull/8985 is accepted - and then use the `navigate(page, url)` helper func instead

fuzeman:
@SchrodingersGat I've rebased on master.

-------------

I tracked down my issue running frontend tests was due to cookies being blocked as `samesite: none` was set in my config.yaml. I see this was previously the default which was later fixed in https://github.com/inventree/InvenTree/commit/914f59c4cc524e96dda55c9eacc6b375b9e9f485

I've found tests appear to be a bit unreliable though - timing out sometimes, and passing other times. Looks like it just stalls during page navigation for some reason.

SchrodingersGat:
@fuzeman yes there are some race conditions due to the tests running in parallel - I am trying to clean it up and get them running more reliably. The more tests we add, the worse the problem is becoming!

SchrodingersGat:
@fuzeman thanks for the addition!

PR #9002, opened 2025-02-01
Report print update

SchrodingersGat:
- Refactors printing calls for Report and Label templates
- Allows printing functions to be called internally (not just via API)
- Enables plugins to (for e.g.) generate reports in response to a particular event


SchrodingersGat:
@tristanle22 you will have to wait for the next release I'm afraid - back-porting these charges are not compatible

PR #9004, opened 2025-02-01
Clean up deprecated module removed in python 3.13

1337joe:
Module `imghdr` is deprecated in python 3.11 and removed in python 3.13: https://peps.python.org/pep-0594/

I removed the validation function because I couldn't find a way to trigger it: I didn't see a call in `inventree-python` and all of the platform ui image update paths I tried rejected non-image files but didn't call the function in question to do so.

If anyone can give me a way to hit the validation function then I'd be happy to put together a replacement using one of the libraries mentioned in the link above.

### DIFF
--- a/src/backend/InvenTree/part/serializers.py
+++ b/src/backend/InvenTree/part/serializers.py
@@ -1,6 +1,5 @@
 """DRF data serializers for Part app."""
 
-import imghdr
 import io
 import os
 from decimal import Decimal
@@ -295,13 +294,6 @@ class PartThumbSerializerUpdate(InvenTree.serializers.InvenTreeModelSerializer):
 
         model = Part
         fields = ['image']
-
-    def validate_image(self, value):
-        """Check that file is an image."""
-        validate = imghdr.what(value)
-        if not validate:
-            raise serializers.ValidationError('File is not an image')
-        return value
 
     image = InvenTree.serializers.InvenTreeAttachmentSerializerField(required=True)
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@1337joe looks like a clean removal to me - the other approach we use elsewhere is the `TestIfImage` function which just calls PIL under the hood. So you could replace this code with a call to the `TestIfImage` func?

src/backend/InvenTree/part/serializers.py
```diff
@@ -298,7 +297,7 @@ class Meta:
 
     def validate_image(self, value):
         """Check that file is an image."""
-        validate = imghdr.what(value)
+        validate = InvenTree.helpers.TestIfImage(value)
         if not validate:
             raise serializers.ValidationError('File is not an image')
```
SchrodingersGat, line 302: While you are here, can you please add a translation wrapper around the error message - as it is a user facing string.

```python
raise serializers.ValidationError(_('File is not an image'))
```

SchrodingersGat (changes requested):

src/backend/InvenTree/part/serializers.py
1337joe: Done

SchrodingersGat (approved):

SchrodingersGat:
@1337joe Thanks for the contribution :)

PR #9006, opened 2025-02-01
Uncomment site_url in default config

SchrodingersGat:
- Re-enable default value for site_url
- Fixes some CI steps which require this to be set


SchrodingersGat:
Manged to fix the broken CI pathways with a different solution

PR #9012, opened 2025-02-02
Debounce `checkLoginState` (testing reliability)

fuzeman:
Had a closer look at the issue I was experiencing with tests stalling in #8992 and tracked it down to a redirection issue during login. I'm not sure if this is the best way to go about fixing this issue, ideally fixing this without a time delay would probably be better. Let me know what you think.

As displayed in the screenshots below: two `checkLoginState` calls are being fired as of 1b5019ba529df4e8a9d17117d4071f5b99d22aab, this can intermittently cause an unexpected redirection back to /home after playwright has clicked a tab on the header which in-turn stalls the test until timeout.

-----------

**master / 1b5019ba529df4e8a9d17117d4071f5b99d22aab** *(test successful)*

2025-02-02 23_45_51-DevTools - localhost_5173_platform_home

**master / 1b5019ba529df4e8a9d17117d4071f5b99d22aab** *(test stalled)*

2025-02-02 23_44_44-DevTools - localhost_5173_platform_home

--------------

**PR / c1804696392897e627a6d4be2d109201cc490029** *(test successful)*

2025-02-02 23_53_26-DevTools - localhost_5173_platform_home

### DIFF
--- a/src/frontend/src/pages/Auth/Logged-In.tsx
+++ b/src/frontend/src/pages/Auth/Logged-In.tsx
@@ -1,5 +1,6 @@
 import { Trans } from '@lingui/macro';
 import { Card, Container, Group, Loader, Stack, Text } from '@mantine/core';
+import { useDebouncedCallback } from '@mantine/hooks';
 import { useEffect } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@fuzeman thanks for looking into this, I really appreciate it. The fix looks pretty clean to me, and having tested it locally it does appear to be more "stable". Fingers crossed that this addresses a good part of our race condition related issues!

@matmair @wolflu05 any thoughts on this one before merge?

matmair (approved):

SchrodingersGat:
@fuzeman thanks for the contribution 😁

fuzeman:
I'll take a look at the other playwright test failures as well. I noticed some test failures had 401 Unauthorized ("Invalid token.") logged in the console, not sure if that's relevant yet though.

matmair:
@fuzeman that is normal for some runs (login tests for example)

Issue #9045, opened 2025-02-07
OpenAPI schema doesn't match data returned from the server


1337joe:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

I can work with pretty much any language I can set up an IDE for, but my language of choice for anything complex is likely to be Java. I've been running SQL queries to pull data directly from my InvenTree database, but that's obviously not ideal.

Given that there's an OpenAPI schema published for InvenTree, I figured it would be straightforward to generate a Java library so I could pull data from my InvenTree instance (particularly `SalesOrder` data). Unfortunately, the schema-generated code doesn't support the response I get back from the server.

Running a query for order 1 against the [`order_so_retrieve`](https://github.com/inventree/schema/blob/de44cc2d4fbbaf658e8f260d04626b0138f6844d/export/309/api.yaml#L11443) operation on my dev server with the test data loaded I get the following error (edited to add line breaks for readability):
```
java.lang.IllegalArgumentException:
The field `notes` in the JSON string is not defined in the `SalesOrder` properties.
JSON: {
    "pk":1,
    "created_by":{"pk":1,"username":"admin","first_name":"","last_name":"","email":""},
    "creation_date":"2021-11-17",
    "start_date":null,
    "target_date":null,
    "description":"Selling some stuff",
    "line_items":1,
    "completed_lines":1,
    "link":"",
    "project_code":null,
    "project_code_detail":null,
    "reference":"SO0001",
    "responsible":6,
    "responsible_detail":{"pk":6,"owner_id":3,"name":"engineer","label":"user"},
    "contact":null,
    "contact_detail":null,
    "address":null,
    "address_detail":null,
    "status":20,
    "status_text":"Shipped",
    "status_custom_key":20,
    "notes":"",
    "barcode_hash":"",
    "overdue":false,
    "customer":30,
    "customer_reference":"",
    "shipment_date":"2022-09-07",
    "total_price":"150.000000",
    "order_currency":null,
    "shipments_count":1,
    "completed_shipments_count":1
}
```

Inspecting the [`SalesOrder`](https://github.com/inventree/schema/blob/de44cc2d4fbbaf658e8f260d04626b0138f6844d/export/309/api.yaml#L32751) schema definition I can verify that `notes` is not in the schema, but I can also see that some fields are null that the schema lists as required (`address_detail`, `contact_detail`, `responsible_detail`).

### Steps to Reproduce

[...]

PR #9046, opened 2025-02-07
Add schema generation hints to resolve warning

1337joe:
As a path to understanding the schema generator for #9045, I'm trying to address at least some of the warnings it currently produces.

`/git/inventree/InvenTree/src/backend/InvenTree/company/serializers.py:107: Warning [CompanyList > CompanySerializer]: unable to resolve type hint for function "address". Consider using a type hint or @extend_schema_field. Defaulting to string.`

String is actually the correct type in this case, but the field also needs to be nullable to be fully correct.

The two changes in this pull request accomplish the same goal: providing the nullable string type hint to the schema generator.

Adding a line to the serializer to explicitly define address produces:
```yaml
address:
  type: string
  readOnly: true
  nullable: true
```

while simply adding the return type hint to the model produces:
```yaml
address:
  type: string
  nullable: true
  description: |-
    Return the string representation for the primary address.

    This property exists for backwards compatibility
  readOnly: true
```

I would argue that the second option (adding type hints to the model) is the better way to go:
  1. It's less serializer-specific code that has to be maintained moving forward for simple fields like this one.
  2. The type definition is in the same place as the value that it matches for ease of maintenance.
  3. It displays the relevant documentation comment in the schema.

It's entirely possible that I'm missing something since I'm just focused on the schema, but the one (admittedly major) downside that I see to adding the hint to the model is that it doesn't match the way the existing serializers are set up.

Which route is preferable moving forward?

### DIFF
--- a/src/backend/InvenTree/company/models.py
+++ b/src/backend/InvenTree/company/models.py
@@ -188,7 +188,7 @@ class Company(
     )
 
     @property
-    def address(self):
+    def address(self) -> str | None:
         """Return the string representation for the primary address.
 
         This property exists for backwards compatibility
--- a/src/backend/InvenTree/company/serializers.py
+++ b/src/backend/InvenTree/company/serializers.py
@@ -161,6 +161,8 @@ class CompanySerializer(
 
         return queryset
 
+    address = serializers.CharField(required=False, allow_null=True, read_only=True)
+
     primary_address = AddressSerializer(required=False, allow_null=True, read_only=True)
 
     url = serializers.CharField(source='get_absolute_url', read_only=True)


SchrodingersGat:
Adding a return type hint to the model seems like the better solution to me. It is a "deeper" solution allowing higher level code (such as the serializers) to introspect. I suspect there are probably a number of other models which will need this attention

1337joe:
> Adding a return type hint to the model seems like the better solution to me.

Perfect. I'll prefer that way when it's an option and I'll migrate from serializer to model where I'm already touching it.

Is this the best way to ask questions about coding direction?

SchrodingersGat:
> Is this the best way to ask questions about coding direction?

GitHub is the best place for these discussions so we can keep everything in the same place :)

SchrodingersGat:
@1337joe are you still taking charge of this one?

1337joe:
That's my intent but I started work on it right before I'm going out of town for a week so I haven't had time to dig into all I wanted to. I'll try to get a first pass pushed tonight or tomorrow with the intent of cleaning it up more when I'm back home.

PR #9087, opened 2025-02-16
feat(backend): Add instance id for linked data plugins

matmair:
Adds an instance id that plugins can use to ensure servers are referenced by the same string across plugins (by default uuid4). This ID can also be exposed (off by default) - this could be interesting for linked data/federation discovery.

### DIFF
--- a/src/backend/InvenTree/InvenTree/api.py
+++ b/src/backend/InvenTree/InvenTree/api.py
@@ -208,6 +208,7 @@ class InfoApiSerializer(serializers.Serializer):
         navbar_message = serializers
 
     server = serializers.CharField(read_only=True)
+    id = serializers.CharField(read_only=True)
     version = serializers.CharField(read_only=True)
     instance = serializers.CharField(read_only=True)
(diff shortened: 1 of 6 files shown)


wolflu05:
This is an interesting idea, do you see any way to have some kind of "shorter identifier" to also include optionally in barcodes?

matmair:
@wolflu05, that is out of the scope of this PR. For linked data / RDF reliable global uniqueness is needed - therefore uuid4 was chosen; it was weighed against uuid7

SchrodingersGat:
@matmair LGTM - do you think it is worth adding some docs notes about this now?

matmair:
@SchrodingersGat as the default behaviour is not changed and this is mostly intended for plugins I thought it was not really necessary. I mainly need this to ensure the inventree-federation plugin is not breaking when someone restores/moves their server as that info might be lost if held in the plugin models/settings

SchrodingersGat:
@matmair fair enough, I am not going to push for it here. Thanks for the addition :)

Issue #9099, opened 2025-02-18
Duplicate PO does not use reference pattern


jacobfelknor:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

When duplicating a PO, the reference pattern should be used to generate a new reference. This is a regression from the CUI as this previously worked.

I am willing to fix this issue and will be creating an MR shortly

### Steps to Reproduce

1. Create a PO
2. Duplicate PO
3. Observe that InvenTree requires manually inputting the reference pattern

### Expected behaviour

1. Duplicate PO
2. Observe InvenTree suggests the next reference for the duplicate

### Deployment Method

- [x] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

InvenTree-Version: 0.18.0 dev
Django Version: 4.2.19
Commit Hash: c314a45
Commit Date: 2025-02-18
Commit Branch: master
Database: postgresql
Debug-Mode: True
Deployed using Docker: False
Platform: Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.36
Installer: GIT
Active plugins: [{"name":"InvenTreeBarcode","slug":"inventreebarcode","version":"2.1.0"},{"name":"InvenTreeCoreNotificationsPlugin","slug":"inventreecorenotificationsplugin","version":"1.0.0"},{"name":"InvenTreeCurrencyExchange","slug":"inventreecurrencyexchange","version":"1.0.0"},{"name":"InvenTreeLabel","slug":"inventreelabel","version":"1.1.0"},{"name":"InvenTreeLabelMachine","slug":"inventreelabelmachine","version":"1.0.0"},{"name":"InvenTreeLabelSheet","slug":"inventreelabelsheet","version":"1.0.0"},{"name":"DigiKeyPlugin","slug":"digikeyplugin","version":"1.0.0"},{"name":"LCSCPlugin","slug":"lcscplugin","version":"1.0.0"},{"name":"MouserPlugin","slug":"mouserplugin","version":"1.0.0"},{"name":"TMEPlugin","slug":"tmeplugin","version":"1.0.0"}]

### Please verify if you can reproduce this bug on the demo site.

- [x] I can reproduce this bug on the demo site.

### Relevant log output

```shell

```

jacobfelknor:
It appears that this issue is not unique to POs and also affects other objects that can be duplicated, like BOs

[...]

PR #9100, opened 2025-02-18
Use ref pattern on PO duplicate

jacobfelknor:
Addresses https://github.com/inventree/InvenTree/issues/9099

This patch fixes the issue specifically for POs. If this strategy is acceptable, I will make the change to other places where duplicating an object should use a reference pattern, like BOs, SOs, etc


jacobfelknor:
It wasn't too hard to go ahead and do this for other order types, so I did.

.pre-commit-config.yaml
matmair, line 1: Did you intend to submit this?

matmair (approved):
LGTM

.pre-commit-config.yaml
jacobfelknor: I did not intentionally change this, so if you'd like it reverted that is totally fine by me

matmair: I would appreciate it, changing dependencies has a lot of potential for unintended consequences. We keep them separate normally

SchrodingersGat (approved):

SchrodingersGat:
@jacobfelknor thanks for the great addition! Your tests introduced some conflicts with other tests (due to the parallel nature in which the tests run) so I had to make some fixes elsewhere to get this to work :)

All good now - and thanks  for the contribution!

PR #9114, opened 2025-02-20
[CI] Segfault fix

SchrodingersGat:
Attempting to address a segfault error which has been introduced into our CI workflow. Seems to be only associated with the unit tests when running inside the production docker image.

The failing test has been located as:

```
plugin.base.label.test_label_mixin.LabelMixinTests.test_printing_endpoints
```

However, the cause of the segfault has not as yet been determined.

### Failing Runs

https://github.com/inventree/InvenTree/actions/workflows/docker.yaml?query=branch%3Amaster+is%3Afailure

### Related Issues

- https://github.com/docker-library/python/issues/993
- https://github.com/prowler-cloud/prowler/pull/6109

It appears that this is an upstream issue in the particular alpine image.


SchrodingersGat:
Closing, replaced by https://github.com/inventree/InvenTree/pull/9118

Issue #7528, opened 2024-06-28
Default language is overwriten by browser language settings


NightSkySK:
### Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find a similar issue

### Describe the bug*

in Inventree config file `/etc/inventree/config.yaml` I've applied language setting:
```yaml
# Select default system language (default is 'en-us')
# Use the environment variable INVENTREE_LANGUAGE
language: en-us
```

How ever once I launch Inventree it's set up language to Polish by default:
![image](https://github.com/inventree/InvenTree/assets/52137446/c1d7c1da-6c1a-4e50-94e7-4e2e503ff3d6)

Even if I save the setting for my account to `English` it return back to `Polish` after some time. 
Also language setting seems to not be stored in account setting as in new browser it display all in `Polish` as well.

### Steps to Reproduce

1. Use browser with primary language different than english
2. Set up config.yaml `language: en-us`
3. Open inventree app and check UI language displayed

### Expected behaviour

1. config.yaml should have higher priority than browser language settings
2. If user decide to use different language the setting should be stored within account settings and all browsers and app (including mobile app ) should launch with the same language version

### Deployment Method

- [ ] Docker
- [ ] Package
- [X] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

# Version Information:
InvenTree-Version: 0.15.4
Django Version: 4.2.12

Database: postgresql
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-5.4.0-186-generic-x86_64-with-glibc2.31
Installer: None

[...]

PR #9116, opened 2025-02-20
feat(backend): Add user profile

matmair:
This PR:
- [x] Adds user profiles containing basic info needed for commercial ops
- [x] Adds endpoints to get users profiles
- [x] Enables persisting of preferred language
- [x] Enables persisting of theme settings
- [x] Adds user and group info pages
- [x] Adds user hover and link to detail pages

UI for "normal" users to discover users and groups
grafik
grafik
grafik
grafik

Hover for user / owner / Group Pill
grafik

To address in later PRs:
- Add avatar pictures to users
- Add avatar picture to groups
- Inline self-service group joining

Closes https://github.com/invenhost/InvenTree/issues/107
Closes https://github.com/invenhost/InvenTree/issues/117
Closes https://github.com/inventree/InvenTree/issues/7528


matmair:
@SchrodingersGat @wolflu05 This adds the following fields for users; anything more required from your pov?
           language  (internal)
           theme (internal)
           widgets  (internal)
           displayname
           status
           location
           active
           contact
           type
           organisation
           primary_group

matmair:
The design of this has one problem currently: if you change widgets, theme or language settings in one session it will be changed in the database to that value. Other sessions might not get the change and if you change things there the changes from session 1 will not be considered.
I do not think this is a huge issue and there are ways to mitigate this but it makes the data flows more complex than they need to be IMO.

matmair:
@SchrodingersGat This is ready for review; i will add some playwright tests but other than that this should be done

SchrodingersGat:
I get a "permission denied" error when I try to edit user profile:

image

src/backend/InvenTree/users/models.py
```diff
@@ -924,3 +925,86 @@ def create_missing_rule_sets(sender, instance, **kwargs):
...
+
+    class UserType(models.TextChoices):
+        """Enumeration for user types."""
+
+        BOT = 'bot', _('Bot')
+        INTERNAL = 'internal', _('Internal')
+        EXTERNAL = 'external', _('External')
+        GUEST = 'guest', _('Guest')
+
+    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name='profile')
```
SchrodingersGat, line 941: All of these fields need translated `verbose_name` and `help_text` attributes - so that they will be auto-translated for the forms UI.

matmair: done

SchrodingersGat:
> @SchrodingersGat thank you for the review - I have addressed the issues you raised. Should we add a title or position field too? I have found that pretty handy in my org of +10k ppl

Yeah, that makes sense, happy for that addition :)

matmair:
@SchrodingersGat done!

SchrodingersGat:
@matmair still looks like the profile settings cannot be edited by non-admin users:

image

matmair:
should be fixed now

SchrodingersGat (approved):

SchrodingersGat:
@matmair nice work here, thanks for making the fixes. This is a great new feature :)

matmair:
@jacobfelknor would you be interested in getting this integrated with the LDAP sync? I have some ideas but would need a user/admin to hash some things out from the consumer side.

jacobfelknor:
I'm happy to take a look. It will probably be a few days before I can commit some time to it

jacobfelknor:
Sorry for the delay on this, but I'm hoping to have some time to take a look over the next couple days

jacobfelknor:
@matmair alright, to answer your questions

> 1. which of the new profile fields you would show on the hover cards

I think the minimal set of name, active, and status message would be all I need/expect. 

I can see how having job title is helpful for large organizations, so that makes sense too if you'd like it. I do like adding whether the user is an administrator as well, and perhaps their primary group would be helpful if set.

>  2. what behaviour you would expect with LDAP sync: should there be a default mapping into profiles
>  3. should synced arguments be locked or still editable; would that make sense as a setting

I tend to follow GitLab's example on these kinds of questions. They keep it pretty simple. They obtain email, name, and group membership from LDAP. Only the email address is locked (since that is the login method). It has a setting for whether users can change their Display Name. Everything else in the user profile (job title, location, etc) can be edited freely and isn't populated directly from LDAP

I have noticed that right now in InvenTree the email from LDAP is attached to the Django user object, but doesn't get populated in the profile email addresses. I think this is the biggest deal for our instance moving forward. I think I add this with the [django_auth_ldap.backend.populate_user](https://django-auth-ldap.readthedocs.io/en/latest/reference.html#django_auth_ldap.backend.populate_user) signal

matmair:
Thank you for your feedback, I will file a PR to sync these fields

PR #9118, opened 2025-02-21
[Docker] Alpine image fix

SchrodingersGat:
Replaces https://github.com/inventree/InvenTree/pull/9114

It appears that the base docker image `python:3.11-alpine3.21` has some issues with segfaulting (see issues referenced below)

Thus, reverting back to `python:3.11-alpine3.20` which does not appear to have this issue.

The **caveat** is that this image only supports postgres:16 - the limitations around this are now referenced more clearly in the documentation.

Note: I have run the failing test (referenced in #9114) multiple times on this new setup, without issue.

### References

- https://github.com/docker-library/python/issues/993
- https://github.com/prowler-cloud/prowler/pull/6109

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -27,7 +27,7 @@ RUN apt update && apt install -y \
    postgresql-client \
    libldap2-dev libsasl2-dev \
    libpango1.0-0 libcairo2 \
-   weasyprint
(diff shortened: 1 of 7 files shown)


matmair (approved):
LGTM; thank you for tackling this- segfaults are to low on the stack for me

PR #9120, opened 2025-02-21
feat(frontend): Add authentication debug window

matmair:
Adds a debug window that shows past authentications (only for superusers)
grafik
And here after a mandatory re-auth to change or view security-relevant information
grafik


matmair:

SchrodingersGat:
1. Is this showing auths for all users, or just the superuser?
2. Can we move this to a separate table in the "admin" section? It feels a bit too prominent for a rarely used feature

matmair:
1. this is showing auth for the current session -> especially helpful to find out why one was requested to provide a password again
2. I do not think this makes sense in the admin section

matmair:

SchrodingersGat:
I still think it would be generally useful to be able to see this in the "admin center" view (still only for superusers) to view all logged in session information. Much like the view we had in the old UI. All the API / data is already there, just needs to be formatted into a superuser-only table?

matmair:
Not sure; login sessions is something different than what it achieved here.
This PR adds debug info for the current session - including multifactor re-authentications. This is usefull to debug various mfa issues

SchrodingersGat:
@matmair this is the last outstanding `1.0.0` issue (currently) - do you want to fix the UI errors and I'll review again? Or, push off?

matmair:
looking into it now

matmair:
@SchrodingersGat can you be more specific what errors you are referring to? It seems to work as intended
grafik

SchrodingersGat:
@matmair it was just failing the frontend tests, seems to be OK now.

I have a [slightly adjusted branch](https://github.com/SchrodingersGat/InvenTree/tree/auth-session) which moves this auth session table into the "user settings / security" panel - would you be OK with this? I think this is much cleaner while retaining functionality

image

matmair:
I guess so. The whole window is only shown for superusers and mostly helpful for quick passkey debugging but 2 extra clicks are probably not too bad in every debug.

SchrodingersGat:
Replaced by https://github.com/inventree/InvenTree/pull/10271

PR #9121, opened 2025-02-21
refactor(frontend): Auth component refactor

matmair:

Refactors the way login/auth components are structured to ensure a consistent look


SchrodingersGat:
@matmair I notice on the "reset password" screen that if you leave everything blank, the errors don't get populated to the right fields in the form:

image

matmair:
@SchrodingersGat I made the confirm button disabled in that case to address this

PR #9122, opened 2025-02-21
feat(frontend): show if redirecting inline

matmair:
Makes it clearer if the user is currently being redirected by showing a loading icon instead of an empty auth window. Especially interesting for the demo as that takes a few seconds to redirect


src/frontend/src/pages/Auth/Login.tsx
```diff
@@ -94,22 +96,28 @@ export default function Login() {
...
-                </Anchor>
-              </Text>
+            {isLoggingIn ? (
+              <Loader />
+            ) : (
+              <>
+                <AuthenticationForm />
+                {both_reg_enabled === false && (
+                  <Text ta='center' size={'xs'} mt={'md'}>
+                    <Trans>Don&apos;t have an account?</Trans>{' '}
```
SchrodingersGat, line 106: Do quote characters not get properly escaped inside the `<Trans>` block?

matmair: Not really sure, this was autoformatted in the prior PR and seems to render fine

SchrodingersGat: If possible I would prefer the literal `Don't` here - it will make the translator's job easier?

SchrodingersGat: Ok, fair enough

matmair:
@SchrodingersGat does this require any changes or can this be merged?

PR #9136, opened 2025-02-21
remove: need for hard coded token

matmair:
codecov supports github OIDC for uplad so we can remove the hard-coded token


matmair:
@SchrodingersGat after merging this we can remove the secret CODECOV_TOKEN

PR #9143, opened 2025-02-22
Schema cleanup tmp

1337joe:
This doesn't cover everything, but it at least hits the endpoints I've written tests for using a opanapi-generated api client in Java. I'll continue working on that as I have time but I wanted to get a first batch committed, especially since I see the auth api just got reworked and I don't want to get merge conflicts from sitting on my changes too long.

The added type hints address warnings thrown by schema generation.

Nullable is added to various fields that my tests get null responses from using the sample data.

Works on #9045 but doesn't fully solve it


matmair (approved):
LGTM

1337joe:
I'll get the failing tests cleaned up tonight. Apparently the nullable type hint that I put in for the schema generator is not at all supported by the testing framework.

SchrodingersGat:
@1337joe getting there! You just need to increment the API version in `./src/backend/InvenTree/InvenTree/api_version.py`

SchrodingersGat:
@1337joe you'll also need to add a descriptive entry into the api version file :)

SchrodingersGat (approved):

1337joe:
I'm guessing there's nothing I can do about the API schema documentation diff being too big for the job to complete?

SchrodingersGat:
@matmair what do you think here? Are we OK to merge, or will this break things?

matmair (approved):
Seems to be covered by tests pretty well; should be ok;
Ready for merge

PR #9150, opened 2025-02-22
Remove 'url' field from serializers

SchrodingersGat:
- Not used in any official client
- Inconsistent use across models

@matmair any thoughts on this one? *Some* models encode the "url" field into the API - but not all models do this. In fact, only a few. The clients (app / web) do not use this. 

If you are in agreement I think we should simply remove this field for a consistent approach.


matmair (approved):

matmair:
I don't think we need them so sure.

matmair (approved):

PR #9158, opened 2025-02-22
Cleanup backend code

SchrodingersGat:
Remove some TODO entries that are out of date or won't be addressed.

### DIFF
--- a/src/backend/InvenTree/InvenTree/settings.py
+++ b/src/backend/InvenTree/InvenTree/settings.py
@@ -1298,9 +1298,7 @@ ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE = True
 
 HEADLESS_ONLY = True
 HEADLESS_TOKEN_STRATEGY = 'InvenTree.auth_overrides.DRFTokenStrategy'
-MFA_ENABLED = get_boolean_setting(
-    'INVENTREE_MFA_ENABLED', 'mfa_enabled', True
-)  # TODO re-implement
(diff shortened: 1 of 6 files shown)


src/backend/InvenTree/InvenTree/settings.py
```diff
@@ -1298,9 +1298,7 @@
 
 HEADLESS_ONLY = True
 HEADLESS_TOKEN_STRATEGY = 'InvenTree.auth_overrides.DRFTokenStrategy'
-MFA_ENABLED = get_boolean_setting(
```
matmair, line 1301: That is still relevant

SchrodingersGat: Ok, sorry about that! Thought you had addressed this in the recent merge.

matmair: I did but there are no satisfactory tests

src/backend/InvenTree/machine/tests.py
```diff
@@ -275,11 +275,6 @@ def test_print_label(self):
         self.assertEqual(self.print_labels.call_args.args[0], self.machine.machine)
         self.assertEqual(self.print_labels.call_args.args[1], template)
 
-        # TODO re-activate test
-        # self.assertQuerySetEqual(
-        #    self.print_labels.call_args.args[2], parts, transform=lambda x: x
-        # )
-
```
wolflu05, line 282: I think this is still relevant, when this is implemented again. Currently somehow this was removed from the function implementation, but I still see high value to have direct access to the class instances of the specific items that where to print.

PR #9191, opened 2025-02-26
[WIP] Call machine func

SchrodingersGat:
Adds framework for calling a specific machine / driver function by string lookup.

Avoids pickling and caching issues when offloading machine tasks to the background worker process.


matmair (commented):
LGTM; plugin and machine registry seem to become more similar over time; this pattern might make sense to unify

matmair (approved):

PR #9197, opened 2025-02-27
Add part full name to supplier part table

SergeoLacruz:
First try on PUI :-)


SchrodingersGat:
Nice work @SergeoLacruz :)

PR #9198, opened 2025-02-27
Show location path in stock tables

gunstr:
A first little PR from me to try out the process and learn a bit.

The full location path were visible in CUI in the stock tables. If the same sub location like "Box1" is used for multiple parents such as ShelfX  the full path is necessary to find the correct part.

I had the intention to also show the full path on the stock item details page. But i have not been able to figure out how to do that so I leave it as is for now.

### DIFF
--- a/src/frontend/src/tables/ColumnRenderers.tsx
+++ b/src/frontend/src/tables/ColumnRenderers.tsx
@@ -70,7 +70,7 @@ export function LocationColumn(props: TableColumnProps): TableColumn {
         );
       }
 
-      return <Text>{location.name}</Text>;
+      return <Text>{location.pathstring}</Text>;
     },
     ...props
   };


matmair:
Not sure about this, long paths break to tables/make them very uncomfortable to use

gunstr:
@matmair With the current implementation I have not found a way to easily find where the parts are for example looking at the "allocated stock" page for a build order. You can click each individaul stock item, look at the breadcrumb and go back but that takes a lot of clicks. As this was the CUI behaviour I thought it would be a reasonable fix.

Do you have a better suggestion to distinguish sub-locations with same name?

SchrodingersGat:
In general we need a solution for "wide columns" in the new tables.  In multiple tables the "description" or "link" column can take up significant screen realestate.

One option might be to "shorten" the string and show the full value on hover? Mantine datatable also allows us to customize column widths. Or perhaps there is a better design pattern to follow here.

gunstr:
Somehow the full path has to be accessible as a hover or any other solution to find the physical location of a part.

I understand that to show the full path everwhere is probably not feasible when working with small screens. This obviosly needs more considerations and a more complex implementation that probably goes beyond my current level of competence. If you do not disagree I assume I better just close this PR and let you guys do whatever you think is best if and when you have the time.

Having that said I see the "Allocated Stock" tab for build orders being a perfect on screen pick list for build orders and as such the full location path should preferably be visible in this particular table without any user interaction. Requires reasonable short location names and / or a large screen of course. For that purpose I will keep this patch on my local production installation for now.

gunstr:
Closing this now as I guess it has to be replaced with someting more advanced

PR #9096, opened 2025-02-18
[Feature] Data export plugins

SchrodingersGat:
This PR adds the ability to define custom plugins which augment data export procedure.

- The "default" export action is offloaded to a builtin plugin.
- Moves "exporter" code out of the "importer" app
- Adds a new "export" output - will allow exports to be processed in the background, in future
- A bunch of refactoring too

### Related Issues

- Closes https://github.com/inventree/InvenTree/issues/8686
- Closes https://github.com/inventree/InvenTree/issues/9078 (by adding a custom BomExport plugin)
- Closes https://github.com/inventree/InvenTree/issues/9218

### TODO

- [x] Documentation for new plugin class
- [x] Add unit tests for BOM exporter
- [x] Playwright tests
- [x] Implement multi-level BOM export
- [x] Export supplier data
- [x] Export manufacturer data
- [x] Export substitute part data
- [x] Export part parameter data
- [x] Offload data export to the background worker (ref: https://github.com/inventree/InvenTree/pull/9199)

PR #9199, opened 2025-02-27
[WIP] Background reports

SchrodingersGat:
Work in progress to allow report generation in the background worker

- Crucial for long-running printing processes which would otherwise timeout the worker

### TODO

- [x] Offload report printing
- [x] Offload label printing
- [x] Playwright testing

### Future Work
- Offload data export - https://github.com/inventree/InvenTree/pull/9096
- Consolidate export output models into single table
- Save error message against printing output object


matmair:
@SchrodingersGat is there a reason to have separate output models and endpoints for reports and labels or can we merge these? We seem to duplicate a lot of functions between the two

SchrodingersGat:
It is a big duplication - and about to become a triplication with the [data export plugins](https://github.com/inventree/InvenTree/pull/9096)

If we moved all this onto one model, and define the "export type" field, then this could be reduced to a single model.

I might look into that after this PR, to reduce the scope

PR #9217, opened 2025-03-02
refactor(frontend): Do not request token in frontend

matmair:
This removes the issuance of tokens to the front end as we do not rely on them anymore with the now session-driven approach.


matmair:
Looks like a potential crsf issue

SchrodingersGat:
@matmair as the CSRF issues are unrelated to your changes here, if you can revert the axios version changes, I'm happy to merge. We will have to continue to look into the CSRF issues separately

matmair:
@SchrodingersGat done

PR #9236, opened 2025-03-05
Check for schema generation state when removing fields

1337joe:
The schema generator doesn't have any tracking for fields that are removed conditionally based on the view type, resulting in the conditionally removed fields not getting included in the schema. This results in messages that don't validate against the schema being produced whenever the fields are actually included.

I added a check in `ready.py` to see when the schema generator is running, then used that to skip field removal in all the serializers that followed that pattern.

Once all the fields were found by the schema, I had to remove the conditionally-removed fields from the schema required fields because most of them are objects, not primitive types, and just setting `allow_null=True` wasn't enough to get the schema to validate when they're outright missing. I used a postprocessing hook into drf-spectacular to accomplish this.

Works on #9045

Replaces #9226


matmair:
Good work here.

I would like to discuss a few things before we merge this in as it raises a few design issues. These are not directly addressed to you @1337joe but a general discussion for you and @inventree/maintainer @inventree/triage:
- Is there a way we can check for such problematic conditions in the CI - ie get notified if we need to add fields to the processing script
- This is a fast fix for a problematic pattern. IMO we should clean up these patterns and stop completely removing fields. Not sure what the correct way to handle this is: maybe we should null the field if the are not used

src/backend/InvenTree/InvenTree/ready.py
```diff
@@ -44,6 +44,11 @@ def isRunningBackup():
     )
 
 
+def isGeneratingSchema():
+    """Return true if schema generation is being executed."""
+    return 'schema' in sys.argv
```
matmair, line 49: We probably need to extend this check to also assess if we are using the included api-doc browser

1337joe:
Dynamically building the list of fields to un-require to the fields that are conditionally removed was probably the one thing the bash scripts from my other PR arguably did better than this. I don't know of a way to automatically check what needs to be in the list of un-required fields.

I think the root of the problems I'm running into is that InvenTree generates the schema but doesn't actually use it for anything, so there's no immediate feedback if the schema definition doesn't actually match the API. I don't have a fix for this. I've been doing all of my schema testing with a Java project that builds from the schema and calls the API, but so far I've just been playing whack-a-mole trying to call the various list and retrieve operations and fix it so the demo dataset objects returned actually parse correctly.

----

In terms of addressing the list of fields removed from the required list: currently all `read_only` fields are automatically marked `required` by drf-spectacular (https://github.com/tfranzel/drf-spectacular/issues/1265). This is appropriate for primary keys, but less so for things like `address_brief`, which may or may not exist but definitely shouldn't be modified.

We can disable that in the settings (`'COMPONENT_NO_READ_ONLY_REQUIRED': True,`) but then all `read_only` fields cannot be marked `required` (rest_framework has a check for `assert not (read_only and required)`) and at least some of them should be (pk being the obvious one). I'm not sure there's harm to them not being required, but I just checked what happens if I change the setting and found it affected 543 individual fields (267 unique field names) and I was hesitant to hit that many things all at once.

Primitive types (int, bool, str) can be marked nullable to get around the `read_only` => `required` link, but fields that contain objects still fail to validate when they're required, nullable, and excluded from the server response; hence the postprocessing hook. I've actually got a few more fields that I added to the postprocessing hook based on testing, but I wanted to get this set of changes reviewed before I made things more complex building on it.

matmair:
So could we do the same as you are doing in the CI? build a client based on the schema and just iterate over all methods?

1337joe:
You certainly could, I just worry it will be fragile to maintain. It seems like a lot of steps: generate schema, generate client against schema, install server, install test dataset, start server, run tests using the generated client.

Actually, with named arguments in Python (Java doesn't have them) it might be manageable.

Regardless, I think getting the schema fixed up to the point that a generated client can call the API is a good place to start.

1337joe:
I've added tests for all I could figure out how to hit. At this point I think it's ready to merge from my end, assuming there aren't any new comments to address.

matmair (approved):

matmair:
@1337joe off-topic this PR is good to merge IMO

would this help with the issue? does you generator support this https://redocly.com/learn/openapi/discriminator

SchrodingersGat:
Hey @1337joe thanks for implementing this. One question - you are hard-coding the list of removed fields, I'm wondering if there is any way to automatically detect these removed fields? Otherwise, this list will have to be maintained as we update the API

matmair:
@SchrodingersGat currently working on an ast finder pattern that will address this; we can run this as part of the ci test

1337joe:
@SchrodingersGat I was trying to narrowly hit the fields that were dynamically removed, which I can't see from downstream. I just realized I can probably detect from the schema generation dict in the post-processing the specific conditions that are likely to break (read-only and object type).

Let me see if that works - would be a more generally-applicable post-processing hook based on annotations instead of a hardcoded list.

matmair:
@1337joe This would be my suggestion to ensure the filters stay up to date
https://github.com/matmair/InvenTree/commit/ab2a54f847c53701a2bbe3a3013f81ab9c93ca2e
(in my fork, can push to this PR if you agree with the approach https://github.com/matmair/InvenTree/tree/pr/1337joe/9236)

1337joe:
> @1337joe This would be my suggestion to ensure the filters stay up to date [matmair@ab2a54f](https://github.com/matmair/InvenTree/commit/ab2a54f847c53701a2bbe3a3013f81ab9c93ca2e) (in my fork, can push to this PR if you agree with the approach https://github.com/matmair/InvenTree/tree/pr/1337joe/9236)

I didn't know you could do that kind of inspection in python, so I appreciate seeing it even if it doesn't get used. Sorry if I wasted your time.

I was too fixated on specifically the set of fields that could be conditionally removed that I didn't consider just testing for the problem state: read-only (and therefore required by drf-spectacular) but also nullable. I went through and annotated the appropriate fields as `allow_null` then just checked for required fields that were nullable and read-only and removed them from the required list.

The changes I just pushed are more flexible and don't require a statically-defined list of fields that are known to need removal from the required list. Let me know if you see any problems with how I changed it.

SchrodingersGat:
@matmair you happy with these changes?

matmair:
@SchrodingersGat yes

SchrodingersGat:
@1337joe thanks for the great work here :)

PR #9239, opened 2025-03-05
Ignore sentry for TemplateSyntaxError

SchrodingersGat:
- Getting flodded with reports of users misapplied template filters


SchrodingersGat:
@matmair failing due to API mismatch but unrelated to this change - we'll have to work out how to fix that.

Issue #9166, opened 2025-02-24
[UI] Create API token from admin center


SchrodingersGat:
Currently user can view API tokens but cannot create a new one (without going to the django interface)

matmair:
See also https://github.com/invenhost/InvenTree/issues/92

matmair:
@SchrodingersGat are you addressing this?

SchrodingersGat:
Not in my sights currently, happy for you to tackle it!

PR #9244, opened 2025-03-05
Feat(frontend): Add token managment tools

matmair:
Fixes https://github.com/inventree/InvenTree/issues/9166
Fixes https://github.com/invenhost/InvenTree/issues/92

### DIFF
--- a/src/backend/InvenTree/users/api.py
+++ b/src/backend/InvenTree/users/api.py
@@ -9,11 +9,11 @@ from django.views.decorators.csrf import ensure_csrf_cookie
 from django.views.generic.base import RedirectView
 
 import structlog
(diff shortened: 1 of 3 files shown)


matmair:

SchrodingersGat:
Looks great in general - some notes below

### Token Reveal

image

- A "copy button" would be nice addition here
- Please use `<StylishText>` for the modal title

### Tokens Table

image

- It would be good to view the actual user details in this table - not just user ID
- Add table filter for "active"
- Allow sorting by "created" / "expiry" / "user

matmair:
I have adressed the raised issues.

As StylishText does not support customisation it should be phased out IMO. Having the UI set to another primary colour than the default one looks irritating.

PR #9277, opened 2025-03-10
Squash migrations

### LINKED ISSUE #5015: Squash migrations for better performance
We currently ship a lot of migrations that are run in order; even on new instances. I suggest shipping 1.0 with [squashed migrations](https://docs.djangoproject.com/en/dev/topics/migrations/#migration-squashing) and removing the old migrations with the next version. There would need to be a check that versions below 0.x can not be upgraded to a version > 1.x

### LINKED ISSUE #6417: Possible breaking changes for 1.0

matmair:
Reduces migrations as far down as possible to reduce migration overhead in tests and new instances
Closes #6417
Closes #5015


SchrodingersGat:
Is this sufficient to close out https://github.com/inventree/InvenTree/issues/5015  ?

matmair:
not in this current iteration; i am playing with ways to make the used process more reliable but it is not there yet

SchrodingersGat:
Not stale. @matmair where are we at with this?

matmair:
Will try another approach; models are too strongly coupled

SchrodingersGat:
It it possible to perform a migration "replacement"? More of a clean break where we describe the current state of the models in a single migration for each app?

matmair:
They are too strongly coupled for that, you need multiple migrations for a few of the apps and that creates problems with circular imports.
We will need to work on the coupling before we can do such a maximal replacement

Issue #5212, opened 2023-07-10
Platform UI


matmair:
Overview for the porting of classic UI to Platform UI. Follow up to #3901 #2789

```[tasklist]
### Workflows
- [x] Run unit tests as part of CI #5011
- [x] Code coverage integration with coveralls
- [x] Enforce code styling (including comments) for tsx code
- [x] Run code style checks as part of CI #5011
- [x] Release: Docker #5011
- [ ] #5277
- [ ] https://github.com/inventree/InvenTree/issues/5278
- [ ] https://github.com/inventree/InvenTree/issues/5282
```

```[tasklist]
### Backend
- [x] Add SPA helpers #5011
- [x] Remove custom redirects #5011
- [x] Simple login #5209 #5207
- [x] Save ui settings https://github.com/inventree/InvenTree/pull/5343
- [ ] https://github.com/inventree/InvenTree/issues/6389
- [ ] https://github.com/inventree/InvenTree/issues/3101
```

```[tasklist]
### Docs
- [x] Frontend architecture
- [x] Code structure
- [x] Getting started for frontend
- [x] Component docs (maybe storybook)
```

```[tasklist]
### Frontend base
- [x] Navigation #5207
- [x] Mega menu #5207
- [ ] https://github.com/inventree/InvenTree/issues/5323
- [x] URL/navigation structure #5207
- [x] Base page layout
- [x] Detail page layout
- [ ] https://github.com/inventree/InvenTree/issues/5354
- [ ] #5218
- [x] Barcode Scanning #5207
- [x] Notifications
- [ ] https://github.com/inventree/InvenTree/issues/5345
- [ ] https://github.com/inventree/InvenTree/issues/6283
- [ ] https://github.com/inventree/InvenTree/issues/6282
- [x] Social Login #5753
- [x] Social Registration #5753
- [ ] #5296
- [ ] https://github.com/inventree/InvenTree/issues/5888
- [ ] #5755
- [ ] https://github.com/inventree/InvenTree/issues/6285
- [ ] https://github.com/inventree/InvenTree/issues/6813
- [x] [PUI] Switch to native Tree component - https://mantine.dev/core/tree/#example-files-tree
- [ ] https://github.com/inventree/InvenTree/issues/7920
```

```[tasklist]
### Frontend pages
- [ ] https://github.com/inventree/InvenTree/issues/5324
- [x] User settings management #5016
- [x] Global settings management #5016
- [x] Part
- [x] Part Category
[...]

PR #9283, opened 2025-03-11
feat: New / Refactor "Nav Mixin"

### LINKED ISSUE #5269: [FR/P-UI] New / Refactor "Nav Mixin"
- [x] Append nav items to general plugin endpoint
- [x] Dynamically render navbar from this endpoints
- [x] Add plugin nav to api state

matmair:
This PR adds a navigation extension for PUI. This is mostly the same as the NavigationMixin with on crucial thing missing: there are no menus on the navigation right now. I did not want to touch that to keep the PR cleaner.  

Fixes #5269
Fixes https://github.com/inventree/InvenTree/issues/9276


matmair:
This is the last thing tracked in https://github.com/inventree/InvenTree/issues/5212 so we might be able to close that after this is merged

matmair:
@SchrodingersGat do we want to support the old NavigationMixin or remove it?

matmair:
@SchrodingersGat  If we do not want to support the NavigationMixin this is ready for merge

SchrodingersGat:
@matmair which parts of the old nav mixin are we "not supporting" in this case?

matmair:
@SchrodingersGat Currently there is no support for the old Mixin. It is not considered by the new integration that completely uses the plugin UI feature

matmair:
@SchrodingersGat they go whereever the link leads; the sample leads nowhere because O did not want to make it more complex but I can add a url mixin if you want

src/frontend/src/components/plugins/PluginUIFeatureTypes.ts
```diff
@@ -73,3 +73,17 @@ export type TemplatePreviewUIFeature = {
...
+export type NavigationUIFeature = {
+  featureType: 'navigation';
+  requestContext: {};
+  responseOptions: PluginUIFeature;
+  featureContext: {
+    ref: HTMLDivElement;
+    registerHandlers: (handlers: {
+      setCode: (code: string) => void;
+      getCode: () => string;
+    }) => void;
```
wolflu05, line 86:
```suggestion

```
 This is not used anywhere.

src/backend/InvenTree/plugin/samples/integration/user_interface_sample.py
```diff
@@ -197,6 +197,18 @@ def get_ui_template_previews(self, request, context, **kwargs):
...
         ]
 
+    def get_ui_navigation_items(self, request, context, **kwargs):
+        """Return a list of custom navigation items."""
+        return [
+            {
+                'key': 'sample-nav-item',
+                'title': 'Sample Nav Item',
+                'icon': 'ti:menu',
+                'source': self.plugin_static_file('sample_navigation.js'),
```
wolflu05, line 207: Where is this source used?

matmair: Nowhere, I got errors previously but have removed it now

matmair: removing the source leads to me having to change the serializer used for all features - is this something that we want?

wolflu05: Mh, thats interesting. But I see. In my initial implementation I have added all of these things where I didn't seen a generic usecase to an options dictionary, but this somehow was changed. Maybe we can improve this for also getting better typing by also adding a typed dict for the options. And regarding the source, best would be to make it mandatory for certain UI features and none for others that dont require a source now...

matmair:
Ready for review @inventree/triage

SchrodingersGat:
Thanks @matmair - getting pretty close to 1.0.0 now

PR #9292, opened 2025-03-13
Static Asset Management

SchrodingersGat:
This PR ensures that the static frontend assets are served entirely by the proxy rather than requiring the python server to issue redirects back to the proxy.

This results in significant reduction in server load and reduces the number of backend requests a lot!

Tested out on the demo server, with much faster performance now.

### DIFF
--- a/contrib/container/Caddyfile
+++ b/contrib/container/Caddyfile
@@ -41,6 +41,14 @@
 		max_size 100MB
 	}
 
+	# Redirect 'asset' requests to the static file server
+	handle_path /assets/* {
+		import cors-headers assets
+
+		root * /var/www/static/web/assets
+		file_server
+	}
+
 	# Handle static request files
 	handle_path /static/* {
 		import cors-headers static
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@matmair @wolflu05 I have discovered here a  massive performance increase, the "assets" directory is being served by the python / gunicorn process (which is super slow, and just performs redirects).

Making the proxy server (caddy / nginx / apache) serve these files is *way* better!

There are two ways to approach this:

1. Update the proxy config to redirect requests from `/assets/` to `/static/web/assets`
2. Change the vite config to look for assets at `/static/web/assets`

This PR currently does both.

However with 2) it breaks the playwright config. Ideally, there's some way to tell playwright to use the "default" `/assets` path, which is slow but not an issue for testing. However I can't work out how to achieve this. Any ideas??

matmair:
@SchrodingersGat looks interesting. Have you checked if the files are still generated in the correct directory for testing/shipping as part of the CI?

matmair:
Might also be an oputunity to clean up this section

SchrodingersGat:
Ok, this change has now been simplified dramatically - looks like specifying a [relative base](https://vite.dev/guide/build.html#relative-base) in the vite config has fixed it up

Issue #1071, opened 2020-10-25
Add "External Build Order"


SchrodingersGat:
Add a concept which (kind of) combines the ideas of a PurchaseOrder and a BuildOrder

- An external supplier provides a "process" which converts one part into another
- e.g. an anodization step turns "un-anodized widget" into "anodized widget"
- e.g. an external PCB assembler loads components onto a board
- Needs to take items out of stock and add new items into stock
- If items are tracked, need to make use of the `install_into` function
- Also retain some of the concepts of a PurchaseOrder

eeintech:
I'm not sure I see the purpose of tracking external builds. It makes sense to track items and processes inside a same entity/company, however outside jobs just turn in as "finished goods" therefore purchase/external build order leads to parts and stocks. Maybe I just don't see what kind of application would this be used for.

SchrodingersGat:
Agreed it's a weird case. This is at the request of a friend of mine who has a very particular use case. I'm not 100% sure what the actual implementation looks like yet, this issue is (for now) mostly just a placeholder until i work out if it necessarily makes sense.

jlaw:
I think I am looking for this use case. I want to use InvenTree to track PCB manufacturing:
1. Order 15 bare boards from PCB fab (supplier+manufacturer).
2. Receive 15.
3. Order 10 assembled boards from PCB assembler (supplier+manufacturer). They require 10 bare boards from me, so I need to supply 10 to them from my 15 stock.
4. Receive 10.
5. Sell 5 to customer.
6. Customer wants 10 more.
7. Place order with PCB assembler for 5 more, taking the remaining 5 bare boards.
8. Receive 5.
9. Take 5 new + 5 existing in inventory to sell to customer.

eeintech:
@jlaw Instead of "selling" your PCB + BOM to your PCB assembler, why not creating build orders for that process?
Granted, your PCB and other parts will still show as "in-stock" in your InvenTree but once they are allocated to the build they become essentially unusable for anything else. So that would be one way I would do it.

jlaw:
Yes, that is one way, but I would essentially have to double enter the order: once as a BO and the second as a PO, which I think this issue was suggesting to combine.

eeintech:
@jlaw Ha right! I kinda lost track of the initial idea :sweat_smile:

SchrodingersGat:
[...]

PR #9312, opened 2025-03-16
[Feature] External build order

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/1071 (4.5 years later...)

This PR implements the concept of an "external" build order

- Build Order and Purchase Order are linked
- Stock is consumed against build order
- Purchase order line items are received into the build order (as outputs)


SergeoLacruz:
-Build Order and Purchase Order are linked
I think this is the most important feature. It reflects our way of working. Actually we just leave comments in PO and BO :-(

-Stock is consumed against build order
Is there a difference from the way it works now? Stock is already consumed.

-Purchase order line items are received into the build order (as outputs)
Means when I receive the line items, BO output is automatically created and I cannot do this in BO? This is not so important, but will ease work and reduce errors :-)

- Add “external” flag to build order
Nice because we also have internal ones :-)

- Add link to build order in line item row
Does that mean each PO line item can be linked to a different BO?  I understand from the other features that on PO is linked against on BO which is quite OK. The other way might make sense. I can imagine multiple POs for one BO in case the volume gets high or we distribute the BO over a longer time frame. 100pcs now, 500pcs a month later. But another BO would also be fine.

PR #9316, opened 2025-03-16
[refactor] Remove django-weasyprint entirely

SchrodingersGat:
- Follow up to https://github.com/inventree/InvenTree/pull/9307
- Previous fix attempt was not sufficient
- Mocking the request object was still causing issues

Upon deeper inspection, now that we offload all printing to the background worker, we no longer have any benefit from the django-weasyprint library. This layer forces us to pass a "request" object, and if this is not present, errors out.

So, remove it entirely! This PR now uses lower level weasyprint calls, and a new import - [pypdf](https://pypi.org/project/pypdf/) to stitch the PDF outputs together

### DIFF
--- a/docs/docs/extend/plugins/label.md
+++ b/docs/docs/extend/plugins/label.md
@@ -148,7 +148,6 @@ class MyLabelPrinter(LabelPrintingMixin, InvenTreePlugin):
         Send the label to the printer
 
         kwargs:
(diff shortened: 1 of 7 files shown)


matmair (approved):
This does work without a background worker present?

SchrodingersGat:
> This does work without a background worker present?

Yes, works the same with or without the background worker process. Except that it blocks if it runs in the foreground process :)

PR #9318, opened 2025-03-17
Schema: Remove hardcoded enum values from customizable fields

1337joe:
In testing against the demo dataset I ran across a handful of enum types in the schema that accept custom values, which of course breaks schema validation because the custom values weren't enumerated at schema compile time.

I did not change the serializer behavior, I simply added `extend_schema` annotations to tell the schema builder to not treat them as enumerable choices.

Further work on #9045


matmair:
Looking at the schema diff, we are losing a lot of information about formatting (currencies are always three letters capitalised, for example), sample values, etc.
Can this info be retained? Not everybody will generate a client - some will just look at the docs / swagger UI.

1337joe:
There are still enum status fields alongside the status_custom_key fields that I stripped the enums off of, but I agree it's less clear what the currency field accepts when it's just a string.

It looks like the `label` and `help_text` fields on the serializers pass through to the web interface, so the only way I can see to add more detail is to annotate in OpenApiExamples, which are at the operation level. I'll try to look into that when I have time.

matmair:
Have you checked if this can be achieved by changing the annotation of those fields?  Maybe together with the schema generator we can find a more integrated approach.
The status fields already are custom fields so we might be able to leverage that too here. Maybe we just strip out the pattern at schema generation.

1337joe:
I think I do have a better way to do this, after figuring out you can use `@extend_schema_field` with a serializer in my last PR. I was planning to make time to revisit this in the next day or two.

PR #9323, opened 2025-03-17
[WIP] Frontend lib

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/8594

This PR is an attempt to refactor the current frontend libraries, allowing us to distribute core UI components via npm.
This  will allow much easier development of plugins with complex UI functionality.

### Structure

- Separate core components out into "lib" directory
- Keep anything not "generic" in the "src" directory

### TODO

Need to work out the following problems

- [x] How to build and distribute the "lib" components (into "dist")
- [x] How to also build the InvenTree UI (using "src" and "lib") - this should be the *default*
- [ ] Remove external packages from the lib mode
- [ ] Distribute to npm
- [ ] Auto-release as part of InvenTree release process
- [x] Investigate [ts path aliases](https://maxmartinez.dev/how-to-configure-tsconfig-path-aliases)
- [x] Split `vite.config` file into `lib` and `src` builds
- [ ] Validate / investigate tree-shaking for library mode
- [ ] Validate / investigation translation within plugins
- [ ] Expose plugin "interface" type to the plugin generator code
- [ ] Default usage of lib for plugin generator
- [ ] Ensure front-end translation still works - even for components in `./lib/`
- [ ] Add frontend README.md describing architecture
- [ ] Update plugin-creator to use support new libs
- [ ] vite coverage must work against both `./lib` and `./src` dirs


SchrodingersGat:
@wolflu05 @matmair before I continue any further with this, let's discuss if this is an appropriate way to tackle this.

In particular, the `// TODO` entries in `vite.config.ts`

src/frontend/lib/components.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  ActionButton,
+  type ActionButtonProps
+} from './components/buttons/ActionButton';
+
+import { Boundary } from './components/Boundary';
+import { ButtonMenu } from './components/buttons/ButtonMenu';
+import { CopyButton } from './components/buttons/CopyButton';
```
wolflu05, line 8: Just as a shortcut, we could also use 
```suggestion
export { CopyButton } from './components/buttons/CopyButton';
```

src/frontend/src/components/barcodes/QRCode.tsx
wolflu05, line 1: That is an interesting question for discussion, where do we draw a line and make a component available in the npm package? I feel like this is not so clear, e.g. this `InvenTreeQrCode` component may be useful for plugins who just want to show a qr code. E.g. for some order simplification stuff, ... (havn't something specific in mind)

src/frontend/vite.config.ts
```diff
@@ -51,8 +52,36 @@ export default defineConfig({
...
+        core: resolve(__dirname, 'lib/core.ts'),
+        functions: resolve(__dirname, 'lib/functions.ts'),
+        tables: resolve(__dirname, 'lib/tables.ts')
+      },
+      name: 'InvenTree',
+      formats: ['es']
+    },
+    rollupOptions: {
+      // TODO: This should be updated for all extrenal deps
+      external: ['React'],
```
wolflu05, line 71: This is a tricky question. I would vote for trying to minimize the external dependencies for the exported components as much as possible and maybe only mark thinks like `react`, `@mantine/*`, `tanstack-query` as external.

src/frontend/vite.config.ts
```diff
@@ -51,8 +52,36 @@ export default defineConfig({
...
+      output: {
+        globals: {
```
wolflu05, line 73: Does this really work that way? Not sure about versions. I would vote for letting the user install its own dependencies, and mark these all as peerdependencies in our package.json. 

E.g. dependency version mismatches between an installed inventree PUI version and the plugin's version of the inventree UI package. This will limit the supported inventree version of a plugin drastically.

src/frontend/vite.config.ts
```diff
@@ -51,8 +52,36 @@ export default defineConfig({
   base: '',
   build: {
     manifest: true,
-    outDir: '../../src/backend/InvenTree/web/static/web',
-    sourcemap: is_coverage
+    // outDir: '../../src/backend/InvenTree/web/static/web',
+    // TODO: outDir should be set to the correct path,
+    // TODO: depending on whether we are building library or UI?
```
wolflu05, line 57: Maybe we can split up the vite config files into multiple or use some arguments supplied by the defineConfig function if we pass a function there to determine the target. ref: https://vite.dev/config/

matmair:
very cool to see movement on this; I think this is an appropriate way to handle the issue;

Not sure how to securely handle publishing in ci/cd and how versioning will work but I will start to read up on those topics.

wolflu05 (commented):
Thank you very much for stepping up and starting the work here. I was at the same point a while ago where I realized its not an easy task, especially with our very complex and tangled code (some generic components import lots of things and dependencies which results in large bundles). As you already found out (and I think this will be the best way - instead of auto exporting everything with some automatic scripts), we need to create a sparate lib folder. The question now is if we only just reexport everything there, or if we move the components over and import them from the lib folder in the codebase. 
1. From what you did, I think you have decided for the second approach right? (No strong opinion from my side here)
2. I guess you only did that for a few components for demonstration unless we have aggreed on something right?
3. We need to see if tree shaking works later. Its not a huge win if someone only uses the  CopyButton and gets the complete datatable code, all tabler icons, ... into their plugin bundle as well which is not needed.
4. We need to work out how translations may work with plugins. Especially with our now provided components used inside of a plugin. 

Additionally it may be useful to look into TS path aliases so that we can avoid `../../../../../...` imports (ref: https://maxmartinez.dev/how-to-configure-tsconfig-path-aliases, not sure about how that works out if we also publish this as an npm package as planned, havent used path aliases with a published package before)

These are just my initial thoughts after taking a first quick review. I really appreciate your work here. Thats a huge step forward to actually releasing v1.0

src/frontend/lib/components.ts
SchrodingersGat: Ok, didn't know that - thanks!

src/frontend/vite.config.ts
SchrodingersGat: Ok, that sounds pretty reasonable :)

src/frontend/vite.config.ts
SchrodingersGat: > Does this really work that way?

🤷 

What I *want* to happen is that we require the user to install their own deps (as you say), and not bundle any external large libs (such as react / mantine) into the distributed libs. I *think* that is what you are also saying here?

SchrodingersGat:
Thanks guys for the early feedback.

> From what you did, I think you have decided for the second approach right? (No strong opinion from my side here)
> I guess you only did that for a few components for demonstration unless we have aggreed on something right?

Yes, I have selected a few "simple" components to try first - basically the ones with zero other interdependencies in the other UI code.

The plan would be that we can discuss which components get exposed to the library. Any "structural" component that does not ever get "reused" will *not* be in here - for example the "global search drawer" is not something we would expose to plugins.

> We need to see if tree shaking works later. Its not a huge win if someone only uses the CopyButton and gets the complete datatable code, all tabler icons, ... into their plugin bundle as well which is not needed.

Yep, this will need to be tested.

> We need to work out how translations may work with plugins. Especially with our now provided components used inside of a plugin.

Good point - no idea how this might work at this stage.

SchrodingersGat:
Ok, bed time. Keen for another review if you have the chance :)

src/frontend/src/components/barcodes/QRCode.tsx
SchrodingersGat: We'll have to assess the borderline components on a case by case basis. Some are going to be quite obvious I think.

SchrodingersGat:
This is turning into a huge effort but is looking promising! A lot of refactoring to do to get a generic forms / tables interface - but I am getting there :)

SchrodingersGat:
Alright I think I am at the limit of what I can achieve here without some insight :)

I have the core components separated out into `./src/frontend/lib` - these are the generic components which are exposed to plugins.

To build the "library" components, run `yarn library`

The components used to build the specific frontend UI are still in `./src/frontend/src`

To build the entire UI, run `yarn build` as per normal.

However, I am having trouble importing the "library" components into a separate plugin.

I am importing the command  `npm link inventree` to make the library available within the plugin, and I can then use the following code within the plugin file:

```typescript
import { useCreateApiFormModal } from 'inventree/forms';
import { apiUrl } from 'inventree/functions';
```

And it even compiles!

However when I try to run it:

```
PluginSource.tsx:18 ERR: Failed to load plugin from http://localhost:8000/static/plugins/wireviz/WirevizPanel.js: TypeError: Failed to resolve module specifier "react". Relative references must start with either "/", "./", or "../".
```

wolflu05:
> The intent is that the plugin code does not bundle react / react-dom - right?

I am not sure if this works (havent tried this yet). Especially due to version mismatches. What definetly should work is:

1. remove the globals from your vite configs (inventree and plugin) and remove the external also from your plugin, but keep it in the lib.
2. then install your version of react into the plugin dir
3. the build should work then

--- 

I recently tried to make the react dev server inside a plugin panel to work with react, but I havent found a solution. Thats why I am using preact for my plugins. Thats a leightweight implementation of react that is compatible 1:1 with react when using the compat layer.

SchrodingersGat:
@wolflu05 interesting - could we use preact for plugins in general, and still make use of the exported components in this library?

wolflu05:
In theory yes. My template editor (the graphical one) also uses preact and mantine. I wouldnt see why this wont work with our custom lib then.

SchrodingersGat:
If you have some small amount of time, I would love to see if you can get one of our exported components (such as `<StylishText>` or `<ProgressBar`> working with live HMR into the main InvenTree UI, using preact.

I am banging my head against the wall trying to get it working (with react). Steps I am following below:

1. Pull down this branch
2. `cd ./src/frontend` and `yarn library`
3. Export lib globally with `npm link`
4. Create a new simple react app with `npm create vite@latest`
5. Import lib components with `npm link inventree`
6. Try to render component as shown below
7. Get error as shown below

### Code Snippet From "Plugin"

```typescript
import { ProgressBar } from 'inventree/components';
import { ApiEndpoints } from 'inventree';
import { apiUrl } from 'inventree/functions';
import { MantineProvider } from '@mantine/core';

function App() {
  const [count, setCount] = useState(0)

  const url = apiUrl(ApiEndpoints.attachment_list, 23);

  return (
    <MantineProvider>
      <div>
        <p>URL: {url}</p>
        <ProgressBar value={50} />
```

### Error

image

### Page with <ProgressBar> removed

```typescript
function App() {
  const [count, setCount] = useState(0)

  const url = apiUrl(ApiEndpoints.attachment_list, 23);

  return (
    <MantineProvider>
      <div>
        <p>URL: {url}</p>
        {/* <ProgressBar value={50} /> */}
        <Text>Hello world!!</Text>
        <a href="https://vite.dev" target="_blank">
          <img src={viteLogo} className="logo" alt="Vite logo" />
        </a>
        <a href="https://react.dev" target="_blank">
          <img src={reactLogo} className="logo react" alt="React logo" />
        </a>
      </div>
```
image

So clearly the imports from the lib are kinda working now, but the "mantine" components are not.

However, if you can get it working with preact, *and* we can use HMR for plugins, that would be great - and I'd suggest we update the plugin template to use preact instead

SchrodingersGat:
My attempt at preact:

```typescript
import { render } from 'preact';

import preactLogo from './assets/preact.svg';
import './style.css';

import { ApiEndpoints } from 'inventree';
import { apiUrl } from 'inventree/functions';
import { ProgressBar } from 'inventree/components';

import { MantineProvider, Text } from '@mantine/core';


export function App() {

	const url = apiUrl(ApiEndpoints.build_line_list, 22);

	return (
		<div>
			<h1>Hello, Preact: URL = {url}</h1>
			<Text size='lg'>Some mantine text!!!</Text>
			<ProgressBar value={99} />
			<a href="https://preactjs.com" target="_blank">
```

image

image

Progress, of sorts, but not sure about this error now

SchrodingersGat:
I'm going to have to put this aside for now, getting very frustrated with it... If you can work out the next piece of the puzzle that would be amazing :)

wolflu05:
That is looking pretty good now. Great that you got through. I also got this error onetime (the __H). A full page reload fixed it for me. If not try to restart the plugin dev server.

I'll see what I can do the next days, currently messing with my LSP... it's looking pretty promising now.

SchrodingersGat:
> A full page reload fixed it for me. If not try to restart the plugin dev server.

No luck on my end I'm afraid! Hopefully you can suggest something :)

wolflu05:
Ok, I kinda got it to work with react (vite + react + typescript + swc) now (I gave up on preact, understanding what is causing issues in the compat layer seems to be too difficult).

The issue with the MantineProvider you were seeing is due to the mantine version bundled into the inventree lib is a different from the one installed locally for the plugin. We need to mark mantine as external (and possibly other mantine libs. These then should be marked as [peerDependencies](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#peerdependencies), also dont forget to mark react and react-dom as peerDep):

`vite.lib.config.ts`
```diff
         },
-        external: ['react', 'react-dom']
+        external: ["react", "react-dom", "@mantine/core", "@mantine/hooks"],
       },
       lib: {
```

This means we have to install `@mantine/core`, ... into the plugin. Also make sure, that youre using react@18 in the plugin as long as we are on react@18 with inventree. 

A feature function can then look like (it would make sense to extract this mantineprovider...):

```tsx
// main.tsx
export const getReportEditor = async ({ inventreeContext, featureContext }: any) => {
  const { createRoot } = await import('react-dom/client');
  const { StrictMode } = await import('react');
  const { MantineProvider } = await import('@mantine/core');
  const { App } = await import('./App.tsx');

  const root = document.createElement('div');
  featureContext.ref.innerHTML = "";
  featureContext.ref.appendChild(root);

  createRoot(root).render(
    <StrictMode>
      <MantineProvider
        theme={inventreeContext.theme}
        forceColorScheme={inventreeContext.colorScheme}
        getRootElement={() => featureContext.ref}
      >
        <App />
      </MantineProvider>
    </StrictMode>
  );
}
```

And lastly the most tricky part for react refresh to work:

In dev mode, we need the following static js (make sure to adjust your correct port):

```tsx
// static/reportEditor.dev.js
import { injectIntoGlobalHook } from "http://localhost:5212/@react-refresh";
injectIntoGlobalHook(window);
window.$RefreshReg$ = () => {};
window.$RefreshSig$ = () => (type) => type;

import "http://localhost:5212/@vite/client";
export { getReportEditor } from "http://localhost:5212/src/main.tsx";
```

We need the async function and async import because they can only be done after the `injectIntoGlobalHook` function is called, but imports dont work in order they are defined, there doesnt run that code inbetween.

Then we can set the feature source in the plugin to:

```py
self.plugin_static_file("reportEditor.dev.js" if is_dev_for_this_plugin else "dist/main.js")
```

It would be nice if we could build some generic system to let the dev set its own port where the plugin server is listening/use the dist if not in dev for this plugin.

I then tried to bundle it to test the prod build. If youre still seeing errors like `Uncaught TypeError: Cannot read properties of null (reading 'useState')` after running `yarn library` and used `npm link` and `npm link inventree` in your repo, thats due to some conflicts with different react copies. You have to use `yarn pack` and then `npm install /path/to/tarball/inventree-v0.1.0.tgz`. Then finally it should show render the UI in prod mode.

This setup is not really ideal as the async imports for prod are not a good idea. But I havent found a better solution yet. Maybe you have an idea now...

SchrodingersGat:
@wolflu05 using your suggestions above, I have at least got it rendering... something... However not really what I expected.

### Code

image

### Output

image

The "size" attribute of the `<StylishText>` does not appear to have any effect - and I currently have no luck getting other components to render.

### Form Issues

Trying to get a form running is even more  difficult. I had to put `lingui` into peer dependencies in "inventree" too. And then wrap 
my code in more providers:

image

When I try to "open" the form:

To be honest this is feeling extremely convoluted, I think we are straying far from something that would be simple to implement, and keep in sync with the code running on the server.

wolflu05:
FYI, I have found some solution about the async thing, as we need this either way:

```
// static/reportEditor.dev.js
import { injectIntoGlobalHook } from "http://localhost:5212/@react-refresh";
injectIntoGlobalHook(window);
window.$RefreshReg$ = () => {};
window.$RefreshSig$ = () => (type) => type;

export const getReportEditor = async (...args) => {
    await import("http://localhost:5212/@vite/client");
    const { getReportEditor } = await import("http://localhost:5212/src/main.tsx");
    return getReportEditor(...args);
};
```

Then we can do everything normal in the `src/main.tsx:gerReportEditor` function and dont need any weird async things.

SchrodingersGat:
Closing - replaced by https://github.com/inventree/InvenTree/pull/9439

PR #9332, opened 2025-03-17
[UI] Fix StylishText

SchrodingersGat:
- Use primary color for StylishText


matmair (approved):
We are losing the fancy gradient but that is probably not possible dynamically

Issue #9256, opened 2025-03-07
[FR] Add oAuth2 for API access


matmair:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

Our current token system always inherits the permissions from the current user. This makes tokens possibly pretty dangerous. You also need to refresh tokens by hand when they expire. 

### Suggested solution

There is a known and widely used protocol to solve these issues: oAuth2. After a few experiments, it seems to integrate pretty nicely with our current stack (mapping the role system).

As a breaking step, I would recommend that we later on remove basic auth and amend the token auth model with the same scopes. Similar to how GitHub handles this.

### Describe alternatives you've considered

-

### Examples of other systems

Microsoft, Google, GitHub all use oAuth2

### Do you want to develop this?

- [x] I want to develop this.

matmair:
See https://github.com/inventree/InvenTree/issues/4562#issuecomment-2704812478

matmair:
In testing how complex this could be I made https://github.com/matmair/InvenTree/pull/308 which is probably already 80% there.

PR #9333, opened 2025-03-17
feat(backend): add oauth2

matmair:
This adds oAuth2 - a widely used technology that enables standardised auth flows, token managment and scoping for both users and machine2machine communication.

Currently, this is additional to already existing auth methods - although I think that it should replace at least basic auth in the future.
The required scopes are all documented in the API schema. This makes required permissions (scopes) a lot more transparent.

A few things are still unclear to me:

- [x] Scope structure: either composite scopes or Jira style [granular scopes](https://developer.atlassian.com/cloud/jira/platform/scopes-for-oauth-2-3LO-and-forge-apps/) --> added granular scopes
- [x] Should we enable (only, not together with oAuth2) OIDC - this requires key material handling
- [x] How integrated should application managment be? currently, it relies on the included templating based managment
- [x] Do we provide a migration path for inventree-python and ship a pre-made client similar to gitea, gitlab and similar for git credential manager?

Contains https://github.com/inventree/InvenTree/pull/9478
Contains https://github.com/inventree/InvenTree/pull/9479
Contains https://github.com/inventree/InvenTree/pull/9488

Closes https://github.com/inventree/InvenTree/issues/9256


matmair:
@SchrodingersGat any opinions on the questions above

matmair:
@p-fruck also intrested in your insights

p-fruck:
Great to see so much progress on this! As a personal suggestion, did you consider setting the oauth endpoints to `/auth/o/` or is it intended to open a new url path for this?

Regarding the implementation: Is this library actually OIDC compatible? To me it seems as if only oauth was supported. Are you planning to support the full OIDC protocol or only oauth?

>  Should we enable (only) OIDC - this requires key material handling

While I am a huge fan of OIDC, I think many admins like to use basic auth at least as a fallback, in case connectivity to the OIDC provider fails.

matmair:
> id you consider setting the oauth endpoints to `/auth/o/` or is it intended to open a new url path for this?

/auth are user authentication endpoints
/o are oAuth / OIDC (as a provider) endpoints - I would not mix the both as they are separate use cases



> Is this library actually OIDC compatible?

They seem to be https://django-oauth-toolkit.readthedocs.io/en/latest/oidc.html, it is not enabled in the code base right now



> While I am a huge fan of OIDC, I think many admins like to use basic auth at least as a fallback, in case connectivity to the OIDC provider fails.

I phrased that unclearly: is there value in having oAuth2 endpoints or should it only be OIDC (which is a bit narrower)

p-fruck:
> is there value in having oAuth2 endpoints or should it only be OIDC

Personally I would simply go with OIDC, as I do not see too much value in plain oauth. However I am not sure if plain oauth can be deactivated. From my understanding, an oauth2 client can always connect to an OIDC server/RP

SchrodingersGat:
What functional differences will this mean for auth flows for:

1. The "normal" web user interface - where users initially authenticate with username / password, and then receive a session token
2. SSO web user interface
3. Mobile app - username / password auth and then user token
4. inventree-python - username / password auth and then user token

matmair:
No changes to all of those, it is additional and on a whole new url path `/o/`. All other auth paths are not affected.

At some point I would suggest to replace the flows 3/4 with oAuth2 as that will enable unified usage of SSO and MFA for token aquisition.

matmair:
Finishing up cleanups / docs and a few additional e2e tests. Should be ready in days.

matmair:
There seems to be some issue with codecov but the rest is running through smoothly.
Ready for review and merge @SchrodingersGat

src/backend/InvenTree/InvenTree/settings.py
```diff
@@ -523,19 +532,30 @@
     }
 ]
 
+OAUTH2_PROVIDER = {
+    # default scopes
+    'SCOPES': oauth2_scopes,
+    # OIDC
+    'OIDC_ENABLED': True,
```
SchrodingersGat, line 539: Do we want to set the `OIDC_ENABLED` attribute based on a run-time configuration parameter e.g. in the config file or env var?

matmair: It is; while it is enabled here, the urlpatterns are flagged. Requiring them to be explicitly be turned on to work.

I also added docs regarding the flags and activation as part of this PR.

SchrodingersGat (approved):

matmair:
@SchrodingersGat merge conflict is addressed now

PR #9334, opened 2025-03-19
[UI] Web Prefix

SchrodingersGat:
- Change "platform" to "web"
- More "standard" prefix
- Lock in before move to 1.0.0

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -549,7 +549,7 @@ jobs:
           invoke migrate
 
   platform_ui:
-    name: Tests - Platform UI
+    name: Tests - Web UI
     runs-on: ubuntu-20.04
     timeout-minutes: 60
     needs: ["pre-commit", "paths-filter"]
@@ -628,7 +628,7 @@ jobs:
           yarn run build
 
   platform_ui_build:
-    name: Build - UI Platform
+    name: Build - Web UI
     runs-on: ubuntu-20.04
     timeout-minutes: 60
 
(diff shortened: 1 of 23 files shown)


SchrodingersGat:
@matmair @wolflu05 any strong preferences here? I would like to make the web prefix more "standard" before we lock in to the 1.0.0 release

matmair:
I would prefer if at least the internal names are technology specific so that bigger migrations like CUI -> PUI are easier the next time around

wolflu05:
Is there a reason, why we can't remove this prefix at all?

SchrodingersGat:
> Is there a reason, why we can't remove this prefix at all?

I'm not opposed to removing it - @matmair any strong reasons to keep?

SchrodingersGat:
> I would prefer if at least the internal names are technology specific so that bigger migrations like CUI -> PUI are easier the next time around

What do you mean by "internal names"?

matmair:
Removing the prefix in the routes is a bad idea, that makes routing stuff in the reverse proxy and then in possible WAFs in front a lot more difficult / unstable without gaining anything. Differentiating the backend and frontend is a lot harder that way.
Also makes subdirectory hosting harder to detect.



> What do you mean by "internal names"?

URL names, variable names, descriptions. Everything a normal user does not touch

SchrodingersGat:
I will not remove the prefix, that's fine :)

@matmair not quite sure if there's any internal names changed here - most of the strings were hard-coded as "platform" around the code. LMK if there's anything you think should be tweaked further.

PR #9346, opened 2025-03-19
[UI] Icons in navbar

SchrodingersGat:
Reintroduces *optional* icons in the navigation bar. Default = OFF (no icons). Per-user configurable.

image

### DIFF
--- a/src/backend/InvenTree/common/setting/user.py
+++ b/src/backend/InvenTree/common/setting/user.py
@@ -183,6 +183,12 @@ USER_SETTINGS: dict[str, InvenTreeSettingsKeyType] = {
         'default': False,
         'validator': bool,
     },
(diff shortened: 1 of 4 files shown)


src/backend/InvenTree/common/setting/user.py
```diff
@@ -183,6 +183,12 @@ def label_printer_options():
         'default': False,
         'validator': bool,
     },
+    'ICONS_IN_NAVBAR': {
```
matmair, line 186: Q: Was this configurable in CUI? If not I would just make it the same as before

matmair (approved):

wolflu05:
That's funny, was thinking about adding the exact same thing today. 🤣

Maybe we could add some way to make it more visually clear what icon belongs to what text, but not sure about a specific solution for this.

PR #9350, opened 2025-03-20
Create allauth email from django_auth_ldap if configured

jacobfelknor:
If the InvenTree instance has LDAP configured, the email obtained from LDAP and attached directly to the Django User object should also get an allauth `EmailAdress` object. 

Only do this when the user is first created. I currently provide no restrictions for removing this email, or means of updating it over time, which could be future improvements.

Inspired by discussion at https://github.com/inventree/InvenTree/pull/9116#issuecomment-2741302883


matmair (approved):
LGTM

Issue #9353, opened 2025-03-20
Contact for Order Detail Copies PK


jacobfelknor:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

The contact field on an order shows the contacts name and a copy button. If you try to copy the value, it copies the PK of the contact object

Additionally, it would be helpful to view the contact's email/phone if it exists

### Steps to Reproduce

1. Go to a Purchase Order
2. Add a contact if it doesn't exist
3. Try to copy the name
4. Observe it copies a number

### Expected behaviour

1. Go to a Purchase Order
2. Add a contact if it doesn't exist
3. Try to copy the name
4. Observe it copies the name

### Deployment Method

- [ ] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

InvenTree-Version: 0.18.0 dev
Django Version: 4.2.20
Commit Hash: cb240e4
Commit Date: 2025-03-20
Commit Branch: master
Database: postgresql
Debug-Mode: True
Deployed using Docker: False
Platform: Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.36
Installer: GIT
Active plugins: [{"name":"InvenTreeBarcode","slug":"inventreebarcode","version":"2.1.0"},{"name":"BOM Exporter","slug":"bom-exporter","version":"1.0.0"},{"name":"InvenTree Exporter","slug":"inventree-exporter","version":"1.0.0"},{"name":"InvenTreeCoreNotificationsPlugin","slug":"inventreecorenotificationsplugin","version":"1.0.0"},{"name":"InvenTreeCurrencyExchange","slug":"inventreecurrencyexchange","version":"1.0.0"},{"name":"InvenTreeLabel","slug":"inventreelabel","version":"1.1.0"},{"name":"InvenTreeLabelMachine","slug":"inventreelabelmachine","version":"1.0.0"},{"name":"InvenTreeLabelSheet","slug":"inventreelabelsheet","version":"1.0.0"},{"name":"DigiKeyPlugin","slug":"digikeyplugin","version":"1.0.0"},{"name":"LCSCPlugin","slug":"lcscplugin","version":"1.0.0"},{"name":"MouserPlugin","slug":"mouserplugin","version":"1.0.0"},{"name":"TMEPlugin","slug":"tmeplugin","version":"1.0.0"}]

[...]

PR #9354, opened 2025-03-20
FIx contact copy, add email/phone if they exist

jacobfelknor:
Addresses #9353

It looks to me that maybe you guys were/are planning on adding a dedicated Contact detail page, given that this was given the "link" type and a URL for this model type existed (`/contact/<pk>/`). Given that, I'm open to different ways of handling this if you don't like me changing this field's type.

However, this should address my user's complaints that they can't view/copy the contact info directly from the Order page


SchrodingersGat (approved):

SchrodingersGat:
@jacobfelknor nice implementation :)

> It looks to me that maybe you guys were/are planning on adding a dedicated Contact detail page

There is no plan for this - it works better just in table form. Probably we should delete that placeholder!

PR #9359, opened 2025-03-21
Fix req file

matmair:
Fix for https://github.com/inventree/InvenTree/pull/9358
Also includes a change to the ci/cd workflows to catch these issues in the future


SchrodingersGat:
Thanks for the fix @matmair

PR #9365, opened 2025-03-23
[UI] Alerts

SchrodingersGat:
Adds system for displaying critical alert information to staff users.

Currently, only displays alert if the background worker is not running:

image


matmair (approved):
LGTM;
Should we add an alert for running in dev mode?

SchrodingersGat:
> LGTM; Should we add an alert for running in dev mode?

Nice idea, I've added some more alert types now

PR #9373, opened 2025-03-24
Show part revision in Build Details panel

### LINKED ISSUE #9343: [FR] Show part revison in Build Details

### Problem statement

The build details panel includes the part Name and IPN fields, If the part has a revision assigned this is also required to identify the specific part to be built and would be good to see also.

Image

### Suggested solution

Show also the revison for parts where it is used.

Image

As I belevie this is a fairly limited change I have implemented this to familiar myself with the frontend code.

If you think this could be useful also for others I would be happy to create a PR.

Follow the part link and check the revision there but it takes a few more clicks.

### Examples of other systems

_No response_

- [x] I want to develop this.

gunstr:
Fixes #9343 

Screenshot 2025-03-24 222021

### DIFF
--- a/src/frontend/src/pages/build/BuildDetail.tsx
+++ b/src/frontend/src/pages/build/BuildDetail.tsx
@@ -102,6 +102,14 @@ export default function BuildDetail() {
         copy: true
       },
       {
+        type: 'string',
+        name: 'part_detail.revision',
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@gunstr thanks for the contribution :)

PR #9389, opened 2025-03-26
[UI] Improve order parts wizard

SchrodingersGat:
- Enhance placeholder text
- Precalculate order quantity
- Ref: https://github.com/inventree/InvenTree/discussions/9379

### Screenshot

image


SchrodingersGat:
@tkuester some great ideas there - I would suggest that you can open new issues for each of these, as some of them will require some additional UI architecture.

tkuester:
> @tkuester some great ideas there - I would suggest that you can open new issues for each of these, as some of them will require some additional UI architecture.

Will do, thanks for the awesome work!

Issue #9345, opened 2025-03-19
Sales Order Report template - How to add created/completed date and customer reference field


drendeux:
### Body of the issue

I'm trying to add a few things to the sales order report template and can't quite figure out how to get them to appear. All the info I'm trying to add appears in the sales order details page in the top right corner.

I'd like to add the created and completed dates, as well as the customer reference field.   I haven't been able to find the documentation that identifies the fields that are available to access for the sales order itself, so adding these hasn't worked.

<img width="404" alt="Image" src="https://github.com/user-attachments/assets/08acb565-4c1b-4c23-b38d-4b35cf666513" />

drendeux:
Can this be reopened? Sergeo's response only answered half my question. I'm still needing to know how to add the customer reference field.

Also, the variables that were suggested order.creation_date and order.target_date aren't on the documentation page in relation to sales orders. There is a creation_date for parts, but nothing else.  Can that be added to the documentation along with any others that are available? This is the page that was referred: https://docs.inventree.org/en/stable/report/context_variables/

wolflu05:
You should get the customer reference via `order.customer_reference`. Sadly this is a very tedious task of keeping that documentation up to date... we are happy to receive pull requests improving it.

drendeux:
The order.creation_date is giving the correct field, but order.target_date is giving "None" for all sales reports, even those that have a completed date. Is there another field that pulls that completed date data?

SchrodingersGat:
> but order.target_date is giving "None" for all sales reports, even those that have a completed date. Is there another field that pulls that completed date data?

`target_date` refers to the target date. NOT hte completion date.

drendeux:
Yes, that's the problem I'm having. I asked originally for completion date variables and was told to use target date, which didn't work. What do I use to get completion date?

SchrodingersGat:
For `SalesOrder` the field is `shipment_date`

drendeux:
@SchrodingersGat  That works perfectly, thank you!

PR #9390, opened 2025-03-26
Add context variables for sales order to the docs

SergeoLacruz:
Asked in #9345
What to do with the section Orders at the end of the file? It seems to be redundant. We can merge this.


SchrodingersGat:
> What to do with the section Orders at the end of the file? It seems to be redundant. We can merge this.

The "TODO" entries in the "Orders" section can be removed entirely. Happy for you to do that as part of this PR.

Additionally, do you want to add in the extra date fields for the other order types?

SergeoLacruz:
Will do so

PR #9393, opened 2025-03-26
User Setting To Search Notes

jacobfelknor:
As requested by my users, add a user setting to include the notes field in searches.

Additionally tie these settings to the default values of the check boxes in the `SearchDrawer`. Changes you make via these check boxes in the drawer are temporary and do not modify your user search setting defaults

image

image

Add corresponding unit test


jacobfelknor:
unsure why my latest commit 14d9a04e937061d34acd17ef770e023e1c3a6cbb to this branch isn't making into the PR....

image

matmair (approved):

matmair:
Seems like something in the gh backend was not computing; I gave it a friendly push

SchrodingersGat (approved):

matmair:
@jacobfelknor could you take another look at the CI tests? They seem to indicate 1 failure

jacobfelknor:
hmm... I'm having trouble replicating the failure shown in CI/CD locally. Weird that it appears to have only failed under the SQLite db

matmair:
it might be one of many things - is there a db connection to the test that is failing?

jacobfelknor:
The test that failed in CI was the new one I wrote. It looked like there was a problem from the response, it wasn't that the counts were wrong. I'm going to see if the error persists after you just updated the fork

jacobfelknor:
I agree. I'm just not seeing the same failure locally so I'm not sure where to move from here. This screenshot from a debug session on the tests. I've also tried with the same command I see in the CI job, 

image

src/backend/InvenTree/InvenTree/test_api.py
```diff
@@ -353,6 +354,69 @@ def test_results(self):
         self.assertNotIn('stockitem', response.data)
         self.assertNotIn('build', response.data)
 
+    def test_search_filters(self):
+        """Test that the regex, whole word, and notes filters are handled correctly."""
+        SEARCH_TERM = 'some note'
+        RE_SEARCH_TERM = 'some (.*) note'
+
+        response = self.post(
```
SchrodingersGat, line 362: Due to how the API pagination works, you need to specify a `limit` parameter. Otherwise the results are *not* paginated. So, the "count" attribute is not returned, instead the results are just a flat list.

So you could do either:

### Option A: Paginate

```python
response = self.post(
    reverse('api-search'),
    {
        'search': SEARCH_TERM,
        'part': {},
        'build': {},
         'limit': 10,
    }
)
```

### Option B: Check Len

```python
        response = self.post(
            reverse('api-search'),
            {'search': SEARCH_TERM, 'part': {}, 'build': {}},
            expected_code=200,
        )

        self.assertEqual(len(response.data), 0)
```

jacobfelknor: Not that I'm aware of. I just added the limit parameter and it still passes locally, so I pushed. I will have to look into why I didn't need it locally, as it definitely wasn't intentional

src/backend/InvenTree/InvenTree/test_api.py
```diff
@@ -353,6 +354,83 @@ def test_results(self):
...
+            reverse('api-search'),
+            {'search': SEARCH_TERM, 'limit': 10, 'part': {}, 'build': {}},
+            expected_code=200,
+        )
+        # No build or part results
+        self.assertEqual(response.data['build']['count'], 0)
```
SchrodingersGat, line 368: I think these lines should be :

````python
self.assertEqual(response.data['count'], 0)
```

SchrodingersGat: Also, the checks are duplicated here for some reason

jacobfelknor: I meant to check the part results in one of the lines... so that was a typo that I will fix.

As far as omitting `'build'` from the indexing, why would it be different here than in the existing (passing) [`SearchTests.test_results()`](https://github.com/jacobfelknor/InvenTree/blob/user_setting_search_notes/src/backend/InvenTree/InvenTree/test_api.py#L312)? I used it as my example

```python
response = self.post(
    reverse('api-search'),
    {'search': 'chair', 'limit': 3, 'part': {}, 'build': {}},
    expected_code=200,
)

# No build results
self.assertEqual(response.data['build']['count'], 0)
```

jacobfelknor: To debug, I'm using the InvenTree tests profile in `launch.json` specifying `SearchTests`

```json
{
    "name": "InvenTree Server - Tests",
    "type": "debugpy",
    "request": "launch",
    "program": "${workspaceFolder}/src/backend/InvenTree/manage.py",
    "args": [
      "test",
      "InvenTree.test_api.SearchTests", // run only a specific test
    ],
    "django": true,
    "justMyCode": true
},
```

According to the [job log](https://github.com/inventree/InvenTree/actions/runs/14208049947/job/39809910686?pr=9393#step:7:1114), the `KeyError` doesn't happen until the last check on line 432. This means it passed the several previous checks using the same `response.data['build']['count']` style indexing

My debug output at that line looks like this:

image

At this point, I'm very stuck and confused,..... Surely I must be missing something, but I can't figure out what

SchrodingersGat:
Ok digging into this one, the combination of `search_regex` and `search_notes` is causing an error:

```
django.db.utils.OperationalError: user-defined function raised exception
```

Commenting out the "regex" component and it works OK:

image

It may well be related to this ticket - https://code.djangoproject.com/ticket/29500 - which would point to a limitation in sqlite

jacobfelknor:
That would explain why I didn't see it locally then since I was in postgres.... Thanks for your help

SchrodingersGat:
@jacobfelknor thanks for the contribution :)

PR #9395, opened 2025-03-27
[CI] Playwright improvements

SchrodingersGat:
An attempt to improve the reliability and speed of our playwright testing in CI.

This PR allows the playwright tests to be run in "production" mode - i.e. build the code first, and do not run the vite server in the middle. This should be faster and also more reliable when running in parallel. We are currently seeing a lot of errors due to vite server / HMR overload.

By default on a PR, the playwright tests will compile the code and run against that with multiple workers.

On a push (commit into master) the playwright tests will revert to "normal" (dev) mode, with the vite server in the middle, and a single worker. This is so that the code coverage runs properly. It will take longer, but only runs on a commit.

Additionally, makes some enhancements to the frontend code to ensure that a valid hostname is always resolved.


matmair:
How are we enforcing minimum coverage after this change

SchrodingersGat:
> How are we enforcing minimum coverage after this change?

Hmm good point. The intent here is to improve overall reliability of the playwright CI as it is wasting so much time having to re-run the tests multiple times on failure. Each run is also ~30 to ~40 minutes - so improving the test time would be good too.

Maybe we can sacrifice this minimum test coverage if we are still running coverage on an push? Happy to discuss options before I merge (assuming I get it working)...

SchrodingersGat:
Even with this "production" testing mode the playwright tests seem very fragile. Maybe we are just overwhelming the web server?

SchrodingersGat:
Awesome, playwright tests are down to 19 minutes now (from > 40)!

matmair:
Tbh not a fan of this change, we will see where coverage goes when it is not highlighted anymore and direct accountability can be assigned

SchrodingersGat:
@matmair happy to discuss how we can fix that too, if you have any ideas. The tests were becoming too fragile, each PR could take multiple test runs at > 45 minutes before it would succeed.

PR #9399, opened 2025-03-27
Use int PK instead of regex string in API endpoints

1337joe:
Currently, a number of endpoints in the API specify the path variable (that maps to a primary key) using regex to indicate a numeric value, which results in the following schema:
```xml
      parameters:
      - in: path
        name: id
        schema:
          type: string
          pattern: ^\d+$
        required: true
```

When generating code from the schema in Java (a strongly typed language) I get the following:
```Java
public SupplierPart companyPartRetrieve(String id) throws ApiException {
   ...
}
```

The problem here is that the primary key of a `SupplierPart` really is an `int`, which leads to an awkward bit of type massaging when calling the API to get details on a part:
```Java
...
SupplierPart listFirst = listQueryResult.get(0);
int supplierPartPk = listFirst.getPk();
SupplierPart result = api.companyPartRetrieve(Integer.toString(supplierPartPk));
```

This draft PR contains the changes to convert the Company API to use int path variables, which generates consistent integer primary keys in Java from this schema:
```xml
      parameters:
      - in: path
        name: id
        schema:
          type: integer
        required: true
```

Before I go through the other APIs making similar changes, is there a reason not to convert the endpoint definitions from `re_path` to `path` in this manner?


matmair:
Should be fine if it is ensured that only int based regex are replaced, we have paths that use other patterns too

SchrodingersGat (approved):

SchrodingersGat:
LGTM :)

1337joe:
I added the API version bump as well as the rest of the regex path variables that resolved to ints. Should be good to go now.

SchrodingersGat:
@1337joe nice work, thanks. Look like the "readthedocs" builds are broken, looking into that.

1337joe:
@SchrodingersGat It looks like readthedocs got a 403 (forbidden) on the API call to get releases from github. Is there an access token that expired or any way to re-trigger that build?

matmair:
looking into it

SchrodingersGat:
@1337joe looking OK now - please fix the conflicts here and we're good to go :)

1337joe:
@SchrodingersGat Done

SchrodingersGat:
Thanks for the contribution @1337joe :)

PR #9419, opened 2025-03-30
Revert some playwright changes

SchrodingersGat:
- Use vite dev server all the time
- Follow-up to https://github.com/inventree/InvenTree/pull/9395
- See if we can retain performance gains with the vite dev server still running


SchrodingersGat:
@matmair I have reverted the playwright tests to use dev mode for coverage. With the other improvements made, there is still a significant time saving (23 minutes vs > 40 minutes)

matmair:
Great, best of both worlds!

PR #9431, opened 2025-03-31
Typed report context

wolflu05:
This PR adds typing and descriptions to report context. These can be used by the documentation or external tools to introspect the available context.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,8 @@ ignore = [
     # - RUF032 - decimal-from-float-literal
     "RUF032",
     "RUF045",
+    # - UP045 - Use `X | None` instead of `Optional[X]`
+    "UP045",
 
     # TODO These should be followed up and fixed
     # - B904 Within an `except` clause, raise exceptions
--- a/src/backend/InvenTree/InvenTree/management/commands/export_report_context.py
+++ b/src/backend/InvenTree/InvenTree/management/commands/export_report_context.py
@@ -35,6 +35,8 @@ def get_type_str(type_obj):
     if type_obj.__module__ == 'builtins':
         return type_obj.__name__
 
+    print(type_obj)
+
     return f'{type_obj.__module__}.{type_obj.__name__}'
 
 
--- a/src/backend/InvenTree/InvenTree/management/commands/export_report_context.py
+++ b/src/backend/InvenTree/InvenTree/management/commands/export_report_context.py
@@ -8,7 +8,6 @@ without having to manually duplicate the information in multiple places.
 """
 
 import json
-from types import NoneType
 from typing import get_args, get_origin, get_type_hints
 
 import django.db.models
@@ -30,7 +29,7 @@ def get_type_str(type_obj):
 
         return f'{origin_str}[{", ".join(get_type_str(arg) for arg in get_args(type_obj))}]'
 
-    if type_obj is NoneType:
+    if type_obj is type(None):
         return 'None'
 
     if type_obj.__module__ == 'builtins':
--- a/src/backend/InvenTree/InvenTree/management/commands/export_report_context.py
+++ b/src/backend/InvenTree/InvenTree/management/commands/export_report_context.py
@@ -1,0 +1,138 @@
+"""Custom management command to export the available report context.
+
+This is used to generate a JSON file which contains all available report
+context, so that they can be introspected by the InvenTree documentation system.
+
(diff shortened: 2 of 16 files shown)


wolflu05:
This is ready for review.

pyproject.toml
```diff
@@ -56,6 +56,8 @@ ignore = [
     # - RUF032 - decimal-from-float-literal
     "RUF032",
     "RUF045",
+    # - UP045 - Use `X | None` instead of `Optional[X]`
```
matmair, line 59: Is this supported by 3.9?

src/backend/InvenTree/InvenTree/management/commands/export_report_context.py
```diff
@@ -0,0 +1,141 @@
...
+        # Report context models
+        for model in report_model_types():
+            model_key = model.__name__.lower()
+            model_name = str(model._meta.verbose_name)
+
+            if (
+                ctx_type := get_type_hints(model.report_context).get('return', None)
+            ) is None:
+                print(
+                    f'Error: Model {model}.report_context does not have a return type annotation'
```
matmair, line 113: Can we use an error code here?

src/backend/InvenTree/InvenTree/management/commands/export_report_context.py
```diff
@@ -0,0 +1,141 @@
...
+
+            attributes = parse_docstring(ctx_type.__doc__).get('Attributes', {})
+            for k, v in get_type_hints(ctx_type).items():
+                context['models'][model_key]['context'][k] = {
+                    'description': attributes.get(k, ''),
+                    'type': get_type_str(v),
+                }
+
+        if is_error:
+            raise CommandError(
```
matmair, line 132: Can we use an error code here?

matmair (approved):
Very cool; is this checking if the returned dicts match the typings?

SchrodingersGat:
@wolflu05 this is awesome!

On the "/docs/report/context_variables" page we have two separate headings for:

- Template Types
- Model Variables

I'm not sure if we need "model variables" any longer? Or should we add the same context data here?

src/backend/InvenTree/InvenTree/management/commands/export_report_context.py
wolflu05: I do not understand really what you mean with error code, can you please make an example?

pyproject.toml
wolflu05: Thats why I ignore them, otherwise it gives me lots of errors that I should replace `Optional[X]` with `X | None` which obviously doesnt work in py3.9 which we still support.

wolflu05:
> is this checking if the returned dicts match the typings?

If you have something like pyright with basic typechecking enabled, yes, this shows if something does not match, or a key is missing.

> I'm not sure if we need "model variables" any longer? 

I think we still need them, the context variables are just references to the part instance, queryset of stock items, ... but then we still need to access the part name via `part.name`, so we have to look at the part model doc and find name there. At some point we may can think about auto generating the model doc too, but thats some more work required.

SchrodingersGat:
> At some point we may can think about auto generating the model doc too, but thats some more work required.

I think this should be on our roadmap, once we have merged these very nice improvements.

What else is needed from the context of merging this PR?

wolflu05:
> What else is needed from the context of merging this PR?

From my side we can merge this, but not sure what @matmair means with error codes.

And yes, this is one of the last piece (with https://github.com/fourdigits/django-template-lsp/pull/103) to have a useable fully integrated template editor with context completions.

matmair:
@wolflu05 we added error codes a while back https://docs.inventree.org/en/latest/settings/error_codes/#inve-e1 using those might be helpful to make these errors easier to understand

matmair:
> > At some point we may can think about auto generating the model doc too, but thats some more work required.
> 
> I think this should be on our roadmap, once we have merged these very nice improvements.

This would be pretty easy following the model used for the export_tags / export_filters commands, most of the deps for that are already in the code

wolflu05:
But this command is just run during CI/docs generation like the other commands, normally a user shouldnt run this command. But if you feel like there still should be one. Please give me a hint how that error code should be named.

matmair:
The next error code is inve-e3; it is much more friendly to new devs to have Google-able error codes so I think we should add them in general for new code that raises errors

SchrodingersGat:
@wolflu05 thanks for the great contribution here

PR #9434, opened 2025-04-01
Template Editor Improvements

wolflu05:
This PR adds an unmount hook and serverContext to the report editor ui feature. Additionally it now supports custom icons for editors/previews using the ApiIcon.


wolflu05:
This is finally ready for review too, not sure what the issue with the test was, but now it works.

SchrodingersGat (approved):

SchrodingersGat:
@wolflu05 thanks for the update :)

PR #9439, opened 2025-04-02
[Plugin] Plugin context

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/pull/9323
- Closes https://github.com/inventree/InvenTree/issues/8594
- Allows plugins to access forms infrastructure

### Related

- Changes to plugin creator - https://github.com/inventree/plugin-creator/pull/35
- `wireviz`plugin migration - https://github.com/inventree/inventree-wireviz/pull/44
- `order-history` plugin migration - https://github.com/inventree/inventree-order-history/pull/22

### TODO

- [x] Finish refactoring components in `./src/frontend/src/` (importing from `./src/frontend/lib/`)
- [x] Expose context "types" via npm package - https://www.npmjs.com/package/@inventreedb/ui
- [x] Update plugin creator code - https://github.com/inventree/InvenTree/issues/8594
- [x] New release (1.3.0) of plugin-creator (once this is merged) - https://github.com/inventree/plugin-creator/releases/tag/1.3.0
- [x] Update documentation for plugin creation
- [x] Check custom dashboard items
- [x] Check custom panel items
- [x] Check custom editor items
- [x] Check it still works with "old" style plugin
- [x] Add system to warn if a plugin is used with a mismatched "plugin context version"
- [x] Provide plugin with a mechanism to "reload" the page instance (e.g. if viewing a part page)
- [x] Expose function to get "detail" URL for a given item (e.g. "part 3")


src/frontend/src/components/plugins/RemoteComponent.tsx
```diff
@@ -57,10 +62,23 @@ export default function RemoteComponent({
...
-              func(componentRef.current, context);
+              // Render the plugin component into the target element
+              // Note that we have to provide the right context(s) to the component
+              createRoot(componentRef.current!).render(
+                <ApiProvider client={queryClient} api={api}>
+                  <MantineProvider
+                    theme={context.theme}
+                    colorSchemeManager={colorSchema}
+                  >
+                    <LanguageContext>{func(context)}</LanguageContext>
```
wolflu05, line 73: Does this mean, we can't use any non-react plugins anymore? I would really not use this. Can't we just pass the createcontext function to the plugin? And also doesn't this break existing plugins?

SchrodingersGat: Will look into passing this through to the plugin.

SchrodingersGat:
Currently working:

- Launch an API form directly from a plugin
- Navigation functionality
- Render mantine components

image

Also: supports translation!

image

src/frontend/src/components/plugins/RemoteComponent.tsx
SchrodingersGat: @wolflu05 ok, I have patched it to support the "old" way of passing context through to the plugin

However, the *new* way is MUCH better and is required to allow any of these new features to work

src/frontend/package.json
```diff
@@ -1,11 +1,29 @@
 {
-    "name": "inventreeui",
+    "name": "inventree",
```
matmair, line 2: is it wise to name this like the project? I think it would be better to use the namespace that you secured and call it something similar to @inventree/lib

matmair (approved):

SchrodingersGat:
@wolflu05 this is getting close to merge ready - can you please check that it works properly with your custom editor plugin?

wolflu05:
Thanks for letting me know, I'll do in the next days.

However I'm still not sure if this is the right approach. Especially when trying to get react refresh to work. I would highly recommend to get that working first before committing to this approach. For some pointers (at least with the approach you would consider legacy), see my last few comments on the old PR.

And then, I won't name the non react way legacy, this implies that it's an old way and won't be supported anymore, but I still think having the ability to use my own framework is something we should support.

SchrodingersGat:
As far as I can tell the improvements made here do not limit or change the functionality of rendering fully externally (as was the case before). I'm happy to support a non-react approach, although I don't think I'll be implementing anything to guide users towards that - e.g. into the plugin creator tool.

Would you be happy with pushing support for "live plugin reload" off to another PR? I feel that would be a large challenge, and I would really like to get this integrated (as there are a number of blocking plugins that I cannot currently implement in the new UI without this)

wolflu05:
> although I don't think I'll be implementing anything to guide users towards that

Definitely, we do not need to provide support or advertise it anywhere.

> Would you be happy with pushing support for "live plugin reload" off to another PR?

If you are 100% sure, that react refresh will ever work if we just invest the required time, yes, but as I am not even sure if it will ever work with this approach, I would not recommend doing so. If I was you, I would at least like to have some kind of POC that it will work, otherwise we base everything on it, and may later realize that the house of cards collapses beneath us and we have to do a lot of work again to refactor everything. But thats your decision.

SchrodingersGat:
> If I was you, I would at least like to have some kind of POC that it will work, otherwise we base everything on it, and may later realize that the house of cards collapses beneath us and we have to do a lot of work again to refactor everything.

I can 100% see your argument here, and I would like this too. However I am finding it difficult to comprehend how to add this functionality in a way that is going to be workable for plugin developers. Or at all, really.

I would really appreciate if you can take a look (based on where this PR is at currently) at whether you can get your ideas working. TBH I don't think I follow your suggestions clearly enough to implement them myself

wolflu05:
Sorry, but I dont have any time right now to look into this.

But I'm really asking me what problem your new approach really solves that could not be solved with #9323. I mean, with both methods we need the providers as both are separate react trees?

wolflu05:
Yes, the bundles will obviously be smaller with this approach. But simpler setup might be subjective. I do not know how scalable this is when components or dependencies (e.g. react versions change). Maybe @matmair has some thoughts on this too?

And for the duplicate copy of react thing, have you made sure that you used yarn pack instead of linking like I described?

Anyway if you feel that's the right approach, go ahead. I unfortunately won't have much time in the coming months.

SchrodingersGat:
Please note that this new approach also leaves the old method available - plugin developers can take full control of the provided root element if they want to.

However if they do, then they will find it much harder to hook into the existing UI framework. I think that's a reasonable trade-off

Issue #9450, opened 2025-04-03
[FR] Disable ignoring warnings on schema generation


matmair:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

We ignore schema generation warnings, leading to suboptimal schema settings that need piecemeal fixes.

### Suggested solution

Recent improvements to the API schema have significantly lowered the number of warnings—it should be possible to remove the rest and remove the ignore flag to ensure no new warnings are introduced.

### Describe alternatives you've considered

N/A

### Do you want to develop this?

- [x] I want to develop this.

algora-pbc:
💎 **matmair** is offering a **$5** bounty for this issue. View and reward the bounty at `app.algora.io/inventree/InvenTree/issues/9450`

👉 Got a pull request resolving this? Claim the bounty by commenting `/claim #9450` in your PR and joining `app.algora.io`

matmair:
see https://github.com/inventree/InvenTree/issues/9045

PR #9452, opened 2025-04-03
Schema: (Redo) Remove hardcoded enum values from customizable fields

1337joe:
I redid my previous enum removal PR with more care taken to improve the schema documentation now that the enumerated values aren't listed in the descriptions.

Unfortunately, some of the `help_text` fields couldn't be overridden by using a serializer in the `@extend_schema_field` annotation, so I ended up adding extra schema-only text using checks against `ready.isGeneratingSchema()`. Let me know if those should be localized strings, but with the schema statically generated by github actions in English I wasn't sure it added value.

Brings warnings/errors down from 29/3 to 16/3 (unique) for #9450

Supercedes #9318

Further work on #9045


src/backend/InvenTree/generic/states/fields.py
```diff
@@ -138,14 +143,21 @@ def add_field(self, cls, name):
         if self.status_class:
             validators.append(CustomStatusCodeValidator(status_class=self.status_class))
 
+        extra_help_text = (
+            '. Can be any of the keys in the corresponding status field, as well as any defined custom keys for the status type.'
+            if InvenTree.ready.isGeneratingSchema()
+            else ''
+        )
```
1337joe, line 150: The python ternary operator feels awkward to me, is there a better way this should have been written?

matmair:
I have checked out the branch to compare it and it seems that we are still losing the descriptions and optional typing
grafik

src/backend/InvenTree/generic/states/fields.py
```diff
@@ -73,6 +77,7 @@ def get_field_info(self, field, field_info):
         return field_info
 
 
+@extend_schema_field(OpenApiTypes.INT)
```
matmair, line 80: This line removes all the choice descriptions; looking into if we can address this in schema generation

1337joe:
> I have checked out the branch to compare it and it seems that we are still losing the descriptions and optional typing

Yes, as soon as we tell the schema builder not to use a `ChoiceField` the enumerated list of options goes away. That's why I added to the description of each field information on where to find the valid values.

What is your preferred result? I was intentionally stripping it down to the choice result (int or string) because that's how the fields are defined. I tried adding the raw type into the `oneOf` list but checking the result against multiple possible types is a pain when what I actually can use is the raw string/int value.

matmair:
The description is what is getting rendered on the api-docs - I do not think the now added docs messags are any help to someone that tries to get started with integrating with InvenTree.

This singificantly lowers the value proporsition of the API docs and needs to be adressed.

1337joe:
Personally, as a consumer of the API when the valid values can't be enumerated in the docs I'd rather have an example and a reference to where to find the list of values instead of an incomplete enumeration of values. If I'm able to query this API I can easily query a different endpoint to retrieve the full list if I know where to look.

That said, I will work to add to the existing documentation instead of replacing it.

I've spent far too much time today trying to hack the `AutoSchema` to add documentation. Regardless of what mechanism I use to override parts during generation, the private `_map_serializer_field()` function generates the metadata containing the `description` text from the base, non-overridden field and accepts schema changes but nothing for `help_text`/`description`. That means I can't get the auto-generated list of values in the description if I'm not letting it generate as a `ChoiceField`, which is when the enum types we don't want get inserted into the schema. I can post-process them back out, but not in a way that feels clean.

@matmair I've commited fixes to restore the description text for currencies. If you think this is an acceptable way to do it I'll extend the pattern to the other fields in the next day or two.

1337joe:
Final documentation changes are as follows (choosing a single representative field for each):

### Currency
```diff
         currency:
-          allOf:
-          - $ref: '#/components/schemas/SalePriceCurrencyEnum'
+          type: string
           description: |-
             Default currency used for this supplier

             * `AUD` - Australian Dollar
             * `CNY` - Chinese Yuan
             * `EUR` - Euro
             * `USD` - US Dollar
+
+            Other valid currencies may be found in the 'CURRENCY_CODES' global setting.
```

### status_custom_key
```diff
         status_custom_key:
+          type: integer
           readOnly: true
           nullable: true
           title: Custom status key
           description: |-
             Additional status information for this item

             * `10` - Pending
             * `20` - Production
             * `25` - On Hold
             * `30` - Cancelled
             * `40` - Complete
-          oneOf:
-          - $ref: '#/components/schemas/StatusCustomKeyC2fEnum'
-          - $ref: '#/components/schemas/NullEnum'
+
+            Additional custom status keys may be retrieved from the corresponding 'status_retrieve' call.
```

### Stock status
```diff
         status:
+          type: integer
           description: |-
             Stock item status code

             * `10` - OK
             * `50` - Attention needed
             * `55` - Damaged
             * `60` - Destroyed
             * `65` - Rejected
             * `70` - Lost
             * `75` - Quarantined
             * `85` - Returned
             * `150` - Incoming goods inspection
-          oneOf:
-          - $ref: '#/components/schemas/Status229Enum'
-          - $ref: '#/components/schemas/BlankEnum'
+
+            Additional custom status keys may be retrieved from the 'stock_status_retrieve' call.
```

matmair:
This looks great! Is ist working with your validator?

1337joe:
> This looks great! Is ist working with your validator?

Yep, the generated code from this passes all the tests that previously failed due to custom status/currency values, and the documentation comments pass through into the generated Java classes. This will be one of the bigger steps towards being able to use the schema straight from the schema repo so far.

SchrodingersGat:
@matmair good to go?

matmair:
@SchrodingersGat yes, good to go!

SchrodingersGat:
@1337joe thanks for another great contribution :)

PR #9460, opened 2025-04-04
Fix icons API endpoint

1337joe:
Calling the /api/icons endpoint was throwing a server error:
`TypeError("'dict_values' object is not subscriptable")`

Quick 1-line fix to get the data in the format the serializer expects. No modification to the schema needed.


SchrodingersGat:
Thanks @1337joe

PR #9477, opened 2025-04-08
Tweak error display

SchrodingersGat:
Small visual updates


src/frontend/src/components/nav/Alerts.tsx
```diff
@@ -117,14 +137,3 @@ export function Alerts() {
     );
   return null;
 }
-export function errorCodeLink(code: string) {
```
matmair, line 120: Is there a reason to remove this? I was already starting to include this for error handling so that I do not have to repeate the url schema

SchrodingersGat: No, didn't mean to remove this. I'll get back to this soon.

PR #9478, opened 2025-04-08
feat(backend): api schema stats

matmair:
Adds a small stats section to the api schema generation to provde better visibility into the api paths (based on the schema).


SchrodingersGat (approved):

matmair:
see https://github.com/inventree/InvenTree/issues/9045

PR #9490, opened 2025-04-09
feat(backend): add importer role

matmair:
Adds an importer role to enable locking import sessions down


matmair:
Yes, that is the idea. I was recently reminded how fast someone can fill an instance if they do not know what they are doing

SchrodingersGat:
Ok, it seems reasonable - especially now that the UI for administering group roles is in place.

We should (as part of this) really enhance our documentation and role management:

- What roles are available
- What operations they control
- How users are assigned "roles" based on the groups they are in

Are you happy to write up some enhanced documentation as part of this?

matmair:
I can do that but it has low priority for me

PR #9492, opened 2025-04-10
Schema: Batch of small fixes

1337joe:
This branch started as a handful of small serialization changes to various endpoints, then got a lot more involved when I hit `VersionTextView`

**API Changes:**
- `AllUnitList` returns an object, not a list. Changed type to `RetrieveAPI` and fixed the serializer to match the data structure
- `LicenseViewSerializer`: licenses are lists of objects, not raw strings: fixed serializer to match data structure
- `VersionView`:
  - removed dead credits link (it was removed in #8811) and references to it
  - allowed `commit_branch` to be null to support detatched HEAD state (checking out a commit instead of branch)
- `VersionTextView`:
  - added missing query parameters
  - removed default pagination
  - fixed response to be a list instead of dict

In the course of testing `VersionTextView` I discovered a number of problems with the version text parsing, first of which was that the format used in `api_version.py` no longer is following the format that the parser expects. I converted a bunch of '-' to the expected '->' rather than changing the parser to handle both cases for simplicity.

I added tests to cover the changes to `version.py` to ensure that moving forward the schema change log file parses correctly.

I tested the Units, License, and Version changes in the UI, but I did not find anywhere that displays the VersionText information (which is probably how we managed to not have the log properly parsing for the last year).

Progress on #9045

### DIFF
--- a/src/backend/InvenTree/InvenTree/api.py
+++ b/src/backend/InvenTree/InvenTree/api.py
@@ -79,8 +79,8 @@ def read_license_file(path: Path) -> list:
 class LicenseViewSerializer(serializers.Serializer):
     """Serializer for license information."""
 
-    backend = serializers.CharField(help_text='Backend licenses texts', read_only=True)
-    frontend = serializers.CharField(
+    backend = serializers.ListField(help_text='Backend licenses texts', read_only=True)
+    frontend = serializers.ListField(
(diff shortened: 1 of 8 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@1337joe thanks again :)

Issue #7998, opened 2024-08-26
[FR] option to only match for IPN in search (or sort results to show direct IPN-matches on top)


simonkuehling:
### Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find a similar feature request

### Problem statement

When searching for a part that I know the IPN of, it is currently sometimes hard to quickly find it through the search panel for certain parts. As an example take an IPN like  ```100``` - which matches a lot of other parts that have a length of 100mm in the title as well. The part with the exact IPN is then often hard to find or way down the list...

### Suggested solution

A solution could be to either
- prioritize IPN matches in the sorting and put them right on top
or
- introduce an option like "only match IPNs" to the search options in the panel

### Describe alternatives you've considered

None that I can think of right now

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
@simonkuehling some enhancements around part search in particular would be good, I often find the same issues, as there are thousands of parts in the database and often finding exactly the one you want is tricky.

Not sure the most "user friendly" way to implement this

simonkuehling:
For context, here are two screenshots as examples. To correct myself, the Onshape filters do not have UI controls like github, but rather show as a tooltip for the search text field - also the actual syntax differs from github:

![grafik](https://github.com/user-attachments/assets/d398a645-002f-498d-96f5-0561dfcea143)


For comparison, here is the github implementation - where there are multiple methods of adding filters or sorting (through the filter dropdown or alternatively through the filter/sorting menu in the results table header)

![grafik](https://github.com/user-attachments/assets/f87ac13e-367f-4992-acc1-1a69e409ff3b)
![grafik](https://github.com/user-attachments/assets/84ac98d4-1a1e-4f87-8d85-80eee302a390)


One nice point to this type of filtering syntax is that the parsing can be done in the backend and the frontend does not need to construct complex queries and does not need to be touched if new parameters are introduced for example (same applies to the app). Also the minimalist Onshape-way is very quick to implement in the frontend as it is merely a tooltip or similar and no actual controls to design (which still can be added later if needed)

wolflu05:
I like the idea of introducing such a search syntax. But this will be a lot of work.

PR #9501, opened 2025-04-13
Refactored PartList to seperate functionality and added ipn sorting

phunprogrammer:
PR to solve #7998 and refactor filter_queryset() to seperate some responsibilities within.

Images shown below using example mentioned in original issue:
**Before:**
image

**After**
image

### DIFF
--- a/src/backend/InvenTree/part/api.py
+++ b/src/backend/InvenTree/part/api.py
@@ -1256,6 +1256,16 @@ class PartList(PartMixin, BulkUpdateMixin, DataExportViewMixin, ListCreateAPI):
 
         queryset = super().filter_queryset(queryset)
 
+        queryset = self.exlude_specific_ids(queryset, params)
+        queryset = self.filter_by_related_parts(queryset, params)
+        queryset = self.filter_by_category(queryset, params)
+        queryset = self.filter_parametric_data(queryset)
+        queryset = self.sort_by_ipn(queryset, params)
+
+        return queryset
+
+
+    def exlude_specific_ids(self, queryset, params):
         # Exclude specific part ID values?
         exclude_id = []
 
@@ -1275,7 +1285,10 @@ class PartList(PartMixin, BulkUpdateMixin, DataExportViewMixin, ListCreateAPI):
                     pass
 
             queryset = queryset.exclude(pk__in=id_values)
+        
+        return queryset
 
+    def filter_by_related_parts(self, queryset, params):
         # Filter by 'related' parts?
         related = params.get('related', None)
         exclude_related = params.get('exclude_related', None)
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@phunprogrammer an interesting approach here, I can see the appeal.

I would like to consider whether this can be made more generic (not tied specifically to the IPN field for example). It feels very specific to your particular requirements / use-case.

matmair:
@phunprogrammer please respond to above question so that we can move forward

phunprogrammer:
@SchrodingersGat Perhaps more functionality could be added to the search options that are already there. Currently, those options seem pretty limited to the three options that are available. A possible solution is to add some sort of selection in those options that could select a part parameter to sort by. Another solution would be to use `ipn:100` in the search to sort by IPN, the use of other tags would be made possible with this pattern. What solution do you think works best in this scenario?

SchrodingersGat:
We would need to consider how it works not just for searching against "Part" instances - as the search also looks for other model types (e.g. sales orders, stock items, etc).

Making this fully "flexible" or "configurable" will present the user with a complex UI to allow them to finely tune their expected results. I'm not immediately sure how best to go about it.

One thing from my persepctive - as we have added more search fields (in response to user requests, mainly) I have found that the search returns too many results, and not the ones I actually want. So I am on board with the goal of what you are proposing here. However I do not want to accept an implementation that is tailored to a specific use-case or preference.

The goal here should be to allow the user to intuitively control how the search results are returned. And also not to cause the back-end DB queries to be too complex and expensive.

Getting search "right" is very hard...

SchrodingersGat:
Thinking about this further, what if we made your approach a bit more generic. We already specify the set of "search_fields" for a given API endpoint.  We could utilize this to also define the "ordering" of the returned results. For example in the `PartList` API:

https://github.com/inventree/InvenTree/blob/afbcfe66bb7c615d74da418e6914ec3195414d4d/src/backend/InvenTree/part/api.py#L1247

we define the `search_fields`:

https://github.com/inventree/InvenTree/blob/afbcfe66bb7c615d74da418e6914ec3195414d4d/src/backend/InvenTree/part/api.py#L1401

```python
    search_fields = [
        'name',
        'description',
        'IPN',
        'revision',
        'keywords',
        'category__name',
        'manufacturer_parts__MPN',
        'supplier_parts__SKU',
        'tags__name',
        'tags__slug',
    ]
```

we could potentially create a custom search filter class (to replace the [DRF SearchFilter class](https://www.django-rest-framework.org/api-guide/filtering/#searchfilter)) which orders the results based on the specified ordering of the search fields.

This would apply the approach across all searchable API endpoints

SchrodingersGat:
If we wanted to go even further we could expose which fields are "searchable" on any given API endpoint - and then the client could potentially pass a list of fields to search when submitting a search. By default, all available fields are searched. This way, we could provision for further customization of search results in the future

PR #9476, opened 2025-04-08
[Refactor] Users and Groups

SchrodingersGat:
### Changes

These changes are necessary to allow user and group administration entirely via the UI

- Refactor user / group / permission management backend code.  No major functionality changes, just cleanup / re-organizing
- Provide enumeration of group roles via the API
- Allow group roles to be edited via the UI
- Allow user groups to be edited via the UI
- Enhanced user role / permission checking in UI
- Bug fixes for various permission checks in backend

### TODO

- [x] Fix over 9,000 errors in the failing CI 
- [x] Extend playwright testing for user permissions
- [x] Extend unit testing (ensure that updating a group applies correct permissions to user)
- [x] Check user-facing User and Group tables
- [x] Enable any authenticated user to "View" groups and users (but not edit / delete)

### Future Ideas
- Display the roles available to the current user on their detail page
- Display the groups the user is a member of on their detail page

### Group Role Editing

image
image

### User Group Selection

image

Issue #7466, opened 2024-06-17
[EPIC] Permission refactor


matmair:
The permission system currently uses a mixture of stock Django permissions, a self-written role-mapping engine and a ownership model for stock. This is not well documented and the behaviour is not consistent - making it harder to understand/predict.

There are a few issues regarding permissions, some quite old

```[tasklist] 
### Related issues
- [ ] https://github.com/inventree/InvenTree/issues/7446
- [ ] https://github.com/inventree/InvenTree/issues/7003
- [ ] https://github.com/inventree/InvenTree/issues/2323
- [ ] https://github.com/inventree/InvenTree/issues/5755
- [ ] https://github.com/inventree/InvenTree/issues/4022
```

## Requirements

TODO: Categorize in needed / good to habe

 Requirements for the overhaul:
1) Interoperable (additional to) with Djangos default system
2) Using existing fine-grained control patterns (Tree per Location -> Part -> Stock items)
3) API-enforced and auto-documented
4) Generic / pluggable so plugins can use the same systems (maybe with a permission register?)
5) Extendable with per-model actions (ie. allocate stock, count stock, create revision, ship order but not change it)
6) Integrate well with (LDAP/SSO synced) groups and be transparent to users

## User stories
 TBD

Feel free to submit your user stories / requirements / issues here - I will update / remove points as consensus is reached on them.

wolflu05:
I think it would be pretty helpful, if we could assign some users only permissions to:
- CRUD parts (including SP, MP, ...) from a specific category (and their nested children categories)
- CRUD locations (including stock items) only from a specific location (and their nested children locations)
- CRUD only specific companies, BOMs ... maybe here it would be helpful to create different "profiles" which can be assigned to a user, and to some models like BOMs, Companies, ...

This would be helpful to my personal inventree instance which I use for my electronics hobby, to add a separate profile for my family so we can manage our cellar, garage, ... too, without them having appearing my electronics inventory when they search for something they have in their garage. I could just run two inventree instances, but 1. this costs double RAM, CPU, ... and 2. I have to use two different logins.

As well, this could also be useful for business who have different departments: manufacturing, selling, .. different categories of products. Where the employees should not be able to mess with the parts, locations in other departments.

SergeoLacruz:
I think it would also be nice be bind this somehow to projects.

* Assign users to projects
* Assign parts/stock whatever to projects

Everyone in the project gets access to every item in the project. User groups already exist and can be used.

[...]

PR #9506, opened 2025-04-15
Rulset simplification

SchrodingersGat:
As part of an effort towards improving / updating / reworking the "permissions" system (with the eventual goal of simplification and allowing plugins to hook into permission checks) - this PR consolidates some of the existing role types:

- Remove `part_category` role - now in `part`
- Remove `stock_location` role - now in `stock`

### References

- https://github.com/inventree/InvenTree/pull/9476
- https://github.com/inventree/InvenTree/pull/9490

### DIFF
--- a/src/backend/InvenTree/InvenTree/permissions.py
+++ b/src/backend/InvenTree/InvenTree/permissions.py
@@ -8,7 +8,11 @@ import users.permissions
 
 
 def get_model_for_view(view):
-    """Attempt to introspect the 'model' type for an API view."""
+    """Attempt to introspect the 'model' type for an API view.
+
+    Arguments:
+        view: The view being accessed
+    """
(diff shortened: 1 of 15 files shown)


matmair:
This makes it such that anyone in an org that should work on parts also can mess up the general structure of the database as they now can also create/remove categories and locations.

Can we hold of on merging this till the mentioned solution to use plugins for permissions is done? This would cause serious headaches in my instances, especially because I would have to patch permissions somehow as the default behavior changes.

SchrodingersGat:
@matmair fair enough, I'll leave this one pending until we have a good solution for exposing to plugins.

matmair:
Maybe we can discuss the overall design shift before implementing it - for example in https://github.com/inventree/InvenTree/issues/7466

PR #9523, opened 2025-04-16
chore(frontend): bump dep resolution

matmair:
To adress GHSA-968p-4wvh-cqc8


matmair:
I was not aware that this kind of error is even something that happens. Will try to figure it out

PR #9527, opened 2025-04-17
feat(backend): clearer behaviour on missing migrations

matmair:
This PR stops a server startup if migrations are missing and logs a clear error with a InvenTree error code.
That makes it much easier for administrators to search what they need to do, as there is only one line of errors and not a huge traceback.

Also adds a decorator that helps log flows of tasks if running in (runner) debug mode.


SchrodingersGat (approved):

SchrodingersGat:
Great idea, thanks @matmair

Issue #9423, opened 2025-03-31
[Report] Generate single report against selected items


SchrodingersGat:
### Discussed in https://github.com/inventree/InvenTree/discussions/9408

<div type='discussions-op-text'>

<sup>Originally posted by **tristanle22** March 30, 2025</sup>
I want to select a few stock items and generate a report. In the report, I want to have a table, with each row for each stock item. I can't find a context variable that represents a list of selected stock items. 

I'm aware of the built-in "Inventree Test Report" generates a new page for each selected item.</div>

### Feature Description

Instead of generating a separate report for each individually selected item (e.g. stock item / order / part), add an option to pass all selected items through to the report context, and then generate a single report from that.

### Implementation Ideas

- Add a "merge" (or similar named) field to the "ReportTemplate" model
- This allows the report template creator to decide whether the report is generated against a single output, or against all selected outputs

The `ReportTemplate` model has an existing `print` method which could be overloaded:

https://github.com/inventree/InvenTree/blob/66d5180d8fa1542891de848655fc77e7ce28ad82/src/backend/InvenTree/report/models.py#L351

Simply replace this iterator:

https://github.com/inventree/InvenTree/blob/66d5180d8fa1542891de848655fc77e7ce28ad82/src/backend/InvenTree/report/models.py#L403

SchrodingersGat:
@tristanle22 FYI

tristanle22:
I'm working on this feature. Just have a dump question as I'm not very familiar with the matter :). Currently I'm adding the "merge" field as a Boolean. If the user want to set this field (or any other field) to True in the report template, how would they do it?

SchrodingersGat:
@tristanle22 great question. Here's a rough sequence of required steps:

1. Add new field to `ReportTemplate` model

https://github.com/inventree/InvenTree/blob/67bdf3162afed8280df3190a6eab9918c1d4a97b/src/backend/InvenTree/report/models.py#L330

2. Run database migrations (to actually add the new field)

`invoke migrate`

3.  Expose new field to the API

Add the new field name to the `Meta.fields` attribute of the `ReportTemplateSerializer` class

https://github.com/inventree/InvenTree/blob/67bdf3162afed8280df3190a6eab9918c1d4a97b/src/backend/InvenTree/report/serializers.py#L61

This will ensure that the new field is exposed to the API

4. Add the new field to the frontend form

Add your new field name to the `additionalFormFields` attribute of the `ReportTemplateTable` component:

[...]

PR #9532, opened 2025-04-18
Report merge

tristanle22:
Allow generation of a signle report PDF when multiple items are selected.

#9423


.pre-commit-config.yaml
matmair, line 1: This seems unrelated, please remove

src/backend/InvenTree/report/models.py
```diff
@@ -170,10 +170,12 @@ class ReportContextExtension(TypedDict):
...
         landscape: Boolean value, True if the report is in landscape mode
+        merge: Boolean value, True if the a single report is generated against multiple items
     """
 
     page_size: str
     landscape: bool
+    merge: bool
 
 
 class ReportTemplateBase(MetadataMixin, InvenTree.models.InvenTreeModel):
```
matmair, line 181: (Blocking): This change requires migrations, which are not included in the PR

src/backend/InvenTree/report/models.py
```diff
@@ -501,6 +518,47 @@ def print(self, items: list, request=None, output=None, **kwargs) -> DataOutput:
...
+                        try:
+                            plugin.report_callback(self, instance, report, request)
+                        except Exception:
+                            InvenTree.exceptions.log_error(
+                                f'plugins.{plugin.slug}.report_callback'
+                            )
+
+                    # Update the progress of the report generation
+                    output.progress += 1
+                    output.save()
```
matmair, line 561: There is a lot duplication between the 2 sections; can you try reducing that? Every change in rendering logic going forward will have to be done twice.

matmair (changes requested):

matmair:
As the API schema is changed you will also need to change the API version - see https://docs.inventree.org/en/stable/develop/contributing/#api-versioning

Let me know if you want more detailed pointers

tristanle22:
> As the API schema is changed you will also need to change the API version - see https://docs.inventree.org/en/stable/develop/contributing/#api-versioning
> 
> Let me know if you want more detailed pointers

Is it just simply updating "src/backend/InvenTree/InvenTree/api_version.py" by bumping up the version and add a new line to INVENTREE_API_TEXT?

matmair:
Yes; Just copy the latest entry and adjust to fit this PR (especially the url)

SchrodingersGat:
@tristanle22 a great start here :) Please begin by addressing the issues raised by @matmair

src/backend/InvenTree/report/models.py
```diff
@@ -233,31 +233,34 @@ def generate_filename(self, context, **kwargs):
...
         return template_string.render(Context(context))
 
-    def render_as_string(self, instance, request=None, **kwargs) -> str:
+    def render_as_string(self, instance, request=None, context=None, **kwargs) -> str:
         """Render the report to a HTML string.
 
         Arguments:
             instance: The model instance to render against
             request: A HTTPRequest object (optional)
+            context: DTL context (optional)
```
matmair, line 242: Question: What does DTL stand for in this context?

src/backend/InvenTree/report/models.py
```diff
@@ -455,7 +479,7 @@ def print(self, items: list, request=None, output=None, **kwargs) -> DataOutput:
         # Start with a default report name
         report_name = None
 
-        report_plugins = registry.with_mixin(PluginMixinEnum.REPORT)
+        # report_plugins = registry.with_mixin(PluginMixinEnum.REPORT)
```
matmair, line 482: Is this meant to be commented out?

matmair (approved):
good reduction of duplicates

matmair:
Looks pretty good now; it mainly is missing the migration I mentioned above and a bit of unit-testing to ensure your changes continue to work in the future;
Do you want pointers to those topics?

tristanle22:
@matmair For migration, there are some new Python files generated in `src/backend/InvenTree/report/migrations/` after running `invoke migrate`, I guess they need to be included in this PR? For the testing part, I'm guessing the report unit tests live here `src/backend/InvenTree/report/tests.py`?

matmair:
Both correct @tristanle22

tristanle22:
Is there a way to run `src/backend/InvenTree/report/tests.py` with VSCode debugger?

SchrodingersGat:
https://docs.inventree.org/en/latest/develop/contributing/?h=dev.test#run-tests-locally

In your case

```bash
invoke dev.test -r report.tests
```

SchrodingersGat:
@tristanle22 some further things I would like to see here:

### Migrations

You have a two migrations here, `0030` and `0031`.

- **0030** shadows an existing migration
- **0031** simply combines the two "0030" migrations

Can you please delete `0031`, and rename `0030` to `0031`.

### Unit Testing

Your unit testing does not actually test for multiple items being combined into a single report as you only send it a single item.

I would suggest that your unit test needs to:

- Supply multiple items
- Ensure that the report output generates a *single report*, not multiple reports stacked together
- You can utilize the "DEBUG MODE" for reports here to return HTML and then you can read the HTML to ensure it is formatting properly

### Documentation

Please add some documentation (in `./docs/`) for this new feature:

- What the "merge" feature does, and how to use it
- How to specify which report templates get merged

tristanle22:
> https://docs.inventree.org/en/latest/develop/contributing/?h=dev.test#run-tests-locally
> 
> In your case
> 
> ```shell
> invoke dev.test -r report.tests
> ```

I setup this command in launch.json, but it doesn't stop at any breakpoints. I want to place breakpoints to play around with the variables to see how the pieces work together. Do you have any suggestions?

matmair:
There is a pre-defined launch definition for this provided in the code base, check out the launch.json file

matmair:
@tristanle22  looking good; could you resolve the merge conflict so we can do a final review round?

wolflu05:
Do I understand it correct, that now we have another "dimension" of possible context generation in a matrix.

So now we have different contexts for:
- type: label / report
- component: part, stocklocation, ...
- merged report: true, false

=> 2 * (~9) * 2 = ~36 different context combinations

Meaning if merge is set, all instance specific context is available in `instances.x.<name>` instead of `<name>`, but e.g. `width` (for labels) is still available as `width` and not `instances.x.width` ?

Just asking, because I have to adjust my [LSP plugin](https://github.com/wolflu05/inventree-report-lsp-plugin) accoringly now.

tristanle22:
> Do I understand it correct, that now we have another "dimension" of possible context generation in a matrix.
> 
> So now we have different contexts for:
> 
> * type: label / report
> * component: part, stocklocation, ...
> * merged report: true, false
> 
> => 2 * (~9) * 2 = ~36 different context combinations
> 
> Meaning if merge is set, all instance specific context is available in `instances.x.<name>` instead of `<name>`, but e.g. `width` (for labels) is still available as `width` and not `instances.x.width` ?
> 
> Just asking, because I have to adjust my [LSP plugin](https://github.com/wolflu05/inventree-report-lsp-plugin) accoringly now.

Yes both correct

tristanle22:
@SchrodingersGat @matmair I think it's ready for a final round of review. Let me know if there's anything else you'd like to add/updated!

SchrodingersGat:
@tristanle22 I'm back from vacation now and ready to review. Can you please look at the conflicts here and I can review again :)

matmair:
@SchrodingersGat I have resolved the conflicts

SchrodingersGat:
### Documentation

The "merge" feature needs some clearer documentation in the "report" page (`./docs/docs/report/report.md`).

In particular, the difference in terminology in the report template, e.g. use `instance.<name>` instead of `<name>`

### Functionality

On testing, it appears to work as expected, nice!

tristanle22:
> ### Documentation
> The "merge" feature needs some clearer documentation in the "report" page (`./docs/docs/report/report.md`).
> 
> In particular, the difference in terminology in the report template, e.g. use `instance.<name>` instead of `<name>`
> 
> ### Functionality
> On testing, it appears to work as expected, nice!

I have a section in `.docs/docs/report/context_variables.md` that explains what you mentioned. I can make it clearer there, instead of in `./docs/docs/report/report.md`

SchrodingersGat:
I would still like to see some more obvious callout from the main `report` docs page :)

SchrodingersGat:
@tristanle22 if you can address the documentation, resolve conflicts and look at the failing tests, I'd love to see this merged!

tristanle22:
Sorry, I've been occupied lately. Will try to resolve the last two items.

tristanle22:
All failing tests are because of the same error:
`FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/InvenTree/media/data_output/output.html'`
This is how I'm grabbing the output html. It worked locally:
`report_path = os.path.join(
            settings.MEDIA_ROOT.parent, response.data['output'].lstrip('/')
        )`

@SchrodingersGat Do you have any suggestion?

SchrodingersGat:
> All failing tests are because of the same error: `FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/InvenTree/media/data_output/output.html'` This is how I'm grabbing the output html. It worked locally: `report_path = os.path.join( settings.MEDIA_ROOT.parent, response.data['output'].lstrip('/') )`
> 
> @SchrodingersGat Do you have any suggestion?

It should not be `MEDIA_ROOT.parent` - this is up one level from the media dir. I think it should be `settings.MEDIA_ROOT` instead

tristanle22:
> > All failing tests are because of the same error: `FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/InvenTree/media/data_output/output.html'` This is how I'm grabbing the output html. It worked locally: `report_path = os.path.join( settings.MEDIA_ROOT.parent, response.data['output'].lstrip('/') )`
> > @SchrodingersGat Do you have any suggestion?
> 
> It should not be `MEDIA_ROOT.parent` - this is up one level from the media dir. I think it should be `settings.MEDIA_ROOT` instead

It has to be `MEDIA_ROOT.parent` because:
```
print(response.data['output']
> /media/data_output/output_qXuMPAB.html
```

SchrodingersGat:
Ah, ok so the "response" object provides the path to the file *on the web server* - NOT where the file resides on disk.

So, the inventree server provides the output at `/media/data_output/some_file.pdf'` - which is a path *relative to the host*.

On disk, it will be somewhere like `/my/path/to/inventree-media/data_output/some_file.pdf/`

So, take the response path, strip the `/media/` off the start, and then join that the `MEDIA_ROOT`

SchrodingersGat:
@tristanle22 thanks for implementing this, it's a great addition! Very nice work :)

tristanle22:
Thanks @SchrodingersGat and @matmair for all the help! :)

PR #9541, opened 2025-04-20
[UI] Add CUI compatibility URLs

SchrodingersGat:
Rerouting for "CUI" (legacy) URLs to new UI patterns.

Allows external links to InvenTree to still work now that we have removed the old UI entirely


matmair (approved):
Lgtm; can we adds setting or flag to disable those? Having unused url paths that I need to consider in the SWG ist kind of cumbersome.
A test or two for this would also be good.

PR #9542, opened 2025-04-20
Reduce notification check frequency

SchrodingersGat:


matmair:
5 minutes seems a bit long, especially in a JIT shop

SchrodingersGat:
1 minute? Or should we make it configurable?

PR #9545, opened 2025-04-20
Docs refactor

SchrodingersGat:
Major documentation refactor in prep for 1.0.0 release

- Updated screenshots (new UI)
- Reorganize docs to match UI structure

### TODO

- [x] Get the "carousel" working again
- [x] Replace "icon" integration in docs (we use "tabler" now) 
- [x] Add shortcuts for icons in docs
- [x] Replace all existing "icon" shortcuts in docs
- [x] Ensure internal docs links still point to the correct locations


docs/docs/security.md
matmair, line 1: We have high level links to /security so removing this would break external sites

docs/mkdocs.yml
```diff
@@ -72,27 +73,58 @@ extra_javascript:
...
+      - Overview: api/python/index.md
+      - Currency Support: api/python/currency.md
+      - Examples: api/python/examples.md
+    - Project Details:
       - Governance: project/governance.md
       - Project Security: project/security.md
       - Resources: project/resources.md
-    - Security: security.md
+      - Privacy: project/privacy.md
+      - Threat Model: project/threat_model.md
```
matmair, line 125: The Thread Model regards individual instances, not the project; not sure if this reordering makes sense

SchrodingersGat: Oh good point. I'll fix that.

docs/mkdocs.yml
```diff
@@ -58,6 +58,7 @@ extra_css:
   - stylesheets/splide.min.css
   - stylesheets/extra.css
   - stylesheets/neoteroi-mkdocs.css
+  - https://cdn.jsdelivr.net/npm/@tabler/icons-webfont@3.31.0/dist/tabler-icons.min.css
```
matmair, line 61: Does this mean docs can not be build offline?

SchrodingersGat: You can build offline, but there will be some elements fetched via CDN when viewing the docs. That is already the case with some elements  (we fetch bootstrap via CDN for example). Do you think we should target fully offline docs builds?

docs/docs/security.md
SchrodingersGat: This is back as a top-level link now

docs/mkdocs.yml
matmair: Fully offline docs would be good, at least from releases. That makes full reproducibility easier

matmair (approved):

docs/mkdocs.yml
SchrodingersGat: To be clear, the docs will run "offline" - your browser will just fetch some icons from a CDN as required.

Issue #9481, opened 2025-04-08
[API] Reduce hidden mutations


SchrodingersGat:
### Body of the issue

A lot of API endpoints can be "mutated" by passing extra parameters (e.g. `?part_detail=true`). These parameters are largely undocumented and should be re-implemented as API filters. This will expose them to the API schema documentation.

matmair:
See https://github.com/inventree/InvenTree/issues/9045

Reza98Sh:
I’d like to take on this issue. Is there any additional context or details I should know before starting?

matmair:
@Reza98Sh With the limited work you have submitted to the backend I am not sure if you have found all the trip-hazards that are hidden in the way our backend api works, but you have shown potential in other work so sure, go ahead. Just let us know if you need pointers.

Just to re-iterate the problem and what would be considered a "solve":
Problem: We currently have parameters in the serializers that adapt the form of the API response.
https://github.com/inventree/InvenTree/blob/0bed5cf5112f57cc7d53e622c4596ac6eb8d0696/src/backend/InvenTree/build/api.py#L377
If you look into the OpenAPI schema you will see that this parameter is not represented there
https://github.com/inventree/schema/blob/38ab7363cf0b1fcedc3e4994fbf5b351db2473bd/export/385/api.yaml#L1964-L2152

There are at least 2 anti-patterns, that should be addressed:
1. parameters that completely change the output schema - probably a lot of work, diving deep into drf serialisation and OpenAPI schema generation
2. parameters that add filters that are not documented - easier as one can use proper django-filter filtersets / fields and that automatically gets picked up by schema generation

Goal: A "solve" would be to eliminate 2 completely and proposes a working solution for addressing 1

The whole issue was discovered by @1337joe as he is generating a client library from the schema and these mismatches lead to missing arguments/broken requests.

matmair:
I think this issue is probably best addressed by a number of small PRs that tackle a few API endpoints at a time, as a large PR that does everything might be time-costly to review and correct.

Reza98Sh:
@matmair 

Thank you. I welcome the challenge.

I agree with your suggestion. I’ll start by tackling the undocumented filter parameters (problem #2) first . Once that’s done, I’ll move on to the more complex schema-changing parameters.

matmair:
I have changed the milestone, I think it is realistic to tackle this with the next feature release

PR #9547, opened 2025-04-21
Schema: Require pagination limit parameter

### LINKED ISSUE #9195: OpenAPI Schema and Pagination

### Describe the bug*

I'm running Inventree 0.17.6 (from the shell script installer), api version `294`.

The OpenAPI definition for most endpoints describe a `Paginated...` object with following structure:
```json
{
  "count": 123,
  "next": "http://...",
  "previous": "http://...",
  "data": []
}
```

however the data returned is simply an array of items, no additional pagination properties. They do appear if I manually supply a `limit=` parameter, but the openapi spec does not seem to reflect that.

Image

Image

### Steps to Reproduce

Access any API list endpoint without providing a limit

### Expected behaviour

OpenAPI schema should match the server output

Image

### Deployment Method

- [ ] Docker
- [ ] Package
- [x] Bare metal

### Version Information

0.17.6

- [x] I can reproduce this bug on the demo site.

### Relevant log output

```shell

```

1337joe:
I'm offering this as a possible solution to the linked issue, I don't know if it's something we want.

**Problem:** If a pagination limit is not provided in a list query, the response comes back as a list of all relevant objects, not a paginated wrapper object as the schema defines.

**Solution:** Require `limit` to be set on list queries. That way if you're following the schema specification when you make the call the response you get also follows the schema specification.

This completely ignores/hides the shape changing nature of the current API for list endpoints.

Fixes #9195

### DIFF
--- a/src/backend/InvenTree/InvenTree/api_version.py
+++ b/src/backend/InvenTree/InvenTree/api_version.py
@@ -1,7 +1,7 @@
 """InvenTree API version information."""
 
 # InvenTree API version
-INVENTREE_API_VERSION = 339
+INVENTREE_API_VERSION = 340
 
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
--- a/src/backend/InvenTree/InvenTree/api_version.py
+++ b/src/backend/InvenTree/InvenTree/api_version.py
@@ -1,13 +1,13 @@
 """InvenTree API version information."""
 
 # InvenTree API version
-INVENTREE_API_VERSION = 340
+INVENTREE_API_VERSION = 341
 
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 
 INVENTREE_API_TEXT = """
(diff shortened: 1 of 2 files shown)


matmair:
Looks interesting; #9481 should address this generally

matmair (approved):

SchrodingersGat:
@1337joe does this change the actual behavior of the API, or just the documentation schema? I ask because there are certainly some API endpoints where we make use of this "default" behavior and do not expect (or desire) any pagination.

1337joe:
@SchrodingersGat I made the changes in the AutoSchema class so that it would only affect the schema. The API itself will remain unchanged, but the schema will be generated with limit required.

SchrodingersGat:
Ok, thanks for the clarification. We should review which endpoints we do not want to be paginated (it is only a limited set that we use that way) and perhaps mark them as such?

However, for this PR we are just updating the documentation schema, so this looks good to go.

PR #9574, opened 2025-04-24
[Plugin] Auto-create builds

SchrodingersGat:
Adds a builtin plugin to allow automatic creation of build orders:

- Create child builds when a build order is issued
- Create build orders when a sales order is issued

This plugin is intended to replace the existing "create_child_builds" functionality when creating a new build order. Because:

- The current implementation is ill considered
- It does not account for current stock levels
- There is no configurability
- It creates the child build orders unconditionally
- It should be called when the build order is *issued* - allowing user time to edit / adjust / schedule

This is a **BREAKING CHANGE** - in that it removes the old method for automatically creating child build orders. However it is much better.

### TODO

- [x] Remove old field which "auto created" child builds for a build order
- [x] Add new event plugin which auto-creates builds when a sales order or build order is issued
- [x] Document new plugin functionality

### References

- https://github.com/inventree/InvenTree/pull/9565


matmair:
Is it intended that there is a large drop in coverage, or will that be addressed?

PR #9575, opened 2025-04-24
Add highlightToday for date inputs

reilly-callaway:
image


SchrodingersGat (approved):

SchrodingersGat:
We also use a `<DateInput>` component in the `<FilterElement>` component (in `./src/frontend/src/tables/FilterSelectDrawer.tsx`) - it would be good to make this change there also

matmair (approved):

PR #9580, opened 2025-04-24
Add revision column to BO table

gunstr:
When revisions of parts are used it would be good to see that field also in the BO table

image


SchrodingersGat:
Thanks for the contribution @gunstr

PR #9582, opened 2025-04-25
Refactor/ubiquitous language

kevbang:
## Refactor: Ubiquitous Language

Hello maintainers,

As part of our assignment, we were tasked with applying Domain-Driven Design (DDD) principles to InvenTree. One aspect of DDD includes ubiquitous language, which refers to shared vocabulary between developers and domain experts. As we read through the codebase, we noticed a frequent usage of the `part` variable.

In the case when a `part` is supplied by a company, it is called a `SupplierPart`. This can be confusing to new contributors because,`part`, as a term is used frequently throughout the entire codebase, but in some instances when the term `part`, is used, it is referring to a `SupplierPart` instead.

This Pull Request refactors instances of `part` in the PurchaseOrder context to `supplier_part` to better reflect the entity's role and improve code readability.  We recognize that adopting this change consistently across the codebase would require significant effort, especially given our limited familiarity with all parts of the project.

As such, this is proposed as a suggestion to improve clarity. We’re happy to explore it further if it aligns with the maintainers’ vision for the codebase.

### DIFF
--- a/src/backend/InvenTree/order/models.py
+++ b/src/backend/InvenTree/order/models.py
@@ -1695,7 +1695,7 @@ class PurchaseOrderLineItem(OrderLineItem):
             return None
         return self.part.part
 
-    part = models.ForeignKey(
+    supplier_part = models.ForeignKey(
         SupplierPart,
         on_delete=models.SET_NULL,
         blank=False,
--- a/src/backend/InvenTree/order/models.py
+++ b/src/backend/InvenTree/order/models.py
@@ -1765,8 +1765,8 @@ class PurchaseOrderLineItem(OrderLineItem):
 
     def update_pricing(self):
         """Update pricing information based on the supplier part data."""
-        if self.part:
-            price = self.part.get_price(
+        if self.supplier_part:
+            price = self.supplier_part.get_price(
                 self.quantity, currency=self.purchase_price_currency
             )
 


SchrodingersGat:
Hi @kevbang while I can appreciate the intent here, changing the underlying database field names will have a significant flow-on effect.

Firstly there will need to be database migrations to change the field name. Not a difficult change but one that must be considered carefully.
 
Secondly there will need to be significant changes to the API. This will impact any existing systems which interface with InvenTree externally. 

Lastly (and most importantly) is that this would break any existing plugins which expect the existence of a field called "part". This change would require changes to many plugins which rely on this.

I agree that the naming is poor. The code has shifted and evolved significantly and this stems from an original design decision which could have been made differently. But any change at this point needs to be weighed carefully.

Perhaps we could add changes along the lines of:

- change the label of the fields in the API to ensure obvious functions 
- add a `supplier_part` getter and setter to the `PurchaseOrderLineItem` model which shadow the `part` field in the database.

matmair:
I have approved a workflow run to gauge the impact on coverage and other parts of the system

kevbang:
I completely understand and acknowledge the impact that changing the underlying database field name would have. These are not trivial concerns, and I agree that any change here would need to be weighed very carefully.

I appreciate your suggestion to explore less disruptive alternatives like adding a `supplier_part` getter/setter or improving API labels. These are great ideas that could potentially be revisited outside the scope of my class, or by other contributors in the future.

Thanks again for your insight, it is much appreciated!

Issue #9579, opened 2025-04-24
[Setup] Double schedule entries on upgrade


Philipboy881:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

i cannot update inventree because of a double schedule name. but i cannot delete the double schedule item, i will return to the home page

### Steps to Reproduce

in cli
sudo inventree run invoke update

### Expected behaviour

the sistem is getting updated to the latest version

### Deployment Method

- [ ] Docker
- [ ] Package
- [x] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

InvenTree-Version: 0.17.0
Django Version: 4.2.17
nullDatabase: sqlite3
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-5.4.0-212-generic-x86_64-with-glibc2.31
Installer: null
nullActive plugins: [{"name":"InvenTreeBarcode","slug":"inventreebarcode","version":"2.1.0"},{"name":"InvenTreeCoreNotificationsPlugin","slug":"inventreecorenotificationsplugin","version":"1.0.0"},{"name":"InvenTreeCurrencyExchange","slug":"inventreecurrencyexchange","version":"1.0.0"},{"name":"InvenTreeLabel","slug":"inventreelabel","version":"1.1.0"},{"name":"InvenTreeLabelMachine","slug":"inventreelabelmachine","version":"1.0.0"},{"name":"InvenTreeLabelSheet","slug":"inventreelabelsheet","version":"1.0.0"},{"name":"DigiKeyPlugin","slug":"digikeyplugin","version":"1.0.0"},{"name":"LCSCPlugin","slug":"lcscplugin","version":"1.0.0"},{"name":"MouserPlugin","slug":"mouserplugin","version":"1.0.0"},{"name":"TMEPlugin","slug":"tmeplugin","version":"1.0.0"},{"name":"Zebra labels","slug":"zebra","version":"0.8.1"},{"name":"inventree_rexel","slug":"inventree_rexel","version":"V0.1.22"}]

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

### Relevant log output

```shell
sudo inventree run invoke update

Updating InvenTree installation...
Installing required python packages from '/opt/inventree/src/backend/requirements.txt'
Requirement already satisfied: pip in ./env/lib/python3.9/site-packages (25.0.1)
[...]

PR #9586, opened 2025-04-25
fix(backend): Make task registering more robust

matmair:
See https://github.com/inventree/InvenTree/issues/9579


matmair:
Not really sure this is the original problem in https://github.com/inventree/InvenTree/issues/9579 but it does not hurt to make this more robust

SchrodingersGat (approved):

PR #9577, opened 2025-04-24
chore: bump zustand

matmair:
Reduced version of #8913
Bumps zustand to version 5, requiring refactor of access patterns.

PR #9589, opened 2025-04-25
bump babel again

matmair:
Follow up to https://github.com/inventree/InvenTree/pull/9523 because https://github.com/inventree/InvenTree/pull/9577 re-introduced a dependency < 7.26.10


matmair:
I now ran the osv scanner locally and it looks good now
grafik

PR #9596, opened 2025-04-28
feat(frontend): add grouping in the admin center

matmair:
This PR adds the option to configure groupings for panels, providing a better overview regarding impacted areas of the panels.
grafik

### DIFF
--- a/src/frontend/src/components/panels/Panel.tsx
+++ b/src/frontend/src/components/panels/Panel.tsx
@@ -13,3 +13,10 @@ export type PanelType = {
   disabled?: boolean;
   showHeadline?: boolean;
 };
+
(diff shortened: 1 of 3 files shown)


matmair:
1. to ensure plugins can not place themself anywhere but are always at the back
2. To ensure each panel is only rendered once

wolflu05:
I would actually see this as an advantage to let plugins place their panels in specific groups, don't you think that?

matmair:
I don’t, that is why I pointed it out as a reason

SchrodingersGat (approved):

SchrodingersGat:
@matmair very nice :)

PR #9602, opened 2025-04-28
feat: improve user/group management actions

matmair:
Adds a few extra user/group management actions:
- open profile
- change password
- lock / unlock user
- ~Invite user~
grafik
This aims to close the last gaps to make user management not rely on the django admin


matmair:
@SchrodingersGat merge conflict resolved

src/frontend/src/tables/settings/UserTable.tsx
```diff
@@ -298,7 +306,43 @@ export function UserTable({
...
+          }
+        },
+        {
+          icon: <IconKey />,
+          title: t`Change Password`,
+          color: 'blue',
+          onClick: () => {
+            setSelectedUser(record.pk);
+            setPassowrd.open();
+          }
```
SchrodingersGat: This action should only be available if the user is a superuser, right?

matmair: I have added a check to hide it; backend requires superuser

SchrodingersGat:
Looks like there's still some issue with the PUI tests

matmair:
There seems to be a known issue with playwright that causes this - I have bumped playwright https://github.com/microsoft/playwright/issues/35183

matmair:
@SchrodingersGat i solved the issue

SchrodingersGat:
Nice work - what was the root cause?

matmair:
The unpinned install pulled an old build of Firefox, upstream addressed that

PR #9604, opened 2025-04-28
!refactor(backend): restructure user endpoints

matmair:
This realigns the user API endpoints by:
- moving endpoints that only return results for the current user to /user/me/
- renaming the list endpoint for tokens from 'tokens' to 'token' as that name is now free

see https://github.com/inventree/InvenTree/pull/11433


src/backend/InvenTree/InvenTree/api_version.py
```diff
@@ -1,11 +1,14 @@
...
 
 # InvenTree API version
-INVENTREE_API_VERSION = 459
+INVENTREE_API_VERSION = 460
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
 
+v460 -> 2026-02-27 : https://github.com/inventree/InvenTree/pull/9604
+    - Refactors the user API endpoint to be less ambiguous in their naming
```
SchrodingersGat: We are going to have to consider how we mix the "next-breaking" and "master" changes to the API - if we do not merge this into master we will run into conflicts.. Any thoughts how to handle this?

matmair: My immediate idea would be to have one API bump per major release that removes all deprecations and just push the API version up by one till merge comes, but I can see that this needs some better considerations

matmair:
moving to draft while I think about the API change that was brought up

matmair:
@SchrodingersGat what do you think about this idea: we could ignore api schema issues on the next-breaking branch and inly introduce one bump one we merge into master before the next stable release is done. This would reduce merge issues and still ensure we bump with the merge

SchrodingersGat:
@matmair I think that's quite sensible

matmair:
@SchrodingersGat I have adjusted the qc checks following above suggestion

matmair:

matmair:
replaced by https://github.com/inventree/InvenTree/pull/11963

PR #9606, opened 2025-04-29
chore(frontend): bump

matmair:
Seeing that https://github.com/inventree/InvenTree/pull/9603 times out, this attempts a partial bump to isolate the issue


SchrodingersGat:
@matmair is this one still being pursued?

matmair:
Certainly something that should be done as the 1.0.x release line will ship these deps a while; I can try to adress this on the weekend

matmair:
Not sure why this React 18-19 and Mantine 7-8 bump is so difficult; closing for now with https://github.com/inventree/InvenTree/issues/10196 as a tracking issue.
Bumping the deps is not mandatory for a stable release

Issue #9268, opened 2025-03-10
[FR] add Can Build to build order overview


mabroens:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

please add can build line to the build order overview. currently the BOM shows how many can be build:

![Image](https://github.com/user-attachments/assets/ee7925f8-a00d-45ad-94a7-e8235976f9c3)

but i would be nice to show this also on the build order:

![Image](https://github.com/user-attachments/assets/69bba58d-7421-4798-b830-d731fded3ec1)

gives better overview.

### Suggested solution

add the line in buid order overview

### Describe alternatives you've considered

none

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
Happy for someone to tackle this - the reason it is not currently there is that the "can build" quantity does not exist easily on the API. In the old UI this was generated via the server-side template

tinpham5614:
I would like to work on this issue to display the "can build" line in the Build Order Detail.

PR #9614, opened 2025-05-02
[FR] Add Can Build to build order overview

tinpham5614:
Fix: #9268 
Add the Can Build line to the Build Order Detail/Overview as in the screenshot.
**Note**: Currently, the "can build" quantity is not in the API yet.

image


SchrodingersGat:
What are your thoughts on calculating the actual quantity here? 

## 1. front-end 

You could potentially grab the BOM data as a secondary API call and calculate the quantity on the front-end.

## 2. Back-end - new endpoint 

You could add a custom endpoint to allow calculation of "can build" for a given part when requested 

## 3. back-end - additional field

Alternatively you could add an extra field to the existing endpoint.

------------------------

I would lean towards 2 here as it:

- provides for reusability on this calculation 
- can be more efficient in its use of database hits 
- calculation is on demand, and can be requested only for the detail view of a part or build order

tinpham5614:
I agree with approach 2:

- It’s accurate and consistent (no duplicate logic)
- It’s efficient (doesn’t slow down other APIs)
- We can reference the existing `Part.can_build` property code.

I can implement this as a new `GET /api/build/<int:pk>/can-build/` endpoint.
Returns: `{ "can_build": 5 }`

Can I please get more guidance on this?

SchrodingersGat:
> Can I please get more guidance on this?

First would be to add a new API endpoint, which requires:

1. Add a new "route" 

See [/src/backend/InvenTree/part/api.py](https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/part/api.py)

The "PartPricingDetail" endpoint is a good one to copy:

https://github.com/inventree/InvenTree/blob/6d47a0dd164f0845d65012891691e7cc24f9f04a/src/backend/InvenTree/part/api.py#L791

2. Add a new "serializer"

e.g.

https://github.com/inventree/InvenTree/blob/6d47a0dd164f0845d65012891691e7cc24f9f04a/src/backend/InvenTree/part/serializers.py#L1348

That should get you started - but LMK if you need any further pointers

tinpham5614:
I'm working on this and will see how it goes.

Thank you.

tinpham5614:
I tried the second approach (custom backend endpoint) but ran into issues. The frontend loads before `partIds` are available, so the `can_build` values are often `null` on the first render. Handling that added complexity to the UI (personally).

Instead, I added a Part `can_build` field directly to the build serializer, which worked much better. It’s simpler, avoids extra API calls, and ensures the value is ready when rendering.

### Results:
**Build Order Overview:**
image

**Build Order Detail:**
image

SchrodingersGat:
@tinpham5614 please refer to my implementation in https://github.com/inventree/InvenTree/pull/9798

There are a couple of issues with your approach here, the major one being that simply adding the `can_build` field to the `BuildSerializer` class results in a *significant* increase in the response time for a list of builds.

A simple comparison with the demo dataset:

URL: `http://localhost:8000/api/build/?format=json&part_detail=true`

| Without 'can_build' | With 'can_build' |
| --- | --- |
| ~ 60ms | ~ 1.1s |

This is because the `can_build` method is not a database annotation - it is called once for each item in the list, and itself makes many database queries. I do not think that the `can_build` number can be easily *annotated* against the queryset. which is why I have suggested implementing a simpler endpoint which is "per part" (so that we can accept a more expensive query)...

> It’s simpler, avoids extra API calls, and ensures the value is ready when rendering.

Simpler, for sure! But, undesirable for the performance issues outlined above.

> I tried the second approach (custom backend endpoint) but ran into issues. The frontend loads before partIds are available, so the can_build values are often null on the first render. Handling that added complexity to the UI (personally).

There are ways around this - in my linked implementation, the value is hidden until the query has completed.

I'm going to close this one out in favor of https://github.com/inventree/InvenTree/pull/9798. Please do not be discouraged from contributing further, the nuances of queryset annotation are a pretty tricky first issue to tackle!

tinpham5614:
Thank you for your feedback. It is very helpful.

Issue #9525, opened 2025-04-16
[Docs] Add missing material for OSSF best practices badge silver


matmair:
### Body of the issue

We are currently fulfilling 93% of the silver level - https://www.bestpractices.dev/en/projects/7179?criteria_level=1

Missing are:
1.  code of conduct: [Contributor Covenant](https://www.contributor-covenant.org/) seems like a very reasonable and non-invasive approach https://www.contributor-covenant.org/version/2/1/code_of_conduct/ see https://github.com/inventree/InvenTree/pull/9568
1. A rough roadmap
1. Architecture docs: working on that
1. A security assurance case: Can be a simple document, working on that

matmair:
@inventree/triage looking for feedback especially on 1 and 2; I will try to tackle 3 and 4 similar to the governance docs

matmair:
I have created a sample for ADRs to fulfill parts of 2 and 3 - https://github.com/matmair/project or https://matmair.github.io/project/

PR #9620, opened 2025-05-02
feat(docs): add stable links for external sites

matmair:
This is useful for the CII best practices badge and other sites where we might need stable links; assuming we maintain control of docs.inventree.org seems like a good bet with current project structures.

### DIFF
--- a/docs/main.py
+++ b/docs/main.py
@@ -93,7 +93,8 @@ def check_link(url) -> bool:
 
     while attempts > 0:
         response = requests.head(url, timeout=5000)
-        if response.status_code == 200:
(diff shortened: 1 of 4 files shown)


matmair:
This is required for example for https://www.bestpractices.dev/en/projects/7179 to ensure the links we provide are stable
part of https://github.com/inventree/InvenTree/issues/9525

matmair:
We seem to be hitting 429's on the GH api right now, might need to wait a bit

SchrodingersGat:
Looks like the 429s continue. Has something changed with  the GitHub API limits?

PR #9626, opened 2025-05-03
Add revision to "Used In" table

gunstr:
Makes it easier to distinguish between multiple revisions of a parts

Screenshot 2025-05-03 230251


SchrodingersGat:
Thanks for the contribution @gunstr

PR #9638, opened 2025-05-07
Update config.md

aluminum-ice:
Added clarification on how to change the timezone correctly. The documentation -- if carefully read -- explains how to change the timezone but it can be easily missed. I also explicitly mentioned the specific variable that must be set to change the timezone from UTC


matmair (approved):

SchrodingersGat:
Thanks for the contribution

PR #9645, opened 2025-05-07
ci: more granular coverage flags

matmair:
Adds more general (and a bit higher) requirements for backend coverage. Industry standard seems to be around 90% so that is the target.


matmair:
This is ready for review and merge

SchrodingersGat (approved):

PR #9655, opened 2025-05-12
feat(frontend): Improve barcode scanning text input on Platform UI. 

willbicks:
This PR adds two improvements to barcode scanning via text input to Platform UI:

- Adds control character handling to barcode text input (used to work in Classic UI).
  - Ports part of https://github.com/inventree/InvenTree/commit/8059fb1e059fcd54bb41eb3ea6f44a8946b67f41 to Platform UI.
  I opted not to port alt-code handling because it seems like the old implementation assumes that your OS doesn't handle alt codes (e.g. works on MacOS, leads to double-inputs on Windows).
  - Fixes https://github.com/inventree/InvenTree/issues/7529 for Platform UI.

- Submit barcode scan on EOT character.
  - Automatically submits 2D barcode scans (e.g. DigiKey, Mouser) when end of transmission character is input.
    This means that popular industrial barcode scanners (e.g. Zebra, Symbol, Motorola) can auto submit 2D barcodes even though they don't terminate scans with 'Enter' out of the box.


SchrodingersGat (approved):

SchrodingersGat:
@willbicks thanks for the submission 😁

SchrodingersGat:
@willbicks thanks for your contribution!

PR #9664, opened 2025-05-15
feat(backend): add typechecking with ty

matmair:
Adds type checking via astral's (experimental) ty which is pretty fast and seems to perform similar enough to mypy for now.
There is no support for django stubs but is a loooot faster so I think it is worth it.


matmair:
@wolflu05 would appreciate a look and feedback on the approach

wolflu05:
Seems like only a small tool change. Looks good in general. The only I idea I have regarding mixins is to use the cast function from typing, but I am not sure if thats a good solution long term.

matmair:
@SchrodingersGat This is ready for a review and possible merge; I have seen that you also started improving typing hints in the last weeks so I think I do not have to lay out the benefits, especially combined with good IDE tooling.

SchrodingersGat (approved):

SchrodingersGat:
Thanks @matmair these all look like great improvements, well worth the effort!

wolflu05:
Is it planned, that ty adds support for Django stubs?

matmair:
One can already install django stubs but there are discussions on the scope of vendored stubs https://github.com/astral-sh/ty/issues/1104

Issue #9684, opened 2025-05-20
[FR] Add check to cli tools to confirm the right commands for deployment method are used


matmair:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

A lot of users seem confused which commands are the correct ones for their enviroment. So they open issues and clog up maintainer time see https://github.com/inventree/InvenTree/discussions/9682 https://github.com/inventree/InvenTree/discussions/9317 

### Suggested solution

Add a check in the tasks that the correct method for the packaging variant is used

### Describe alternatives you've considered

tell ppl to read the docs but that seems to not work

### Examples of other systems

netbox warns when common pitfalls are detected

### Do you want to develop this?

- [x] I want to develop this.

SchrodingersGat:
Great idea!

PR #9687, opened 2025-05-20
feat(backend): Add check to cli tools to confirm the right commands for deployment method are used

matmair:
Fixes https://github.com/inventree/InvenTree/issues/9684


SchrodingersGat (approved):

matmair:
We will see @SchrodingersGat. 1.0 has better error codes, a lot of pit-fall detections and clearer architecture docs. Maybe that helps

PR #9704, opened 2025-05-27
fix(backend): record export is empty when include permissions flag is used

### LINKED ISSUE #9702: Exporting records with permissions results in an empty file

### Describe the bug*

If using the `--include-permissions` when running `invoke export-records` the resulting export file is an empty json array.

Changed in this commit https://github.com/inventree/InvenTree/commit/59c78460c8a6ba9eb4ed0644ec236ea7eb29ad38 a check to exclude temporary settings was added inside the check to remove permissions. It also changed to storing all output data in a new list inside this check.
When wanting to include permissions this new list is never written to.

To get any output containing permissions one need to retain the temporary file and use that instead, potentially including temporary settings.

The `if` checking if permissions should be removed or not and appending the output data, also where the output is written using the `data_out` variable:
https://github.com/inventree/InvenTree/blob/f9675fc93b9dd700ebb6a3511227a8273f0fb489/tasks.py#L877-L898

### Steps to Reproduce

1. Run `invoke export-records --include-permissions -f data.json`
2. Check file and it will only contain `[]` 

### Expected behaviour

To be able to export data with permissions.

### Deployment Method

- [x] Docker
- [ ] Package
- [ ] Bare metal

### Version Information

# Version Information:
InvenTree-Version: 0.17.12
Django Version: 4.2.17
Commit Hash: 9b759d2
Commit Date: 2025-05-21

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.1.21-v8+-aarch64-with
Installer: DOC

Active plugins: [{'name': 'InvenTreeBarcode', 'slug': 'inventreebarcode', 'version': '2.1.0'}, {'name': 'InvenTreeCoreNotificationsPlugin', 'slug': 'inventreecorenotificationsplugin', 'version': '1.0.0'}, {'name': 'InvenTreeCurrencyExchange', 'slug': 'inventreecurrencyexchange', 'version': '1.0.0'}, {'name': 'InvenTreeLabel', 'slug': 'inventreelabel', 'version': '1.1.0'}, {'name': 'InvenTreeLabelMachine', 'slug': 'inventreelabelmachine', 'version': '1.0.0'}, {'name': 'InvenTreeLabelSheet', 'slug': 'inventreelabelsheet', 'version': '1.0.0'}, {'name': 'DigiKeyPlugin', 'slug': 'digikeyplugin', 'version': '1.0.0'}, {'name': 'LCSCPlugin', 'slug': 'lcscplugin', 'version': '1.0.0'}, {'name': 'MouserPlugin', 'slug': 'mouserplugin', 'version': '1.0.0'}, {'name': 'TMEPlugin', 'slug': 'tmeplugin', 'version': '1.0.0'}]

### Relevant log output

```shell

```

GoryMoon:
This fix moves the if statement inside the for loop to allow the removal of temporary settings to still run.

Fixes #9702

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -876,22 +876,22 @@ def export_records(
 
     data_out = []
 
-    if include_permissions is False:
-        for entry in data:
-            model_name = entry.get('model', None)
+    for entry in data:
+        model_name = entry.get('model', None)
 
-            # Ignore any temporary settings (start with underscore)
-            if model_name in ['common.inventreesetting', 'common.inventreeusersetting']:
-                if entry['fields'].get('key', '').startswith('_'):
-                    continue
+        # Ignore any temporary settings (start with underscore)
+        if model_name in ['common.inventreesetting', 'common.inventreeusersetting']:
+            if entry['fields'].get('key', '').startswith('_'):
+                continue
 
+        if include_permissions is False:
             if model_name == 'auth.group':
                 entry['fields']['permissions'] = []
 
             if model_name == 'auth.user':
                 entry['fields']['user_permissions'] = []
 
-            data_out.append(entry)
+        data_out.append(entry)
 
     # Write the processed data to file
     with open(target, 'w', encoding='utf-8') as f_out:


SchrodingersGat (approved):

matmair:
Can we add test cases that test that all paths function as documented/intended?

SchrodingersGat:
@GoryMoon thanks for the contribution!

Issue #9710, opened 2025-05-29
[FR] Currency search should also match ISO 4217


Jorropo:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

When selecting a currency by habit I always type `USD` or `CNY` or `EUR`.
However (except € due to luck) `CNY` and `USD` brings no result.
That is because the search bar is matching against the strings `US Dollar` and `Chinese Yuan`.

### Suggested solution

The search should search not only against the currency strings but also their ISO 4217 codes.

### Describe alternatives you've considered

We could rename all currencies to their ISO 4217 codes, but someone could argue it's ugly or that it breaks their habits and I don't care about what text is displayed.

### Do you want to develop this?

- [x] I want to develop this.

SchrodingersGat:
@Jorropo this is a great idea and should be implemented. Would you like to tackle this?

matmair:
We use Django-money which uses babel for localized formatting. Not sure what the performance implication of doing something like this are but it should be a focus of the implementation to keep impact on api calls (especially OPTIONS) low

SchrodingersGat:
Currency selection relies on the text displayed in the `<select>` element - so simply appending the currency code to the text might be sufficient here. We should be able to implement as a pure front end change.

Jorropo:
After sleeping on it, I like the magic part of just typing `USD` and having `US Dollar` showing up but it's poor discoverability, new users would need to guess or randomly stumble onto that feature.

I think it makes more sense if instead we do `$CURRENCY_NAME ($CURRENCY_ISO4217)` so in the list it would show as `US Dollar (USD)`, `Chinese Yuan (CNY)`, ...

matmair:
@Jorropo are you developing this (ie can we assign this to you) and do you need pointers to starting material?

Jorropo:
No, I've been busy actually using inventree.
I'll do it if no one else have done it within a couple weeks.
Anyone else feel free to do earlier.

SchrodingersGat:
See incoming PR - https://github.com/inventree/InvenTree/pull/9717

PR #9717, opened 2025-06-02
Include currency code in display text

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/9710

e.g.

image


matmair:
This might break automations due to the api changes and should be called out in the release notes

PR #9718, opened 2025-06-02
[Feature] Override global settings

SchrodingersGat:
Allow settings values to be specified via the environment variables or config file.

- Override setting values
- Prevent editing by users at runtime


src/backend/InvenTree/InvenTree/settings.py
```diff
@@ -1395,6 +1395,18 @@
...
+# If provided, these values will override any "global" settings (and prevent them from being set)
+GLOBAL_SETTINGS_OVERRIDES = get_setting(
+    'INVENTREE_GLOBAL_SETTINGS', 'global_settings', typecast=dict
+)
+
+if len(GLOBAL_SETTINGS_OVERRIDES) > 0:
+    logger.info('Global settings overrides: %s', str(GLOBAL_SETTINGS_OVERRIDES))
+    for key, value in GLOBAL_SETTINGS_OVERRIDES.items():
+        # Set the global setting
+        logger.debug('- %s = %s', key, value)
```
matmair: We might expose secrets via this; maybe some masking should be applied

src/backend/InvenTree/common/apps.py
```diff
@@ -37,3 +42,24 @@ def clear_restart_flag(self):
                     set_global_setting('SERVER_RESTART_REQUIRED', False, None)
         except Exception:
             pass
+
+    def override_global_settings(self):
```
matmair, line 46: Wouldn't it be saver / easier to check for this when settings are read? The proposed change could make the system quite hard to debug.

src/backend/InvenTree/InvenTree/settings.py
SchrodingersGat: good point - this has been fixed now.

src/backend/InvenTree/common/apps.py
SchrodingersGat: This approach is used because often the settings are bulk read from the database, which bypasses a lot of the python model functionality. So, by using this approach we ensure that the settings are "overridden" in the database itself.

matmair: I think this has great potential to confuse the user. The docs currently just state that one might override settings globally at 2 places and then point at each other in a circular fashion.
Maybe we need to restructure the settings docs to be more clear about how configuration values are read and how to debug them.

SchrodingersGat: I've made some improvements to the docs here

PR #9721, opened 2025-06-02
Notification Improvements

SchrodingersGat:
This PR makes some adjustments / improvements to the notifications system.

- Allow disabling of notification system entirely
- Allow (some) selection of which notifications get sent
- Improved documentation
- Fill last remaining "placeholder" panel in the UI


src/backend/InvenTree/part/test_part.py
```diff
@@ -811,15 +811,19 @@ def _notification_run(self, run_class=None):
         self.part.set_starred(self.user, True)
         self.part.save()
 
-        # There should be 1 (or 2) notifications - in some cases an error is generated, which creates a subsequent notification
-        self.assertIn(NotificationEntry.objects.all().count(), [1, 2])
+        # There should be 0, 1 or 2 notifications - in some cases an error is generated, which creates a subsequent notification
+        self.assertIn(NotificationEntry.objects.all().count(), [0, 1, 2])
```
matmair: Doesn't this defeat the purpose of the assertation?

src/frontend/src/components/nav/Header.tsx
```diff
@@ -145,24 +145,26 @@ export function Header() {
...
-                <ActionIcon
-                  onClick={openNotificationDrawer}
-                  variant='transparent'
-                  aria-label='open-notifications'
-                >
-                  <IconBell />
-                </ActionIcon>
-              </Tooltip>
-            </Indicator>
+            {globalSettings.isSet('NOTIFICATIONS_ENABLE') && (
```
matmair, line 148: How are we notifying superusers about errors with this disabled?

src/frontend/src/pages/Index/Settings/SystemSettings.tsx
```diff
@@ -113,15 +110,13 @@ export default function SystemSettings() {
...
-          <Stack>
-            <Alert
-              color='teal'
-              title={t`This panel is a placeholder.`}
-              icon={<IconInfoCircle />}
-            >
-              <Text c='gray'>This panel has not yet been implemented</Text>
-            </Alert>
-          </Stack>
+          <GlobalSettingList
```
matmair, line 113: Where are the user specific notification settings?

src/frontend/src/pages/Notifications.tsx
```diff
@@ -125,6 +129,11 @@ export default function NotificationsPage() {
     ];
   }, [unreadTable, readTable]);
 
+  if (!globalSettings.isSet('NOTIFICATIONS_ENABLE')) {
```
matmair, line 132: How are we communicating system errors without this?

matmair (changes requested):

matmair:
This PR removes (indirectly) the option to notify superusers about errors that occurred.
The settings pane is missing user specific settings so not really solving the placeholder.

SchrodingersGat:
> This PR removes (indirectly) the option to notify superusers about errors that occurred.

Good point. How should we handle this?

1. "error" notifications are always sent, ignoring the "global notifications" rule
2. Add a "error notifications" category, which can be turned on while all others are turned off
3. Another approach?

SchrodingersGat:
> The settings pane is missing user specific settings so not really solving the placeholder.

That's in the "User Settings" page already:

image

src/backend/InvenTree/part/test_part.py
SchrodingersGat: I  think this check is no longer needed.

src/frontend/src/pages/Notifications.tsx
SchrodingersGat: They are still going to be visible in the "errors" table in the UI

matmair: Yes but one has to explicitly know that it exhausts and that there might be errors as there are no indicators before you explicitly enter that pane in admin ui.
Not sure this is helpful to new users

matmair:
I would say option 1

I am not really grasping the value that this change intends to deliver as notifications can be granulated turned off already

matmair:
Looking at this again - this panel is still missing
grafik

SchrodingersGat:
> Looking at this again - this panel is still missing

Yes, this has never yet been implemented in the new user interface. I'll have to look into that.

SchrodingersGat:
Closing this, replaced by https://github.com/inventree/InvenTree/pull/9735

PR #9728, opened 2025-06-02
Fix style

SchrodingersGat:
Small style fix


matmair:
@SchrodingersGat can we merge this?

PR #9730, opened 2025-06-03
[feature] Disable pricing task

SchrodingersGat:
This PR adds option to disable automatic update of part pricing. When this option is disabled, pricing is only recalculated when requested by the user. Default = `True` (preserving current behavior)

This is the first part of an "overhaul" of the pricing system.

### TODO

- [x] Update documentation


jack-fdrv:
In the ideal scenario it will be cool to do not add those task to the q rather then just add an empty task, which we later skip to execute. 
Right now after we do some get all parts and get all BOMs we see 2-3000 bg tasks spikes. And most of them are notifications.

SchrodingersGat:
@jack-fdrv that is the intent here - with this "PRICING_AUTO_UPDATE" setting disabled, pricing tasks would never get generated, and pricing would have to be manually refreshed when required.

Issue #9571, opened 2025-04-24
[Bug] Transition Method Issues


SchrodingersGat:
## Problem description

Plugins which register a "TransitionMethod" do not need to be enabled for that method to be active. The transitionmethod is applied even if the plugin is disabled.

### Steps to Reproduce

1. Activate the `SampleTransitionPlugin` sample plugin - and then disable it.
2. Try to mark a ReturnOrder as "complete"
3. Cannot "complete" the ReturnOrder - as the provided `ReturnChangeHandler` prevents it

### Expected behaviour

A plugin which provides this functionality must be *enabled* to do so!

## Discussion 

This issue raises some other points about how "Transition Methods" are implemented which I am not happy about.

### Activation

Plugins should *NOT* be able to have any effect on the system if they are not activated.

### Registration

To *register* a transition method, we simply loop through all available classes which inherit from the `TransitionMethod` class. This bypasses any plugin activation checks, and allows *any* code on the system to inject custom state transition behavior without observability.

### Collection

Transition methods are only collected at process startup, in `InvenTree/apps.py`. So, if we fix the issues above, we also need to ensure that we re-collect the methods whenever we reload the plugin registry.

### Mixin

I think that we should refactor this and adda new *mixin* class for plugins and collect transition methods by explicitly iterating through active plugins which implement that mixin type.

This is:

a) In-line with how we implement other plugin functionality
b) Safer as it validates plugin active status first

## Notification Method Registration

I see that the same issues exist with `NotificationMethod` - so we will have to fix problems there too..

This is a critical issue we need to address for 1.0 - and will be a breaking change for plugins which implement these features currently.

## Target Goals

I think we should:

- Refactor the `TransitionMethod` and `NotificationMethod` classes to be plugin mixins
- Load them with the "normal" plugin mixin loading strategy

matmair:
At least I and the default notification plugin provide more than 1 method per plugin so the normal mixin route is not really feasible

SchrodingersGat:
Either we could allow the "mixin" plugin class to return a set of multiple methods, perhaps?

Or a single "handler" could have logic for handling multiple transition types...

Either way, not insurmountable

PR #9735, opened 2025-06-03
[Refactor] Notification plugins

SchrodingersGat:
### Breaking Change

This PR represents a *breaking change* to any custom plugin which implements notification functionality. To continue to function, any such plugin must be reworked to follow the new approach:

- Each custom notification method must be implemented as a separate plugin
- Each custom notification method must inherit the `NotificationMixin` class
- Refer to the 3x built-in notification plugins for examples

### Details

This PR is a refactor of how notification methods are defined and collected.

Currently, it is a separate collection mechanism to plugins, and loading the notification methods is complex and somewhat "brittle" (compared to the plugin registry which has been optimized),

The intent here is to just make each notification method a plugin, which is then in line with our general concept of plugins. It also ensures that notification methods cannot be "collected" from plugins which are *installed* but not *active* (which is unfortunately the case currently).

### Feature - Plugin User Settings

It introduces a new concept, a "per user" setting structure against plugins. This replaces the "NotificationSetting" which was stored once per user against each type of notification. The new approach is more generic as *any* plugin can store per-user settings.

### Enhancements

- Improve query efficiency when loading and checking plugin registry
- Fix for "stock required for build" event - pass to the notification systems now

### Tasks

- [x] Split built-in notification methods into separate plugins
- [x] Documentation for new built-in plugin types
- [x] Reimplement solution for per-user-per-plugin settings
- [x] Fix unit testing
- [x] Check that email notifications still work
- [x] Check that UI notification still work
- [x] API unit tests for new plugin-user-setting 
- [x] Playwright tests for configuring plugin user settings
- [x] Fix remaining "notification settings" user settings panel

### Related Issues

- https://github.com/inventree/InvenTree/issues/9571
- Replaces https://github.com/inventree/InvenTree/pull/9721


src/backend/InvenTree/common/notifications.py
```diff
@@ -451,122 +262,26 @@ def trigger_notification(
...
-
-        elif hasattr(method, 'send'):
-            for target in method.targets:
-                if method.send(target):
-                    success_count += 1
-                else:
-                    success = False
+        # Skip if the plugin is *not* in the "delivery_methods" list?
+        if delivery_methods and plugin.slug not in delivery_methods:
+            continue
```
matmair: ie every trigger now needs to list all intended notification methods -> all plugins that trigger notifications need to be refactored because there is no generic filtering peruser/global anymore?

matmair (commented):
TBH there are so many lost / fundamentally changed mechanisms that I can not compare this to the current implementation as it just shares some entry points and names.
This means a new write of every even slightly complex notification method plugin and refactors for all plugins that trigger notifications.

SchrodingersGat:
@matmair fair comments - I'll see if I can adjust this to reduce the number of changes required

src/backend/InvenTree/common/notifications.py
SchrodingersGat: If "delivery_methods" is not provided, then the default is to send the notification to all plugins. You only need to provide that argument if you want to restrict the plugins which receive the notification.

matmair:
Not sure why, but we are losing a bunch of coverage on core plugin and part code

SchrodingersGat:
I think the difference in code coverage here is due to deleted lines in the codebase....

In any case, @matmair this is ready for review now. The key points are:

- Notification methods are now delivered by "normal" plugin approach
- Yes, this is a breaking change (which is why I want to get it done now, before 1.0.0)
- Plugins can now specify "user-specific settings". This replaces the concept of "notification setting" which was specific to a notificationmethod

src/backend/InvenTree/templates/email/build_order_required_stock.html
```diff
@@ -25,9 +25,9 @@
         <a href='{{ line.link }}'>{{ line.part.full_name }}</a>{% if line.part.description %} - <em>{{ line.part.description }}</em>{% endif %}
     </td>
     <td style="text-align: center;">
-        {% decimal line.required %}
+        {% decimal line.required %} {% if part.units %} [{{ part.units }}]{% endif %}
     </td>
-    <td style="text-align: center;">{% decimal line.available %}</td>
+    <td style="text-align: center;">{% decimal line.available %} {% if part.units %} [{{ part.units }}]{% endif %}</td>
```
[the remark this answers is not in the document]
SchrodingersGat: "part" is made available as part of the global context for this report type.

matmair:
TBH I am still not a fan of this change, I feel like notification settings should be a first-party component just like other settings. Moving them to the plugins makes them far more obscure.
The rest (except coverage) looks fine.

matmair:
The new plugin user settings seem to not be covered fully, there seems to be stuff missing on the APIs

SchrodingersGat:
> TBH I am still not a fan of this change, I feel like notification settings should be a first-party component just like other settings. 

I see what you mean, but the plugin specific notification settings are controlled exclusively by the plugins which implement them.

What about a compromise:

### First-Party Settings

- Add some specific "first party" user settings for notifications (such as "enable emails" / "enable UI notifications")
- First party plugins reference these settings, rather than their own "plugin specific" settings
- These settings are placed in the "user notification settings" panel
- Any other "third party" plugins which provide notification support will have their settings appear in the "plugin settings" panel

### Edit

I came up with a cleaner way - display all user settings for notification plugins in the "user notifications" tab:

image

This approach also works nicely for global plugin settings:

image

SchrodingersGat:
@matmair coverage has been increased across new features, and the notification settings now appear in their own panel

Issue #8562, opened 2024-11-26
Parameter Table Improvements


matmair:
Enhance the Parameter Table to enable dynamic filtering through parameters and stock information to enable dynamic part selection. Support for searching and filtering with an intuitive search syntax.

Issue #7346, opened 2024-05-26
Search by Parameter Value


SchrodingersGat:
I'm a bit confused, and maybe just misunderstanding, but I still cannot properly filter on parameter values. Both on my own server, and on the demo server, if I search for text that's in one of the parameter columns (but NOT in the part name/description), then those parts don't come up.

For example, on the demo server if you search for 'canvas' on the Parameters page, only one result comes up that has 'canvas' in the name. However there are multiple parts that have canvas as a parameter value. Clicking the sort arrow on a column properly sorts, but I can't figure out how to filter down by specific  parameter values.

There are also no params available in the filter dropdown, as eeintech showed in their screenshot above.

_Originally posted by @second-string in https://github.com/inventree/InvenTree/issues/4823#issuecomment-2132244334_

SchrodingersGat:
The idea here is to add the ability to search by "parameter value" in the parameteric part table

LavissaWoW:
Yes! Preferably, I'd love if we make it so that it's like any type of webshop where you get a list of filters containing all parameters of that type.

So, say you have a parameter called "Screw size", then instead of having to type "M3", "4/40" etc, you can just select it from a table that's compiled from available parameter values.

second-string:
@LavissaWoW's comment is a definite 'nice to have' for me as well, but I absolutely am just okay with a raw text input that matches on all parameters for any parts in the filtered list.

SchrodingersGat:
I'm going to open this one up for funding if people want to support getting this feature available sooner

wolflu05:
> @LavissaWoW's comment is a definite 'nice to have' for me as well, but I absolutely am just okay with a raw text input that matches on all parameters for any parts in the filtered list. 

That sounds like #5576

mylesagray:
I threw $50 in on this issue, would love to see this!

SchrodingersGat:
@mylesagray appreciate it! Currently a couple of other features I am giving priority to but this is also high on my list.

SchrodingersGat:
Hey @mylesagray unfortunately polar.sh have removed their issue funding feature - meaning that the funding you committed against this issue has been "removed". This issue is still in the pipeline, perhaps you would still be willing to sponsor via other methods (but no pressure)

Issue #4851, opened 2023-05-18
[Feature] Filter based on parameter value


SchrodingersGat:
After https://github.com/inventree/InvenTree/pull/4823 is completed, add ability to define custom filters for the 'parametric part table' based on parameter value

e.g.

- `?param_3_gt=9000` : Find parts with length > 9,000mm
- `?param_4=0402` : Find components with package `0402`

Support operators like:

-eq (equality)
- lt (less than)
- gt (greater than)

Would have to use some complex queryset annoations on the server (much like in https://github.com/inventree/InvenTree/pull/4823) to make this performant

agooddaytowork:
Hi @SchrodingersGat 
Have you been working on this? I really need this feature to be implemented.

In case you have not worked on this feature, I would like to implement this my self locally first. Could you drop more guidelines on where I should start?

SchrodingersGat:
@agooddaytowork no work on this to date, so if you want to take it on, that's great!

### References

As a starting point, look at these recent PRs which have added parameter filtering, and parameter units:

- https://github.com/inventree/InvenTree/pull/4823
- https://github.com/inventree/InvenTree/pull/4935
- https://github.com/inventree/InvenTree/pull/4854

Currently we support:

- Sorting parts by specific parameter value
- Filtering parts which have a particular parameter value associated

also check out the `filter_parametric_data` function which was implemented in the PRs above:

https://github.com/inventree/InvenTree/blob/20b59c35752fdcae25578ab5399efaa96080f27f/InvenTree/part/api.py#L1217

Let me know if you need any other pointers! Be very keen to see this feature get off the ground.

SchrodingersGat:
Not stale. @agooddaytowork did you ever get anywhere with this? Would be keen to see this working :)

agooddaytowork:
@SchrodingersGat

I made some updates but not going anywhere near what I want. I'm snuggling to get used to the UI part. 
you could take a look at this PR. https://github.com/agooddaytowork/InvenTree/pull/1/files

matmair:
@agooddaytowork the API work looks intresting.
Do you mind opening a PR against (this) upstream repo so we can discuss and maybe build a UI around it?

SchrodingersGat:
Not stale. @agooddaytowork any updates on this from your end?

SchrodingersGat:
Not stale. Adding to horizon milestone

PR #9739, opened 2025-06-05
[Feature] Filter by parameter

SchrodingersGat:
This PR introduces the feature to allow parts to be filtered by the value of the associated parameter(s)

### Related Issues

- Closes https://github.com/inventree/InvenTree/issues/8562
- Closes https://github.com/inventree/InvenTree/issues/4851
- Closes https://github.com/inventree/InvenTree/issues/7346


tasks.py
matmair, line 1: Was this change intended to be included?

SchrodingersGat: Just as a convenience, `invoke dev.shell` rather than having to navigate to the django dir. It's "hidden" behind the dev flag.

second-string:
Thank you!!

Issue #619, opened 2020-01-23
Export cascading BOM


SchrodingersGat:
Currently a BOM export is flat (only show top-level BOM items).

However it would be useful to include sub-assembly information in the BOM

Make this a selectable option?

sihawken:
This would be a fantastic addition to InvenTree. Definitely something I'd really like to have. 

I'm currently working for a start-up. We do a lot of prototyping, 3D printing, etc. and we've been using InvenTree to organize all of our part numbers and generate the bill of materials for manufacturing.  It's been great, super easy to use and to organize the information. 

A multi-level BOM is something I really, really am aching for. Especially for the engineering team looking at different projects, and at a glance seeing what the project contains with the relevant information. This would make finding the information much more efficient.

SchrodingersGat:
@sihawken - see https://github.com/inventree/InvenTree/pull/641

PR #9759, opened 2025-06-11
feat: Cascading BOM with export option #619

nlaubal:
Creation of a button to generate a bill of materials in which all levels are visible from a single element. It has an export option that allows you to select a CSV format in which you can choose the columns to export, as well as an option to ‘consolidate’ the bill of materials, which means that you export a bill of materials without duplication, but which adds the number of times the object is required.
This BOM system then allows you to anticipate potential shortages, better organise your stock and have a better overview of the creation process.

### DIFF
--- a/src/frontend/src/locales/en/messages.po
+++ b/src/frontend/src/locales/en/messages.po
@@ -10821,3 +10821,99 @@ msgstr "Read the docs"
 #: src/views/MobileAppView.tsx:35
 msgid "Ignore and continue to Desktop view"
 msgstr "Ignore and continue to Desktop view"
+
+#: src/tables/part/BomTable.tsx
+msgid "Part is Locked"
+msgstr "Part is Locked"
+
+#: src/tables/part/BomTable.tsx
+msgid "BOM cannot be edited - part is locked"
+msgstr "BOM cannot be edited - part is locked"
+
+#: src/tables/part/BomTable.tsx
+msgid "Flat BOM View"
+msgstr "Flat BOM View"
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@nlaubal the existing export system already allows for exporting a complete "top to bottom" BOM for an assembly - including subassemblies.

Can you please show some screenshots of the proposed UI changes?

Noowaah:
home
flatbom
export

matmair:
1) Please remove the translation changes, that is handled automatically
2) Please add docs for the feature you added

SchrodingersGat:
@nlaubal the ability to "expand" the BOM is definitely needed here (we had it in the legacy UI for e.g) however I would request some changes here:

1. Remove the custom "export" functionality in the code. There is a custom (server side) export framework which already allows top-to-bottom BOM export. **And** it can be extended via user plugins if required.

2. Display nested BOMs as nested items

Refer to the docs here - https://icflorescu.github.io/mantine-datatable/examples/nested-tables/

It should be possible to expand each BOM row to show the subassembly BOM. Then, the user can also "collapse" that row. And for convenience we can add an "expand all subassemblies" button.

Basically, let's use the UI functionality already provided! Additionally, expanding the subassemblies as required is *much* more performant than fetching all BOM data. For a BOM with a very deep heirarchy this could be a very long time spent requesting bulk data.

SchrodingersGat:
Replaced by https://github.com/inventree/InvenTree/pull/10210

Issue #9760, opened 2025-06-11
[FR] Allow opening creation modal from related model field


wolflu05:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

Its always time consuming to create a new related model when I am just in a form with a related model field and I want to select the newly created.

### Suggested solution

When using a RelatedModelField in PUI, this should render a + button to open an dialog and create a new related model directly from the current form without needing to go the the corresponding page and open the dialog there.

![Image](https://github.com/user-attachments/assets/2ec63302-1545-40e2-839a-d13625a6adad)

### Describe alternatives you've considered

Always going to the correct part in PUI where we can add the model and then try that it refetches the newly created data.

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
@wolflu05 nice idea :) Not sure how easy it will be to implement but I'd support it for sure!

You want to tackle this one?

wolflu05:
Yeah, thats the not so easy part.

No, unfortunately I have no time to implement this now, but I hope I'll get the supplier thing done.

PR #9761, opened 2025-06-11
Supplier Mixin

wolflu05:
This Adds a supplier mixin so that plugin can provide functions to import data from supplier APIs.

## TODO

- [x] initial import wizard
- [x] make sure that it works again after closing
- [x] initial stock creation
- [x] allow importing only MP and SP
- [x] check todo comments
- [x] sample plugin
- [x] docs
- [x] tests

## Future Ideas

- Caching
- Handling of rate limits
- PO synchronization
- normalize parameter values
- allow creation of parameter templates (#9760)


src/frontend/src/forms/StockForms.tsx
```diff
@@ -100,6 +111,25 @@ export function useStockFields({
...
+    // Find the highest price break that is less than or equal to the quantity
+    const priceBreak = Object.entries(pricing)
+      .sort(([a], [b]) => Number.parseInt(b) - Number.parseInt(a))
+      .find(([br]) => quantity >= Number.parseInt(br));
+
+    if (priceBreak) {
+      setPurchasePrice(priceBreak[1][0]);
+      setPurchasePriceCurrency(priceBreak[1][1]);
+    }
+  }, [pricing, quantity]);
```
wolflu05, line 125: @SchrodingersGat do I understand this wrong how this form framework works? I want to hook up the provided price breaks to the purchase price input field. Meaning if I change the quantity, it should auto fill the purchase price from the provided price breaks. I have debugged this issue like 2h now and still cannot figure out why the price does not get updated in the field, even now the field definition updates the value correctly (after the change in `ApiForm.tsx`). Do you have any idea here?

wolflu05: You can test this now with the integrated sample plugin very easily. Just go to a part table and open the new import wizard there next to the create new part button. Then search for e.g. `M5` and import any bolt. When you are at the stock step, click the plus icon and update the quantity and see that the pricing does not change. However when logging the stockItemFields to the console, the value actually updates.

wolflu05:
And do you have any idea, why the PUI tests fail, I havent touched them in a while and I do net get anything from the error. There seems to be tests failing where I havent changed anything... I would really appriciate some pointers on how to fix them.

Otherwise this PR is ready for a first review.

matmair:
without looking at it deeply - there is a lot of coverage on the backend missing

src/backend/InvenTree/plugin/base/supplier/api.py
```diff
@@ -0,0 +1,200 @@
...
+from plugin import registry
+from plugin.plugin import PluginMixinEnum
+
+from .serializers import (
+    ImportRequestSerializer,
+    ImportResultSerializer,
+    SearchResultSerializer,
+)
+
+# from .supplier import ImportParameter, PartNotFoundError
```
matmair: Q(nb): is this intended to be commented out?

src/backend/InvenTree/plugin/base/supplier/mixins.py
```diff
@@ -0,0 +1,217 @@
...
+            'model_filters': {'is_supplier': True},
+            'required': True,
+        }
+
+    @property
+    def supplier(self):
+        """Return the supplier company object."""
+        return company.models.Company.objects.get(
+            pk=self.get_setting('SUPPLIER', cache=True)
+        )
```
matmair: C: this will fail with a big error if the setting is wrong

src/backend/InvenTree/plugin/base/supplier/mixins.py
```diff
@@ -0,0 +1,217 @@
...
+        """Data class to represent a parameter for a part during import."""
+
+        name: str
+        value: str
+        on_category: Optional[bool] = False
+        parameter_template: Optional[part_models.PartParameterTemplate] = None
+
+        def __post_init__(self):
+            """Post-initialization to fetch the parameter template if not provided."""
+            if not self.parameter_template:
```
matmair: Q: Why are these in the mixin itself and not module context? Is there a need to have separate instances per plugin?

src/backend/InvenTree/plugin/base/supplier/mixins.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def create_related_parts(
+        self, part: part_models.Part, related_parts: list[part_models.Part]
+    ):
+        """Create relationships between the given part and related parts."""
+        for p in related_parts:
+            try:
+                part_models.PartRelated.objects.create(part_1=part, part_2=p)
+            except ValidationError:
+                pass  # pass, duplicate relationship detected
```
matmair, line 177: Q(nb): do these need to be overrideable or could they be placed in the module scope?

matmair (approved):
Very good quality, as always. A few questions / nitpicks; nothing really blocking.
I think a few lines of user docs on this would be good too - if you have time.

This is a great start, there are a few things that I think could be more sketched out (in future PRs, by anyone):
- Caching
- Handling of rate limits
- oAuth (digikey and a few other have that)
- parameter normalizing among suppliers
- data structures to handle multiple supplier accounts

src/backend/InvenTree/plugin/base/supplier/mixins.py
wolflu05: As python classes inside classes, are nothing more special than just having the abillity to call them by `SupplierMixin.ImportParameter`, I have thought this is just an easy way to expose them via the plugin interface without needing to export a bunch of things. If you dont like that approach, please show me in detail, where I should export them, and make sure, that this export does not break in the future.

src/backend/InvenTree/plugin/base/supplier/mixins.py
wolflu05: Not really, just thought this might be some helpers to simplify the importing process for plugin developers. Do you have a specific place in mind where I can place them?

src/backend/InvenTree/plugin/base/supplier/api.py
wolflu05: no, good catch, this is a leftover from me copy pasting and making it work in inventree core, have removed it.

src/backend/InvenTree/plugin/base/supplier/mixins.py
wolflu05: Good catch. I have catched it now and raise a PartImportError

src/backend/InvenTree/part/api.py
```diff
@@ -1700,6 +1701,23 @@ class PartParameterList(PartParameterAPIMixin, DataExportViewMixin, ListCreateAP
     ]
 
 
+class PartParameterBulkCreate(CreateAPIView):
+    """Bulk create part parameters.
```
SchrodingersGat: @wolflu05 it might be worth looking into a `BulkCreateMixin` class here, to make this a generic approach.

I recently added in a `BulkUpdateMixin` - https://github.com/inventree/InvenTree/pull/9313 - which employs a very similar approach. The intent here is to:

1. Reduce API calls for better throughput and atomicity
2. Utilize the already defined serializer classes for back-end validation

So, thoughts? A `BulkCreateMixin` would complement the `BulkUpdateMixin` and `BulkDeleteMixin` classess nicely!

src/backend/InvenTree/part/serializers.py
```diff
@@ -472,6 +472,29 @@ def __init__(self, *args, **kwargs):
...
 
+class TemplateUniquenessListSerializer(serializers.ListSerializer):
+    """List serializer that validates unique fields for bulk create."""
+
+    validators = [
+        InvenTree.serializers.ListUniqueValidator(unique_field_names=['template'])
+    ]
+
+
+class PartParameterBulkSerializer(InvenTree.serializers.InvenTreeModelSerializer):
```
SchrodingersGat: If we employ the `BulkCreateMixin` these extra classes may not be required...

src/backend/InvenTree/plugin/base/supplier/api.py
```diff
@@ -0,0 +1,200 @@
...
+        ],
+        responses={200: SearchResultSerializer(many=True)},
+    )
+    def get(self, request):
+        """Search parts by supplier."""
+        supplier_slug = request.query_params.get('supplier', '')
+
+        supplier = None
+        for plugin in registry.with_mixin(PluginMixinEnum.SUPPLIER):
+            if plugin.slug == supplier_slug:
```
SchrodingersGat: Is the "supplier" supposed to be the name of the supplier (e.g. `digikey`) or the slug for the plugin?

src/backend/InvenTree/plugin/base/supplier/mixins.py
```diff
@@ -0,0 +1,217 @@
...
+            'model': 'company.company',
+            'model_filters': {'is_supplier': True},
+            'required': True,
+        }
+
+    @property
+    def supplier(self):
+        """Return the supplier company object."""
+        return company.models.Company.objects.get(
+            pk=self.get_setting('SUPPLIER', cache=True)
```
SchrodingersGat: This will throw an error if a supplier is not linked - or has been deleted -  are you happy with that?

You could alternatively call `objects.filter(...).first()` to return `None` if there is no match

src/frontend/src/tables/purchasing/SupplierPartTable.tsx
```diff
@@ -175,13 +178,25 @@ export function SupplierPartTable({
...
         onClick={() => addSupplierPart.open()}
         hidden={!user.hasAddRole(UserRoles.purchase_order)}
+      />,
+      <ActionButton
+        key='import-part'
+        icon={<IconPackageImport />}
+        color='green'
+        tooltip={t`Import supplier part`}
+        onClick={() => importPartWizard.openWizard()}
+        hidden={!user.hasAddRole(UserRoles.part) || !params?.part}
```
SchrodingersGat: I think that this button should only be visible if there is an import plugin associated with the linked supplier.

src/frontend/src/forms/StockForms.tsx
SchrodingersGat: @wolflu05 I'm not setup to test this right now but the code you have written here looks like it should work. This is the approach we use elsewhere...

wolflu05:
Thanks for that feedback, that are some good ideas. I have collected them in the initial PR description now, but I want to make sure that I full understand them:

> parameter normalizing among suppliers

What exactly do you mean by that?

> data structures to handle multiple supplier accounts

I think that should be the goal of the SettingsMixin and UIMixin, also with the oauth stuff.

And still, I would really appreciate some help with the PUI tests, do you have any idea @SchrodingersGat ?

src/frontend/src/tables/purchasing/SupplierPartTable.tsx
wolflu05: Good idea to hide the button if there is no supplier mixin plugin available.

src/backend/InvenTree/plugin/base/supplier/mixins.py
wolflu05: Good catch, yes I'll raise a specific error instead now, as this is crucial for a supplier part creation in the import process.

src/backend/InvenTree/part/api.py
wolflu05: Good idea, I'll see what I can do and if I need any pointers.

SchrodingersGat: Can you submit that as a separate PR first? I'd like to be able to review that separately

src/backend/InvenTree/plugin/base/supplier/mixins.py
matmair: This way, we have multiples of each of these classes in the server process when more than 1 supplier plugin is used (which I think is the whole idea behind the mixin). Seeing how the plugin mechanism works that might be a problem and make interactions with classes unreliable as any plugin might override them and lead to hard to debug bugs.

I would move them into the module context of this file and export them in the `plugin.mixin` space, just like `BulkNotificationMethod`.

>  make sure, that this export does not break in the future.

I am not the one introducing unstable interfaces that are relied upon like they are stable. If this way breaks on the regular, I know who designed them

wolflu05:
@matmair Maybe you can add some clarification on my above questions?

30350n:
>> parameter normalizing among suppliers

> What exactly do you mean by that?

Different supplier parts (even from the same supplier) can feature vastly different formats for the same parameter:

image

This is a whole can of worms and could probably eat up an inifinite amount of time, so I never really looked into this much when developing `inventree-part-import`. Although I offer the option for users to manually implement things like this as part of their configuration, as I for example did [here](https://github.com/30350n/inventree_part_import_config/blob/c4fc6dc6d57fea1050977813f29da982b5d645a4/hooks.py#L105-L115).

Another thing that relates to this is the `Enforce Parameter Units` setting. I'm not sure how it works *now* exactly, but back in the day this would disrupt part creation for any part that failed the "unit check". (This would already include things like the `'0.125W, 1/8W'` in the above screenshot).

wolflu05:
If I understand that correct, this is already implemented. You may can test that using the static sample plugin.

matmair:
> What exactly do you mean by that?

https://en.wikipedia.org/wiki/Canonical_form#Computing ensuring that parameters and parameter units from different suppliers are handled in a standardized form

> If I understand that correct, this is already implemented. You may can test that using the static sample plugin.

I seem to have overlooked that, could you point that out?

> > data structures to handle multiple supplier accounts
> 
> I think that should be the goal of the SettingsMixin and UIMixin, also with the oauth stuff.

That would mean every plugin implements this itself, probably. In the future, I would prefer to have that as a core feature to ensure that the interface is the same between plugins and to make portability between plugins easier, might they become unmaintained / insecure.

src/backend/InvenTree/plugin/base/supplier/mixins.py
matmair: see https://github.com/inventree/InvenTree/pull/9761#discussion_r2148843962

wolflu05:
I see, have misunderstood this first, no this is not implemented, but a good idea for the future, I have added it to the list. For now, the process looks like this:

https://github.com/user-attachments/assets/1acc6e14-1693-49ce-8fdb-34a3c968e302

matmair:
Sorry about that, I thought it was unambiguous enough.

This is already great progress as is.

src/backend/InvenTree/plugin/base/supplier/api.py
wolflu05: That was the idea, at least for the key used here in the api to determine the correct plugin. Maybe later we can display some more pretty name in the UI for it instead of just the plugin name.

SchrodingersGat:
@wolflu05 sorry I have not yet had sufficient time to properly review this. I would like to push this to `1.1.0` as it requires a lot of testing and I'm keen to get our stable `1.0.0` release out soon.

wolflu05:
Too bad to hear; this is actually done, it just misses the bulk creation api refactor for part parameter bulk creation. I tried to look into it, but I don't think I have the deep knowledge to implement something generic here. I would be really happy if you could help me there.

And then there are the pui tests that I never really got working and know why they are failing as it's really annoying to push and wait for the ci, because I never have got them to run locally with that new auth cache system..

wolflu05:
Now that #10146 is merged, a BulkAddMixin is already there. I now need to find some time to refactor the current bulk creation mechanism of part parameters and fix the frontend tests. But this will take me some time (at least ~1 month) until I get there. If someone would like to help, I would be happy if someone could work on the bulk parameter creation in a separate PR.

src/backend/InvenTree/plugin/base/supplier/mixins.py
wolflu05: I have moved them now

src/frontend/src/forms/StockForms.tsx
wolflu05: I have tested this again a long time, but cannot get this to work, I would really appreciate any help here, as this is the last thing blocking this PR.

src/backend/InvenTree/plugin/base/supplier/mixins.py
wolflu05: After taking a closer look, I think these are fine here as helpers, like we already do with e.g. the render helpers for label printing plugins, ...

wolflu05 (commented):
This PR is done now from my side, I would appreciate if you could take a second review.

matmair:
@SergeoLacruz This would bring things to core that you provide via plugin(s); do you want and have time to review and maybe give feedback from your unique perspective given that you have written and run a similar integration? Would this reduce the effort you have to spend to maintain your plugin?

SergeoLacruz:
Thanks for the info. I will have a look. It might take some days.

SergeoLacruz:
Sorry for the long post. 
I had a look, read the docs, the example. Did not try it.
As far as I understand it eases the process of adding parts from a supplier by offering a unified way to do so. It allows to add parts, supplierparts, parameters and prices. This is a nice idea and helps for such integrations as I did with Mouser, Digikey and Farnell.
During my work most of the time was spent to figure out how the API of the supplier works, handling the access system, analyze how data comes back and bring this to a unified form. Adding data to Inventree is simple because it is well documented. We did a lot of experiments with adding parts automatically. The only method that turned out reliable is the SKU. Other searches show unreliable results. And thats the point: Parts in Inventree need to be 100% error free. Thats why our part
creation process for Inventree, schematic and layout is more or less manual and follows a strict workflow with several checks defined in Jira. Only the supplier part can be added automatically based on the supplier SKU. 
Adding parameters from a supplier is dangerous. They tend to return funny formats and units. Different values may show up for different suppliers. The only ground truth for part parameters is the manufacturer Datasheet or the manufacturers WEB API if there is one like Murata. There are solutions that analyze Datasheets and extract values using AI but we did not try reliability so far.
Whats the benefit?
Part creation is seldom and done by a small amount of people <<10. Normal users who use Inventree on a daily bases are many more >100. Thats why we decided to stop all automatic part creation trials except prices from existing SKUs
which are regularly updated which is already complicated enough when suppliers add geofencing for currencies. It is more cost effective to spend work on features the help the majority of the users.
The situation is different when you already have a large database, may be transferred from another system, and need to add supplier info to all of them. We had this nightmare years ago...
Wishes?
The plugin settings. For an easy supplier like Mouser you need already two keys.  All settings are global and so there is only one Mouser account for the whole company. This tends to get chaotic when shopping carts are created. So user specific plugin settings would be nice.  Some settings, like proxy accounts, shall be user specific. 
Just proceed with it. For certain usage scenarios it is a real benefit. Finally the responsibility for the data added to the system is up to the user.

matmair:
Thank you for your insights @SergeoLacruz?

wolflu05:
Thanks for your insights. It's already possible to just import a spp. I have also already build the system with later synchronization in mind. That's why there is a separate function for importing different parts. I imagine in future development to add some kind of periodic tasks that can sync things. But from my understanding this solves a different usecase currently than yours?

SergeoLacruz:
Syncronization is really helpful and finally needed to keep things up to date. I wrote something for that.
https://github.com/SergeoLacruz/inventree-supplier-sync
For supplierparts it syncs prices and lifecycle. I also reports if additional parts in the Mouser data base might match your part and reports those. But the final decision for adding and checking is up to the libarian. Also here the majority of the work is in the code that handles the supplier API.
IMHO Inventree cannot help with that part because it is very supplier specific and there are many... Mouser and Digikey might be special cases but anyhow I do not think that Inventree should prefer one by adding supplier specific features. 
This is up to the user and region specific. 
For lifecycle it is not perfect because this shall come from the manufacturer. But there are even more manufacturers out there than suppliers and only a few have APIs. We have manual checks here. 
Our major usecase is ordering. That is done very often compared to part creation.

matmair:
This would be worthy of a changelog entry too

wolflu05:
I have added it to the changelog now.

CHANGELOG.md
```diff
@@ -30,3 +31,4 @@ Our blog holds [a few articles](https://inventree.org/blog/2024/09/23/ui-roadmap
 
 
 Specific entries to the changelog will be kept for all stable channel minor releases, for changes in 1.0 please refer to the [blog posts](https://inventree.org/blog/2025/09/15/1.0.0) and the [milestone](https://github.com/inventree/InvenTree/milestone/17)
+p
```
matmair:
```suggestion

```
Not sure if this was intended?

matmair (approved):
LGTM; great work

CHANGELOG.md
wolflu05: Thanks, I think I accidentally typed this while merging

matmair:

SchrodingersGat:
@wolflu05 I'll get onto this in the next couple of days. Sorry, super swamped!

SchrodingersGat:
@wolflu05 firstly, great work, this is a huge undertaking and I think a very useful feature.

As this is such a big feature, there are guaranteed to be some issues we will have to track down over the next few release cycles. I hope that you can take charge of these?

SchrodingersGat:
Regarding documentation:

While you have documented the plugin mixin type itself, I think it would be good to also add a general documentation page regarding importing parts from a supplier. This should be in the "parts" documentation category, and provide some generic documentation about the requirements and process for importing parts:

- Install a plugin with the right mixin type
- Associate plugin with the supplier
- Where to find the UI elements
- What the process does

Happy for you to add that now, or we can make it a target for an additional PR?

SchrodingersGat:
Overall, happy with this, I think that (after maybe some additional docs) we can merge in and make sure it gets tested thoroughly before the next release. Hopefully we can also provide some supplier-specific plugins in the future?

wolflu05:
Thanks for your review. I have added some user docs now. Just let me know if there is something else we need to change.

SchrodingersGat (approved):

SchrodingersGat:
Nice addition, thanks! I apologise in advance for any changes that I will request after this is merged - it's hard to catch everything on such a large feature addition.

wolflu05:
That makes sense. Hope that I will find some time then, because I will be pretty busy for the next weeks.

PR #9768, opened 2025-06-12
feat: Add Delivery Address field to Purchase Orders

### LINKED ISSUE #9303: [FR] Delivery Address on Purchase Orders

### Problem statement

It would be great to allow a 'Delivery Address' field for purchase orders.

This field would allow you to then select any companies address and could then be used to generate a delivery address in the purchase order report.

As discussed here https://github.com/inventree/InvenTree/discussions/9219

### Suggested solution

The existing address field in POs should be renamed to something like 'Suppliers Address' and the new field should be something like 'Delivery Address'.

I'm currently getting around this by adding a 'shipping address' to the suppliers addresses list. I then select this suppliers 'shipping address' in the purchase order 'address' field drop down.

### Examples of other systems

_No response_

- [ ] I want to develop this.

Reza98Sh:
This pull request implements the feature requested in #9303 by adding a dedicated `delivery_address` field to Purchase Orders.

### Changes Implemented

**Backend:**
- [x] Created a new database migration (`0110_purchaseorder_delivery_address.py`) to apply the schema change.
- [x] Added a `delivery_address` `ForeignKey` field to the `PurchaseOrder` model, linking to `company.Address`.
- [x] Exposed the `delivery_address` field in the `PurchaseOrderSerializer` to make it available via the API.
- [x] Included `delivery_address` in the `autocomplete_fields` in `PurchaseOrderAdmin` for a better user experience in the admin panel.

**Frontend:**
- [x] Added the `delivery_address` field to the `usePurchaseOrderFields` hook, making it visible and editable in the Purchase Order forms.

Closes #9303


Reza98Sh:
If this method is not considered best practice, please let me know.
I am doing my best to improve it.

matmair:
@Reza98Sh there is a conflict in the migrations; please delete the ones you added and regenerate them

src/backend/InvenTree/order/models.py
```diff
@@ -666,6 +666,16 @@ def company(self):
...
     )
 
+    delivery_address = models.ForeignKey(
+        Address,
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+        verbose_name=_('Delivery Address'),
+        help_text=_('Delivery address for this order'),
+        related_name='+',
```
matmair: This is not practical for future use; please use a related name that makes sense when using the Adress model

matmair (changes requested):
Looks very good for a first PR

SchrodingersGat:
An implementation question: where / how do the users define an address for "themself" which can be used here?

One approach might be to create a "company" object which is the "internal" company, which can then have one or more addresses associated with it?

Any other ideas?

src/backend/InvenTree/order/models.py
Reza98Sh: I have changed it

Reza98Sh:
> An implementation question: where / how do the users define an address for "themself" which can be used here?
> 
> One approach might be to create a "company" object which is the "internal" company, which can then have one or more addresses associated with it?
> 
> Any other ideas?


Sorry I haven't any serious experience with InvenTree but I think It's good if there is option to create address without define company

SchrodingersGat:
@matmair thoughts on this? Does a set of "internal addresses" for "me" (which are not linked to a company object) make the most sense?

matmair:
There is no real inherent connection between Addresses and Companies but the way the current data model works there are a few functions that would need rethinking.

Mainly:
- shipping notes
- primary addresses

Having (1 or more) internal companies as references would be nice for a few of the open EPICs.
My suggestion for now: creating a special "locked/protected" company that represents the org that owns the InvenTree instance and link the addresses there.

matmair (approved):

matmair:
Still relevant

SchrodingersGat:
@Reza98Sh I'm moving this to the `1.2.0` release target - there are still some questions to address here, I will get back to review this soon

gunstr:
I have the need to add our own company in the header for Sales Ordrers and found this PR. Would it make sense to make it generic and create an internal company how "owns" the database and expose the same to all reports? I can see the need also for other kind of reports (e.g. stock reports) where it would be beneficial if the report included the company it's valid for in the header.

PR #9769, opened 2025-06-12
refactor (backend): move config files out of the source directories

### LINKED ISSUE #9756: moving config files out of the source directories
mixing config and sources in one dir makes change diffs more difficult and is confusing

matmair:
This adds a new error code if files are not in the recommended config directory and moves default configs out into a separate folder located in the root.

For now, it maintains compatibility but the next major should remove those.

Fixes #9756


matmair:
@SchrodingersGat ready for a review and merge with the lower baseline from https://github.com/inventree/InvenTree/pull/9849

SchrodingersGat:
Nice, this is an important improvement for 1.0.0

SchrodingersGat:
@matmair actually it looks like the warnings might be a little aggressive here?

e.g. in the devcontainer setup:

```bash
2025-06-24 17:13:50,433 WARNING INVE-W10 - Config for 'INVENTREE_SECRET_KEY_FILE' not in recommended directory '/home/inventree/config'.
2025-06-24 17:13:50,434 WARNING INVE-W10 - Config for 'INVENTREE_CONFIG_FILE' not in recommended directory '/home/inventree/config'.
2025-06-24 17:13:50,435 WARNING INVE-W10 - Config for 'INVENTREE_PLUGIN_FILE' not in recommended directory '/home/inventree/config'.
2025-06-24 17:13:50,436 WARNING INVE-W10 - Config for 'INVENTREE_OIDC_PRIVATE_KEY_FILE' not in recommended directory '/home/inventree/config'.
```

And the paths won't be set to that for the docker image either..

SchrodingersGat:
I have a further thought to this - can we change it so that there is a warning if any config file is located inside the source directory, but remove the warnings based on specific external location? The intent here is that config files should *not* be inside the source dir, and the warning should reflect that

matmair:
> @matmair actually it looks like the warnings might be a little aggressive here?

It warns once per wrong file destination, I do not think that is too aggressive. The other option is the novice user needing 4 attempts to clear this error

> And the paths won't be set to that for the docker image either..

That can easily be added; where it the canonical location for docker image configs?

> I have a further thought to this - can we change it so that there is a warning if any config file is located inside the source directory, but remove the warnings based on specific external location? The intent here is that config files should _not_ be inside the source dir, and the warning should reflect that

I can change that; the idea was to also reduce the maintainer burden of having to answer where the file should be moved, as it is now obviously communicated

SchrodingersGat:
The "canonical" paths are:

- **docker**: `/home/inventree/data/`
- **devcontainer:** `/home/inventree/dev/`

Although perhaps it would be a good idea to make devcontainer paths the same as for the default docker image?

SchrodingersGat:
> I can change that; the idea was to also reduce the maintainer burden of having to answer where the file should be moved, as it is now obviously communicated

We could say, that *if* the file is located within the source dir, we warn the user, and *also* provide information on the precise path where the file should be moved

Issue #8695, opened 2024-12-17
[PUI] Purchase order total cost not updated when adding extra line items


sur5r:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

When adding an extra line item to a purchase order, the total cost is not updated unless the pages is reloaded.

### Steps to Reproduce

1. Go to a purchase order
2. Add an extra line item with a price
3. Click on "Order details"
4. Observe the total cost not including the price of the extra line item

### Expected behaviour

The total cost in "Order details" gets updated without a page reload.

### Deployment Method

- [ ] Docker
- [ ] Package
- [x] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

InvenTree-Version: 0.18.0 dev
Django Version: 4.2.17
Commit Hash: ff69cf6
Commit Date: 2024-12-17
Commit Branch: master
Database: postgresql
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-6.1.0-13-cloud-amd64-x86_64-with-glibc2.36
Installer: GIT
nullActive plugins: [{"name":"InvenTreeBarcode","slug":"inventreebarcode","version":"2.1.0"},{"name":"InvenTreeCoreNotificationsPlugin","slug":"inventreecorenotificationsplugin","version":"1.0.0"},{"name":"InvenTreeCurrencyExchange","slug":"inventreecurrencyexchange","version":"1.0.0"},{"name":"InvenTreeLabel","slug":"inventreelabel","version":"1.1.0"},{"name":"InvenTreeLabelMachine","slug":"inventreelabelmachine","version":"1.0.0"},{"name":"InvenTreeLabelSheet","slug":"inventreelabelsheet","version":"1.0.0"},{"name":"DigiKeyPlugin","slug":"digikeyplugin","version":"1.0.0"},{"name":"LCSCPlugin","slug":"lcscplugin","version":"1.0.0"},{"name":"MouserPlugin","slug":"mouserplugin","version":"1.0.0"},{"name":"TMEPlugin","slug":"tmeplugin","version":"1.0.0"},{"name":"Brother Labels","slug":"brother","version":"1.0.0"}]

### Please verify if you can reproduce this bug on the demo site.

- [x] I can reproduce this bug on the demo site.

### Relevant log output

```shell

```

sur5r:
[...]

PR #9773, opened 2025-06-13
fix(frontend): Reload order details when changing line items (Closes: #8695)

sur5r:
Closes https://github.com/inventree/InvenTree/issues/8695

### DIFF
--- a/src/frontend/src/tables/sales/SalesOrderLineItemTable.tsx
+++ b/src/frontend/src/tables/sales/SalesOrderLineItemTable.tsx
@@ -57,7 +57,7 @@ export default function SalesOrderLineItemTable({
   editable
 }: Readonly<{
   orderId: number;
-  orderDetailRefresh: () => void
+  orderDetailRefresh: () => void;
   currency: string;
   customerId: number;
   editable: boolean;


SchrodingersGat:
@sur5r nice :) Can I request the following:

1. Also perform this update for the "extra line" table
2. Also implement for sales orders
3. Also implement for return orders

It's a nice clean fix so makes sense to fix it all at once :)

sur5r:
True, will do.

src/frontend/src/tables/general/ExtraLineItemTable.tsx
```diff
@@ -27,11 +27,13 @@ import {
 export default function ExtraLineItemTable({
   endpoint,
   orderId,
+  orderDetailRefresh,
   currency,
   role
 }: Readonly<{
   endpoint: ApiEndpoints;
   orderId: number;
+  orderDetailRefresh: any;
```
SchrodingersGat: One small thing here regarding typing:

Instead of:

```typescript
orderDetailRefresh: any;
```

it should be

```typescript
orderDetailRefresh: () => void
```

(please roll this out to all tables where you have added this parameter

SchrodingersGat (changes requested):

SchrodingersGat:
@sur5r very nice :) Just one small change needed regarding typing

src/frontend/src/tables/general/ExtraLineItemTable.tsx
matmair: I have added this

SchrodingersGat:
@sur5r thanks for this contribution 😁

sur5r:
And thanks @matmair for finishing it. Family vacation is a tricky situation to find time for oneself...

PR #9779, opened 2025-06-14
[bug] Status fix

SchrodingersGat:
Fixes issue with setting "status" for multiple items at once.

Instead of "fixing" the faulty serializer it has been replaced with a `BulkUpdate` mixin to remove unnecessary code duplictation.

Closes https://github.com/inventree/InvenTree/issues/9157

Issue #9157, opened 2025-02-22
Custom states does not work as expected


gunstr:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

I have created some custom states with OK as Logical State.

![Image](https://github.com/user-attachments/assets/d0c050dd-2e53-4483-9373-9e840e8f4626)

If I start with a part that have OK as state:

![Image](https://github.com/user-attachments/assets/abfb45c1-9f38-429d-b355-3f321f7b61d0)

and changed that to Lab Item for example the state is changed but the stock is indicated as Not Available

![Image](https://github.com/user-attachments/assets/04b8fc7b-9241-49cd-8688-3dfe13ab3218)

If I open the stock item in the admin interface I can see the custom status key entry is blank.

![Image](https://github.com/user-attachments/assets/d8ed9c4b-5906-48da-8ca3-0a1d301f4ea9)

If I enter the custom state value (102 in this case) in the field and go back to the stock list the part is available again and it all looks OK.

![Image](https://github.com/user-attachments/assets/602e8f6b-8c53-4a6b-92be-3f61da060081)

It's strange that the Custom State is rendered correctly even if the custom status key is blank but I guess there are more fields involved that I'm not aware of. The manual edititing I did in the admin interface might be completely wrong, but it seems to work... 

### Steps to Reproduce

Se above

Tried to reproduce on the Demo site but the database seems to be corrupted

### Expected behaviour

A Custom State linked to the OK status should be considered as available

### Deployment Method

- [ ] Docker
- [ ] Package
- [x] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

![Image](https://github.com/user-attachments/assets/76676817-d34e-4475-91c1-3ed33ef63297)

### Please verify if you can reproduce this bug on the demo site.

- [ ] I can reproduce this bug on the demo site.

### Relevant log output

```shell

```

gunstr:
Just found out that this happens when changing state directly from the stock item list (it was discovered when I did a bulk change). If I open each stock item individually and change state from the "edit" modal the Custom status key is updated as it should.

matmair:
I will look into this, I don’t think we have tests that cover this scenario.

gunstr:
Not really, the same happens also if just one stock item is selected from the table.

matmair:
[...]

PR #9781, opened 2025-06-14
[bug] stock status fix

SchrodingersGat:
- Alternative solution to https://github.com/inventree/InvenTree/pull/9779
- Prevent stock status from being changed directly on the "stock item" modal
- Closes https://github.com/inventree/InvenTree/issues/9157
- Closes https://github.com/inventree/InvenTree/pull/9779

### DIFF
--- a/src/backend/InvenTree/stock/models.py
+++ b/src/backend/InvenTree/stock/models.py
@@ -1333,7 +1333,7 @@ class StockItem(
         self.location = location
 
         if status := kwargs.get('status'):
-            self.status = status
+            self.set_status(status)
             tracking_info['status'] = status
 
         self.save()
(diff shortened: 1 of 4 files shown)


gunstr:
@SchrodingersGat very appreciated the you address this bug but

> Prevent stock status from being changed directly on the "stock item" modal

Does that mean you can no longer change status when you select "Edit" from a stock item details page or does the "stock item" modal refer to someting else?

image

Most importat is of course the fix to avoid a corrupted database when changing status from the stock table but it would be a bit of a limitation if the status can no longer be changed when editing a single stock item.

SchrodingersGat:
Closed by https://github.com/inventree/InvenTree/pull/9858

PR #9791, opened 2025-06-16
[UI] Query Improvements

SchrodingersGat:
This PR makes a number of changes to the various UI components which load data from the server, to reduce the number of queries which are fired off to the server on a page load.

## Stock Operations Modal

Fixes broken `useQuery` hook in the `stockOperationModal` component.

Any page that used this modal would (prior to this fix) submit a query requesting *all* stock information - which block the server for potentially multiple seconds.

- Fixes so that the query does not automatically fire
- Only fires when the modal is opened
- Only fire if a valid part ID is provided

## Serial / Batch Generators

- Fix these so that the queries only fire *if* the modal is opened
- There is no point running them if the user never opens the modal

## Table Loading

- Fix bug which caused each table to be fetched twice on initial load


matmair (approved):
LGTM; having Sentry or tracing somewhere in the QC stack would probably help discover these things. Should we look into that?

PR #9802, opened 2025-06-17
fix: package - various changes

matmair:
This fixes a few nit-picks I found when looking into recent reports regarding the package.

Some of these have already been addressed in master, I have tested this branch on my distribution channel - https://packager.io/gh/matmair/InvenTree/refs/installer-fail-on-update

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -65,7 +65,7 @@ jobs:
 
   javascript:
     name: Style - Classic UI [JS]
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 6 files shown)


SchrodingersGat (approved):

matmair:
@SchrodingersGat the jobs won't run anymore because of the depreciation of those actions runners; should we patch this out for the last few point releases?

SchrodingersGat:
@matmair yes I think we should fix it

SchrodingersGat (approved):

SchrodingersGat:
@matmair thanks for the backport of this

PR #9804, opened 2025-06-18
Stylish text component

SchrodingersGat:
Expose `<StylishText>` component to the plugin library, allowing plugins to easily recreate the visual styling based on user selected theme


SchrodingersGat:
Closing this out, the experiment failed and calling components like this does not work well. Will need further consideration.

Issue #8670, opened 2024-12-16
[1.0.0] Refactor "part scheduling" into plugin


SchrodingersGat:
Moving to a plugin allows for more configurability and offloads a messily integrated function which is not useful for all applications.

WIP Here: https://github.com/inventree/inventree-scheduling/

PR #9811, opened 2025-06-19
[Breaking] Remove part scheduling feature

SchrodingersGat:
The existing "part scheduling" feature has always been a "tacked on" feature which has not received great support and is also not what I would consider "general purpose". Some users may want to forecast differently, and it is not "one size fits all".

Thus, it is being removed in preparation for the `1.0.0` release. The functionality will instead be supported via the [stock forecasting plugin](https://github.com/inventree/stock-forecasting/pull/3)

Closes https://github.com/inventree/InvenTree/issues/8670


matmair (approved):
LGTM: This should be mentioned in the release notes

PR #9817, opened 2025-06-20
Company reports

SchrodingersGat:
Adds the ability to generate reports against a Company instance


matmair (approved):

src/backend/InvenTree/company/models.py
```diff
@@ -56,9 +54,21 @@ def rename_company_image(instance, filename):
...
 
+class CompanyReportContext(report.mixins.BaseReportContext):
+    """Report context for the Company model.
+
+    Attributes:
+        name: The name of the Company
+        primary_address: The primary address associated with the Company
+    """
+
+    name: str
```
wolflu05, line 65: Some attributes are missing here, and why not expose the complete company object?

src/backend/InvenTree/company/models.py
```diff
@@ -56,9 +54,21 @@ def rename_company_image(instance, filename):
...
+
+
 class Company(
     InvenTree.models.InvenTreeAttachmentMixin,
     InvenTree.models.InvenTreeNotesMixin,
+    report.mixins.InvenTreeReportMixin,
```
wolflu05, line 71: When adding this mixin, a short code for a company report needs to be defined.

wolflu05 (commented):
Unfortunately I had no time before to add those comments....

SchrodingersGat:
@wolflu05 ah, yes that was quite the oversight. Fix incoming in https://github.com/inventree/InvenTree/pull/9819

src/backend/InvenTree/company/models.py
SchrodingersGat: Isn't the short code for the barcode mixin?

src/backend/InvenTree/company/models.py
SchrodingersGat: See https://github.com/inventree/InvenTree/pull/9819

PR #9819, opened 2025-06-20
Fix missing context items

SchrodingersGat:
Bug fix for missing context items in Company reports.

Ref: https://github.com/inventree/InvenTree/pull/9817

### DIFF
--- a/src/backend/InvenTree/InvenTree/models.py
+++ b/src/backend/InvenTree/InvenTree/models.py
@@ -948,7 +948,7 @@ class InvenTreeBarcodeMixin(models.Model):
         This is used to generate a efficient QR code for the model type.
         It is expected to match this pattern: [0-9A-Z $%*+-.\/:]{2}
 
-        Note: Due to the shape constrains (45**2=2025 different allowed codes)
+        Note: Due to the shape constraints (45**2=2025 different allowed codes)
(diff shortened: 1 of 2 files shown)


src/backend/InvenTree/InvenTree/models.py
```diff
@@ -948,7 +948,7 @@ def barcode_model_type_code(cls):
         This is used to generate a efficient QR code for the model type.
         It is expected to match this pattern: [0-9A-Z $%*+-.\/:]{2}
 
-        Note: Due to the shape constrains (45**2=2025 different allowed codes)
+        Note: Due to the shape constraints (45**2=2025 different allowed codes)
```
wolflu05, line 951: This method is still not overridden... or do I oversee something?

matmair (approved):

src/backend/InvenTree/InvenTree/models.py
SchrodingersGat: That is required for barcode support, not for the ReportMixin. I think you are getting the two confused?

wolflu05: Oh yes your right, we don't need barcodes for a company. Sorry

wolflu05 (approved):

PR #9826, opened 2025-06-22
[bug] Fixes for plugin loading mechanism

SchrodingersGat:
Recently discovered than an error during plugin loading could cause the mutex to remain locked, meaning that any subsequent plugin loading attempts would fail.

### DIFF
--- a/src/backend/InvenTree/plugin/registry.py
+++ b/src/backend/InvenTree/plugin/registry.py
@@ -324,7 +324,11 @@ class PluginsRegistry:
             logger.debug('Skipping reload - plugin registry is currently loading')
             return
 
-        if self.loading_lock.acquire(blocking=False):
+        if not self.loading_lock.acquire(blocking=False):
+            logger.exception('Could not acquire lock for reload_plugins')
+            return
+
+        try:
             logger.info(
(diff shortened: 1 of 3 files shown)


src/backend/InvenTree/users/apps.py
```diff
@@ -35,11 +35,15 @@ def ready(self):
                 rebuild_all_permissions()
             except (OperationalError, ProgrammingError):
                 pass
+            except Exception as e:
```
matmair, line 38: This is a very broad catch

src/backend/InvenTree/plugin/registry.py
```diff
@@ -317,14 +318,23 @@ def reload_plugins(
...
-        if self.loading_lock.acquire(blocking=False):
+        if not self.loading_lock.acquire(blocking=False):
+            logger.exception('Could not acquire lock for reload_plugins')
+            return
+
+        # Reset the loading error state
+        if clear_errors:
+            self.errors = {}
+
+        try:
```
matmair, line 337: There is no coverage on these new guards that seem pretty vital. Any idea how we could cover them in a test? We are slowly but steadily losing coverage on the plugin mechanisms.

PR #9828, opened 2025-06-22
refactor(docs): restructure barcode plugin section

matmair:
make it clearer that these only support barcode parsing


SchrodingersGat (approved):

SchrodingersGat:
Thanks for the clarification

Issue #9827, opened 2025-06-18
Plugin Configuration documentation


bdmartens:
I am new to Inventree, trying to sort my way through getting everything up and running.  
It appears the plugin configuration documentation is scant, or at least incongruent with my install  (Docker, 0.17.13)
I'm trying to get the Digikey barcode plugin (?renamed to just "DigiKey Plugin") configured.
The stable docs don't mention the plugin at all.
The latest does:
https://docs.inventree.org/en/latest/plugins/builtin/digikey/
It took a while to see that the word 'supplier' was a link.
It is then implied that in order to scan barcodes, you need to create a supplier, then select it.

I then do this, create a supplier (website, phone number, email, etc - no API info), go back to plugin settings for the digikey plugin, set the supplier to digikey.  No futher fields are opened up to put in the ClientID or Client Secret.

Looking at other documentation for the inventory plugin, there apparently are configuration parameters for these.

I don't see forums elsewhere where this process has been answered already.  I don't see any youtube videos on this, and generic google searches don't seem to turn up relevant information.  As populating inventory with parts is a (presumably) common task, it would be nice if these docs could get fleshed out a bit more.  I'm not sure how new this version of the plugins may be.

Is it possible to get the digikey (& mouser, etc) plugin process fleshed out a bit more, including this hangup on how to activate the plugin to get barcode features working?

Once I get things working, being fresh with the install experience, I'm happy to suggest specific content for the docs, should that be helpful (and correct)

(Separately, for those who like to create youtube content, this could be a useful area to address.  There are overviews, and KiCad detailed videos, but nothing going through a basic setup, creating categories, suppliers, enabling barcode scanning, and adding parts)

SchrodingersGat:
> Looking at other documentation for the inventory plugin, there apparently are configuration parameters for these.

Please provide links here. None of our builtin plugins support API calls to a  supplier.

matmair:
I have transferred this as it has nothing to do with the website. @bdmartens please answer the question, the issue will auto close in about 3 months without answer

bdmartens:
[...]

PR #9829, opened 2025-06-23
refactor(backend)  rename barcode plugins

matmair:
Following development in https://github.com/inventree/InvenTree/pull/9761 and feedback in https://github.com/inventree/InvenTree/issues/9827 this renames the current barcode plugins.
Note: this only changes names on new instances of the plugin config.

The docs were already addressed in https://github.com/inventree/InvenTree/pull/9828

### DIFF
--- a/src/backend/InvenTree/plugin/builtin/suppliers/digikey.py
+++ b/src/backend/InvenTree/plugin/builtin/suppliers/digikey.py
@@ -13,6 +13,7 @@ class DigiKeyPlugin(SupplierBarcodeMixin, SettingsMixin, InvenTreePlugin):
     """Plugin to integrate the DigiKey API into InvenTree."""
 
     NAME = 'DigiKeyBarcodePlugin'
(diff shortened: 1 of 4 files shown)


30350n:
FYI, I choose the names like that back then to leave the option for these plugins to implement other supplier related functionality/interfaces ... so pretty much exactly what's being implemented in #9761 rn. I would generally be interesting in looking into implementing them for atleast some of the suppliers, so I'm unsure how much sense this change would make now.

matmair:
It leads to confusion right now and the function is not there. We are gearing up for 1.0 so I’d like to get this sorted before

SchrodingersGat:
Won't renaming the slugs mean that any users who have this already "enabled" will find these plugins are now disabled on update?

matmair:
These are the current slugs, they were added to ensure that settings etc stay consistent

SchrodingersGat:
@matmair ah, on a second read through I see the logic :)

PR #9833, opened 2025-06-24
Plugin url fix

SchrodingersGat:
Adds error separation between plugins which register URLs. If a single plugin provides a broken URL path, it no longer will result in all the plugin URLs from breaking

### DIFF
--- a/src/backend/InvenTree/plugin/urls.py
+++ b/src/backend/InvenTree/plugin/urls.py
@@ -2,8 +2,10 @@
 
 from django.conf import settings
 from django.urls import include, re_path
+from django.views.generic.base import RedirectView
 
 from common.validators import get_global_setting
+from InvenTree.exceptions import log_error
 from plugin import PluginMixinEnum
 
 PLUGIN_BASE = 'plugin'  # Constant for links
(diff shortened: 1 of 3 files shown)


src/backend/InvenTree/plugin/urls.py
```diff
@@ -16,8 +19,35 @@ def get_plugin_urls():
...
+        for plugin in registry.with_mixin(PluginMixinEnum.URLS):
+            try:
+                if plugin_urls := plugin.urlpatterns:
+                    # Check if the plugin has a custom URL pattern
+                    for url in plugin_urls:
+                        # Attempt to resolve against the URL pattern as a validation check
+                        try:
+                            url.resolve('')
+                        except Resolver404:
+                            pass
```
matmair, line 31: What are you trying to protect again bat here? This statement has seldom an effect

SchrodingersGat: It will at least ensure that each returned URL is a valid URLpattern, is the thinking

wolflu05:
This PR removes (changes?) the names for the endpoints registered by plugins. Previously they where available as `plugin:<slug>:<name>`. This caused my tests to fail, see https://github.com/wolflu05/inventree-bulk-plugin/actions/runs/15944072339/job/44975897357#step:6:36 is there any reason to break this?

SchrodingersGat:
@wolflu05 that wasn't intentional. I can't see from my changes how that has happened?

wolflu05:
I can't see that either, but git bisect has returned this commit to be the first faulty commit. @matmair do you have any idea? Has this also broken some of your plugins?

matmair:
There is too much movement in the code base at the moment for me to be confident in it, I have frozen my nightly deployments till it stabilizes again and therefore no insights.

PR #9842, opened 2025-06-24
Adjust "child_items" annotation for stockitem

SchrodingersGat:
- Show the direct children only
- Makes more a much more efficient DB query
- In a mysql setup, reduced from  >20 seconds to < 200ms

This is a subtle change in the returned data - the  "child_items" field now shows the *direct* children of a stock item, rather than all the top-to-bottom children.

This makes no change to the user interface or behaviour. The "child items" table in the stock item view will still show all the children underneath a given item, no matter the "level".

However it makes a significant difference if running on a mysql DB which cannot run OuterRef queries as efficiently.

### DIFF
--- a/src/backend/InvenTree/stock/serializers.py
+++ b/src/backend/InvenTree/stock/serializers.py
@@ -604,7 +604,7 @@ class StockItemSerializer(
         queryset = queryset.annotate(installed_items=SubqueryCount('installed_parts'))
 
         # Annotate with the total number of "child items" (split stock items)
-        queryset = queryset.annotate(child_items=stock.filters.annotate_child_items())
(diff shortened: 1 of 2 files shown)


src/backend/InvenTree/stock/serializers.py
```diff
@@ -604,7 +604,7 @@ def annotate_queryset(queryset):
         queryset = queryset.annotate(installed_items=SubqueryCount('installed_parts'))
 
         # Annotate with the total number of "child items" (split stock items)
-        queryset = queryset.annotate(child_items=stock.filters.annotate_child_items())
```
matmair, line 607: Do we need the annotate function for something else too?

matmair (approved):

Issue #8671, opened 2024-12-16
[1.0.0] Refactor "part stock history" into plugin


SchrodingersGat:
The "stocktake" part panel can be moved to a plugin, to allow for greater customization and removing a complex piece of code which some users may not want anyway.

matmair:
@SchrodingersGat what specific function are you referring to here? The ability to do a stock take or stock change tracking in general?

matmair:
@SchrodingersGat you still see this as a 1.0 feature?

SchrodingersGat:
Yes, currently working on this. The [inventree-stock-forecasting](https://github.com/inventree/inventree-stock-forecasting) plugin is done, the "stock history" is a bit more complex

SchrodingersGat:
Closed in https://github.com/inventree/InvenTree/pull/10124

PR #9852, opened 2025-06-25
[breaking] Remove stocktake code

SchrodingersGat:
Removes the internal "stocktake" functionality - will be replaced by the [stock history plugin](https://github.com/inventree/inventree-stock-history)

- Closes https://github.com/inventree/InvenTree/issues/8671

## Breaking Change

This is a *significant* breaking change - in that it removes entirely the old "stocktake" functionality from the core code. It is replaced by the incoming "stock history" plugin which provides:

- A "smaller target" of functionality
- Cleaner implementation

Part of the legacy stocktake functionality (generating "stock reports") is now provided by the generic data exporter - exporting "stock on hand" counts for a given part can be done with the generic exporter.

### Tasks

- [x] Keep old "stocktake" table in database to allow data migration to plugin
- [ ] Update documentation
- [ ] Wait for release and test of new plugin
- [x] Ensure "stocktake" export functionality is fully supported by data export plugins
- [x] Add new export plugin to provide stocktake functionality
- [ ] Unit testing for new exporter plugin
- [x] Delete old table *after* data has been imported into new plugin (only if ORM no longer exists)
- [ ] Add entry to "changelog"
- [ ] Add entry to `1.0.0` release notes blog post


src/backend/InvenTree/part/migrations/0136_remove_partstocktakereport_user_and_more.py
```diff
@@ -0,0 +1,27 @@
...
+        migrations.RemoveField(
+            model_name="part",
+            name="last_stocktake",
+        ),
+        migrations.DeleteModel(
+            name="PartStocktake",
+        ),
+        migrations.DeleteModel(
+            name="PartStocktakeReport",
+        ),
```
matmair: that might be a data-loss problem

SchrodingersGat: Yes - I need to work out a solution for that.

matmair:
I think this would be a good candidate for an entry in the changelog.md

SchrodingersGat:
I am replacing this with https://github.com/inventree/InvenTree/pull/10124 - it was proving impossible to retain the stock history data, so I am simplifying the approach.

Issue #8539, opened 2024-11-23
How to obtain output id for `/api/build/$buildId/complete/` endpoint?


jack-fdrv:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

To use endpoint `/api/build/$buildId/complete/` we need to provide output ids.
But there is no way from docs to obtain output ids some how.
It should be returned when we create it with endpoint `/api/build/$buildId/create-output/` 

### Steps to Reproduce

1. Create build with POST to `/api/build/`
2. Then create build output with POST to '/api/build/$buildId/create-output/'
3. Then try to complete the build with POST to `/api/build/$buildId/complete/`

### Expected behaviour

From step 2 we should obtain output id which later we can use in completing build request  `/api/build/$buildId/complete/` which required output id. 

### Deployment Method

- [ ] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

# Version Information:
InvenTree-Version: 0.16.3
Django Version: 4.2.15
Commit Hash: b0b05e4
Commit Date: 2024-09-21

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.8.0-1019-aws-x86_64-with
Installer: DOC

[...]

PR #9857, opened 2025-06-25
[API] Return stock item list when creating multiple items

SchrodingersGat:
- Closes https://github.com/inventree/inventree/issues/8539
- When performing an API operation which creates *multiple* items, return the list of items
- e.g. "stock item serialize" / "create build outputs"

### DIFF
--- a/src/backend/InvenTree/stock/api.py
+++ b/src/backend/InvenTree/stock/api.py
@@ -129,7 +129,12 @@ class StockItemSerialize(StockItemContextMixin, CreateAPI):
         items = serializer.save()
 
         response = StockSerializers.StockItemSerializer(
-            items, many=True, context=self.get_serializer_context()
+            items,
+            many=True,
+            context=self.get_serializer_context(),
+            part_detail=False,
+            location_detail=False,
+            supplier_part_detail=False,
         )
 
         return Response(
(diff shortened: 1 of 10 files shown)


src/backend/InvenTree/stock/api.py
```diff
@@ -1010,6 +1025,10 @@ def create(self, request, *args, **kwargs):
         # Check if a set of serial numbers was provided
         serial_numbers = data.pop('serial_numbers', '')
 
+        # Exclude 'serial' from submitted data
+        # We use 'serial_numbers' for item creation
+        data.pop('serial', None)
+
```
jacobfelknor, line 1031: Just a heads up, I got burned by this change because my clients used the `serial` keyword when using this API to create a single stock item. 

I have no problem updating my clients to use `serial_numbers` instead. However, `serial` still appears in the API documentation.  I saw the release blog post mention the return type being a list, so I probably should have seen this coming, but I overlooked it. I may not be the only one hitting this, took me a second to find this change and determine that's why it stopped working.

Would the correct thing be to remove `serial` from the `StockItemSerializer`? Or would this have other unintended consequences?

SchrodingersGat: @jacobfelknor good question, and a fair point that we should address.

The behavior here is not well documented, as the `serial` field applies to the serial number for a single stock item, but the `serial_numbers` is a write-only (and only on initial item creation) that allows you to specify multiple serial numbers for creating multiple stock items.

PR #9860, opened 2025-06-25
[bug] Fix for tracing init

SchrodingersGat:
- Circular include means that settings.DB_ENGINE may not be available
- Fixes bug introduced in https://github.com/inventree/InvenTree/pull/9808
- Currently killed the demo server

image


matmair (approved):
Sorry for that, fix looks good

PR #9887, opened 2025-06-27
[Bug] Report fix

SchrodingersGat:
Fixes an issue with a report template tag `render_date` which was [broken a while ago](https://github.com/inventree/InvenTree/pull/8932)

Also significantly improves error reporting around report generation

### DIFF
--- a/src/backend/InvenTree/InvenTree/middleware.py
+++ b/src/backend/InvenTree/InvenTree/middleware.py
@@ -169,17 +169,12 @@ class InvenTreeExceptionProcessor(ExceptionProcessor):
 
     def process_exception(self, request, exception):
         """Check if kind is ignored before processing."""
-        kind, info, data = sys.exc_info()
+        kind, _info, _data = sys.exc_info()
 
         # Check if the error is on the ignore list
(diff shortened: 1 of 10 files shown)


src/backend/InvenTree/InvenTree/middleware.py
```diff
@@ -194,15 +189,10 @@ def process_exception(self, request, exception):
...
-            html=ExceptionReporter(request, kind, info, data).get_traceback_html(),
-            path=path,
-            info=info,
-            data='\n'.join(traceback.format_exception(kind, info, data)),
-        )
+        # Pass off to the exception reporter
+        from InvenTree.exceptions import log_error
 
-        error.save()
+        log_error(path)
```
matmair, line 195: This might be a dumb question but are we retaining error objects with full HTML tracebacks this way?

SchrodingersGat: `log_error` has access to the exception info internally, this is how it runs everywhere else in the code. You don't need to pass the traceback through to the function.

PR #9894, opened 2025-06-28
[UI] Refactor useQuery hooks

SchrodingersGat:
Throughout the UI we were misusing the `useQuery` hook. By handling any API errors internally, we prevented the "retry on failure" mechanism (provided in the `useQuery` hook) from working.

This PR refactors a lot of our uses of `useQuery` to allow automatic retry of failed queries. It makes a lot of the UI components resiliant against query failure for whatever reason.

Reference: https://tanstack.com/query/v4/docs/framework/react/reference/useQuery


matmair (approved):
Lgtm; maybe we should cache these option queries using react query. Tables and forms could share these definitions in theory

PR #9895, opened 2025-06-28
[FR] Add proactive system check for site_url

### LINKED ISSUE #7847: [FR] Add proactive system check for site_url

### Problem statement

Having the wrong site_url set can create hard-to-debug problems (probably https://github.com/inventree/InvenTree/issues/7826). This issue is not actively raised to server admins so it is often lost.

### Suggested solution

Add a simple middleware that checks if the currently used domain is:
1. available in the sites
2. the same as site_url if multiple sites are not enabled

+ a check for the system status

Add just a system status check

### Examples of other systems

_No response_

- [X] I want to develop this.

matmair:
Fixes #7847


src/backend/InvenTree/InvenTree/middleware.py
```diff
@@ -223,3 +223,38 @@ def process_response(self, request, response):
...
+            return render(request, 'config_error.html', {'error_message': msg})
+
+        # Check trusted origins
+        if not any(
+            accessed_scheme.startswith(origin)
+            for origin in settings.CSRF_TRUSTED_ORIGINS
+        ):
+            msg = f'INVE-W11: The used path `{accessed_scheme}` is not in the TRUSTED_ORIGINS'
+            logger.error(msg)
+            return render(request, 'config_error.html', {'error_message': msg})
```
matmair: Note: This check will be improved before it is ready for merge

SchrodingersGat (approved):

SchrodingersGat:
@matmair this seems like an effective way to do this, nice implementation :)

SchrodingersGat:
@matmair it appears that this might break in cases where the server is behind a proxy, and the passed host header is something like `127.0.0.1`?

Edit: Nevermind, tested this now, and as long as the proxy is properly forwarding the host headers it should work just fine.

matmair:
@SchrodingersGat that should be accounted for; maybe we need another check for improperly configured reverse proxies?

PR #9898, opened 2025-06-29
Fix for caching table options

SchrodingersGat:
- The sync: false option does not perform properly
- Sometimes, the data is simply not loaded from localStorage at all
- Fixes implementation provided in https://github.com/inventree/InvenTree/pull/9873

### DIFF
--- a/src/frontend/src/hooks/UseTable.tsx
+++ b/src/frontend/src/hooks/UseTable.tsx
@@ -6,6 +6,12 @@ import type { FilterSetState } from '@lib/types/Filters';
 import type { TableState } from '@lib/types/Tables';
 import { useFilterSet } from './UseFilterSet';
 
+// Interface for the stored table data in local storage
+interface StoredTableData {
+  pageSize: number;
+  hiddenColumns: string[] | null;
+}
+
 /**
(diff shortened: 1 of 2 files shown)


src/frontend/src/hooks/UseTable.tsx
```diff
@@ -65,37 +71,71 @@ export function useTable(tableName: string, idAccessor = 'pk'): TableState {
...
+              // We only do this on first load, to avoid live syncing between tabs
+              // Note: The 'sync: false' option is not used, it does not perform as expected
+              setTableData(tableData);
+            }
+            return true;
+          });
+        }
+        return tableData;
+      }
+    });
```
matmair, line 114: this looks pretty similar to a zustand store; could we use that (as it is proven to work)

matmair (approved):

src/frontend/src/hooks/UseTable.tsx
SchrodingersGat: Yeah we should look into that as an alternative option to the `useLocalStorage` hook from mantine. This is working OK now but might be cleaner to migrate to zustand.

PR #9910, opened 2025-06-30
fix(backend) : use correct settings patching

matmair:
This ensures that error states do not cause issues in other tests and makes the CSRF test (#9895) a bit more robust


matmair:

SchrodingersGat (approved):

SchrodingersGat:
Much cleaner test code!

PR #9911, opened 2025-06-30
Introducing Aggregate Reports

### LINKED ISSUE #9774: [FR] Report for all Parts

### Problem statement

Hi,

I started using inventree recently.

I already created a template for rendering a report for each storage location. Love the Jinja option to create custom templated report 🤩 

However, I am missing the option to create a templated report **over all parts**. Basically, I want to create  a report with my own jinja template for all parts. I noticed that you can export all parts by csv.

Image

How can I create a report where I can cycle through all parts like

```
{% for p in all_parts %}
    {{ p.full_name }}
{% endfor %}

````
 I found the sample report for [bill of material](https://docs.inventree.org/en/latest/report/samples/#bill-of-materials-report) which let me select multiple parts and create one item report per page. 

Image

But I want to create a report similar to the view above.

Is this possible somehow? Or does it need to be implemented.

Thanks a lot, for the entire project 👍 
gitbock.

### Suggested solution

I would like to be able to click a button in the "all parts" view where I can select a report template that accepts "all_parts" context.

A workaround where I can access "all_parts" in a report.

### Examples of other systems

_No response_

- [ ] I want to develop this.

### PR DESCRIPTION
**Pull request by gitbock**
This is a proposal to fix #9774: Introducing of aggregate reports.

Aggregate Reports are not tied to a specific model type, but can be used to report over all instances of a model type. Like this we can overcome the pagination limitation for reporting, where we cannot select **all** parts in the table.

In the commit, I created the first aggregate report which allows to report over all parts, named "allparts"

Backend Changes:
- Report API accepts either PKs of instances or string model_type to enable aggregate report
- new template for allparts report

Frontend Changes:
- In Parts Table enable the report printing icon by default, but distinguish in sub items for selected items and aggregate report over all items
- Adapt Template Editor for aggregate report: no dropdown for instance selection

### CODE REVIEW

File: .pre-commit-config.yaml
Comment (matmair) [inline .pre-commit-config.yaml:1]: this is unrelated to the stated change; please remove

File: src/backend/InvenTree/part/models.py [hunk 1]
```diff
@@ -376,27 +376,31 @@ class PartReportContext(report.mixins.BaseReportContext):
...
         test_templates: Dict object of test templates associated with the Part
+        total_in_stock: Total stock quantity for the Part
     """
 
     bom_items: report.mixins.QuerySet[BomItem]
     category: Optional[PartCategory]
     description: str
     IPN: Optional[str]
     name: str
+    notes: str
```
matmair, line 395: these might be rather large

src/backend/InvenTree/part/models.py
```diff
@@ -489,13 +493,16 @@ def report_context(self) -> PartReportContext:
...
+            'notes': self.notes,
             'parameters': self.parameters_map(),
             'part': self,
             'qr_data': self.barcode,
             'qr_url': self.get_absolute_url(),
             'revision': self.revision,
             'test_template_list': self.getTestTemplates(),
             'test_templates': self.getTestTemplateMap(),
+            'total_stock': self.total_stock,
+            'units': self.units,
```
matmair, line 505: this is not documented

matmair:
i have submitted a fix for the pre-commit file to let ci run through

SchrodingersGat:
I agree with @matmair that a cleaner approach would be an option (when printing) to supply *all* items to the report template rather than just the selected items.

This could be done at the user level when they press the "generate report" button. We currently send a list of ID values in the API request. We could add an "all_items=true" arg (or similar).

Or a better option might be to have an option to include the query filters which are currently active in the given table, and then the outcome is "generate a report for all items in this table, even the ones not currently displayed". This is how the data export process currently functions so I think it would make sense to use a similar approach here.

gitbock:
> Some general notes:
> 
> * this is missing backend unit and frontend tests
> * there is an unrelated file, please keep PRs clean
> * the files do not follow the style guide, please [set up pre-commit](https://docs.inventree.org/en/stable/develop/contributing/?h=pre-commit#setup-devtools)

Thanks. The pre-commit stuff is sometimes messing up on my side and allowed me only to commit if i add the pre-commit file 🤔  Never mind, will keep an eye on this.


> I agree with @matmair that a cleaner approach would be an option (when printing) to supply _all_ items to the report template rather than just the selected items.
> 
> This could be done at the user level when they press the "generate report" button. We currently send a list of ID values in the API request. We could add an "all_items=true" arg (or similar).

> 
> Or a better option might be to have an option to include the query filters which are currently active in the given table, and then the outcome is "generate a report for all items in this table, even the ones not currently displayed". This is how the data export process currently functions so I think it would make sense to use a similar approach here.

All right. I am closing this PR right now, as it was only a proposal. Perhaps I will create a new one when I finally understood all comments by looking at the source.

PR #9912, opened 2025-06-30
feat: speed up docker publishing

matmair:
by spliting tests tasks for docker publishing into paralell also makes releasing safer as no tests are run in the then published image and the permissions are slimer


SchrodingersGat (approved):

matmair:
@SchrodingersGat I do not want this to screw up the 1.0 release, so I have moved it to 1.1
It is otherwise ready

matmair:

SchrodingersGat:
Nice, this is significantly faster overall

matmair:
very low hanging fruit; next step would be to centralize the environment preparation step in qc.yaml

PR #9916, opened 2025-07-01
Load calendar on first render

SchrodingersGat:
Ensure that calendar views submit an API request on first load


matmair:
LGTM; maybe we should put the date in the url so it is shareable?

SchrodingersGat:
> maybe we should put the date in the url so it is shareable?

Nice idea, like we do with the tables already

PR #9917, opened 2025-07-01
Dashboard widgets refresh automatically

SchrodingersGat:


src/frontend/src/components/dashboard/widgets/QueryCountDashboardWidget.tsx
```diff
@@ -39,6 +39,7 @@ function QueryCountWidget({
     queryKey: ['dashboard-query-count', modelType, params],
     enabled: user.hasViewPermission(modelType),
     refetchOnMount: true,
+    refetchInterval: 5 * 60 * 1000, // 5 minutes
```
matmair, line 42: Does this mean it also refetches when not in view?

SchrodingersGat: Ah yes. I'll address that.

PR #9918, opened 2025-07-01
Enhance "path" columns

SchrodingersGat:
- Show "short" version
- Hovercard for full "pathstring"

### DIFF
--- a/src/frontend/tests/pages/pui_build.spec.ts
+++ b/src/frontend/tests/pages/pui_build.spec.ts
@@ -80,7 +80,7 @@ test('Build Order - Basic Tests', async ({ browser }) => {
   await page.getByText('Quantity: 25').waitFor();
   await page.getByText('Continuity Checks').waitFor();
   await page
-    .getByRole('row', { name: 'Quantity: 16 No location set' })
+    .getByRole('row', { name: 'Quantity: 16' })
     .getByRole('button')
     .hover();
   await page.getByText('Add Test Result').waitFor();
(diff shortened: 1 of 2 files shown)


src/frontend/src/tables/stock/StockItemTable.tsx
```diff
@@ -222,9 +222,14 @@ function stockItemTableColumns({
       accessor: 'batch',
       sortable: true
     },
-    LocationColumn({
-      hidden: !showLocation,
-      accessor: 'location_detail'
+    // LocationColumn({
+    //   hidden: !showLocation,
+    //   accessor: 'location_detail'
+    // }),
```
matmair: Is this meant to be left in?

SchrodingersGat: Nope, thanks! Will remove

gunstr:
@SchrodingersGat Are you against making this settable? I still feel that the full path is more useful if you keep them reasonable short.

SchrodingersGat:
I'd be OK with a user setting something like "SHOW_FULL_PATH_IN_TABLES" - default = False - then I think everyone is happy.. Would you like to submit this?

gunstr:
I'm afraid I do not have a clue how to add a setting :-(
If you are aware of a similar PR I can use as template I would be happy to try, but else I do not know where to start...

wolflu05:
Why do we change that behavior that’s been around for quite a while in both PUI and CUI, see the tests for reference. From my perspective, this feels like a big step backward.
Ideally, this should be configurable. As it is now, it’s become difficult for me to use, since I have many locations simply named “D1” (for “Drawer 1”), and the surrounding context, like which shelf, room, or building it’s in, is really important. Without that context, the naming loses its meaning.

PR #9922, opened 2025-07-01
Replace the url with a clickable icon for link columns in tables

gunstr:
Closes #9901


SchrodingersGat:
Nice simple fix! Do you think that adding a [tooltip](https://mantine.dev/core/tooltip/) around the icon would be helpful?

gunstr:
The url is there already as a sort of tooltip when you hover over the icon.

SchrodingersGat:
Ok, I have checked it locally and it looks good. Thanks for the fix!

PR #9930, opened 2025-07-01
fix(frontend): logic error in display of registration option

### LINKED ISSUE #9921: New Platform UI shows register option even though Enable registration is disabled

### Describe the bug*

I have disabled the `Enabled Resgistration` in the Setting page. 

Image

While the old UI, doesn't shows the registration option, but the new Platform UI does. 
`Don't have an account? Register`

Image

I don't want the users to register at all. 

Is it a bug or i have missed some settings?

### Steps to Reproduce

1. Disable `Enable Registration` in Login Settings
2. Go to the new Platform UI
3. It shows `Don't have an account? Register` option
4. In the old UI, same option doesn't exist.

### Expected behaviour

I don't want the `Don't have an account? Register` to be show in the Platform UI, same as the old UI. 

### Deployment Method

Docker

### Version Information

Inventree: Stable
Postgres: 16

I did not try to reproduce

Reproducible

### Relevant log output

```shell

```

matmair:
Fixes #9921

### DIFF
--- a/src/frontend/src/pages/Auth/Login.tsx
+++ b/src/frontend/src/pages/Auth/Login.tsx
@@ -109,7 +109,7 @@ export default function Login() {
             ) : (
               <>
                 <AuthenticationForm />
-                {both_reg_enabled === false && (
+                {both_reg_enabled != false && (
                   <Text ta='center' size={'xs'} mt={'md'}>
                     <Trans>Don&apos;t have an account?</Trans>{' '}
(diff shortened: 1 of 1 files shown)


SchrodingersGat (approved):

SchrodingersGat:
Thanks for the quick fix @matmair

matmair:
simple issue

PR #9931, opened 2025-07-01
Adjust UI wording

SchrodingersGat:
- Separation between "items" and "stock"


src/frontend/playwright.config.ts
```diff
@@ -72,15 +66,15 @@ export default defineConfig({
       timeout: 120 * 1000
     },
     {
-      command: WEB_SERVER_CMD,
+      command: 'invoke dev.server -a 127.0.0.1:8000',
```
matmair: Did we not experience problems with running parallel tests with the dev server?

matmair (approved):

PR #9934, opened 2025-07-02
[bug] Logic fix for plugins

SchrodingersGat:
- Prevent tasks being run for disabled plugins


src/backend/InvenTree/report/tasks.py
```diff
@@ -68,7 +68,7 @@ def print_labels(
     model = template.get_model()
     items = model.objects.filter(pk__in=item_ids)
 
-    plugin = registry.get_plugin(plugin_slug)
+    plugin = registry.get_plugin(plugin_slug, active=True)
```
matmair, line 71: Comment: would it make sense to filter to active plugins by default? Makes unintended use less likely

matmair (approved):

src/backend/InvenTree/report/tasks.py
SchrodingersGat: Good point, I'll change that too

PR #9937, opened 2025-07-02
[UI] Refactor stock adjustment actions

SchrodingersGat:
- Refactor stock adjust actions into a common hook
- Enable stock actions for items allocated to build order
- Enable stock actions for items allocated to build order
- Refactor all other stock action instances

### DIFF
--- a/src/backend/InvenTree/stock/serializers.py
+++ b/src/backend/InvenTree/stock/serializers.py
@@ -307,41 +307,6 @@ class StockItemTestResultSerializer(
         return data
 
 
-class StockItemSerializerBrief(
-    InvenTree.serializers.NotesFieldMixin,
-    InvenTree.serializers.InvenTreeModelSerializer,
-):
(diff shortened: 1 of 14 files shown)


src/backend/InvenTree/stock/serializers.py
```diff
@@ -307,41 +307,6 @@ def validate(self, data):
         return data
 
 
-class StockItemSerializerBrief(
```
matmair: If I remember correctly, there were some performance issues with using normal serializer everywhere; are those resolved?

SchrodingersGat:
> New hook looks clean! Have you checked if this impacts rendering performance?

As far as I can tell it's better overall. The hook rendering dependencies have been fixed, they were far too "wide"

src/backend/InvenTree/stock/serializers.py
SchrodingersGat: Yes, this is OK now

PR #9070, opened 2025-02-12
[UI] Settings to control editing of "completed" orders

matmair:
This re-implements the locking of completed orders - this time in the backend.

Todo:
- [x] also lock line items
- [x] exclude status itself from being locked

Fixes #8976

PR #9948, opened 2025-07-03
Logic fix for LOCK_SETTING

SchrodingersGat:
- Because the setting is to "allow editing", it is really an UNLOCK setting
- This setting had to be DISABLED (false) to allow editing of completed orders
- Ref: https://github.com/inventree/InvenTree/pull/9070


SchrodingersGat:
@matmair can you please check over this one? I think the logic of the original implementation was flipped

PR #9960, opened 2025-07-05
Add setting to show location path as plain text in tables

gunstr:
Follow up to #9918 

I think I figured out how to add a user setting to control the visibility of the location path.

@SchrodingersGat I'm not sure why you decided to use different entry points for the column rederer for the different tables. I changed back to using the same entry point (LocationColumn) for all tables. Let me know if this was the wrong decision.

### DIFF
--- a/src/backend/InvenTree/common/setting/user.py
+++ b/src/backend/InvenTree/common/setting/user.py
@@ -225,6 +225,14 @@ USER_SETTINGS: dict[str, InvenTreeSettingsKeyType] = {
         'default': False,
         'validator': bool,
     },
+    'SHOW_FULL_LOCATION_PATH_IN_TABLES': {
+        'name': _('Show Stock Item Location Path in Tables as text'),
+        'description': _(
+            'Disabled: The full location path is displayed as a hover tooltip. Enabled: The full location path is displayed as plain text.'
+        ),
+        'default': False,
+        'validator': bool,
+    },
     'NOTIFICATION_ERROR_REPORT': {
         'name': _('Receive error reports'),
         'description': _('Receive notifications for system errors'),
--- a/src/frontend/src/tables/ColumnRenderers.tsx
+++ b/src/frontend/src/tables/ColumnRenderers.tsx
@@ -19,7 +19,10 @@ import { TableStatusRenderer } from '../components/render/StatusRenderer';
 import { RenderOwner, RenderUser } from '../components/render/User';
 import { formatCurrency, formatDate } from '../defaults/formatters';
 import { resolveItem } from '../functions/conversion';
-import { useGlobalSettingsState } from '../states/SettingsStates';
+import {
+  useGlobalSettingsState,
+  useUserSettingsState
+} from '../states/SettingsStates';
 import type { TableColumn, TableColumnProps } from './Column';
 import { ProjectCodeHoverCard, TableHoverCard } from './TableHoverCard';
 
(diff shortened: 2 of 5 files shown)


SchrodingersGat:
@gunstr nice work, this is a clean implementation.

> I'm not sure why you decided to use different entry points for the column renderer for the different tables. I changed back to using the same entry point (LocationColumn) for all tables. Let me know if this was the wrong decision.

Good call. It was probably a merge conflict that got solved incorrectly. It was not intentional.

One suggest change here - can you make the setting more generic for "path display" and have it apply to part categories too?

gunstr:
@SchrodingersGat I will try. However I would suggest to implement a separate setting for the Part Category as the use cases are slightly different.

- Stock Location path: Mainly used to find the physical location - the user will read whats on the screen and then pick up the item
- Part Category: Mainly used to find similar items in the database - the user will read whats on the screen and then navigate using the mouse to that category.

As an addition I would also suggest to apply the "Stock Location setting" also to the Default Location column for parts. I would assume most users leave that column hidden but for consistency I think it make sense to apply the same appearance for Stock Location strings throughout the application.

Do you agree?

SchrodingersGat:
Happy for two separate settings, and to apply globally to all "location" or "category" columns in tables

gunstr:
@SchrodingersGat I think this is ready now except one thing. I cannot figure out why the Category Column does not wrap around when rendering the full path. For the Location Path it works as expected and the same renderer is used for both. Any hints?

image

src/backend/InvenTree/common/setting/user.py
```diff
@@ -225,6 +225,22 @@ def label_printer_options():
         'default': False,
         'validator': bool,
     },
+    'SHOW_FULL_LOCATION_PATH_IN_TABLES': {
+        'name': _('Show Stock Item Location Path in Tables as text'),
```
SchrodingersGat: please simplify the name to "Show full stock location in tables"

src/backend/InvenTree/common/setting/user.py
```diff
@@ -225,6 +225,22 @@ def label_printer_options():
...
+        'description': _(
+            'Disabled: The full location path is displayed as a hover tooltip. Enabled: The full location path is displayed as plain text.'
+        ),
+        'default': False,
+        'validator': bool,
+    },
+    'SHOW_FULL_CATEGORY_PATH_IN_TABLES': {
+        'name': _('Show Part Category Path in Tables as text'),
```
SchrodingersGat: Please simplify to "Show full part categories in tables"

src/frontend/src/tables/ColumnRenderers.tsx
```diff
@@ -114,24 +117,97 @@ export function PathColumn(props: TableColumnProps): TableColumn {
...
 }
 
+export function PathColumnPlainText(props: TableColumnProps): TableColumn {
+  return {
+    ...props,
+    accessor: props.accessor ?? 'path',
+    render: (record: any) => {
+      const instance = resolveItem(record, props.accessor ?? '');
+
+      if (!instance || !instance.name) {
```
SchrodingersGat: I think this should be 

```typescript
      if (!instance || !instance.pathstring) {
```

gunstr:
@SchrodingersGat I have changed as per your suggetsions. First time I changed based on comments... Maybe I should push the changes first and mark as resolved after that? 

Regaring the wrap around issue it seems to be a more generic thing as wrapping text is only done on whitespace. One potential solution would be to add white space around the path dividers (like Mechanical / Enclosures) but I would suggest to handle that as a separate issue/PR it you think it's a good idea.

src/backend/InvenTree/common/setting/user.py
SchrodingersGat: I still think the "name" attributes here could be clearer. Can you please change this to "Stock full stock location in tables"? And same for the part category option.

gunstr: This is the first time I change based on comments and obviously I did someting in the wrong order. I have changed according to your suggestions and pushed the commits yesterday already - but it seems you cannot see the updated code. If I look at the changed files in the PR the changes are there.

I did the changes, commited, marked the comment as resolved and finally pushed. Is that not how it should be done? What should I do to make the changes visible?

gunstr: Ahh, sorry - I just realized you alos suggest changes to the 'name'. I will fix that.

SchrodingersGat:
@gunstr thanks for the contribution :)

PR #9969, opened 2025-07-07
Fix annotations for returning serialized StockItems as lists

1337joe:
The `@extend_schema` annotations added in #9857 were put on `create` methods, but that annotation only works on `post`, `get`, etc methods or on the class itself. Here's the actual change in schema as a result of that pull request (as pulled into my java api project): https://github.com/1337joe/inventree-sdk-java/commit/fdc4efdd1867108dd5bc75060ed710fb83bf1155#diff-d6a362a495478710a98099c50875c627ed6dc6552e56e87d1869004c5b8cc3e4

I simply moved the annotations on `StockItemSerialize` and `BuildOutputCreate` (and removed their pagination classes because paging through a create result doesn't work), but that didn't work for `StockList`: I couldn't remove the paginating because that would break the `StockList#GET` functionality. I worked around that by directly modifying the extended schema generator, which works but isn't as clean as I'd prefer. I'm open to any other way to get the annotation to stick that actually works.

I also noticed that `serial_numbers` was used in `StockList#create` but didn't exist in `StockItemSerializer` so I added it there.

With the schema as generated by these changes I was able to verify the workflow detailed in #8539.


src/backend/InvenTree/InvenTree/schema.py
```diff
@@ -101,6 +101,16 @@ def get_operation(
                         f'{parameter["description"]} Possible fields: {", ".join(ordering_fields)}.'
                     )
 
+        # Change return to array type, simply annotating this return type attempts to paginate, which doesn't work for
+        # a create method and removing the pagination also affects the list method
+        if self.method == 'POST' and type(self.view).__name__ == 'StockList':
```
matmair, line 122: Q: This seems very finicky; can this be addressed in/on the view itself?

src/backend/InvenTree/stock/api.py
```diff
@@ -1167,7 +1168,7 @@ def create(self, request, *args, **kwargs):
 
                 item.save(user=user)
 
-                response_data = serializer.data
+                response_data = [serializer.data]
```
matmair, line 1186: C: This is breaking

src/backend/InvenTree/InvenTree/api_version.py
```diff
@@ -1,12 +1,15 @@
...
 # InvenTree API version
-INVENTREE_API_VERSION = 364
+INVENTREE_API_VERSION = 365
 
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
 
+v365 -> 2025-07-06 : https://github.com/inventree/InvenTree/pull/9969
+    - Correctly apply changes listed in v358
```
matmair, line 10: C (blocking): Please update this text to mention that it is containing a breaking change for the response of StockItem requests

matmair (changes requested):
LGTM in general; the change text fails to list the severity of a broken response format. Any usage of the stock list endpoint will need adjustment in production

src/backend/InvenTree/InvenTree/schema.py
1337joe: Not without significant changes:
- The view is the list view, so retrieving a list requires pagination to be enabled
- Pagination is not compatible with the list of results returned by creating items because additional calls to get later pages of results would create more items.

The only way I could see to do it would be to create a separate view (similar to the `BuildOutputCreate` or `StockItemSerialize` classes), but I don't know the implications of splitting it out.

SchrodingersGat:
As this is a breaking change it would be great to get this sorted out before the 1.0.0 release :)

1337joe:
> we will need to change the frontend behavior when creating a new stock item. The form expects a single object to be returned - not a list:
> 
> https://github.com/inventree/InvenTree/blob/6e81fa2e7b36a078446f09a839d7e0c73b092f61/src/frontend/src/tables/stock/StockItemTable.tsx#L510

@SchrodingersGat I copied the behavior from StockDetail.tsx#serializeStockItem: on success navigate to the details page of the first response.

I don't see that the playwrite tests or docs build are failing for anything I've changed, if that's not the case a pointer on what to look into would be helpful.

SchrodingersGat:
@1337joe please fix the API version conflicts so we can merge

1337joe:
@SchrodingersGat I just merged master and fixed the version_api conflict.

SchrodingersGat:
Thanks! I think it's just waiting on acceptance from @matmair now.

src/backend/InvenTree/InvenTree/schema.py
matmair: tbh I would prefer that approach in the future, this is a very untransparent way of handling this and very hard to discover

matmair (approved):
I am not really happy with this from a data flow perspective, but it is better than what is currently in `master`

PR #9973, opened 2025-07-07
[bug] Data import fix

SchrodingersGat:
Fixes issue where data import will not work with background worker

- The frontend component was not correctly "requesting" a refreshed status of the import session
- Worked OK if the background worker was NOT running (as in our CI tests) - because the import session was blocking


SchrodingersGat:
Replaced by https://github.com/inventree/InvenTree/pull/9992 - the commit history got out of whack here

PR #9979, opened 2025-07-08
[bug] Tree fix

SchrodingersGat:
This is a significant bug which was discovered as part of general hardening of tests for the upcoming 1.0.0 release

### Background

We make use of the [django-mptt](https://django-mptt.readthedocs.io/en/latest/) library throughout the code, to realize a database-efficient "tree" structure for the following models:

- PartCategory
- Part
- StockLocation
- StockItem
- Build

Of these, the `PartCategory` and `StockLocation` models have special consideration (via the `InvenTreeTree` model) which handles node deletion and re-parenting of child nodes.

The other models (`Part`, `StockItem` and `Build`) inherit directly from the `MPTTModel` class, without this mixin shim.

**However** - it turns out that by default, `django-mptt` does not handle tree rebuilds correctly when a node is deleted from the tree (*at least, not in the way that we need it to*).

So, as per the current code base, if a `StockItem` (or `Part` or `Build`) is deleted, and it has child nodes attached, those nodes are effectively orphaned, as they are not correctly re-parented to the upstream node. As these child nodes retain the same `tree_id` value, this creates a corrupted tree, which requires a full rebuild of the model structure (which can be extremely expensive for a large dataset).

### Changes

This PR removes the direct inheritance from the `MPTTModel` class - instead ensuring that all our "tree" structures inherit from the `InvenTreeTree` class - which correctly handles node deletion, re-parenting, etc.

Additionally it adds a significant amount of unit test code to ensure that tree operations are sufficiently covered for all model types, to ensure we do not get a regression here.

To achieve this, the `InvenTreeTree` model had to be refactored somewhat, and some of the functionality has been split out into another mixin which is relevant only for the `StockLocation` and `PartCategory` models.

There are no migrations or changes to the API.

### Backporting

If it can be backported cleanly...

### TODO

- [x] Update `Build` model
- [x] Update `StockItem` model
- [x] Update `Part` model
- [ ] Ensure all unit tests run

### DIFF
--- a/src/backend/InvenTree/InvenTree/models.py
+++ b/src/backend/InvenTree/InvenTree/models.py
@@ -554,12 +554,17 @@ class InvenTreeTree(MetadataMixin, PluginValidationMixin, MPTTModel):
 
         order_insertion_by = ['name']
 
-    def delete(self, delete_children=False, delete_items=False):
+    def delete(self, delete_children: bool = False, delete_items: bool = False):
         """Handle the deletion of a tree node.
 
-        1. Update nodes and items under the current node
-        2. Delete this node
-        3. Rebuild the model tree
+        Arguments:
+            delete_children: If True, delete all child nodes (otherwise, point to the parent of this node)
+            delete_items: If True, delete all items associated with this node (otherwise, point to the parent of this node)
+
+        Order of operations:
+            1. Update nodes and items under the current node
+            2. Delete this node
+            3. Rebuild the model tree
         """
         tree_id = self.tree_id if self.parent else None
 
--- a/src/backend/InvenTree/InvenTree/models.py
+++ b/src/backend/InvenTree/InvenTree/models.py
@@ -625,8 +625,8 @@ class InvenTreeTree(MetadataMixin, PluginValidationMixin, MPTTModel):
         Removing a tree "node" from the database must be considered carefully,
         based on what the user intends for any items which exist *under* that node.
 
-        - "children" are any nodes which exist *under* this node (e.g. PartCategory)
-        - "items" are any items which exist *under* this node (e.g. Part)
+        - "children" are any nodes (of the same type) which exist *under* this node (e.g. PartCategory)
+        - "items" are any items (of a different type) which exist *under* this node (e.g. Part)
 
         Arguments:
             delete_children: If True, delete all child items
@@ -645,29 +645,30 @@ class InvenTreeTree(MetadataMixin, PluginValidationMixin, MPTTModel):
         # - Delete all items at any lower level
         # - Delete all descendant nodes
         if delete_children and delete_items:
-            self.get_items(cascade=True).delete()
+            self.delete_items(cascade=True)
             self.delete_nodes(child_nodes)
 
         # Case B: Delete all child nodes, but move all child items up to the parent
         # - Move all items at any lower level to the parent of this item
         # - Delete all descendant nodes
         elif delete_children and not delete_items:
-            self.get_items(cascade=True).update(**{self.ITEM_PARENT_KEY: self.parent})
-
+            if items := self.get_items(cascade=True):
+                items.update(**{self.ITEM_PARENT_KEY: self.parent})
             self.delete_nodes(child_nodes)
 
         # Case C: Delete all child items, but keep all child nodes
         # - Remove all items directly associated with this node
         # - Move any direct child nodes up one level
         elif not delete_children and delete_items:
-            self.get_items(cascade=False).delete()
+            self.delete_items(cascade=False)
             self.get_children().update(parent=self.parent)
 
         # Case D: Keep all child items, and keep all child nodes
         # - Move all items directly associated with this node up one level
         # - Move any direct child nodes up one level
         elif not delete_children and not delete_items:
-            self.get_items(cascade=False).update(**{self.ITEM_PARENT_KEY: self.parent})
+            if items := self.get_items(cascade=False):
+                items.update(**{self.ITEM_PARENT_KEY: self.parent})
             self.get_children().update(parent=self.parent)
 
     def delete_nodes(self, nodes):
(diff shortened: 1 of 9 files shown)


matmair:
Looking interesting @SchrodingersGat; can I review once it is fully done?

SchrodingersGat:
@matmair I'll leave this un-merged until you have had a change to properly review.

SchrodingersGat:
@matmair good to review :)

src/backend/InvenTree/order/test_api.py
```diff
@@ -1919,6 +1919,12 @@ def test_download_csv(self):
 class SalesOrderAllocateTest(OrderTest):
     """Unit tests for allocating stock items against a SalesOrder."""
 
+    @classmethod
+    def setUpTestData(cls):
+        """Init routine for this unit test class."""
+        super().setUpTestData()
+        Part.objects.rebuild()  # Ensure part variant trees are correct
```
matmair: Is this not an indication that we have an issue in the model create/update setup?

src/backend/InvenTree/part/models.py
```diff
@@ -467,6 +467,11 @@ class MPTTMeta:
         # For legacy reasons the 'variant_of' field is used to indicate the MPTT parent
         parent_attr = 'variant_of'
 
+    @property
+    def parent(self):
+        """Return the template part for this Part instance."""
+        return self.variant_of
```
matmair: Q: This a change in behavior, right?

src/backend/InvenTree/stock/models.py
```diff
@@ -622,9 +630,7 @@ def _create_serial_numbers(cls, serials: list, **kwargs) -> QuerySet:
         StockItem.objects.bulk_create(items)
 
         # We will need to rebuild the stock item tree manually, due to the bulk_create operation
-        InvenTree.tasks.offload_task(
-            stock.tasks.rebuild_stock_item_tree, tree_id=tree_id, group='stock'
-        )
+        stock.tasks.rebuild_stock_item_tree(tree_id)
```
matmair: Have you tested what this means for endpoint performance?

src/backend/InvenTree/stock/models.py
```diff
@@ -2085,10 +2091,17 @@ def merge_stock_items(self, other_items, raise_error=False, **kwargs):
         self.save()
 
         # Rebuild stock trees as required
+        rebuild_result = True
         for tree_id in tree_ids:
-            InvenTree.tasks.offload_task(
-                stock.tasks.rebuild_stock_item_tree, tree_id=tree_id, group='stock'
+            if not stock.tasks.rebuild_stock_item_tree(tree_id, rebuild_on_fail=False):
+                rebuild_result = False
```
matmair, line 2115: Have you tested what this means for endpoint performance?

src/backend/InvenTree/stock/models.py
```diff
@@ -2150,7 +2163,6 @@ def splitStock(self, quantity, location=None, user=None, **kwargs):
 
         # Update the new stock item to ensure the tree structure is observed
         new_stock.parent = self
-        new_stock.level = self.level + 1
```
matmair, line 2153: Q: Is this now done automatically?

src/backend/InvenTree/stock/tasks.py
```diff
@@ -7,23 +7,53 @@
...
 
-    The StockItem tree uses the MPTT library to manage the tree structure.
+    Arguments:
+        tree_id (int): The ID of the StockItem tree to rebuild.
+        rebuild_on_fail (bool): If True, will attempt to rebuild the entire StockItem tree if the partial rebuild fails.
+
+    Returns:
+        bool: True if the partial tree rebuild was successful, False otherwise.
+
+    - If the rebuild fails, schedule a rebuild of the entire StockItem tree.
```
matmair, line 42: C: This would be a good metric to track on the demo instance

src/backend/InvenTree/InvenTree/models.py
```diff
@@ -526,41 +526,36 @@ def create_attachment(self, attachment=None, link=None, comment='', **kwargs):
 
 
 class InvenTreeTree(MetadataMixin, PluginValidationMixin, MPTTModel):
-    """Provides an abstracted self-referencing tree model for data categories.
+    """Provides an abstracted self-referencing tree model for data categories."""
```
matmair: C: This description could be more general

src/backend/InvenTree/InvenTree/models.py
```diff
@@ -526,41 +526,36 @@ def create_attachment(self, attachment=None, link=None, comment='', **kwargs):
 
 
 class InvenTreeTree(MetadataMixin, PluginValidationMixin, MPTTModel):
```
matmair: Q: Is there a technical reason to have the Validation Mixin here? If not, I would recommend removing it to reduce unintended nesting

matmair (approved):
LGTM in general, a few minor questions / comments

matmair:
Comment on coverage: we seem to lose some coverage on part/api.py - not sure why
There are also non-covered paths in InvenTreeTree, which seems pivotal for ensuring correct tree building

src/backend/InvenTree/order/test_api.py
SchrodingersGat: Good point - this is because some of our fixtures contain corrupted trees - I'll need to work to fix that. But, I'll ensure that all calls to "rebuild" are removed from the test code.

SchrodingersGat: Also going to add some unit tests to ensure that the fixture data itself is correct.

src/backend/InvenTree/part/models.py
SchrodingersGat: That was only a helper func in there for testing. I've removed it now.

src/backend/InvenTree/stock/tasks.py
SchrodingersGat: I've just added sentry reporting for errors in tree rebuild

src/backend/InvenTree/stock/models.py
SchrodingersGat: Yes, this is now handled correctly by the custom tree class

src/backend/InvenTree/InvenTree/models.py
SchrodingersGat: Done, improved now.

src/backend/InvenTree/InvenTree/models.py
SchrodingersGat: No, there's no reason. I have now flattened it out

src/backend/InvenTree/stock/models.py
SchrodingersGat: As the tree issues are now fixed, the rebuild is limited to only the single tree which is "pre-filled" as much as possible. It's quite performant, and does not seem to add significant overhead

PR #9982, opened 2025-07-08
fix error if missing license info

matmair:
Already fixed in master


matmair:
Not sure why these checks are even running, they are normally skipped https://github.com/inventree/InvenTree/pull/9850 @SchrodingersGat

SchrodingersGat:
@matmair unsure, agreed they should not be running

PR #9994, opened 2025-07-09
Refactor more UI components out into lib directory

SchrodingersGat:
- Expose more functionality / components to the npm library
- Enhances UI development with plugins


matmair (approved):
LGTM; we should think about what kind of API stability we want to guarantee and test for with all these components.

PR #10000, opened 2025-07-10
Plugin source URLs

SchrodingersGat:
- Improve handling of URLs for remote plugin components
- Use proper URL methods rather than naive string comparisons


SchrodingersGat:
Whoa, PR 10,000

PR #10001, opened 2025-07-10
fix(backend): missing scope

matmair:
The importer sessions are missing oAuth2 scopes.
This also adds a check to ensure this does not happen again


SchrodingersGat (approved):

SchrodingersGat:
Thanks for the fix, and the additional checks!

PR #10026, opened 2025-07-15
Logic fix for plugin.is_active

SchrodingersGat:
- Previously were not considering the "mandatory" attribute


matmair:
@SchrodingersGat do we have docs on mandatory plugins anywhere?

SchrodingersGat:
Yes it is discussed here - https://docs.inventree.org/en/latest/plugins/builtin/

PR #10051, opened 2025-07-21
fix(backend): remove coreapi

matmair:
Fixes coreapi warning Closes #10021

PR #10052, opened 2025-07-21
!backend(refactor): move to default engine

matmair:
Follow up to https://github.com/inventree/InvenTree/pull/10051 - moves file name rending to the default rendering engine instead of jinja2
This might be breaking in some cases

### DIFF
--- a/src/backend/InvenTree/common/setting/system.py
+++ b/src/backend/InvenTree/common/setting/system.py
@@ -1,7 +1,6 @@
 """System settings definition."""
 
 import json
-import logging
 import os
 import re
 import uuid
@@ -13,6 +12,8 @@ from django.core.validators import MaxValueValidator, MinValueValidator, URLVali
 from django.template import Context, Template
 from django.utils.translation import gettext_lazy as _
 
+import structlog
+
 import build.validators
 import common.currency
 import common.validators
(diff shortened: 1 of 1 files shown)


matmair:
@inventree/triage I can not remember why we used Jinja2 instead of the normal templating engine - this removes our direct dependency on jinja2. This be breaking in some edge cases

SchrodingersGat:
I have a vague recollection that it was due to how we provide context for simple string rendering (e.g. for batch codes, order reference strings, filenames for generated reports). It looks like that is where the PR is failing currently.

If we can work around this and get those string rendering functions working, I'm happy to remove the dep and make this change

matmair:
I will have to take a deeper look, puhsing this back

SchrodingersGat:
One difficulty here is that the jinja syntax we use in `PART_NAME_FORMAT` is incompatible with the django template syntax - so we will need to make a breaking change to that format string

src/backend/InvenTree/common/setting/system.py
```diff
@@ -9,18 +10,18 @@
...
-
 import build.validators
 import common.currency
-import common.models
 import common.validators
 import order.validators
 import report.helpers
 from common.setting.type import InvenTreeSettingsKeyType
 
+logger = logging.getLogger('inventree')
```
matmair: I think this should be structlog

matmair:
This was addressed in https://github.com/inventree/InvenTree/pull/11400 by @SchrodingersGat

PR #10056, opened 2025-07-22
[Refactor] BOM Validation

SchrodingersGat:
### Features

- Updated API endpoints
- Bring BOM validation information into the new interface
- Cache the "bom validated" values (and recalculate when changes occur)
- Allows parts to be filtered by "BOM validated" status (without expensive runtime calculations)
- Add dashboard widget

### Tasks

- [ ] Update unit tests
- [ ] Update demo dataset (BOM validation)
- [ ] Playwright testing for new features


src/backend/InvenTree/InvenTree/urls.py
matmair, line 1: Why is this changed? I see no real functional change with this approach

SchrodingersGat: This was an attempt to work out the circular import issues - which ended up being the plugin initialization sequence bug. 

I'll revert this change, thanks for catching it.

PR #9899, opened 2025-06-29
Remove circular import vector

SchrodingersGat:
- Sometime causes issues with circular imports on `invoke migrate`

PR #10057, opened 2025-07-22
Plugin sequence fix

SchrodingersGat:
Fixes a significant logic bug which was discovered as part of https://github.com/inventree/InvenTree/pull/10056

### Problem Description

When launching the InvenTree project, the various `apps.py` models contain a `ready` method - this is code which is run when all apps are loaded.

In `plugin/apps.py` we explicitly load the plugin apps - until this point in the code, the plugin registry is initialized but none of the plugins have been discovered yet.

However, *other* apps can trigger code which can (directly or indirectly) call a function which requires plugin functionality. 

In the existing code (prior to this patch) this can result in the plugin registry being "reloaded" (by the other app) before it has actually been fully loaded yet (depending on which app "ready" function is called first).

This can lead to very complex and hard to debug circular imports, or other issues where apps are only partially initialized.

### Solution

This PR introduces a single "ready" flag for the plugin registry. Until that flag is set, plugins cannot be accessed and the registry cannot be loaded.

The ready flag is only set once - in `apps.py` 'ready' method.

### References

- Likely cause of https://github.com/inventree/InvenTree/pull/9899

### DIFF
--- a/src/backend/InvenTree/plugin/registry.py
+++ b/src/backend/InvenTree/plugin/registry.py
@@ -105,6 +105,11 @@ class PluginsRegistry:
         self.installed_apps = []  # Holds all added plugin_paths
 
     @property
+    def is_ready(self) -> bool:
+        """Return True if the plugin registry is ready to be used."""
+        return self.ready
+
+    @property
     def is_loading(self) -> bool:
         """Return True if the plugin registry is currently loading."""
         return self.loading_lock.locked()
--- a/src/backend/InvenTree/plugin/registry.py
+++ b/src/backend/InvenTree/plugin/registry.py
@@ -369,7 +369,7 @@ class PluginsRegistry:
             logger.info('Plugin Registry: Loaded %s plugins', len(self.plugins))
 
         except Exception as e:
-            logger.exception('Expected error during plugin reload: %s', e)
+            logger.exception('Unexpected error during plugin reload: %s', e)
             log_error('reload_plugins', scope='plugins')
 
         finally:
--- a/src/backend/InvenTree/plugin/registry.py
+++ b/src/backend/InvenTree/plugin/registry.py
@@ -68,6 +68,8 @@ class PluginsRegistry:
         'parameter-exporter',
     ]
 
+    ready: bool = False  # Marks if the registry is ready to be used
+
     def __init__(self) -> None:
         """Initialize registry.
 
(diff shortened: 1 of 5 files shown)


src/backend/InvenTree/plugin/registry.py
```diff
@@ -176,6 +200,10 @@ def set_plugin_state(self, slug: str, state: bool):
             slug (str): Plugin slug
             state (bool): Plugin state - true = active, false = inactive
         """
+        if not self.is_ready:
+            logger.warning('registry.set_plugin_state: Plugin registry is not ready')
+            return
```
matmair: There seems to be a bit of duplication - could this be a case for a decorator?

matmair:
We should maybe test that stuff is correctly initialized to ensure this vital mechanism does not break

SchrodingersGat:
> We should maybe test that stuff is correctly initialized to ensure this vital mechanism does not break

What extra tests would you like to see?

src/backend/InvenTree/plugin/registry.py
SchrodingersGat: Nice idea, have implemented this now

src/backend/InvenTree/plugin/registry.py
```diff
@@ -905,6 +960,10 @@ def check_reload(self):
 
         Returns True if the registry has changed and was reloaded.
         """
+        if not self.is_ready:
+            logger.warning('registry.check_reload: Plugin registry is not ready')
+            return False
+
```
matmair: if we add a flag if the check_reload function should be loaded to the decorator this could also be in the decorator

src/backend/InvenTree/plugin/registry.py
```diff
@@ -323,6 +374,10 @@ def reload_plugins(
             clear_errors (bool, optional): Clear any previous loading errors. Defaults to False.
             _internal (list, optional): Internal apps to reload (used for testing). Defaults to None
         """
+        if not self.is_ready:
+            logger.warning('Plugin registry is not ready - cannot reload plugins')
+            return
```
matmair: With a flag on the decorator this could be avoided

src/backend/InvenTree/plugin/registry.py
SchrodingersGat: Done

src/backend/InvenTree/plugin/registry.py
SchrodingersGat: Implemented :)

matmair:
> > We should maybe test that stuff is correctly initialized to ensure this vital mechanism does not break
> 
> What extra tests would you like to see?

Checks for the conditions that are now handled additionally.
Looking at the full coverage file, it seems that mostly only the intended path is tested. That leaves the potential that we over/under catch things

SchrodingersGat:
@matmair agreed there are some areas in the plugin registry which could see some increased test coverage

Issue #9788, opened 2025-06-16
[FR] Multiple Part Pictures


Islendur:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

some parts its hard to decide which picture represents the part the best so you have to decide which angle you want to show in the thumbnail. this can sometimes make it hard to distinguish between parts without having to look it up manually 

### Suggested solution

for some parts it would be benefitial to be able to upload a few pictures from diffrent angles. 

### Describe alternatives you've considered

uploading extra picture files in attachment, but this is not as intuitive. 

### Examples of other systems

in a typical webshop for example there is a field to the left or under the picture where you can click the other pictures available 

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
We have received similar requests from other users too. It would require someone to step up and develop it. Is this something you are interested in?

VMT-Solutions:
i will look into it or aquire the help needed.

Reza98Sh:
@SchrodingersGat 
Can I work on this issue ?

SchrodingersGat:
Sure, happy for you to work on it! Please LMK how you are intending to approach it as it will require some significant API and UI changes

Reza98Sh:
Of course!  
I think there are **two approaches** for implementing this:

1. **Move the image field to a separate table and connect it to the Part table with a ForeignKey.**  
   In this approach, you need to write a manual migration file to migrate the existing images.

2. **Add multiple image fields to the Part model (for example, three image fields).**  
   This approach doesn’t require a manual migration, but the number of photos is limited, and it will create a lot of empty fields for parts that don't have multiple images.

---

One other thing I noticed is that InvenTree uses [django-stdimage](https://pypi.org/project/django-stdimage), which has been deprecated.  
Does InvenTree team have any plans to use an alternative for this package?

matmair:
I would like to introduce a variant of 1: move the image (and the stdimage logic) to the attachments table and enable selecting a primary and associated attachments. That would make re-using images possible - see #5532

Regarding django-stdimage; I looked into its suggested replacement this spring:
- There are dependency issues
- I am vary of changing the image format to something only browsers really understand

SchrodingersGat:
[...]

PR #10069, opened 2025-07-24
Feature/issue9788

Reza98Sh:
Refactoring how "Images" are stored, moving to a single table approach.

### Tasks

- [x] Create new table for uploaded images
- [x] Add data migration for migrate old images into new upload image table
- [x] Remove old tables
- [x] Expose new model to API
- [x] Add a mixin class for models which will have image associated with them
- [x] Mixin class must provide method for accessing image
- [x] Add previous Image functionalities  to new image model 
- [x] Add images carousel for multiple part pictures 
- [x] Update Image UI for parts and company

- [ ]  Update unit tests
- [ ]  Extract user information on upload

### Demo

[Screencast from 2025-07-24 16-00-28.webm](https://github.com/user-attachments/assets/91c2b757-3631-4eac-955b-b156407e423e)

---

### Next Steps & Request for Feedback

This PR covers the core implementation of the new image storage architecture. The main logic, database changes, and UI updates are in place.

Before I proceed with updating the unit tests and update test database data, I would like to request a review of the current approach.

Could you please review the architecture and implementation to confirm if this is the desired direction?

Once the approach is approved, I will finalize the PR by completing the remaining tasks and ensuring full test coverage.

Thank you

### Related Issues

[[FR] Multiple Part Pictures #9788 ](https://github.com/inventree/InvenTree/issues/9788)


src/backend/InvenTree/common/migrations/0041_migrate_part_images_to_uploadedimage.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+
+def forwards(apps, schema_editor):
+    Part = apps.get_model('part', 'Part')
+    UploadedImage = apps.get_model('common', 'UploadedImage')
+
+    for part in Part.objects.all():
+        if not part.image:
+            continue
```
matmair: Please just filter for parts that have images

src/backend/InvenTree/common/migrations/0041_migrate_part_images_to_uploadedimage.py
```diff
@@ -0,0 +1,55 @@
...
+        except Exception:
+            # If file is missing or corrupted, skip this Part
+            continue
+
+        # Read binary data and close the file
+        data = part.image.read()
+        part.image.close()
+
+        # Extract the base filename to avoid path issues
+        filename = os.path.basename(part.image.name)
```
matmair: Please use pathlib

src/backend/InvenTree/common/migrations/0042_migrate_company_images_to_uploadedimage.py
```diff
@@ -0,0 +1,53 @@
...
+from django.db import migrations
+from django.core.files.base import ContentFile
+
+def forwards(apps, schema_editor):
+    Company = apps.get_model('company', 'Company')
+    UploadedImage = apps.get_model('common', 'UploadedImage')
+
+    for company in Company.objects.all():
+        if not company.image:
+            continue
```
matmair: Please use a filter on the object instead

src/backend/InvenTree/common/migrations/0042_migrate_company_images_to_uploadedimage.py
```diff
@@ -0,0 +1,53 @@
...
+        except Exception:
+            # If file is missing or corrupted, skip this Company
+            continue
+
+        # Read binary data and close the file
+        data = company.image.read()
+        company.image.close()
+
+        # Extract the base filename to avoid path issues
+        filename = os.path.basename(company.image.name)
```
matmair: Please use pathlib

src/backend/InvenTree/common/admin.py
```diff
@@ -33,6 +36,25 @@ def formfield_for_dbfield(self, db_field, request, **kwargs):
...
+            db_field.choices = common.validators.get_model_options(
+                InvenTree.models.InvenTreeImageUploadMixin
+            )
+
+        return super().formfield_for_dbfield(db_field, request, **kwargs)
+
+    list_display = ('model_type', 'model_id', 'image', 'primary')
+
+    search_fields = ('image',)
+    list_filter = ('model_type',)
```
matmair: We do not promote usage of the django admin anymore, please remove this

src/backend/InvenTree/common/api.py
```diff
@@ -783,6 +786,54 @@ def destroy(self, request, *args, **kwargs):
...
+
+    filterset_fields = ['model_type', 'model_id']
+
+    search_fields = ['model_type', 'image']
+
+    def list(self, request, *args, **kwargs):
+        """Serialize the available UploadedImage entries for Parts.
+
+        - Images may be used for multiple parts!
+        """
```
matmair: This is also the text that is used in the API docs, please rewrite the text to be more fitting for that.

src/backend/InvenTree/common/api.py
```diff
@@ -783,6 +786,54 @@ def destroy(self, request, *args, **kwargs):
...
+        # Aggregate by image file, counting how many parts use each image
+        agg = qs.values('image').annotate(count=Count('pk')).order_by('-count')
+
+        page = self.paginate_queryset(agg)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
+
+        serializer = self.get_serializer(agg, many=True)
+        return Response(serializer.data)
```
matmair, line 841: question: Not really sure we need this; what deeper purpose does the `count` field serve?

src/backend/InvenTree/common/helpers.py
```diff
@@ -0,0 +1,27 @@
...
+
+    TODO: Future work may be needed here to support other storage backends, such as S3
+    """
+    part_image_directory = os.path.abspath(
+        os.path.join(settings.MEDIA_ROOT, UPLOAD_IMAGE_DIR)
+    )
+
+    # Create the directory if it does not exist
+    if not os.path.exists(part_image_directory):
+        os.makedirs(part_image_directory)
```
matmair: Please use pathlib

src/backend/InvenTree/common/models.py
```diff
@@ -1882,6 +1885,82 @@ def rename_attachment(instance, filename: str):
     )
 
 
+UPLOAD_IMAGE_DIR = 'upload_images'
+
+
+@cleanup.ignore
```
matmair, line 1920: Question: How are images / thumbnails cleaned up then?

src/backend/InvenTree/common/models.py
```diff
@@ -1882,6 +1885,82 @@ def rename_attachment(instance, filename: str):
...
+class UploadedImage(models.Model):
+    """Class which represents an uploaded image for Part and Company.
```
matmair: (Optional): I would prefer this to be more general, please remove the models

src/backend/InvenTree/common/models.py
```diff
@@ -1882,6 +1885,82 @@ def rename_attachment(instance, filename: str):
...
+    """
+
+    model_type = models.CharField(
+        max_length=100,
+        validators=[common.validators.validate_upload_image_model_type],
+        verbose_name=_('Model type'),
+        help_text=_('Target model type for image'),
+    )
+
+    model_id = models.PositiveIntegerField()
```
matmair: This is re-inventing the wheel, please use https://docs.djangoproject.com/en/5.2/ref/contrib/contenttypes/#generic-relations

src/backend/InvenTree/common/serializers.py
```diff
@@ -583,6 +590,146 @@ class Meta:
...
+        Handles the existing_image field.
+        If this is the first image for (model_type, model_id), marks it as primary.
+        """
+        instance = super().save()
+        data = self.validated_data
+
+        existing_image = data.pop('existing_image', None)
+
+        if existing_image:
+            img_path = os.path.join(common_helper.UPLOAD_IMAGE_DIR, existing_image)
```
matmair: Please use pathlib

src/backend/InvenTree/common/serializers.py
```diff
@@ -583,6 +590,146 @@ class Meta:
...
+
+        if remote_img and instance:
+            fmt = remote_img.format or 'PNG'
+            buffer = io.BytesIO()
+            remote_img.save(buffer, format=fmt)
+
+            # Construct a simplified name for the image
+            filename = f'part_{instance.pk}_image.{fmt.lower()}'
+
+            instance.image.save(filename, ContentFile(buffer.getvalue()))
```
matmair: Please wrap this in a try/catch to ensure issues with the buffer/io issues do not create deeper issues and raise an appropriate exception if something occurs

src/backend/InvenTree/common/serializers.py
```diff
@@ -583,6 +590,146 @@ class Meta:
...
+        if not img:
+            return img
+
+        img = img.split(os.path.sep)[-1]
+
+        # Ensure that the file actually exists
+        img_path = os.path.join(common_helper.get_part_image_directory(), img)
+
+        if not os.path.exists(img_path) or not os.path.isfile(img_path):
+            raise ValidationError(_('Image file does not exist'))
```
matmair: Please use pathlib

src/backend/InvenTree/common/serializers.py
```diff
@@ -583,6 +590,146 @@ class Meta:
...
+        fields += ['existing_image']
+
+        return fields
+
+
+class UploadedImageThumbSerializer(serializers.Serializer):
+    """Serializer for a thumbnail of an uploaded image."""
+
+    image = serializers.URLField(read_only=True)
+    count = serializers.IntegerField()
```
matmair: For what do we need this count field? There are no docs around it.

src/backend/InvenTree/company/models.py
```diff
@@ -269,11 +262,11 @@ def get_absolute_url(self):
...
-    def get_image_url(self):
-        """Return the URL of the image for this company."""
-        if self.image:
-            return InvenTree.helpers.getMediaUrl(self.image.url)
-        return InvenTree.helpers.getBlankImage()
+    # def get_image_url(self):
+    #     """Return the URL of the image for this company."""
+    #     if self.image:
+    #         return InvenTree.helpers.getMediaUrl(self.image.url)
+    #     return InvenTree.helpers.getBlankImage()
```
matmair: Question: Why is this not rewritten to provide the primary image?

src/backend/InvenTree/company/serializers.py
```diff
@@ -117,7 +125,7 @@ class CompanySerializer(
 
     export_exclude_fields = ['primary_address']
 
-    import_exclude_fields = ['image']
+    # import_exclude_fields = ['image']
```
matmair: Question: Have you tested that the importer still works?

src/backend/InvenTree/part/api.py
```diff
@@ -479,6 +479,7 @@ class PartTestTemplateList(PartTestTemplateMixin, DataExportViewMixin, ListCreat
     ordering = 'test_name'
 
 
+# TODO: We don't need to this view after implement part image in UploadImage Table
 class PartThumbs(ListAPI):
```
matmair: Comment: Then please remove it

src/backend/InvenTree/part/helpers.py
```diff
@@ -81,20 +77,20 @@ def render_part_full_name(part) -> str:
...
 
-    # Create the directory if it does not exist
-    if not os.path.exists(part_image_directory):
-        os.makedirs(part_image_directory)
+#     # Create the directory if it does not exist
+#     if not os.path.exists(part_image_directory):
+#         os.makedirs(part_image_directory)
 
-    return part_image_directory
+#     return part_image_directory
```
matmair: Please remove code that is not needed anymore

src/backend/InvenTree/part/models.py
```diff
@@ -940,18 +903,6 @@ def get_absolute_url(self):
...
-        """Return the URL of the image for this part."""
-        if self.image:
-            return helpers.getMediaUrl(self.image.url)
-        return helpers.getBlankImage()
-
-    def get_thumbnail_url(self) -> str:
-        """Return the URL of the image thumbnail for this part."""
-        if self.image:
-            return helpers.getMediaUrl(self.image.thumbnail.url)
-        return helpers.getBlankThumbnail()
```
matmair, line 958: Please re-implement these functions

src/backend/InvenTree/part/serializers.py
```diff
@@ -1102,9 +1070,9 @@ def create(self, validated_data):
                 instance.notes = original.notes
                 instance.save()
 
-            if duplicate.get('copy_image', False):
-                instance.image = original.image
-                instance.save()
+            # if duplicate.get('copy_image', False):
+            #     instance.image = original.image
+            #     instance.save()
```
matmair: Please remove code that is not needed anymore

src/frontend/src/components/render/Order.tsx
```diff
@@ -24,7 +24,7 @@ export function RenderPurchaseOrder(
         status: instance.status_custom_key,
         type: ModelType.purchaseorder
       })}
-      image={supplier.thumbnail || supplier.image}
+      image={supplier?.image.thumbnail || supplier?.image.image}
```
matmair: Does this mean there is no placeholder image anymore?

src/frontend/src/pages/part/PartDetail.tsx
```diff
@@ -611,32 +604,29 @@ export default function PartDetail() {
...
-            <Grid.Col span={{ base: 12, sm: 8 }}>
-              <DetailsTable fields={tl} item={data} />
-            </Grid.Col>
-          </Grid>
-          {enableRevisionSelection && (
-            <Stack gap='xs'>
-              <Text>{t`Select Part Revision`}</Text>
-              <RevisionSelector part={part} options={partRevisionOptions} />
-            </Stack>
-          )}
```
matmair, line 760: Why is this removed?

src/frontend/src/pages/sales/SalesOrderDetail.tsx
```diff
@@ -584,8 +583,8 @@ export default function SalesOrderDetail() {
         <Stack gap='xs'>
           <PageDetail
             title={`${t`Sales Order`}: ${order.reference}`}
-            subtitle={subtitle}
-            imageUrl={order.customer_detail?.image}
+            subtitle={order.description}
```
matmair: Why is this changed?

src/frontend/src/tables/part/PartThumbTable.tsx
```diff
@@ -99,30 +101,44 @@ function PartThumbComponent({
...
+    })
+    .then((response) => {
+      // Update the image in the parent component
+      modals.closeAll();
+      onSuccess(response.data.image);
+      return response;
+    })
+    .catch((error) => {
+      // Handle error
+      console.error('Error setting new image:', error);
```
matmair: Please remove this

src/frontend/src/tables/ColumnRenderers.tsx
```diff
@@ -34,10 +34,18 @@ export function PartColumn({
   part: any;
   full_name?: boolean;
 }) {
+  const get_image_src = () => {
+    // Get primary image source
+    const primary_image = part?.images?.find(
+      (img: any) => img.primary === true
+    );
+    return primary_image?.thumbnail ?? primary_image?.image;
+  };
```
matmair: Question: Did you look into performance with many images?

matmair (commented):
This is a good start; a few comments:
- I think `UploadedImage` is a misleading name; all images are uploaded in some form so it should be just `Image` (maybe `InvenTreeImage`)
- I think it would be better to use generic-relations instead of building your own choice and validation mechanisms as that enables easier use within plugins
- the removal of thumbnail in the apis makes no sense to me - why not link the thumbnail of the primary image and break less stuff?
- In a few places code is commented out instbeing beeing removed; not a fan of that, please clean that up
- os.path can have issues when switching between operating system, please use pathlib.Path - here is a conversion table https://docs.python.org/3/library/pathlib.html#corresponding-tools


Things to polish this a bit more:
- add a bit of docs for how images work
- add a control to the Admin Center to view and possibly remove images instead of the old django admin page
- add unit test (you already have that listed)
- update the description of your PR to describe how this change the api for users

matmair:
Also: This needs a new api revision - see https://docs.inventree.org/en/stable/develop/contributing/#api-versioning - and CI won't run till the conflicts are resolved

src/backend/InvenTree/InvenTree/models.py
```diff
@@ -1061,6 +1061,62 @@ def delete(self, *args, **kwargs):
...
+        Note: abstract must be true, as this is only a mixin, not a separate table
+        """
+
+        abstract = True
+
+    def delete(self):
+        """Custom delete method for InvenTreeImageUploadMixin.
+
+        - Before deleting the object, check if there are any uploaded images associated with it.
+        - If so, delete the images first
```
SchrodingersGat: Can an image be uploaded against multiple objects simultaneously? Or is it a single image-to-object association?

src/backend/InvenTree/common/admin.py
SchrodingersGat: I don't think "not promoting" means "not support" here - for ~20 LOC we get a low-level admin interface for free which does come in handy sometimes?

src/backend/InvenTree/common/models.py
SchrodingersGat: e.g. 

```python
"""Class which represents an uploaded image linked to another model instance."""
```

src/backend/InvenTree/common/models.py
```diff
@@ -1882,6 +1885,82 @@ def rename_attachment(instance, filename: str):
...
+        with transaction.atomic():
+            # Delete the instance
+            super().delete(*args, **kwargs)
+
+            # If it was primary, promote the next candidate
+            if self.primary:
+                next_img = (
+                    UploadedImage.objects.filter(
+                        model_type=self.model_type, model_id=self.model_id
+                    )
```
SchrodingersGat, line 2009: You probably also want to exclude the current ID from this filter - inside an atomic transaction, can we guarantee that the "self" ID hasn't been removed yet?

src/backend/InvenTree/common/serializers.py
```diff
@@ -583,6 +590,146 @@ class Meta:
...
+
+        if len(self.fields['model_type'].choices) == 0:
+            self.fields['model_type'].choices = common.validators.get_model_options(
+                InvenTreeImageUploadMixin
+            )
+
+    image = InvenTree.serializers.InvenTreeImageSerializerField(
+        required=False, allow_null=True
+    )
+    thumbnail = serializers.CharField(source='get_thumbnail_url', read_only=True)
```
SchrodingersGat: This may cause issues because it is calling a function on each item in the dataset - we will need to carefully test that it does not execute a high number of database queries when performing a LIST API operation.

src/backend/InvenTree/common/migrations/0042_migrate_company_images_to_uploadedimage.py
```diff
@@ -0,0 +1,53 @@
...
+Migration to copy the image file from the Company model into the UploadedImage model.
+'''
+
+import os
+from django.db import migrations
+from django.core.files.base import ContentFile
+
+def forwards(apps, schema_editor):
+    Company = apps.get_model('company', 'Company')
+    UploadedImage = apps.get_model('common', 'UploadedImage')
```
SchrodingersGat: This migration must be in the "company" app, and it **must** be before the new `0076_remove_company_image` migration file. Otherwise, the images may be deleted *before* this migration runs!

Alternatively, add *this* migration as a requirement for the `0076` migration in "company" app.

src/backend/InvenTree/company/serializers.py
```diff
@@ -50,15 +47,26 @@ class Meta:
...
             'currency',
             'tax_id',
         ]
         read_only_fields = ['currency']
 
-    image = InvenTreeImageSerializerField(read_only=True)
+    # image = InvenTreeImageSerializerField(read_only=True)
+    image = serializers.SerializerMethodField()
+
+    def get_image(self, obj):
```
SchrodingersGat: This approach is going to really slow down the API when you list *all* company objects. For each company it will need to go back to the database (potentially multiple times).

The implementation must annotate the primary image against the model instance in a single DB query. This should be made generic so we can apply the same annotation across multiple model serializers as required. Look at `part/filters.py` for how we have been implementing generic annotations like this.

src/backend/InvenTree/part/migrations/0140_remove_part_image.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.22 on 2025-07-22 18:58
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('part', '0139_remove_bomitem_overage'),
```
SchrodingersGat: Make sure the dependencies here are such that this migration DOES NOT RUN until the images have been migrated correctly.

src/backend/InvenTree/part/migrations/0140_remove_part_image.py
```diff
@@ -0,0 +1,17 @@
...
+    ]
+
```
SchrodingersGat: You should add some migration unit testing around this, as it is a *significant* change to the database.

src/backend/InvenTree/part/models.py
```diff
@@ -530,42 +529,6 @@ def delete(self, **kwargs):
 
         super().delete()
 
-    def save(self, *args, **kwargs):
```
SchrodingersGat, line 538: Not all the code you have removed here was associated with the image validation.

Please ensure that you are not deleting code that is still useful. Please revert the changes here that are not related to the images.

src/backend/InvenTree/part/serializers.py
```diff
@@ -367,10 +363,9 @@ def __init__(self, *args, **kwargs):
         read_only=True, allow_null=True
     )
 
-    image = InvenTree.serializers.InvenTreeImageSerializerField(
-        read_only=True, allow_null=True
+    images = common_serializers.UploadedImageSerializer(
+        source='get_images', many=True, read_only=True, allow_null=True
```
SchrodingersGat: This is (once again) going to cause significant API query speed issues due to database hits.

We will need to think very carefully about query annotations before this is mergable.

src/backend/InvenTree/common/migrations/0041_migrate_part_images_to_uploadedimage.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/migrations/0041_migrate_part_images_to_uploadedimage.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/migrations/0042_migrate_company_images_to_uploadedimage.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/migrations/0042_migrate_company_images_to_uploadedimage.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/admin.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/api.py
Reza98Sh: The sorting by count is to show the most used images first in the selection UI for better usability.

The count value is also displayed directly in the UI (inside parentheses) to show how many times each image is in use.

 [see here](https://github.com/inventree/InvenTree/blob/6e81fa2e7b36a078446f09a839d7e0c73b092f61/src/frontend/src/tables/part/PartThumbTable.tsx#L41). 
 
Screenshot from 2025-07-25 13-13-25

src/backend/InvenTree/common/helpers.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/models.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/models.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/models.py
Reza98Sh: Good point. I was initially following the pattern from the Attachment model, but I agree that GenericForeignKey is a better approach. I’ll make the change.

src/backend/InvenTree/common/serializers.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/serializers.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/serializers.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/common/serializers.py
Reza98Sh: To display that count in the selection UI:

Screenshot from 2025-07-25 13-13-25

src/backend/InvenTree/company/models.py
Reza98Sh: get_image_url is no longer applicable because for delete action you need to access  image id .

src/backend/InvenTree/company/serializers.py
Reza98Sh: You’re right, I missed testing this. I’ll fix it in the next commit.

src/backend/InvenTree/part/api.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/part/helpers.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/part/models.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/part/serializers.py
Reza98Sh: thanks! Got it. Will update soon.

src/frontend/src/pages/part/PartDetail.tsx
Reza98Sh: You’re right, this was removed by mistake during a local merge. I’ll restore it immediately. Sorry about that.

src/frontend/src/pages/sales/SalesOrderDetail.tsx
Reza98Sh: You’re right, this was changed by mistake during a local merge. I’ll restore it immediately. Sorry about that.

src/frontend/src/tables/part/PartThumbTable.tsx
Reza98Sh: thanks! Got it. Will update soon.

src/frontend/src/tables/ColumnRenderers.tsx
Reza98Sh: I see two options to optimize this:

1. Add the primary image URL directly from the backend serializer.
2. Use useMemo on the frontend to avoid re-computation.

I believe option 1 is the superior approach as it’s more performant and keeps the frontend logic cleaner. 

useMemo would be more of a patch. Do you agree? I’ll proceed with the backend change.

src/backend/InvenTree/InvenTree/models.py
Reza98Sh: The implementation allows for a many-to-many association, meaning an image can indeed be linked to multiple objects simultaneously via the “select existing image” feature. 
It doesn't save duplicate images.

src/backend/InvenTree/common/models.py
Reza98Sh: Good catch. You’re right, I should exclude self.pk from the query to prevent a potential race condition. I’ll fix that. Thanks!

src/backend/InvenTree/common/serializers.py
Reza98Sh: You’re right to point out the risk of an N+1 query problem on list views.

I will investigate this.

src/backend/InvenTree/common/migrations/0042_migrate_company_images_to_uploadedimage.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/company/serializers.py
Reza98Sh: Thanks for the crucial feedback. You’re right.

I understand the need to use a generic annotation here instead of a SerializerMethodField. I will look at the example in part/filters.py and implement a  solution based on that pattern to solve this properly.

src/backend/InvenTree/part/migrations/0140_remove_part_image.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/part/migrations/0140_remove_part_image.py
Reza98Sh: thanks! Got it. Will update soon.

src/backend/InvenTree/part/models.py
Reza98Sh: thanks! Got it. I Will double check this out.

src/backend/InvenTree/common/serializers.py
SchrodingersGat: Happy to provide some pointers or assistance in this area, it's a very nuanced concept

SchrodingersGat:
I'm going to push this out to the next release cycle, as I can see it taking quite some time to get it perfect.

src/backend/InvenTree/common/serializers.py
Reza98Sh: Thanks a lot, I really appreciate the offer! I’ll take an initial look myself, and will definitely ping you if I have any questions or could use some pointers. My initial thought was to use a tool like django-silk to profile the queries and check for optimization opportunities.

matmair: Our dev setup already shops with a variety of tools and should alert you once you add a large enough test-dataset; they run during backend tests

src/backend/InvenTree/common/api.py
matmair: We will have to talk about this specific implementation (the whole function, not the annotation), there are a lot of ways this can break. Either it needs to be unit tested  or the annotation needs to be moved out

src/backend/InvenTree/common/models.py
matmair: As discussed in the original issue, attachments are not really a perfect design themself, this is one of those points.

src/backend/InvenTree/company/models.py
matmair: That does not really answer my question; there might be (I at least have some) plugin code that uses this function - so is there a good reason to remove it? It has uses in report-generation plugins for example

src/frontend/src/tables/ColumnRenderers.tsx
matmair: We use useMemo a lot, it is the right approach if you only can / want to touch the frontend.
As pointed out in the backend comments, I think the old (primary) image and thumbnail fields should be preserved so this would not be needed.

Reza98Sh:
@SchrodingersGat 

Excuse me, I have a trouble with images fields in PartBriefSerializer, 
This serializer has been used in other serializers in several place. 

My question is: What is the best practice in this scenario to resolve the N+1 issue? Should I go through every view/serializer that uses PartBriefSerializer and modify its queryset to use

Reza98Sh:
@matmair, @SchrodingersGat 

I am planning to create a new test file at common/test_api.py to house the tests for our image management API. My goal is to move the existing image-related tests from the Part application to this new central file.

Does this approach seem reasonable to you?

src/backend/InvenTree/common/models.py
Reza98Sh: In the last update of the PR, I rewrote StdImageField to clean images if all references become 0.

Reza98Sh:
@matmair, @SchrodingersGat,

I’m looking into updating  our tests for the image management API. I see two possible approaches, and I’d like to get your opinion on which is better.

Option 1: We could update the existing image-related tests within the Part application to be more comprehensive.

Option 2: We could create a new, centralized test file at common/test_api.py and move all image-related tests there, making them more generic.

What do you think is the best path forward?

src/backend/InvenTree/common/serializers.py
Reza98Sh: I’m encountering a test failure in part.test_api.PartDetailTests.test_part_operations related to an excessive number of database queries.

The failure occurs during the PATCH request to rename a part:

response = self.patch(url, {'name': 'a new better name'})

The test expects the operation to complete in fewer than 250 queries, but the test runner reports 346 queries, causing an AssertionError.

Here’s the interesting part: to investigate this, I manually performed the same part rename operation on a development server with django-silk enabled for profiling. Silk reported only 92 queries for the entire request.

This is a significant discrepancy, and it suggests the issue might be specific to the test environment’s query counting mechanism rather than the core logic of the rename operation itself.

matmair: The test environment is staged with a reasonable amount of data to find o+1 issues. You will need to run silk with the test running to reproduce this; certainly sounds like something is iterating over parts.

matmair:
If the refactored PR is generic, it should have tests in common/; if there is still image-related code in part/ those parts should be covered by tests in the part/ directory.

this will be easier to judge when there are no conflicts anymore and the CI is able to run through. You will be provided with a visual coverage breakdown and a clear list what parts of code still require coverage here in the PR by a bot

src/backend/InvenTree/common/serializers.py
Reza98Sh: I found the problem it related to django-silk. :)
In test environment it count silk query also that cause excessive number of database queries

Reza98Sh:
@SchrodingersGat  @matmair 

Thank you for the feedback!
I learned a lot from your comments and have refactored  based on them.

Here’s a summary of the changes:

- Replaced all os path operations with pathlib
- Refactored migrate_company_images.py and migrate_part_images.py for better efficiency. During this refactoring, I discovered that Part images were stored in duplicate. To resolve this, I implemented a sha256–based mechanism to prevent duplicate image storage.Later, I thought we could adopt this mechanism more broadly instead of the current existing_image check — by storing the image hash and comparing it when uploading new images to avoid duplicates.
- Improved function docstrings for better API documentation generation.
- Implemented a custom field CustomStdImage(StdImageField) to automatically delete unused images.
- Renamed the UploadedImage class to InvenTreeImage, and using GenericForeignKey for more flexible model relations.
- Added image_url and thumbnail_url fields, along with an InvenTreeImageSerializerMixin to make image serialization more reusable.
- Reworked the part duplication mechanism to properly handle new images.
- Restored code that was unintentionally removed during a previous merge.
- Reviewed all new and modified API endpoints for potential N+1 query issues and optimized them where needed.
- Added dedicated API tests for InvenTreeImage.
- Created TestLegacyImageMigration to test old image migration functionality.
- Addressed several other minor issues mentioned in the feedback.

matmair:
@Reza98Sh can you please address the merge conflicts before we starts another review round?
You can do this by pulling from the master branch

Reza98Sh:
Sure, I’ll address the merge conflicts and pull from master as soon as possible.

Reza98Sh:
Conflicts resolved and branch updated.

matmair:
Thank you; if will wait for the CI and then run some on-prem tests

Reza98Sh:
It seems some tests have failed. My plan was to review and update all other the tests in the next PR, after the proposed implementation for images is approved.

Reza98Sh:
@matmair @SchrodingersGat

Excuse me, I am preparing for Hacktoberfest and I have several PRs planned for October (including this one).

I would like to close this PR for now, fix all test failures, and then open a clean PR in October, so it will count for Hacktoberfest together with my other planned PRs.

Would that be okay for you?

matmair:
Sure, go ahead

PR #10082, opened 2025-07-26
[plugin] Exchange rate host plugin

SchrodingersGat:
Re-add option for exchangerate.host currency exchange, as a builtin plugin.


SchrodingersGat:
@matmair I'm not sure whether it makes more sense to introduce this as a builtin or standalone plugin. The exchangerate.host provides more currency rates than the frankfurter service - but requires an API key to work.

matmair:
I think it should be a standalone plugin, it requires payment/registration to function and makes initial setup more confusing

src/backend/InvenTree/plugin/builtin/integration/currency_exchange.py
```diff
@@ -47,3 +47,64 @@ def update_exchange_rates(self, base_currency: str, symbols: list[str]) -> dict:
...
+    DESCRIPTION = _('Fetches currency exchange rates from the exchangerate.host API')
+    VERSION = '1.0.0'
+
+    SETTINGS = {
+        'API_KEY': {
+            'name': _('API Key'),
+            'description': _('API key for exchangerate.host'),
+            'default': '',
+        }
+    }
```
matmair, line 72: This should probably be marked as secret to not leak keys to every user

SchrodingersGat:
> I think it should be a standalone plugin, it requires payment/registration to function and makes initial setup more confusing

Ok, I'll make it a separate plugin.

PR #10086, opened 2025-07-26
Update to Bare Metal installation docs

gunstr:
Some fixes to the Bare Metal installation instruction.

Ref. #10033

### DIFF
--- a/docs/docs/start/install.md
+++ b/docs/docs/start/install.md
@@ -163,6 +163,14 @@ The PostgreSQL python binding must also be installed (into your virtual environm
 pip3 install psycopg pgcli
 ```
 
+#### Install Postgresql client
+
+If PostgreSQL and InvenTree are installed on separate servers / containers the PostgreSQL client has to be installed also where InvenTree is running.
+
+```
+sudo apt-get install postgresql-client
+```
+
 ### MySQL / MariaDB
 
 #### Install Backend
--- a/contrib/deploy/supervisord.conf
+++ b/contrib/deploy/supervisord.conf
@@ -23,7 +23,7 @@ port = 127.0.0.1:9001
 ; InvenTree Web Server Process
 [program:inventree-server]
 user=inventree
-directory=/home/inventree/src/InvenTree
+directory=/home/inventree/src/src/backend/InvenTree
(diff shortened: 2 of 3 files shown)


gunstr:
@SchrodingersGat Is the Supervisor config file used in any other installation setups where the path might be different?

docs/docs/start/install.md
```diff
@@ -163,6 +163,14 @@ The PostgreSQL python binding must also be installed (into your virtual environm
 pip3 install psycopg pgcli
 ```
 
+#### Install Postgresql client
+
+If PostgreSQL and InvenTree are installed on separate servers / containers the PostgreSQL client has to be installed also where InvenTree is running.
```
SchrodingersGat, line 168: Not quite correct - the client libs are required even if connecting to postgres on a localhost.

SchrodingersGat:
> @SchrodingersGat Is the Supervisor config file used in any other installation setups where the path might be different?

No, I don't think so

docs/docs/start/install.md
gunstr: Is that really so? When I had problems with the containerized installation I did a test to install InvenTree and Postgres on the same VM and as I recall nothing else than the Postgres installation command `sudo apt-get install postgresql postgresql-contrib libpq-dev` was required. I belevie the client libs are then installed as a dependency.

SchrodingersGat: > I belevie the client libs are then installed as a dependency.

Yes this is correct. My clarification was more that the client libs are required even if you are connecting to a local db instance, not just a remote one.

So your text here could be more general, like "The postgresl client libraries need to be installed, if they are not already. This applies whether you are connecting to a local database instance, or a remote one"?

gunstr: I get your point but as this written at the end of the PostgreSQL section the user have just followed the database installation instruction. If done locally the command to install the complete database including the client library has just been execued a few lines above in the docs why an action from the user is only required in case of a remote database?

SchrodingersGat: Ok, no that's fair. In that case I think that your docs as written are fine :)

PR #10088, opened 2025-07-27
Transition plugin

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/9571
- Move "transition methods" to the plugin ecosystem entirely
- Ensure that transition methods are loaded by the plugin registry
- Prevent disabled plugins from registering transition methods

### Breaking

This is a *breaking change* for any plugins which register state transition methods.

Previously, registration was implicit - the plugin only needed to define a state transition class.

Now, registration requires that the plugin:

- Inherit from the `TransitionMixin` plugin mixing class
- Define a list of transition handler class instances

### Tasks

- [x] Update unit testing
- [x] Add documentation for new plugin mixing type
- [ ] add entry to release blog


SchrodingersGat:
@matmair  I think this is a reasonably minimal change which brings the "transition handlers" inline with the updated and "hardened" plugin system. This is (I think) the last change to the plugin system required before 1.0.0 release

From your comments in the linked issue, I have made this as close to "drop in" as possible. Any plugins just need to inherit from the new mixin class, and return a list of transition handler classes

matmair:
This is a small (LOC) but massive change that has good potential to completely brick plugin-based security and compliance plugins. I will have to assess the changes and the operations implications once I am done with my notification backlog.

SchrodingersGat:
> This is a small (LOC) but massive change that has good potential to completely brick plugin-based security and compliance plugins. I will have to assess the changes and the operations implications once I am done with my notification backlog.

Understood. My intent here is to make these breaking changes before we release 1.0.0. After which point, it becomes even harder to make these changes. 

I'm happy to give you time to assess and potentially refactor plugins as needed, and hold off on the next release. But I do want to ensure that "rough edges" on the plugin system are fixed before we move forward.

SchrodingersGat:
@matmair thoughts on this one now that #10094 is merged in?

PR #10092, opened 2025-07-28
Add missing nullable annotations, search_field schema documentation

1337joe:
A couple quick fixes before I get back to working on #9969:

- Fix a few details fields that crept in without `allow_null=True` (so they're required by the schema but the fields aren't present unless requested)
- Add the list of search fields to the `search` parameter comments in the schema (similar to the `ordering` parameter handling)

### DIFF
--- a/src/backend/InvenTree/InvenTree/schema.py
+++ b/src/backend/InvenTree/InvenTree/schema.py
@@ -103,6 +103,16 @@ class ExtendedAutoSchema(AutoSchema):
                         f'{parameter["description"]} Possible fields: {", ".join(ordering_fields)}.'
                     )
 
+        # Add valid search fields to the search description.
+        search_fields = getattr(self.view, 'search_fields', None)
+        if search_fields is not None:
+            parameters = operation.get('parameters', [])
+            for parameter in parameters:
+                if parameter['name'] == 'search':
+                    parameter['description'] = (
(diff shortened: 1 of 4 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@1337joe thanks for your continued improvement in this area 😃

SchrodingersGat:
@1337joe can you please have a look at this again - your change here is causing API schema errors in https://github.com/inventree/InvenTree/pull/10094

The issue is that the search field annotation is not sorted - and because the fields are sometimes provided in a different order, it can throw a "difference" in the API schema.

So perhaps, the search fields just need to be sorted to ensure a consistent result

src/backend/InvenTree/InvenTree/schema.py
```diff
@@ -103,6 +103,16 @@ def get_operation(
                         f'{parameter["description"]} Possible fields: {", ".join(ordering_fields)}.'
                     )
 
+        # Add valid search fields to the search description.
+        search_fields = getattr(self.view, 'search_fields', None)
```
SchrodingersGat, line 107: Please sort this :)

SchrodingersGat:
I have submitted a fix here - https://github.com/inventree/InvenTree/pull/10095

PR #10094, opened 2025-07-28
[plugin] Mandatory plugins

SchrodingersGat:
### Related PRs

- https://github.com/inventree/InvenTree/pull/9718
- https://github.com/inventree/InvenTree/pull/9339

### Overview

This PR adds a function to specify in the server configuration that certain plugins are considered "mandatory". These plugins cannot be disabled, allowing a particular configuration to be "locked down" and prevent certain functionality from being disabled.

This is a follow-on from #9718 which provided a way to override global settings values in the config file, and #9339 which added the concept of marking some "builtin" plugins as mandatory

### Tasks

- [x] Implement "mandatory" functionality
- [x] Add unit testing to ensure mandatory plugins are enabled on startup
- [x] Add unit testing to ensure that these plugins cannot be disabled
- [x] Ensure checks for "mandatory" / "active" states are not offloaded to the plugin code
- [x] Update documentation around mandatory plugins
- [x] Increase code coverage for plugin loading / unloading mechanisms
- [x] Ensure "enable" / "disable" via API is correctly disabled by backend setting


matmair:
This looks like it solves some mayor concerns with https://github.com/inventree/InvenTree/pull/10088, so far I have not been able to break it without using root (which is meaningless as you have lost at that point already)
One thing I am still to tinker with is name confusion / typo squatting like attacks

SchrodingersGat:
@matmair that's good to hear. Please LMK if you find any issues, or have some suggestions for more points of test coverage

matmair (approved):
This is a great addition; this should mitigate most concerns I have. I am looking into a follow-up that ensures there can be less publisher confusion

src/backend/InvenTree/InvenTree/test_api.py
matmair, line 1: nice efficiency gains

SchrodingersGat: Now I have to work out why the tests are all failing on the github CI - "it works on my machine"

src/backend/InvenTree/InvenTree/test_api.py
```diff
@@ -603,7 +607,16 @@ def test_info_view(self):
         # Check for active plugin list
         self.assertIn('active_plugins', data)
         plugins = data['active_plugins']
-        self.assertGreater(len(plugins), 0)
+
+        # Check that all active plugins are listed
+        N = len(plugins)
```
matmair: has the auth state of the test changed? There should be no plugins listed in an anon connection

src/backend/InvenTree/InvenTree/unit_test.py
```diff
@@ -372,25 +372,46 @@ def assertNumQueriesLessThan(
         self.assertLess(n, value, msg=msg)
 
 
-class InvenTreeTestCase(ExchangeRateMixin, UserMixin, TestCase):
-    """Testcase with user setup build in."""
-
-
-class InvenTreeAPITestCase(ExchangeRateMixin, TestQueryMixin, UserMixin, APITestCase):
-    """Base class for running InvenTree API tests."""
+class PluginRegistryMixin:
```
matmair, line 375: Would it be possible to flag this to ensure it only introduces a performance hit if plugins are required for tests?

src/backend/InvenTree/InvenTree/test_api.py
SchrodingersGat: It appears to be the same on `master`. Should I fix it as part of this PR?

SchrodingersGat: I have now ensured that this is list only available to staff users

src/backend/InvenTree/InvenTree/unit_test.py
SchrodingersGat: Yep, have refactored this now so it only loads when required

src/backend/InvenTree/InvenTree/test_api.py
SchrodingersGat: For reference, I have been testing with `--keepdb` (for speed) which was not removing the old `PluginConfig` objects

src/backend/InvenTree/InvenTree/test_api.py
matmair: I have checked again and you are right, I must have confused something. IMO we can roll back the change or keep it, the filed is not used by any function I know off

PR #10107, opened 2025-07-31
feat: Add support for cache password in Redis configuration

the78mole:
This pull request introduces support for using a password in the Redis cache configuration. The changes include adding a new utility function to fetch the cache password and updating the Redis URL construction logic to include the password if provided.

### Enhancements to Redis cache configuration:
* `src/backend/InvenTree/InvenTree/cache.py`:
  - Added a new function `cache_password()` to retrieve the cache password from settings.
  - Updated the `get_cache_config` function to construct the Redis URL with an optional password, improving security by supporting password-protected Redis instances.


SchrodingersGat:
@the78mole thanks for the contribution, this is a nice addition to redis support

PR #10119, opened 2025-08-01
feat(backend): more error messages

matmair:
adds a few error messages


SchrodingersGat:
This is a solid improvement, nice 👍

PR #10121, opened 2025-08-03
chore(backend): increase coverage

matmair:
Increase coverage on various areas:
- cleans up notification method that is not used anymore
- cleans up startup method that is never called
- adds more tests for notifications
- adds migrations test for currency migrations
- adds e2e plugin install/uninstall test via api

### DIFF
--- a/src/backend/InvenTree/common/apps.py
+++ b/src/backend/InvenTree/common/apps.py
@@ -56,7 +56,9 @@ class CommonConfig(AppConfig):
 
                 if current_value != value:
                     logger.info(
-                        'Overriding global setting: %s = %s', value, current_value
+                        'INVE-I1: Overriding global setting: %s = %s',
+                        value,
+                        current_value,
                     )
                     set_global_setting(key, value, None, create=True)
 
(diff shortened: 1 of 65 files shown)


src/backend/InvenTree/common/apps.py
```diff
@@ -40,28 +35,5 @@ def clear_restart_flag(self):
...
-            try:
-                current_value = get_global_setting(key, create=False)
-
-                if current_value != value:
-                    logger.info(
-                        'INVE-I1: Overriding global setting: %s = %s',
-                        value,
-                        current_value,
-                    )
-                    set_global_setting(key, value, None, create=True)
```
matmair, line 63: @SchrodingersGat this part of the code is never reachable, as the value returned `get_global_setting` will always be overridden by the override value it is compared to- there can be no difference at this point. Therefore, I have removed the whole section

SchrodingersGat: Ok, thanks for the catch

SchrodingersGat (approved):

matmair:
This PR should bring coverage back up over the current targets

matmair:
@SchrodingersGat ready for review, seems like Postgres is more caring about order than other dbms

SchrodingersGat:
@matmair nice work, this is a good improvement :)

PR #10136, opened 2025-08-06
feat(contrib): pin packages in docker

matmair:
WIP Currently the apk packages in the docker-file are not pinned and therefore builds not reproducible
This is a test to see how pinning them might work


matmair:
After reading a lot more about alpine it seems like building reproducible alpine is a lot harder than reproducible debian

PR #10140, opened 2025-08-06
feat(backend): add storages to make usage of s3/sftp easier

matmair:
WIP

### DIFF
--- a/docs/docs/start/config.md
+++ b/docs/docs/start/config.md
@@ -378,6 +378,32 @@ Database and media backups **require** a local directory for storage. This direc
 
 Alternatively this location can be specified with the `INVENTREE_BACKUP_DIR` environment variable.
 
+
+### Storage backends
+
+It is also possible to use alternative storage backends for static and media files, at the moment there is direct provide direct support bundled for S3 and SFTP. Google cloud storage and Azure blob storage would also be supported by the [used library](https://django-storages.readthedocs.io), but require additional packages to be installed.
+
+| Environment Variable | Configuration File | Description | Default |
+| --- | --- | --- | --- |
+| INVENTREE_STORAGE_TARGET | storage.target | Storage target to use for static and media files, valid options: local, s3, sftp | local |
+| INVENTREE_S3_ACCESS_KEY | storage.s3.access_key | Access key | *Not specified* |
+| INVENTREE_S3_SECRET_KEY | storage.s3.secret_key | Secret key |
+| *Not specified* |
+| INVENTREE_S3_BUCKET_NAME | storage.s3.bucket_name | Bucket name, required by most providers |
+| *Not specified* |
+| INVENTREE_S3_DEFAULT_ACL | storage.s3.default_acl | Default ACL for uploaded files, defaults to provider default if not set | *Not specified* |
+| INVENTREE_S3_REGION_NAME | storage.s3.region_name | S3 region name |
+| *Not specified* |
+| INVENTREE_S3_ENDPOINT_URL | storage.s3.endpoint_url | Custom S3 endpoint URL, defaults to AWS endpoints if not set |
+| *Not specified* |
+| INVENTREE_S3_VERIFY_SSL | storage.s3.verify_ssl | Verify SSL certificate for S3 endpoint | True |
+| INVENTREE_S3_LOCATION | storage.s3.location | Sub-Location that should be used | inventree-server |
+| INVENTREE_SFTP_HOST | storage.sftp.host | SFTP host | *Not specified* |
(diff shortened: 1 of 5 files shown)


matmair:
@SchrodingersGat this would be ready for a review as well; I would wait for #10484 for a final update and merge to make that one a bit easier

src/backend/InvenTree/InvenTree/settings.py
```diff
@@ -1560,3 +1561,76 @@
 
 if SITE_URL and not TESTING:  # pragma: no cover
     SPECTACULAR_SETTINGS['SERVERS'] = [{'url': SITE_URL}]
+
```
SchrodingersGat, line 1567: This is starting to bloat the `settings.py` file - can we push these out into a `storages.py` file - much like we do with other large settings groups?

src/backend/InvenTree/InvenTree/settings.py
```diff
@@ -1560,3 +1561,76 @@
...
+# Storage backends
+STORAGE_TARGET = get_setting(
+    'INVENTREE_STORAGE_TARGET', 'storage.target', 'local', typecast=str
```
SchrodingersGat: We should validate that the provided `INVENTREE_STORAGE_TARGET` variable is one of the valid options we actually support.

docs/docs/start/config.md
```diff
@@ -380,6 +380,35 @@ Database and media backups **require** a local directory for storage. This direc
...
 
+
+### Storage backends
+
+It is also possible to use alternative storage backends for static and media files, at the moment there is direct provide direct support bundled for S3 and SFTP. Google cloud storage and Azure blob storage would also be supported by the [used library](https://django-storages.readthedocs.io), but require additional packages to be installed.
+
+| Environment Variable | Configuration File | Description | Default |
+| --- | --- | --- | --- |
+| INVENTREE_STORAGE_TARGET | storage.target | Storage target to use for static and media files, valid options: local, s3, sftp | local |
+| INVENTREE_S3_ACCESS_KEY | storage.s3.access_key | Access key | *Not specified* |
```
SchrodingersGat, line 396: Can we split the `_S3_` keys into their separate sub-table, showing that they only apply if `s3` is selected

docs/docs/start/config.md
```diff
@@ -380,6 +380,35 @@ Database and media backups **require** a local directory for storage. This direc
...
+| INVENTREE_S3_REGION_NAME | storage.s3.region_name | S3 region name |
+| *Not specified* |
+| INVENTREE_S3_ENDPOINT_URL | storage.s3.endpoint_url | Custom S3 endpoint URL, defaults to AWS endpoints if not set |
+| *Not specified* |
+| INVENTREE_S3_LOCATION | storage.s3.location | Sub-Location that should be used | inventree-server |
+| INVENTREE_S3_DEFAULT_ACL | storage.s3.default_acl | Default ACL for uploaded files, defaults to provider default if not set | *Not specified* |
+| INVENTREE_S3_VERIFY_SSL | storage.s3.verify_ssl | Verify SSL certificate for S3 endpoint | True |
+| INVENTREE_S3_OVERWRITE | storage.s3.overwrite | Overwrite existing files in S3 bucket | False |
+| INVENTREE_S3_VIRTUAL | storage.s3.virtual | Use virtual addressing style - by default False -> `path` style, `virtual` style if True | False |
+
```
SchrodingersGat, line 410: Can we split the `_SFTP_` keys into their separate subtable, showing that they only apply if `sftp` is selected

src/backend/InvenTree/part/api.py
```diff
@@ -513,9 +513,7 @@ def list(self, request, *args, **kwargs):
         queryset = self.filter_queryset(self.get_queryset())
 
         # Return the most popular parts first
-        data = (
-            queryset.values('image').annotate(count=Count('image')).order_by('-count')
-        )
+        data = queryset.annotate(count=Count('image')).order_by('-count')
```
SchrodingersGat: Is this change relevent here?

src/backend/InvenTree/InvenTree/helpers.py
```diff
@@ -176,6 +176,8 @@ def constructPathString(path: list[str], max_chars: int = 250) -> str:
 
 def getMediaUrl(filename):
     """Return the qualified access path for the given file, under the media directory."""
+    if settings.STORAGE_TARGET == 's3':
```
SchrodingersGat: I suggest that we add an enum for the valid `STORAGE_TARGET` values - so that we are not performing string comparisons against raw values in multiple places across the codebase.

src/backend/InvenTree/part/api.py
matmair: I came across it when profiling the code, I will remove it

matmair:
@SchrodingersGat I have addressed the review comments, lmk if there is anything else

SchrodingersGat (approved):

SchrodingersGat:
Nice work :)

PR #10145, opened 2025-08-07
Docker fixes again

SchrodingersGat:
- Relax query requirements for unit test

### DIFF
--- a/src/backend/InvenTree/InvenTree/unit_test.py
+++ b/src/backend/InvenTree/InvenTree/unit_test.py
@@ -8,6 +8,7 @@ import re
 import time
 from contextlib import contextmanager
 from pathlib import Path
+from typing import Optional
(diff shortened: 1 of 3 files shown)


src/backend/InvenTree/InvenTree/unit_test.py
```diff
@@ -25,6 +26,33 @@
 from plugin.models import PluginConfig
 
 
+@contextmanager
+def count_queries(
+    msg: Optional[str] = None, log_to_file: bool = False, using: str = 'default'
+):  # pragma: no cover
```
matmair, line 32: Should this be added to the contribution docs?

SchrodingersGat: I've added a small note about this now

Issue #956, opened 2020-09-03
Bulk add test results


jnewlands:
Given a multi-selection of stock items, should be able to bulk add test results. This could be a newly named test result or against any of the test templates applicable to the common ancestor part of the selection.

PR #10146, opened 2025-08-07
Bulk add test results

SchrodingersGat:
Allow test results to be added in "bulk" against multiple stock items.

To achieve this, adds a `BulkCreate` mixin class on the backend - allowing multiple items to be created with a single API call.

### Breaking Change

This PR removes the (old, legacy, undocumented) API function where a test result could be uploaded against a test template that did not already exist, and a new test template could be automatically generated based on the provided test name. This has not been used in the API / UI for nearly two years, and is being removed before stable release

### Related Issues

- Closes https://github.com/inventree/InvenTree/issues/956

### DIFF
--- a/docs/docs/develop/contributing.md
+++ b/docs/docs/develop/contributing.md
@@ -208,6 +208,21 @@ coverage html -i
 
 The coverage database is also generated in the CI-pipeline and exposd for 14 days as a artifact named `coverage`.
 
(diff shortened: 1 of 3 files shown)


matmair:
@SchrodingersGat it seems like the schema did not change with this PR but the version was bumped; is the api schema reflecting the possible inputs correctly?

PR #10171, opened 2025-08-14
 Fix N+1 Query for primary_address in Company API

Reza98Sh:
This PR resolves an N+1 query problem in the primary_address property for Company objects.

Previously, retrieving companies triggered an extra query for each one to load its primary address.

Before: Executing this query resulted in 33 database queries.
After: Reduced to 6 queries by using  in prefetch_related annotate_queryset.

Screenshot from 2025-08-14 10-01-28

---
If the team prefers a different approach, I can update the implementation accordingly.


SchrodingersGat:
@Reza98Sh this looks like a great addition, and nice performance gain - thanks for looking into this.

SchrodingersGat (approved):

matmair:
@SchrodingersGat ist there an easy way to find out if we can lower some of the db request limits in tests?

SchrodingersGat:
Rather than breaking the API (by adding the `primary_address` field) can you please change this to keep the existing `address` naming.

SchrodingersGat:
> @SchrodingersGat ist there an easy way to find out if we can lower some of the db request limits in tests?

Most of these issues only get revealed when we test with a very large dataset - which we would need to to add for each test case

Reza98Sh:
> Rather than breaking the API (by adding the `primary_address` field) can you please change this to keep the existing `address` naming.

You are right, I mixed up the descriptions. primary_address is the full object and address is the string representation.

Reza98Sh:
I think for the list API, we can call it with different “limit” query parameters. If the number of DB queries changes, it means there is an issue.

SchrodingersGat:
@1337joe can you please look at the deltas in the API schema here? It looks like the ordering of the extra attributes needs to be fixed. The differences seem to be only in field ordering.

1337joe:
@SchrodingersGat I haven't found the code that causes this, but it looks like drf-spectacular adds the attributes in a different way for extend schema annotations that results in the order change.

To prevent the schema from changing it looks like you can move the `allow_null=True` back into the original serializer and leave the extended type specification empty. A quick diff of what I mean for `primary_address`:

```diff
-    primary_address = serializers.SerializerMethodField()
+    primary_address = serializers.SerializerMethodField(
+        allow_null=True,
+    )

-    @extend_schema_field(AddressSerializer(allow_null=True))
+    @extend_schema_field(AddressSerializer())
    def get_primary_address(self, obj):
```

Reza98Sh:
Do I need to do something ?

src/backend/InvenTree/company/serializers.py
```diff
@@ -161,17 +163,36 @@ def annotate_queryset(queryset):
...
         return queryset
 
-    address = serializers.CharField(
+    address = serializers.SerializerMethodField(
         label=_(
             'Return the string representation for the primary address. This property exists for backwards compatibility.'
-        ),
-        allow_null=True,
-        read_only=True,
+        )
```
1337joe:
```suggestion
        ),
        allow_null=True,
```

src/backend/InvenTree/company/serializers.py
```diff
@@ -161,17 +163,36 @@ def annotate_queryset(queryset):
...
     )
-
-    primary_address = AddressSerializer(allow_null=True, read_only=True)
+    primary_address = serializers.SerializerMethodField(allow_null=True)
+
+    @extend_schema_field(serializers.CharField(allow_null=True))
```
1337joe:
```suggestion
    @extend_schema_field(serializers.CharField())
```

matmair:

matmair:
@1337joe thank you!

SchrodingersGat:
Nice work,.thanks everyone

PR #10178, opened 2025-08-15
ci: fix wrong invoke environment

daniil4udo:
Fix annoying error during update.

The function should return early when in Docker environment

image


SchrodingersGat:
> Fix annoying error during update.

I'm assuming you are using docker - at what stage of the update process do you see this error?

daniil4udo:
During execution `docker compose run --rm inventree-server invoke update` after image has been pulled

SchrodingersGat:
@daniil4udo thanks for the improvement

PR #10180, opened 2025-08-15
feature(FE): add pnpm package manager

daniil4udo:
As a modern alternative add PNPM, in case you'll want to switch soon

### DIFF
--- a/contrib/container/Dockerfile
+++ b/contrib/container/Dockerfile
@@ -112,9 +112,16 @@ RUN apk add --no-cache --update nodejs npm yarn bash
 RUN npm install -g --ignore-scripts n@10.1.0
 RUN bash -c "n lts"
 RUN yarn config set network-timeout 600000 -g
+RUN npm install --global corepack@latest --force && \
+    corepack enable pnpm && \
+    corepack prepare pnpm@latest-10 --activate
+
 COPY src ${INVENTREE_HOME}/src
 COPY tasks.py ${INVENTREE_HOME}/tasks.py
-RUN cd ${INVENTREE_HOME} && invoke int.frontend-compile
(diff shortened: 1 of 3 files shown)


matmair:
What improvement does this bring compared to yarn?

daniil4udo:
This is the [benchmarks](https://pnpm.io/benchmarks).

Also in the Docker you use Yarn v1 (which was release in 2016), alot has been changed. 
**pnpm** outperforms both npm and Yarn in speed, thanks to its shared package store and linking mechanism.

matmair:
Is there any indication that these performance gains are actually realized in the isolated docker build process? As far as I read the literature provided it would not have any meaningful impact on ci times and add another dependency without removing another

daniil4udo:
You're absolutely right—there’s only a slight performance boost. My main motivation is to stay aligned with modern technologies, especially when the transition is seamless.

I don’t mind sticking with my custom Docker setup if you feel this change doesn’t add enough value.

SchrodingersGat:
Unless this is fixing a bug, or providing a significant improvement, this feels like needless churn. Changing it here will no doubt lead to other changes, and the current frontend packaging is called very infrequently anyway

matmair:
tbh i do not see enough reason to change the frontend build/package pipeline here.

daniil4udo:
np. will use it internally

daniil4udo:
Although, as you can see, I am not replacing Yarn as the package manager — I’m simply enabling (not even installing) pnpm as an additional package manager. It’s up to you which one you want to use.

matmair:
Yes, you are adding another package that will need to be kept updated, as the image shows up as vulnerable if there is an issue with that package. Without adding any real benefit to the general user or dev.

PR #10181, opened 2025-08-15
[UI] Search shortcut

SchrodingersGat:
Enable `/` shortcut to open the global search drawer - as is the "standard" on a lot of websites.


matmair:
Great addition!

PR #10202, opened 2025-08-19
chore(frontend): upgrade react to 19

matmair:


matmair:
@SchrodingersGat so Copilot was able to solve this and seems to make a pretty good job at it too

SchrodingersGat:
Nice, good to get this in before the release :)

Issue #10197, opened 2025-08-19
Report filters: Problem with comma seperated list


SergeoLacruz:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

I tried to use the report filters to filter the report templates of a PO. It works fine as long as I add just one supplier to the filter. But a second one does not work. Following setup:

<img width="777" height="756" alt="Image" src="https://github.com/user-attachments/assets/05ae81f9-edd5-4aed-a575-237f0863806f" />

the template is only shown for a PO for the supplier with the pk 11. Did I understand something wrong? 

### Steps to Reproduce

Set filters for PO template
Select PO
Print report

### Expected behaviour

The template should be visible for both suppliers.

### Deployment Method

Bare metal

### Version Information

<img width="779" height="707" alt="Image" src="https://github.com/user-attachments/assets/6c037da2-d848-4657-9f4c-48dc2c476000" />

### Try to reproduce on the demo site

I did not try to reproduce

### Is the bug reproducible on the demo site?

Not reproducible

### Relevant log output

```shell

```

SchrodingersGat:
@SergeoLacruz the filters cannot be "repeated" like this, the key values must be unique.

matmair:
And please provide the version information as text; there is a dedicated “copy version information” button

SergeoLacruz:
So no OR combination. What a pity...

SchrodingersGat:
@SergeoLacruz please see https://github.com/inventree/InvenTree/pull/10203 - I have added support for list based filtering

SergeoLacruz:
This works quite great :-)
Many thanks for the quick help.

PR #10203, opened 2025-08-19
List filters

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/10197

Adds support for list filtering for label and report templates


matmair (approved):
lgtm

PR #10213, opened 2025-08-23
Disable paging on purchaseOrderReceive API call

1337joe:
#10174 changed the annotations for the schema to return a list using type `PaginatedStockItemList`

Since you can't really page through create results without creating more results on each call, the paging needs to be disabled so a raw array of `StockItem` is returned.


SchrodingersGat (approved):

SchrodingersGat:
@1337joe thanks again, this is a good addition :)

PR #10214, opened 2025-08-23
Refactor (backend):  Improve BuildItemList API filters 

Reza98Sh:
This PR addresses issue #9481 for the BuildItemList endpoint by removing undocumented query parameter and optimizing database queries.

Removed Manual Parameter Parsing: The logic for manually handling `output`  has been removed. This is a step towards replacing them with proper, documented API filters that will appear in the OpenAPI schema.

Improved Query Performance: Added extensive select_related and prefetch_related calls to get_queryset. This reduces database queries.

Related Issue
[#9481](https://github.com/inventree/InvenTree/issues/9788)


src/backend/InvenTree/build/api.py
```diff
@@ -850,6 +850,24 @@ def filter_location(self, queryset, name, location):
         locations = location.get_descendants(include_self=True)
         return queryset.filter(stock_item__location__in=locations)
 
+    output = rest_filters.NumberFilter(
+        label=_('Output'),
+        method='filter_output',
+        help_text=_(
+            "Filter by output stock item ID. Use 'null' to find uninstalled build items."
```
SchrodingersGat, line 861: Have you verified that a "null" value is actually handled by the `NumberFilter` type? I believe that it might actually throw an error if a non-numeric value is passed to it. Please make sure there are at least some unit tests for this.

SchrodingersGat:
@Reza98Sh as the public-facing API has now changed you will need to add an update entry in `api_version.py`

src/backend/InvenTree/build/api.py
Reza98Sh: I double-checked using Postman and confirmed that ‘null’ values are correctly handled in my local setup.

I will add unit tests to explicitly cover this scenario.

SchrodingersGat: Thanks, that will be great for ensuring there are no regressions.

Reza98Sh:
@SchrodingersGat 
you were correct that the previous approach wasn’t properly handling the null value.

I’ve now updated the implementation to use a custom filter field which correctly processes null, and I’ve also added dedicated unit tests to ensure this scenario is fully covered.

SchrodingersGat:
Thanks for looking further into this @Reza98Sh 

You will have to address the issues with the test fixtures. Likely the failing tests do not include the "stock" fixtures

SchrodingersGat:
@Reza98Sh thanks for this contribution. This new filter could be used in other places in the API too, for the same purpose.

PR #10215, opened 2025-08-23
Fix extra query in company/part/ endpoint when manufacturer_detail=true

Reza98Sh:
This patch improves performance by preventing an unnecessary database query triggered when manufacturer_detail=true in the company/part/ endpoint.

The extra query was causing avoidable overhead; now it has been eliminated.

Screenshot from 2025-08-23 20-11-59


SchrodingersGat:
@Reza98Sh thanks for the improvement 😁

Issue #10208, opened 2025-08-20
[docker]musl libc ignored edns0 option causing the microsoft login to fail


tamisoft:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

The docker images are built using musl libc, and that[ ignores in the /etc/resolv.conf](https://git.musl-libc.org/cgit/musl/tree/src/network/resolvconf.c#n41) the `options edns0 no-aaaa`  line. This will cause name resolution error on domain names that return an answer larger than 512 bytes. For example: login.microsoftonline.com
This will cause the oauth login to fail due to name resolution timeout.

This issue could cause weird problems with other components as well that rely on cloud infrastructure with various hosts providing the service (I could think of S3 from AWS for example).

Currently I am using a workaround, where the local `bind/named` server got an option: `minimal-responses yes;` and the answer this way is only 401 bytes, so it is working again. 

One of the potential solutions would be to build the images also with a glibc based image and add it as a -glibc, or -ubuntu tag.

The cross test to find this problem was to execute a fedora:41 container in the same network namespace and that had no problem with the name resolution when the `options` line was in the `resolv.conf`, but would experience the same behavior if it was removed. After this I was packet capturing the `podman1` interface to verify that this is indeed the case.

I think the fix would be to change [this line](https://github.com/inventree/InvenTree/blob/e3a1c5fa20c2395789505deb195a4d1aaf489b7c/contrib/container/Dockerfile#L12C6-L12C16) to `FROM python:3.11-trixie AS inventree_base` and update the installed packages to use the debian equivalents.

### Steps to Reproduce

1. install inventree as docker image
2. start the container
3. podman exec -it inventree-server ping login.microsoftonline.com

### Expected behaviour

the name resolution would resolve answers larger than 512 bytes

### Deployment Method

Docker

### Version Information

# Version Information:
InvenTree-Version: 0.17.11
Django Version: 4.2.17
Commit Hash: e2a092e
Commit Date: 2025-04-22

Database: mysql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.15.4-200.fc42.x86_64-x86_64-with
Installer: DOC

[...]

PR #10227, opened 2025-08-25
Debian docker image

SchrodingersGat:
- Swap from alpine to debian slim-trixie
- Refactor Dockerfile
- Optimize image size
- Reduce @vitejs/plugin-react version
- Closes https://github.com/inventree/InvenTree/issues/10208

### Overview

This PR refactors the docker image to use a debian base rather than alpine.

Advantages:

- Better broad compatibility than alpine which uses musl builds
- Similar setup to the installer package (debian base)
- Similar setup to the devcontainer image
- Fixes some ongoing issues with alpine based image

### Comparison

The size of the new image is less than the previous alpine image after trimming:

image


wolflu05:
Interesting, why is our current alpine based image so big? I thought it was around <750MB where I last touched it and reduced it from >1.2GB (the old Debian based)?

contrib/container/docker-compose.yml
```diff
@@ -42,7 +42,7 @@ services:
     # Database service
     # Use PostgreSQL as the database backend
     inventree-db:
-        image: postgres:16
+        image: postgres:17
```
matmair, line 45: this might be dangerous; have you checked this does work as a drop-in?

contrib/container/dev-docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
     # Use PostgreSQL as the database backend
     # Note: This can be changed to a different backend if required
     inventree-dev-db:
-        image: postgres:13
+        image: postgres:17
```
matmair, line 21: This is a very large jump, I do not think auto-migrations work with that many versions

matmair (approved):
LGTM

contrib/container/dev-docker-compose.yml
SchrodingersGat: This change does not actually force any change on the user.

- If they already have a working docker compose file, they will not need to change (and their db will remain pinned at `postgres:13`) 
- If they are starting a new docker setup, they will get the latest `postgres:17` image

Additionally, they  can always edit their docker compose file as it is intended to be copied outside of the InvenTree source code anyhow

matmair:
@SchrodingersGat great to see; this should prop be called out in the breaking section

Issue #10228, opened 2025-08-25
Login issues after update form 0.16.5 to 0.17.14


melektron:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [x] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

I am not sure if this is a bug or an install issue, so I have selected install issue for now.

I have updated my inventree installation from version 0.16.5 to 0.17.14 using the migration workflow described [in the docs ](https://docs.inventree.org/en/latest/start/migrate/#migrating-data-to-newer-version). To make sure I have the latest static files, I deleted the static files, caddy files and config.yaml file for the new version to regenerate them with the correct latest version from my understanding. I have never modified config.yaml and I have compared the versions and don't see any significant differences relevant to my setup. 

The migration seems to have worked fine, and I can start the new server and log in to CUI successfully with my old accounts. However, when logging into PUI with an old user, it always fails with this message:

<img width="449" height="73" alt="Image" src="https://github.com/user-attachments/assets/a3080b3c-8928-4dde-a9ee-21a0115e3b69" />
On this request:
<img width="577" height="766" alt="Image" src="https://github.com/user-attachments/assets/08394b34-378f-4a63-bbd7-e1746640f0e3" />
The `/api/auth/login` request before seems to work...

When I then manually navigate to some PUI Path, I am logged in and can access everything, it's just the login that's broken.

To debug this I created another superuser using `invoke superuser`. Surprisingly, I was able to successfully log in with the newly created user. I then also created one in PUI, and I am able to successfully log in with the new users, just not with the existing ones form 0.16.5.

The most closely matching existing issue I found was #10048, but it doesn't talk about the same problem.
#9030 also points to #8269, but the there suggested solution was already active, since I used the new config.yaml file. (I also tried it with the old config file at one point with the same results, although I don't know if new users would have worked there).

### Steps to Reproduce

I am not sure if this is reproducibe or just a bug stemming from my old data, but I would assume the following steps should mimic what I have done: 

1. Setup an inventree 0.16.5 instance with docker compose and create some users.

Exact version info:
```
# Version Information:
InvenTree-Version: 0.16.5
Django Version: 4.2.15
Commit Hash: 6e37f0c
Commit Date: 2024-10-07

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
[...]

PR #10233, opened 2025-08-26
[bug] Authentication endpoint may return expired token

melektron:
This PR fixes issue #10228.

PUI login via the `/api/auth/login` endpoint is handeled by a (modified) dj_rest_auth `LoginView`, which uses the TOKEN_CREATOR to get the token to return. The configured token creator [`users.models.create_token_expiry()`](https://github.com/inventree/InvenTree/blob/bca375dae5ac1d49bb5388393360c461639dbbb8/src/backend/InvenTree/users/models.py#L59) doesn't account for expired tokens and returns the first matching token for the user, even if it may be expired.

In contrast, the [`users.api.GetAuthToken()`](https://github.com/inventree/InvenTree/blob/bca375dae5ac1d49bb5388393360c461639dbbb8/src/backend/InvenTree/users/api.py#L312) view handles this correctly and creates a new token for authenticated users if no vaild token exists.

This is fixed by consolidating the logic for retrieving or creating an `ApiToken` into a single classmethod.

### DIFF
--- a/src/backend/InvenTree/users/api.py
+++ b/src/backend/InvenTree/users/api.py
@@ -1,6 +1,5 @@
 """DRF API definition for the 'users' app."""
 
-import datetime
 import logging
 
 from django.contrib.auth import authenticate, get_user, login, logout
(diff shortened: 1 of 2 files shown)


melektron:
I targeted this PR at stable as per the contribution guidelines for bugfixes, but this might apply to master as well.

Also, the `GetAuthToken` view I modified clearly states [in the comments](https://github.com/melektron/InvenTree/blob/2d43539e4f4eef9c3be4254fc65a143a980c1da1/src/backend/InvenTree/users/api.py#L321) that it shouldn't return existing tokens, yet the previous implementation did (which I think makes sense). I kept the existing functionality in my change, should I modify/remove this comment?

matmair:
Thank you for the submition @melektron ! This will not need a merge into current dev, we have removed dj_rest_auth

PR #10260, opened 2025-09-03
fix a few docker security holes

matmair:
fixes a few security scanner results from https://github.com/inventree/InvenTree/pull/10227


SchrodingersGat:
Nice improvements :)

Issue #9382, opened 2025-03-25
Cannot import BOM -- Invalid Type


galmiklos:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

Trying to import a BOM to an assembly part from a csv file. I have all the parts added to the inventory, with the manufacturer part number as "Part". All rows show "Incorrect type. Expected pk value, received str."

### Steps to Reproduce

1. Create a part, assembly type.
2. Click the "Import BOM Data" icon.
3. Select csv file (attached for reference) and Submit.
[SmartPrintCoreH7x_BOM.csv](https://github.com/user-attachments/files/19458327/SmartPrintCoreH7x_BOM.csv)
5. Map the "Manufacturer_Part_Number" column to Component, and the "Qty" column to Quantity. 
![Image](https://github.com/user-attachments/assets/cbeb72e3-fd4d-41f3-b364-d9cf914d58e0)

5. Click Accept Column Mapping

All entries in the Component column are marked incorrect.

![Image](https://github.com/user-attachments/assets/535bb0b8-788e-4ff3-b02c-e6845b994351)

### Expected behaviour

I expect the BOM import to work with this csv file.

### Deployment Method

- [ ] Docker
- [ ] Package
- [x] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

InvenTree-Version: 0.18.0 dev
Django Version: 4.2.19
Commit Hash: 06961c6
Commit Date: 2025-03-07
Commit Branch: master
Database: mysql
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-6.8.0-52-generic-x86_64-with-glibc2.35
Installer: GIT
[...]

PR #10268, opened 2025-09-04
BOM import: add field selection for item reference (ID/IPN/Name)

### LINKED ISSUE #7574: [Importer] Natural Key Support
When importing a dataset with "foreign key" relations, these may be specified with a natural key (e.g. "name" / "sku" / etc) - rather than the internal InvenTree ID - especially when importing from a different dataset.

So, add some options to lookup using other fields in the related model, rather than just "pk" value. "pk" is used as a priority, when a valid pk is provided.

For example when importing rows for a purchase order, supplier parts could be looked up by "sku" - this will be present when importing a purchase order dataset from an external supplier.

### LINKED ISSUE #9248: [FR] Add IPN as matching field for BoM import

### Problem statement

I have used IPN as part identifier for BoM imports from KiCad. I just realized that the PUI BoM importer do not support that. In the latest [documentation](https://docs.inventree.org/en/latest/build/bom_import/#select-fields) it is stated that IPN is the prefered way. _Specifying the Part_IPN field matching is very powerful as it allows to create direct pointers to InvenTree parts_ 

Image

### Suggested solution

Would be nice to get IPN matching possibility also in PUI.

Easily solved with a bit of processing in Excel between the KiCad BoM and an Inventree part export, but  it would be better to avoid this step.

### Examples of other systems

Inventree 17.x / CUI

- [ ] I want to develop this.

TehDmitry:
#9248, #9382 and partly #7574
inspired by #2349

vivaldi_Yn0jh0W0yL

edit matmair:
Closes #9248
Closes #7574

### DIFF
--- a/src/backend/InvenTree/importer/models.py
+++ b/src/backend/InvenTree/importer/models.py
@@ -4,6 +4,7 @@ import json
 from collections import OrderedDict
 from typing import Optional
 
+from django.apps import apps
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError as DjangoValidationError
 from django.core.validators import FileExtensionValidator
@@ -312,6 +313,24 @@ class DataImportSession(models.Model):
                 available_fields=available_fields,
                 commit=False,
             )
+
+            for key in row.data:
+                if f'_reference_{key}' in self.field_overrides:
+                    reference = self.field_overrides.get(f'_reference_{key}')
+                    Part = apps.get_model('part', 'Part')
+                    parts = []
+
+                    if reference == 'ipn':
+                        part_ipn = row.data.get(key)
+                        parts = Part.objects.filter(IPN=part_ipn)
+                    elif reference == 'name':
+                        part_name = row.data.get(key)
+                        parts = Part.objects.filter(name=part_name)
+
+                    if parts.count() == 1:
+                        part = parts.first()
+                        # Update the 'part' value in the row
+                        row.data[key] = part.pk if part is not None else None
 
             row.valid = row.validate(commit=False)
             imported_rows.append(row)
--- a/src/backend/InvenTree/importer/models.py
+++ b/src/backend/InvenTree/importer/models.py
@@ -315,22 +315,23 @@ class DataImportSession(models.Model):
             )
 
             for key in row.data:
-                if f'_reference_{key}' in self.field_overrides:
-                    reference = self.field_overrides.get(f'_reference_{key}')
-                    Part = apps.get_model('part', 'Part')
-                    parts = []
+                if f'_import_as_{key}' in self.field_overrides:
+                    row_value = row.data.get(key)
+                    import_as = self.field_overrides.get(f'_import_as_{key}')
+                    results_with_pk = []
 
-                    if reference == 'ipn':
-                        part_ipn = row.data.get(key)
-                        parts = Part.objects.filter(IPN=part_ipn)
-                    elif reference == 'name':
-                        part_name = row.data.get(key)
-                        parts = Part.objects.filter(name=part_name)
+                    if import_as == 'Part.IPN':
+                        Part = apps.get_model('part', 'Part')
+                        results_with_pk = Part.objects.filter(IPN=row_value)
 
-                    if parts.count() == 1:
-                        part = parts.first()
-                        # Update the 'part' value in the row
-                        row.data[key] = part.pk if part is not None else None
+                    elif import_as == 'Part.name':
+                        Part = apps.get_model('part', 'Part')
+                        results_with_pk = Part.objects.filter(name=row_value)
+
+                    if results_with_pk.count() == 1:
+                        results_with_pk = results_with_pk.first()
+                        # Update the value in the row
+                        row.data[key] = results_with_pk.pk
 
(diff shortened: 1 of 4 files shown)


SchrodingersGat:
@TehDmitry this is an interesting idea, but your approach is very much hard-wired for "part" import. The importer works for many different models, and so the approach should be generic too.

I would suggest that to make this generic, we add a list of possible "Import ID" fields to the underlying import model class. So, for the "Part" class, you might add:

```python
class Part(...):
    
    def import_id_fields(self):
        """Return a list of potential ID fields for import."""
        return ['pk', 'IPN', 'name']
       
```

If this method is not specified, then the importer just uses the 'pk' field (as a default).

The import process then matches the provided field values, to the "names" of the fields (which are translated already, so we don't need to replicate that).

Then, the import system exposes the available import field types for any "foreign key relation" to the frontend.

I like the frontend implementation, that's a clean way of handling this.

Overall, this is a really nice enhancement - we just need to make it generic!

src/frontend/src/components/importer/ImporterColumnSelector.tsx
```diff
@@ -126,6 +127,60 @@ function ImporterDefaultField({
...
+          // TODO: Error message?
+        });
+    },
+    [column]
+  );
+
+  return (
+    <Select
+      data={[
+        { value: 'id', label: 'Import by ID' },
```
SchrodingersGat: This should be generic - the available field / name combinations can be supplied via the API

src/frontend/src/components/importer/ImporterColumnSelector.tsx
```diff
@@ -126,6 +127,60 @@ function ImporterDefaultField({
...
+        { value: 'ipn', label: 'Import by IPN' },
+        { value: 'name', label: 'Import by Name' }
+      ]}
+      value={partReference}
+      onChange={onChange}
+      placeholder='Select reference'
```
SchrodingersGat: Any user-facing text must be translated:

```tsx
placeholder={t`Select reference`}
```

src/frontend/src/components/importer/ImporterColumnSelector.tsx
```diff
@@ -150,7 +205,12 @@ function ImporterColumnTableRow({
         </Group>
       </Table.Td>
       <Table.Td>
-        <Text size='sm'>{column.description}</Text>
+        <Flex justify='space-between' align='center'>
+          <Text size='sm'>{column.description}</Text>
+          {column.field === 'sub_part' && (
```
SchrodingersGat: This is not generic. The conditions for displaying the `<ReferenceSelect>` component should be:

1. The field is a "related" field
2. The field exposes multiple options for the lookup reference field (default = 'pk', singular)

src/backend/InvenTree/importer/models.py
```diff
@@ -313,6 +314,24 @@ def import_data(self) -> None:
                 commit=False,
             )
 
+            for key in row.data:
+                if f'_reference_{key}' in self.field_overrides:
+                    reference = self.field_overrides.get(f'_reference_{key}')
+                    Part = apps.get_model('part', 'Part')
+                    parts = []
+
+                    if reference == 'ipn':
```
SchrodingersGat: See notes in PR review.

SchrodingersGat (changes requested):

src/backend/InvenTree/importer/models.py
```diff
@@ -313,6 +314,25 @@ def import_data(self) -> None:
                 commit=False,
             )
 
+            for key in row.data:
```
SchrodingersGat, line 317: Looks like you still have some hard-coded fields here - the "import_as" fields should not have to be called out by name.

src/backend/InvenTree/part/models.py
```diff
@@ -2674,6 +2674,11 @@ def is_part_low_on_stock(self):
         """Returns True if the total stock for this part is less than the minimum stock level."""
         return self.get_stock_count() < self.minimum_stock
 
+    @staticmethod
+    def import_as_fields():
+        """Return a list of potential ID fields for import."""
+        return ['id', 'IPN', 'name']
```
SchrodingersGat, line 2680: Nice, this is a simple way to specify the import-as options

src/frontend/src/components/importer/ImporterColumnSelector.tsx
```diff
@@ -126,14 +126,79 @@ function ImporterDefaultField({
   );
 }
 
+function ImporterPartImportAsSelect({
```
SchrodingersGat, line 129: This should be `ImporterImportAsSelect` - it is not specific to importing parts, it should be generic across any type of import

SchrodingersGat (changes requested):

SchrodingersGat:
@TehDmitry have you had any chance to review this further?

Badokas:
Any updates on this? We need this functionality, maybe I can help?

SchrodingersGat:
@TehDmitry still around and planning on pushing this ahead?

SchrodingersGat:
Closing this, https://github.com/inventree/InvenTree/pull/11288 provides a much cleaner approach

PR #10279, opened 2025-09-06
Refactor (backend): Improve BuildItemList API filters

Reza98Sh:
This PR addresses issue https://github.com/inventree/InvenTree/issues/9481 for the StockList endpoint by removing undocumented query parameters.

- Refactors 'exclude_tree', 'cascade', and 'location' filters in StockList API endpoint
- Update related unit tests and fixtures 
- Update api_version.py

### DIFF
--- a/src/backend/InvenTree/InvenTree/api_version.py
+++ b/src/backend/InvenTree/InvenTree/api_version.py
@@ -9,7 +9,6 @@ INVENTREE_API_TEXT = """
 
 v391 -> 2025-09-06 : https://github.com/inventree/InvenTree/pull/10279
     - Refactors 'exclude_tree', 'cascade', and 'location' filters in StockList API endpoint
-    - Removes undocumented 'filter_queryset' method in StockList view
 
 v390 -> 2025-09-03 : https://github.com/inventree/InvenTree/pull/10257
     - Fixes limitation on adding virtual parts to a SalesOrder
--- a/src/backend/InvenTree/InvenTree/api_version.py
+++ b/src/backend/InvenTree/InvenTree/api_version.py
@@ -7,7 +7,7 @@ INVENTREE_API_VERSION = 391
 
 INVENTREE_API_TEXT = """
 
-v391 -> 2025-09-06 : https://github.com/inventree/InvenTree/pull/10259
+v391 -> 2025-09-06 : https://github.com/inventree/InvenTree/pull/10279
     - Refactors 'exclude_tree', 'cascade', and 'location' filters in StockList API endpoint
     - Removes undocumented 'filter_queryset' method in StockList view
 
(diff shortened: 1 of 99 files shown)


src/backend/InvenTree/InvenTree/api_version.py
```diff
@@ -1,11 +1,16 @@
...
-INVENTREE_API_VERSION = 390
+INVENTREE_API_VERSION = 391
 
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
+
+v391 -> 2025-09-06 : https://github.com/inventree/InvenTree/pull/10279
+    - Refactors 'exclude_tree', 'cascade', and 'location' filters in StockList API endpoint
+    - Removes undocumented 'filter_queryset' method in StockList view
```
matmair: I think this lines is not interesting for API consumers, we should consider if we remove this

matmair (approved):

src/backend/InvenTree/InvenTree/api_version.py
Reza98Sh: Just to confirm — you mean we don’t need to update the api_version for this change, right?

matmair: Thank you for asking; only the last line. My statement was not formulated clearly enough

SchrodingersGat (approved):

src/backend/InvenTree/stock/fixtures/stock.yaml
matmair, line 1: Why is this changed? It seems to cause problems in the mysql CI checks

src/backend/InvenTree/InvenTree/api_version.py
matmair:
```suggestion
```

src/backend/InvenTree/stock/fixtures/stock.yaml
Reza98Sh: For the `exclude_tree` test, I had to update it to use it in the test.
 In my local setup, the test worked.

matmair: did you test with mysql? The CI seems to fail

Reza98Sh: No I only tested with postgre. I’ll test it with mysql and update the code if I find any issues.

Reza98Sh: I have fixed it.

This happened because StockItem has a self-referential ForeignKey (parent_id).
When using MySQL, deleting all rows with .delete() fails if child rows still reference a parent row, as MySQL enforces ON DELETE restrictions differently than PostgreSQL

matmair (approved):
LGTM

SchrodingersGat:
@Reza98Sh thanks for the contribution :)

PR #10282, opened 2025-09-08
fix(backend): re-add active plugins to anon status

matmair:
This was disabled due to wrong feedback by me, common debug tools and desktop client discovery does not work because of this.


matmair:
I added a small fix to make the mfa test more robust

PR #10287, opened 2025-09-08
[UI] Fix Quantity calculation in OrderPartsWizard

TehDmitry:
Previously, the Quantity field was calculated as 0 when opening the Order Parts Wizard from certain stock views, e.g.:
- Stock → Stock Details → Order
- Stock → Stock Items → Order
- etc.

The Part object used in the wizard did not contain the required fields for quantity calculation.
Now the wizard fetches the required data directly from the API instead of relying on incomplete Part object data.

vivaldi_XSrhh3ffvc

### DIFF
--- a/src/backend/InvenTree/part/serializers.py
+++ b/src/backend/InvenTree/part/serializers.py
@@ -1263,6 +1263,7 @@ class PartRequirementsSerializer(InvenTree.serializers.InvenTreeModelSerializer)
             'allocated_to_build_orders',
             'required_for_sales_orders',
             'allocated_to_sales_orders',
+            'minimum_stock',
         ]
 
     total_stock = serializers.FloatField(read_only=True, label=_('Total Stock'))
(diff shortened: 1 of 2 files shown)


TehDmitry:
Draft PR for review

After creating some Build Orders and trying to create Purchase Orders, I realized that my previous idea was incorrect.
Because there are many possible usage scenarios, it’s not possible to calculate a single “right” value.

I think the Wizard should display part stock data, requirements, and data from the Build Order (default quantity), and give the user the ability to choose.
I’m not completely sure which fields should be shown, though.

vivaldi_KdtMpJ4Mzh

SchrodingersGat:
### Popover

image

Instead of a Popover please look at using a [HoverCard](https://mantine.dev/core/hover-card/) as we have done for other hover-over UI elements in the interface.

Additionally, please clean up the text presented in this hovercard:

- There is a mix of capitalization
- Some entries use an underscore
- Not all items are translated

PR #10304, opened 2025-09-11
Release version checker

SchrodingersGat:
- Refactor / update script for checking InvenTree version
- Preparing for 1.0.0 rc release


SchrodingersGat:
@matmair please review this, the workflow for extracting API version for checking has been changed slightly

SchrodingersGat:
Some tests might be a good idea here - not sure how we would go about testing it though

PR #10306, opened 2025-09-11
feat(frontend): Add samples for dashboard

### LINKED ISSUE #9990: Add default widgets when none are defined
### Discussed in https://github.com/inventree/InvenTree/discussions/9981

<sup>Originally posted by **matmair** July  8, 2025</sup>
current a new user is presented with a pretty empty screen - 
grafik
this is not very friendly and also not showing the possibilities. I would like to add a set of default widgets consisting of the following widgets
grafik

matmair:
Closes #9990

### DIFF
--- a/src/frontend/src/components/dashboard/DashboardLayout.tsx
+++ b/src/frontend/src/components/dashboard/DashboardLayout.tsx
@@ -1,5 +1,13 @@
 import { t } from '@lingui/core/macro';
-import { Alert, Card, Center, Divider, Loader, Text } from '@mantine/core';
+import {
+  Alert,
+  Card,
+  Center,
+  Divider,
+  Loader,
+  Space,
+  Text
+} from '@mantine/core';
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@matmair nice work, my only thought is that the "clear widgets" button now does not do what it says. Is there a way that we can make this "default layout" only appear a first time for the user? If they manually clear it, it stays cleared?

matmair:
We could persist that information in the session storage (LocalStateProps) similar to navigation and hostlist info - would this be sufficient? Another option would be to widen the use of the user profile for that.

matmair:
I have added the code for a session-based state

matmair:
@SchrodingersGat is this how you envisioned it?

SchrodingersGat (approved):

SchrodingersGat:
@matmair thanks, I think that's pretty reasonable

PR #10316, opened 2025-09-13
Add duplicatePart to parts list

### LINKED ISSUE #10276: [FR] Add more actions to PartTable

### Problem statement

Using Version Docker Latest with the new PUI.

In the part table listing, add more row actions. Two I can think of "Delete Item" and "Duplicate Item".

### Suggested solution

Reduce clicks. Improves UX. Especially when just starting out and there is lots of information to input to the system.

M0r clix

### Examples of other systems

_No response_

- [ ] I want to develop this.

wolfpackmars2:
Resolves #10276


src/frontend/src/tables/part/PartTable.tsx
```diff
@@ -349,6 +357,19 @@ export function PartListTable({
 
   const [selectedPart, setSelectedPart] = useState<number>(-1);
 
+  const {
```
SchrodingersGat: I think you do not need to separately fetch the part instance - it's already available in the "record" instance when the action is triggered.

src/frontend/src/tables/part/PartTable.tsx
```diff
@@ -380,10 +449,17 @@ export function PartListTable({
             setSelectedPart(record.pk);
             editPart.open();
           }
+        }),
+        RowDuplicateAction({
+          hidden: !can_add,
+          onClick: () => {
+            setSelectedPart(record.pk);
```
SchrodingersGat: You can store the provided `record` instance here, rather than fetching it again

SchrodingersGat (changes requested):

SchrodingersGat:
@wolfpackmars2 some changes needed here (please see my comments)

wolfpackmars2:
> @wolfpackmars2 some changes needed here (please see my comments)

My apologies, I'm struggling to follow the changes requested. This is my first foray into tsx template files. I am happy to learn. And I especially approve of reducing unnecessary function calls and writing code that is as efficient as possible.

My understanding is that "const duplicatePart..." sets up the duplicate part popup form, using data from the "const part..." object.

The Row Edit form uses the primary key ID to pull the part details into the popup form. Conversely, the "part" object loads fetches details about the part that aren't already available to the table, for the purpose of copying those details to a new record. The Duplicate Part form doesn't have a provision for pulling details with just a primary key.

I agree that the edit form and duplicate form should be able to use the same object as the source of its data, but I'm not sure what that looks like.

src/frontend/src/tables/part/PartTable.tsx
```diff
@@ -349,6 +357,19 @@ export function PartListTable({
 
   const [selectedPart, setSelectedPart] = useState<number>(-1);
```
SchrodingersGat: @wolfpackmars2 I would suggest that you can refactor the line above to:

```typescript
const [selectedPart, setSelectedPart] = useState<any>({});
```

And then we can store the entire part instance (not just the primary key) locally.

Refer to my comments below for further information on how this would be handled.

src/frontend/src/tables/part/PartTable.tsx
SchrodingersGat: This whole `useInstance` query can be removed.

src/frontend/src/tables/part/PartTable.tsx
```diff
@@ -357,6 +378,53 @@ export function PartListTable({
...
+      }
+    };
+  }, [createPartFields, globalSettings, part]);
+
+  const duplicatePart = useCreateApiFormModal({
+    url: ApiEndpoints.part_list,
+    title: t`Add Part`,
+    fields: duplicatePartFields,
+    initialData: {
+      ...part,
```
SchrodingersGat:
```typescript
initialData: {
   ...selectedPart,
   active: true,
   locked: false
},
```

src/frontend/src/tables/part/PartTable.tsx
```diff
@@ -357,6 +378,53 @@ export function PartListTable({
...
 
+  const createPartFields = usePartFields({ create: true });
+
+  const duplicatePartFields: ApiFormFieldSet = useMemo(() => {
+    return {
+      ...createPartFields,
+      duplicate: {
+        children: {
+          part: {
+            value: part.pk,
```
SchrodingersGat: Replace `part.pk` with `selectedPart.pk` here.

src/frontend/src/tables/part/PartTable.tsx
```diff
@@ -380,10 +449,17 @@ export function PartListTable({
             setSelectedPart(record.pk);
```
SchrodingersGat: Here, instead of storing just the `pk` value, store the entire record:

```typescript
setSelectedPart(record);
```

src/frontend/src/tables/part/PartTable.tsx
SchrodingersGat:
```typescript
setSelectedPart(record);
```

src/frontend/src/tables/part/PartTable.tsx
```diff
@@ -349,6 +357,19 @@ export function PartListTable({
...
+    pk: selectedPart,
+    params: {
+      path_detail: true
+    },
+    refetchOnMount: true
+  });
+
   const editPart = useEditApiFormModal({
     url: ApiEndpoints.part_list,
     pk: selectedPart,
```
SchrodingersGat: Here, as we have changed `selectedPart` to be the entire record now, you will need to change this line to:

```typescript
pk: selectedPart.pk,
```

src/frontend/src/tables/part/PartTable.tsx
```diff
@@ -357,6 +378,53 @@ export function PartListTable({
...
+        children: {
+          part: {
+            value: part.pk,
+            hidden: true
+          },
+          copy_image: {
+            value: true
+          },
+          copy_bom: {
+            value: part.assembly && globalSettings.isSet('PART_COPY_BOM'),
```
SchrodingersGat: Same here, `part` can become `selectedPart`

wolfpackmars2:
I think that should do it. Ran into a few snags throughout the process. Probably pebcak or due to my codespace environment being a few weeks old.

Had to manually update the SITE_URL variable from "http://localhost:8000" to "https://localhost:8000" in order to login to the test server. To this point, not sure where, in the codespace, I would configure this for the debug server.

Ran into issue with pre-commit hook where the process was using the wrong version of go - needed 1.23.8 but used 1.23.6. The problem was related to gitleaks

wolfpackmars2:
I'm going to create a new clean PR with just the changes to PartTable.tsx. It was not my intention to modify .pre-commit-config.yaml. I have discovered that running _invoke dev.setup-dev -s_ modifies the yaml file when using codespaces.

SchrodingersGat:
@wolfpackmars2 there is no need to commit changes to the yaml file in that case!

PR #10335, opened 2025-09-16
fix(backend): better siteurl testing in middleware

matmair:
heavy wip, needs security review


matmair:
I think I have found the source for the issues with multi-homed setups; this is not ready for prod - DO NOT COPY THIS CODE INTO AN INTERNET EXPOSED INSTANCE

SchrodingersGat (approved):

matmair:
I have run various tests and coordinated with more knowledgeable security engineers, the abroach should be fine. Need to figure out what is different between CI and my local machine before merge is possible

SchrodingersGat:
@matmair LMK when you think this is ready for review. I'm keen to get a bugfix patch out

matmair:
I think this should be fine once CI runs through

PR #10374, opened 2025-09-22
Refactor query parameters handling for SupplierPart and StockList endpoints

Reza98Sh:
This PR addresses [issue #9481](https://github.com/inventree/InvenTree/issues/9481) for the SupplierPartList and SupplierPartDetail API endpoints by removing undocumented query parameters.

Changes

- Refactored parameters: part_detail, supplier_detail, manufacturer_detail, and pretty in the SupplierPartList API endpoint.
- Updated unit tests to reflect the changes.
- Updated api_version.py.
- Added a custom OutputOptionsMixin and schema_for_view_output_options decorator that automatically:
        -     Reads available output_options from the API view.
        -     Generates the corresponding OpenAPI query parameters from these options.
        -     Ensures consistent parameter handling and serializer configuration for both list and retrieve views.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -129,7 +129,7 @@ jobs:
 
     steps:
       - name: Check out repo
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # pin@v5.0.0
         with:
           persist-credentials: false
       - name: Run Migration Tests
@@ -153,11 +153,11 @@ jobs:
 
     steps:
       - name: Check out repo
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # pin@v5.0.0
         with:
           persist-credentials: false
       - name: Set Up Python ${{ env.python_version }}
-        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # pin@v5.6.0
+        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # pin@v6.0.0
         with:
           python-version: ${{ env.python_version }}
       - name: Version Check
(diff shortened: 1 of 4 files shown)


src/backend/InvenTree/InvenTree/fields.py
```diff
@@ -212,3 +212,28 @@ def __init__(self, **kwargs):
...
+
+    def __init__(self, description: str, flag: str, default=None):
+        """Initialize the output option."""
+        self.description = description
+        self.flag = flag
+        self.default = default
+
+
+class OutputConfiguration:
+    """Contains all available output options for the SupplierPart endpoints."""
```
matmair: This should be more generic

src/backend/InvenTree/InvenTree/apps.py
```diff
@@ -82,6 +82,11 @@ def ready(self):
         # used for account updates whereas the receiver is used for the initial account creation.
         from InvenTree import sso
 
+        try:
+            from . import schema  # noqa: F401
```
matmair: Can you explain why this is needed?

Reza98Sh: As per DRF Spectacular docs (Step 5: Extensions), importing schema here ensures custom extensions are registered. 

https://drf-spectacular.readthedocs.io/en/latest/customization.html#step-5-extensions

matmair: ah that makes it more clear; can you add this link so that the next one reading the line is not as confused as me?

src/backend/InvenTree/company/api.py
```diff
@@ -345,17 +375,9 @@ def get_queryset(self, *args, **kwargs):
...
-            )
-            kwargs['pretty'] = str2bool(params.get('pretty', None))
-        except AttributeError:
-            pass
+        params = self.request.query_params
 
+        kwargs.update(self.output_options.format_params(params))
         kwargs['context'] = self.get_serializer_context()
 
         return super().get_serializer(*args, **kwargs)
```
matmair: Could we move this section to generalized mixin that we can add to views? Not sure if we need the context part, but the rest seems like it would make sense in a mixin, maybe even as part of the general views.

matmair:
I really like this addition, I think this could be the final piece to address https://github.com/inventree/InvenTree/issues/9481 broadly

Reza98Sh:
I have discovered a more concise approach that eliminates the need for the DRF Spectacular extension, relying solely on `extend_schema_view`.

matmair:
Looks good; can you update the PR description?

Reza98Sh:
Updated the PR description.

Thanks for your review.
I’ll be submitting the remaining PRs for [#9481](https://github.com/inventree/InvenTree/issues/9481) very soon!

SchrodingersGat:
@Reza98Sh thanks for these changes!

1337joe:
For what it's worth, as a consumer of the API I like the schema generated by this change. I appreciate that the options are now exposed via schema and the comments are exactly what I need to see (including the default value).

PR #10381, opened 2025-09-23
Machine properties and periodic ping

wolflu05:
This PR adds machine properties to the machine details page. This is useful to display e.g. model name, firmware version, printed pages, ... 

image

If you want to try this out yourself, either activate the sample plugin and create a machine using its driver or if you own a zebra printer, you can try out this plugin: https://github.com/wolflu05/inventree-zebra

## TODO

- [x] fix machine status label not showing
- [x] fix redis cache settings on macos
- [x] machine properties
- [x] periodic ping task
- [x] docs
- [x] tests


wolflu05:
This PR is ready for review

src/backend/InvenTree/machine/machine_type.py
```diff
@@ -48,6 +48,24 @@ class MachineStatus(StatusCode):
...
+        value: Value of the property (required)
+        group: Grouping of the property
+        type: Type of the property (required, one of 'str', 'bool', 'progress')
+        max_progress: Maximum value for progress type (required if type is 'progress')
+    """
+
+    key: str
+    value: Union[str, bool, int]
+    group: str
+    type: Literal['str', 'bool', 'progress']
```
matmair: Would having support for int make sense here?

wolflu05: Sure we could add that, but do you have any special formatting other than str on the frontend side in mind?

wolflu05:
Thanks for you quick review.

> Maybe you could add a line to the CHANGELOG mentioning the new feature - that think it is worth being called out

Is that that `CHANGELOG.md` file in the root of the repo? How would such a entry look like, maybe you could help me here? Currently there is nothing in there. And why does it say `[Unreleased - 1.0.0 ] - 2025-07-xx` this doesnt seem like its being actively used?

SchrodingersGat (approved):

SchrodingersGat:
@wolflu05 thanks for implementing this.

@matmair any more changes required from your review?

matmair:
REviewing now; looks good overall

src/backend/InvenTree/machine/machine_type.py
matmair: I would love to see this extended to include `int` and `float` and use `string` and `formatDecimal` (with 4 Points of precision as that is the machinist default) for display

bigger picture idea: use the status as stable interface for assertions about a machine in other plugins

matmair:
> Thanks for you quick review.
> 
> > Maybe you could add a line to the CHANGELOG mentioning the new feature - that think it is worth being called out
> 
> Is that that `CHANGELOG.md` file in the root of the repo? How would such a entry look like, maybe you could help me here? Currently there is nothing in there. And why does it say `[Unreleased - 1.0.0 ] - 2025-07-xx` this doesnt seem like its being actively used?

Under the Added section something like:

```
- Added much more detailed status information for machines to the API endpoint (including backend and frontend changes) in [#10381](https://github.com/inventree/InvenTree/pull/10381)
```

src/backend/InvenTree/machine/serializers.py
```diff
@@ -30,9 +45,10 @@ class Meta:
             'machine_errors',
             'is_driver_available',
             'restart_required',
+            'properties',
         ]
 
-        read_only_fields = ['machine_type', 'driver']
+        read_only_fields = ['machine_type', 'driver', 'properties']
```
matmair, line 52: Is this correctly applied in all consuming APIViews? The rendered API diff seems to indicate that `properties` field is required in various manipulation endpoints
https://github.com/inventree/InvenTree/actions/runs/18153725568/attempts/1#summary-51669080779
As far as I understand it, the properties are not to be manipulated via API.

Am I misunderstanding the behavior of the field, or is there an issue with the APIs / schema generation?

wolflu05: I think youre anderstanding it correctly, properties is always readonly, it is ment to be set by InvenTree plugins on the server side, not by API consumers. Not sure why it shows up there, maybe the API schema generation is broken? All sub-serializers should inherit the read_only_fields.

src/backend/InvenTree/machine/machine_type.py
wolflu05: Sure, I'll add that tomorrow even though I dont really have the need for it yet.

wolflu05: I have added them now.

src/backend/InvenTree/machine/serializers.py
matmair: i think we can address that in a follow up if it becomes an issue

wolflu05:
Is there anything needed from me before merging this PR? The coverage report seems to be broken...

matmair:
@wolflu05 the CI coverage issue is addressed now

SchrodingersGat:
@wolflu05 thanks for the improvements here!

Issue #10442, opened 2025-10-01
[PUI] Line Item Total Price does not update after editing Purchase Price


J0hnny007:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

When editing the _Purchase Price_ of a line item, the _Total Price_ does not update accordingly. When editing the price again, the calculated _Total Price_ of the old new _Purchase Price_ is shown instead of the value it should be after the second edit.

Correct values are shown after a page reload.

After first edit:
<img width="256" height="164" alt="Image" src="https://github.com/user-attachments/assets/2bb97633-6850-43c6-af89-b54c067704ca" />

After second edit:
<img width="228" height="180" alt="Image" src="https://github.com/user-attachments/assets/da7dffd3-9706-4653-9853-a7cfdd62609a" />

### Steps to Reproduce

1. Go to a _Purchase Order_
2. Have a _Line Item_ with a _Unit Price_ and with a quantity > 1 (I used the Paint as an example on the demo site)
4. Edit the _Unit Price_
5. Observe _Total Price_ not changing

### Expected behaviour

After editing a _Line Items_ price, the _Total Price_ in the chart should update to the correct calculated price without a page reload.

### Deployment Method

Package

### Version Information

InvenTree-Version: 1.0.2
Django Version: 4.2.24
Commit Hash: 1d2700fa
Commit Date: 2025-09-29T04:01:43Z
Commit Branch: stable
Database: django.db.backends.sqlite3
Debug-Mode: False
Deployed using Docker: False
Platform: Linux-6.1.0-37-amd64-x86_64-with-glibc2.36
Installer: PKG
Target: debian:11
[...]

PR #10450, opened 2025-10-02
[UI] Fix total price for Purchase Order table

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/10442


matmair:
seems like a memo bug that could effect more surfaces that use that column

SchrodingersGat:
@matmair I believe the issue is that when you edit a single row in a table, it does not re-fetch the entire table - and some of the annotated data (such as "total price") is not properly re-fetched. So in this case it is easier to calculate dynamically, and this keeps it in line with how the other order tables handle this

PR #10454, opened 2025-10-02
feat(backend): releax protocol check

matmair:
Recent issues show that correct (secure) proxy setup is proving difficult for users, especially getting protocols right. The new INVENTREE_SITE_LAX_PROTOCOL option that defaults to true addresses this and should ease the burden of linking to the docs for the thrown error code


SchrodingersGat:
@matmair thanks for the quick attention

matmair:
@SchrodingersGat this should pass in a few minutes and be ready

gid204:
@SchrodingersGat I can see this PR has been added to milestone 1.1.0

Is there any indication when this will be added to an official release?
I spent about 3 hours trying to get my access working with no luck...
I'm exposing my server through Pangolin.

SchrodingersGat:
@gid204 we'll have a new point release out in the next couple of days. Just waiting on another crucial bugfix to be addressed

PR #10457, opened 2025-10-02
chore: cleanup changelog for 1.0

matmair:
It was recently pointed out that we never filed out the 1.0 release, so this declares that entries will start with 1.1 as I do not see much value in back-filling this after it is already released


SchrodingersGat (approved):

SchrodingersGat:
Nice catch, thanks

PR #10462, opened 2025-10-02
fix: log ENV key as it needs to be set

matmair:
Currently the error if you are missing important database keys is listing different names than the environment values - that is confusing. This simple PR fixes that.


SchrodingersGat (approved):

matmair:
@SchrodingersGat ready for review - this one is only cosmetic ;-)

PR #10479, opened 2025-10-03
refactor(backend): more types

matmair:
removes global ignore of another typing rule


matmair:
this was made a few weeks ago, seems like there are stil a few things missng

PR #10486, opened 2025-10-04
Add to serializers to fix missing return types in config/flags schema

1337joe:
I noticed that a couple of list/retrieve endpoints (`/api/admin/config/{key}`, `/api/flags/{key}`) were missing return types in the schema and eventually tracked it back to custom serializers that didn't conform to the pattern drf-spectacular can parse when it generates the schema.

Additionally, trying to page through the flags failed because `settings.FLAGS` wasn't a list so I disabled pagination on that field to match admin/config.

For reference, here's the difference in schema for the `/api/admin/config` and `/api/admin/config/{key}` endpoints:
```diff
       responses:
         '200':
-          description: No response body
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Config'
+          description: ''
...
       responses:
         '200':
-          description: No response body
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Config'
+          description: ''
```

### DIFF
--- a/src/backend/InvenTree/common/api.py
+++ b/src/backend/InvenTree/common/api.py
@@ -624,6 +624,9 @@ class FlagList(ListAPI):
     serializer_class = common.serializers.FlagSerializer
     permission_classes = [AllowAnyOrReadScope]
 
+    # Specifically disable pagination for this view
+    pagination_class = None
+
 
 class FlagDetail(RetrieveAPI):
     """Detail view for an individual feature flag."""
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
Hi @1337joe please can you address merge conflicts here :)

SchrodingersGat:
Thanks for the contribution @1337joe and the fix-up @matmair

PR #10497, opened 2025-10-05
add-minimum-stock-to-brief-part-serializer

neilbeukes:
Adding `minimum_stock` param to `PartBriefSerializer`. 

See this discussion for context:
https://github.com/inventree/InvenTree/discussions/10458#discussioncomment-14596310

### DIFF
--- a/src/backend/InvenTree/part/serializers.py
+++ b/src/backend/InvenTree/part/serializers.py
@@ -336,6 +336,7 @@ class PartBriefSerializer(InvenTree.serializers.InvenTreeModelSerializer):
             'locked',
             'assembly',
             'component',
+            'minimum_stock',
             'is_template',
             'purchaseable',
             'salable',
@@ -362,6 +363,8 @@ class PartBriefSerializer(InvenTree.serializers.InvenTreeModelSerializer):
     category_default_location = serializers.IntegerField(
         read_only=True, allow_null=True
     )
+
+    minimum_stock = serializers.FloatField(required=False, default=0)
 
     image = InvenTree.serializers.InvenTreeImageSerializerField(
         read_only=True, allow_null=True


src/backend/InvenTree/part/serializers.py
```diff
@@ -363,6 +364,8 @@ def __init__(self, *args, **kwargs):
         read_only=True, allow_null=True
     )
 
+    minimum_stock = serializers.FloatField(required=False, default=0)
```
SchrodingersGat: You should not have to add a field definition here - it is taken from the underlying model. So, please remove this line. The addition to the "fields" list is sufficient.

SchrodingersGat:
As this changes the API you will need to update `/src/backend/InvenTree/InvenTree/api_version.py`

SchrodingersGat:
As this changes the API you will need to update `/src/backend/InvenTree/InvenTree/api_version.py`

src/backend/InvenTree/part/serializers.py
neilbeukes: Thanks. I tested it and the default provided format was actually a string. So 0 would be "0.00000" 

So thought id update it to be a float. Will remove it.

neilbeukes:
Also @SchrodingersGat. 

I seem to run into this issue when commiting. On codespaces + local setup. 

image

Any idea? I dont think I should have to update dependencies or installed packages on dev server.

src/backend/InvenTree/part/serializers.py
neilbeukes: Done

SchrodingersGat:
@neilbeukes not sure about the go related error. I have not seen that and I am constantly running the devcontainer / vscode setup

SchrodingersGat (approved):

SchrodingersGat:
@neilbeukes note that for any other (more significant) changes to the API, you can also create your own custom API endpoints via plugins

PR #10498, opened 2025-10-05
refactor(backend): shift filterable serializer responses to a more introspection friendly model

matmair:
Follow-up to https://github.com/inventree/InvenTree/issues/9481

Not fully sure about the approach. In the end, I would really like to have a style where we mark a field on a serializer and the annotations and all necessary steps in processing in all data interfaces (API, exports, imports) "just work".

Having things like custom `get_queryset` or `__init__` methods everywhere introduces too ways to make errors IMO.

TODO:
- [x] Performance testing
- [x] INVE-T1 support difference between field and filter name
- [x] INVE-T1 support controlling multiple fields with one filter
- [x] Add a test to ensure all filtered fields (on the serializes) appear in annotations - in the init
- [x] Ensure all annotations are used in serializes - as a CI test

Future work:
Add dev docs how we want to treat filtered endpoints - see https://github.com/inventree/project-adr/issues/4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added much more detailed status information for machines to the API endpoint (including backend and frontend changes) in [#10381](https://github.com/inventree/InvenTree/pull/10381)
 - Added ability to partially complete and partially scrap build outputs in [#10499](https://github.com/inventree/InvenTree/pull/10499)
 - Added support for Redis ACL user-based authentication in [#10551](https://github.com/inventree/InvenTree/pull/10551)
+- Expose stock adjustment forms to the UI plugin context in [#10584](https://github.com/inventree/InvenTree/pull/10584)
 
 ### Changed
 
--- a/src/frontend/CHANGELOG.md
+++ b/src/frontend/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 This file contains historical changelog information for the InvenTree UI components library.
 
+### 0.7.0 - October 2025
+
+Exposes stock adjustment forms to plugins, allowing plugins to adjust stock adjustments using the common InvenTree UI form components.
+
+### 0.6.0 - September 2025
+
+Updated underlying Mantine library versions.
+
 ### 0.5.0 - August 2025
 
 This release updates the base `react` major version from `18.3.1` to `19.1.1`. This change may introduce breaking changes for plugins that rely on the InvenTree UI components library (plugin developers should test their plugins against this new version).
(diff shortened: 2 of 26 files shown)


matmair:
@SchrodingersGat @wolflu05 this would be ready for a review. The interesting things are in `src/backend/InvenTree/InvenTree/serializers.py`, the rest is mostly getting the codebase into line with the new approach to manage serializers that can be filtered

src/backend/InvenTree/build/test_api.py
```diff
@@ -1189,6 +1189,21 @@ def test_sub_builds(self):
...
+    def test_output_options(self):
+        """Test the output options for BuildOrderList list."""
+        self.run_output_test(
+            self.url,
+            [
+                'part_detail'
+                # TODO re-enable ('project_code_detail', 'project_code'),
+                # TODO re-enable 'project_code_detail',
+                # TODO re-enable ('user_detail', 'responsible_detail'),
+                # TODO re-enable ('user_detail', 'issued_by_detail'),
```
matmair, line 1201: These are disabled as they are not working in `master` right now too - this would be a follow-up PR

src/backend/InvenTree/part/test_api.py
```diff
@@ -1395,6 +1395,20 @@ def test_part_notes(self):
...
+    def test_output_options(self):
+        """Test the output options for PartList list."""
+        self.run_output_test(
+            reverse('api-part-list'),
+            [
+                ('location_detail', 'default_location_detail'),
+                'parameters',
+                ('path_detail', 'category_path'),
+                # TODO re-enable ('pricing', 'pricing_min'),
+                # TODO re-enable ('pricing', 'pricing_updated'),
```
matmair, line 1407: These are disabled as they are not working in `master` right now too - this would be a follow-up PR

src/backend/InvenTree/part/test_api.py
```diff
@@ -2668,7 +2682,13 @@ def test_output_options(self):
         """Test that various output options work as expected."""
         self.run_output_test(
             reverse('api-bom-item-detail', kwargs={'pk': 3}),
-            ['can_build', 'part_detail', 'sub_part_detail'],
+            [
+                'can_build',
+                'part_detail',
+                'sub_part_detail',
+                # TODO re-enable 'substitutes',
+                # TODO re-enable ('pricing', 'pricing_min'),
```
matmair, line 2690: These are disabled as they are not working in `master` right now too - this would be a follow-up PR

docs/docs/settings/error_codes.md
```diff
@@ -183,5 +183,12 @@ Overriding a global setting with a different value than the current one.
 
 See [Override global settings](../settings/global.md#override-global-settings) for more information.
 
+#### INVE-I2
```
SchrodingersGat, line 186: Nice, this is a great addition and should hopefully lead to better debugging

SchrodingersGat (approved):
I like this approach overall, great work!

SchrodingersGat:
@wolflu05 any further thoughts? Otherwise I am happy to merge this

matmair:
Thank you, I think this is a good step to get more functions from the individual serializers into a framework of sorts that we can more easily test and performance optimize instead of the large surface of all the use case specific serializers

PR #10521, opened 2025-10-06
fix: correct admin credential msg

matmair:
Fixes an issue where empty access credentials are shown.


matmair:
@SchrodingersGat this is the second fix mentioned before; I would appreciate a quick patch release as the installer is currently broken and requires a manual `sudo inventree run cli update`after every update.

SchrodingersGat:
@matmair no worries I'll make a point release when the backport is merged

matmair:
any idea why the backport is not triggering?

SchrodingersGat:
@matmair takes a few minutes sometimes

PR #10529, opened 2025-10-07
allow mobile check ignore

wolflu05:
This PR adds a new setting to the `FRONTEND_SETTINGS` which enables me to allow to use the PUI interface, of course at my OWN RISK also on mobile devices.

```env
INVENTREE_FRONTEND_SETTINGS='{"mobile_mode": "allow-ignore"}'
```

There are 3 modes now:
- `default` - (default) no changes, works like it worked before
- `allow-ignore` - shows a ignore banner with a link to ignore this warning
- `allow-always` - skips the mobile check and allows mobile devices (of course at the server admins OWN RISK)

### DIFF
--- a/src/frontend/src/main.tsx
+++ b/src/frontend/src/main.tsx
@@ -35,6 +35,7 @@ declare global {
       api_host?: string;
       sentry_dsn?: string;
       environment?: string;
+      mobile_mode?: 'default' | 'allow-ignore' | 'allow-always';
     };
     react: typeof React;
     React: typeof React;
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
@wolflu05 seems reasonable, although I would like this to be documented somewhere!

wolflu05:
Makes sense! I have added some docs now, also for all the other, currently undocumented, frontend settings.

SchrodingersGat:
Nice work on the docs, thanks!

PR #10530, opened 2025-10-07
Generic status endpoint fixes

1337joe:
1. `/api/generic/status` had no return type in the schema, so code generated from the schema wouldn't read the response from the call.

I resorted to returning "Object" here. When I tried to create a serializer that captured the structure of the data the fields would have to be dynamically generated based on the list of class names that inherit from `StatusCode`. There may be a way around this, but drf-spectacular really wants to know the field names of the data to define the structure, not just the type of data contained in the fields.

Resulting diff:
```diff
       responses:
         '200':
-          description: No response body
+          content:
+            application/json:
+              schema:
+                type: object
+                additionalProperties: {}
+          description: Mapping from class name to GenericStateClass data
```

2. When I tried to use a `StatusCode` class name to query the `/api/generic/status/{statusmodel}` it always failed. I fixed the error messages to actually provide the name queried, then added a bit of logic to find the class by name (if it's in the list of valid classes) because I can't pass python classes in via API call.

### DIFF
--- a/src/backend/InvenTree/generic/states/api.py
+++ b/src/backend/InvenTree/generic/states/api.py
@@ -4,6 +4,7 @@ import inspect
 
 from django.urls import include, path
 
+from drf_spectacular.types import OpenApiTypes
 from drf_spectacular.utils import OpenApiResponse, extend_schema
 from rest_framework import serializers
 from rest_framework.generics import GenericAPIView
@@ -14,6 +15,7 @@ import common.serializers
 import InvenTree.permissions
 from data_exporter.mixins import DataExportViewMixin
 from InvenTree.filters import SEARCH_ORDER_FILTER
+from InvenTree.helpers import inheritors
 from InvenTree.mixins import ListCreateAPI, RetrieveUpdateDestroyAPI
 from InvenTree.serializers import EmptySerializer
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@1337joe I think while not optimal, it's better than what we had before. I'm happy to accept these changes - thanks for the continued enhancements :)

SchrodingersGat:
It looks like the changes have triggered some errors in the CI though, please look into that

1337joe:
Yeah, I'll get the tests fixed when I get back to my computer in a couple hours. Simple fix, I just ran out of time to get it pushed.

matmair:
Have you verified that turning off pagination does not break the frontend?

1337joe:
@matmair 
> Have you verified that turning off pagination does not break the frontend?

It appears that paging is implicitly disabled by `serializer_class = EmptySerializer`; having it explicitly set is leftover from when I was trying to build a serializer that could represent the data properly for the schema.

I don't know how to view it in the web interface, but `GlobalStatusState.tsx` is the only use of that API call that I found and it doesn't try to page.

With the current return type being a dictionary, the typical limit/offset paging scheme doesn't directly work anyway. If you like I can revert that line to minimize changes, or I can leave it in to make it more clear that paging isn't done on this endpoint.

PR #10532, opened 2025-10-08
Fix last missing returns

1337joe:
run tests


SchrodingersGat:
Thanks again @1337joe

PR #10549, opened 2025-10-11
CreateSalesOrderLineItem: auto-apply sales price based on part price_breaks, currency, and quantity

KaranLala:
This PR includes: 

**Backend Update – PartsList Endpoint**

- Added a new query parameter for the PartsList api: price_breaks.
- When `price_breaks=True`, the endpoint includes each part’s price_breaks data in the response.
- Default behavior: `price_breaks=False`
- Added corresponding tests to verify both functionality and performance.

**Frontend Update – Automatic Sales Price Calculation**

- Automatically sets the sales_price in the create form (not in the edit form).
- Calculation is based on the part’s currency and quantity.
- Improves the user experience by pre-filling the appropriate price during creation.

**Bug Fix – ApiForm.tsx Field Value Update**

- Fixed an issue where field values were not correctly updating when set via ApiFormFields.
- Root cause:
The defaultValues variable is memoized and recalculates when props.fields changes. However, it was reading from the component’s internal fields state, which hadn’t yet synced with props.fields. As a result, defaultValues used outdated values, and the subsequent setFields call overwrote the updated props.fields.value, preventing value updates.
- Fix:
defaultValues now recalculates based on props.fields instead of fields, ensuring it reflects the latest data and correctly updates the “defaultValue.”
- Note:
The defaultValues variable currently handles both value and defaultValue, which may be confusing. It could be clearer if these concepts were separated, though the existing complexity makes it difficult to fully map all edge cases.

### DIFF
--- a/src/backend/InvenTree/InvenTree/api_version.py
+++ b/src/backend/InvenTree/InvenTree/api_version.py
@@ -1,14 +1,17 @@
 """InvenTree API version information."""
 
 # InvenTree API version
-INVENTREE_API_VERSION = 411
+INVENTREE_API_VERSION = 412
 
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
 
-v411 -> 2025-10-13: https://github.com/inventree/InvenTree/pull/10549
+v412 -> 2025-10-19 : https://github.com/inventree/InvenTree/pull/10549
     - added a new query parameter for the PartsList api: price_breaks (default: false)
+
+v411 -> 2025-06-19 : https://github.com/inventree/InvenTree/pull/10630
+    - Editorialy changes to machine api - no functional changes
 
 v410 -> 2025-06-12 : https://github.com/inventree/InvenTree/pull/9761
     - Add supplier search and import API endpoints
--- a/src/backend/InvenTree/machine/api.py
+++ b/src/backend/InvenTree/machine/api.py
@@ -166,7 +166,7 @@ class MachineTypesList(APIView):
 
     @extend_schema(responses={200: MachineSerializers.MachineTypeSerializer(many=True)})
     def get(self, request):
-        """List all machine types."""
+        """List of all machine types."""
         machine_types = list(registry.get_machine_types())
         results = MachineSerializers.MachineTypeSerializer(
             machine_types, many=True
--- a/src/backend/InvenTree/machine/machine_type.py
+++ b/src/backend/InvenTree/machine/machine_type.py
@@ -48,6 +48,9 @@ class MachineStatus(StatusCode):
     """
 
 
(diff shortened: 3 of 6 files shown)


KaranLala:
> LGTM; you will need to bump the api_version as there is a change to the schema

Is this something you'd like me to do? To clarify the INVENTREE_API_VERSION = 407 in api.version.py, so you want me to bump it up to 408 and write a note re what it is?

Something like 

INVENTREE_API_VERSION = 408

v408 -> 2025-10-11: https://github.com/inventree/InvenTree/pull/10549
    - added a new query parameter for the PartsList api: price_breaks (default: false)

SchrodingersGat:
@KaranLala that's correct - every functional change to the API requires an entry in this file

SchrodingersGat:
@KaranLala very nice work here, this is a great improvement to the user experience.

SchrodingersGat:
Please note that the `api_version.py` file needs updating - as we have since merged another update to the API version. Sorry about that!

KaranLala:
nbd! Lmk if you need any changes here. I'm considering adding company based price breaks for my personal instance, so lmk if you'd like me to implement that in this branch. It will include an update to the sales price panel in part to include a company dropdown (null = default) and the company management page will have a panel to manage it as well (only for sales orders). 

edit: oh and I'm also considering adding a company global discount if set, so prices for the company will be discounted based on the percentage set. Helpful if you have a deal with a certain company, franchise partners etc. Again this is planned only for sales orders and not purchase orders for my instance.

SchrodingersGat:
Looks like only a small formatting issue now. You can install our git precommit hooks to automatically format your code for you!

SchrodingersGat:
@KaranLala looks like the web / frontend tests are what is failing here

KaranLala:
> @KaranLala looks like the web / frontend tests are what is failing here

okay will review

KaranLala:
Some updates: 

1. The Part Parameter form was incorrectly passing the boolean false instead of the string "false" because the adjust value function was not being invoked when initialized. To fix this:
- Updated ApiForm to correctly propagate field.value and field.default when it has falsy values, i.e. false, or 0. 
- Updated ApiFormField to eagerly call adjustValue when first setting the field. This helps when the default value is the same as the form element, which prevents the onChange function from firing (like in the case of switches that are false by default)

2. Updated the pui_tables.spec.ts tests to identify the pagination text using a range (26-29)

All other tests passed when run locally for me. I will revisit when the tests run through to check and fix any issues.

SchrodingersGat:
@KaranLala it also looks like some more merge conflicts have crept in here

KaranLala:
merged! fingers crossed!

SchrodingersGat:
@KaranLala great work and thanks for the addition :)

PR #10556, opened 2025-10-12
[UI] Adjust login error messages

SchrodingersGat:
- Improve login error messages in UI
- Don't hard-code cookie mode into template
- Revert to the "default" values (which are the same)
- Should be useful in helping to determine login issues

### DIFF
--- a/src/frontend/tests/pui_login.spec.ts
+++ b/src/frontend/tests/pui_login.spec.ts
@@ -9,8 +9,8 @@ import { doLogin } from './login.js';
 test('Login - Failures', async ({ page }) => {
   const loginWithError = async () => {
     await page.getByRole('button', { name: 'Log In' }).click();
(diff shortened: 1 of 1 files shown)


src/frontend/src/components/forms/AuthenticationForm.tsx
```diff
@@ -76,12 +76,6 @@ export function AuthenticationForm() {
             followRedirect(navigate, location?.state);
           } else if (success) {
             // MFA login
-          } else {
-            showLoginNotification({
-              title: t`Login failed`,
-              message: t`Check your input and try again.`,
-              success: false
-            });
```
matmair, line 84: why was this removed?

SchrodingersGat: Added back in now, I removed it for testing.

PR #10558, opened 2025-10-12
fix(installer): make VERSION information accessible in invoke calls

### LINKED ISSUE #10554: [installer] environment variables not properly set

### Describe the bug*

When installing via the installer, it appears that the `INVENTREE_PKG_INSTALLER` var is not actually set.

This env-var is required for detecting the proper install method:

```python
def inventreeInstaller() -> Optional[str]:
    """Returns the installer for the running codebase - if set or detectable."""
    # First look in the environment variables, e.g. if running in docker

    installer = os.environ.get('INVENTREE_PKG_INSTALLER', '')

    if installer:
        return str(installer)
```

However as this is not set, the installer method is not correctly detected. This means that in the logs we get messages like:

```
INVE-W10 - Config for 'INVENTREE_CONFIG_FILE' not in recommended directory '/opt/inventree/config'.
```

### Deployment Method

Package

### Version Information

latest

matmair:
This fixes an edge case where arguments in the VERSION file are needed for correct operations.
Closes #10554

### DIFF
--- a/src/backend/InvenTree/InvenTree/version.py
+++ b/src/backend/InvenTree/InvenTree/version.py
@@ -269,7 +269,7 @@ def inventreeBranch():
     branch = os.environ.get('INVENTREE_PKG_BRANCH', '')
 
     if branch:
-        return branch
+        return ' '.join(branch.splitlines())
 
     if main_branch is None:
         return None


SchrodingersGat:
@matmair thanks for the quick fix!

Issue #10548, opened 2025-10-11
[FR] Non-intuitive behaviour of various UI elements and some glitches


MR-DOS:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

Since the 1.0.0, the new skin is enforced. Since the old skin is no longer an option (which I was using), I would like to point out several UX/UI issues which go against usual UI conventions or against human visual pattern recognition.

1. Clickable areas with no indication that they are clickable. No border indicating where that area ends (pointer does not change, too). <img width="469" height="331" alt="Image" src="https://github.com/user-attachments/assets/5a2d38af-6f07-422a-8a35-be4a80dfc866" />

1. File selection element which disguises itself as text field. <img width="799" height="358" alt="Image" src="https://github.com/user-attachments/assets/cca48f46-5e37-4283-90bf-a650acaa1264" />

1. Many menu items are not divided from others by any visual clue. In single-line items, this may be fine, but in multi-line items, this tends to look like chaos. When hovering over items, no highlighting appears which would clarify what the user is actually pointing at. It appears when hovering over the "Stock Items" above, so the UI has highlighting feature. <img width="1185" height="877" alt="Image" src="https://github.com/user-attachments/assets/a3cc15d6-73c6-4007-9236-b7d1298b092a" />
 
1. Parameter view shows "Part" column for every entry though the UI does not seem to allow having parameters from several parts under one part (the part field is not clickable when editing). Maybe it can be changed when the part is a "variant of..." something? <img width="917" height="642" alt="Image" src="https://github.com/user-attachments/assets/395cd90f-c316-4d57-ac59-dc449f033680" /> <img width="805" height="548" alt="Image" src="https://github.com/user-attachments/assets/d384fe85-df08-4167-a094-3a8ce5f5231e" />

 
1. Parameter view now hides edit/delete buttons under Actions button, requiring user to click twice instead of just once. <img width="555" height="178" alt="Image" src="https://github.com/user-attachments/assets/702aee1d-350d-4ca7-bc39-910d7f6d3128" />

[...]

PR #10577, opened 2025-10-13
[UI] Fix broken dashboard link

SchrodingersGat:
Ref: https://github.com/inventree/inventree/issues/10548


src/frontend/src/defaults/actions.tsx
```diff
@@ -27,7 +27,7 @@ export function getActions(navigate: NavigateFunction) {
       id: 'dashboard',
       label: t`Dashboard`,
       description: t`Go to the InvenTree dashboard`,
-      onClick: () => {}, // navigate(menuItems.dashboard.link),
+      onClick: () => navigate('/'),
```
matmair, line 30: Are we sure the dashboard will always be at this location?

PR #10602, opened 2025-10-17
refactor(backend): filtered endpoints - generic testing and small fixes

matmair:
This PR:
- Adds generic testing for the Filtered* Mixins
- Enables filtering on query parameters (on by default)
- Enables tests for that behavior

TODO:
- [x] Improve docstrings for query parameters
- [x] make query parameter parsing controllable on a by-field level
- [x] do we need better enforcing of OpenAPI doc strings?

Follow up to https://github.com/inventree/InvenTree/issues/9481


SchrodingersGat:
@matmair please address conflicts here :)

matmair:

src/backend/InvenTree/build/serializers.py
```diff
@@ -163,6 +164,17 @@ class Meta:
         filter_name='project_code_detail',
     )
 
+    project_code = enable_filter(
```
SchrodingersGat, line 167: @matmair I had to remove this addition in https://github.com/inventree/InvenTree/pull/10605 as it causes 500 errors in the UI tests.

The `project_code` field itself (which is just the numerical PK of the project code) is not dependent on the `project_code_detail` flag - please remove this addition

matmair: Are you sure? This current change reflects the original behviour https://github.com/inventree/InvenTree/pull/10498/files#diff-e0e8be7ad726594bcd2004c2b47a241d9b97458fbb0e032f71ed56ebef635d8dL192-L195

SchrodingersGat: If you look at [the stable version](https://github.com/inventree/InvenTree/blob/stable/src/backend/InvenTree/build/serializers.py) the `project_code` field is not dependent on the `project_code_detail=true` property:

image

I discovered that the addition caused errors in playwright tests and fixed it (be removing the added code) in https://github.com/inventree/InvenTree/pull/10605

matmair: If I look into stable, it clearly is dependent
https://github.com/inventree/InvenTree/blob/c471a1cd386a9290c55c94d904187ca56f365676/src/backend/InvenTree/build/serializers.py#L175-L194
See the variable set in line 175 controlling the removal of the project_code field in line 192

I have removed the section now, but this is clearly a change in behavior

matmair: Exactly this kind of confusion is the reason for the refactor

SchrodingersGat: > See the variable set in line 175 controlling the removal of the project_code field in line 192

Ah, yes, my mistake! This field *is* dependent. In that case, please keep it this way, and I'll try to reproduce the error again and provide a workaround.

Sorry for the mistake!

matmair: Having to search through the init is certainly not transparent, I understand the confusion. I have reverted the removeal

matmair:
The requested change is now in @SchrodingersGat

SchrodingersGat:
Cheers @matmair

Issue #10603, opened 2025-10-17
PO report printing failes after update


mabroens:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

Hi, 

aftre the update last week my custom reports no longer work. ie i get:

<img width="438" height="92" alt="Image" src="https://github.com/user-attachments/assets/2206eabd-5e06-4b0c-897c-d90b1a3251be" />

and than nothing but also no error under error reports.
printing the the default InvenTree Purchase Order doeswork. so i must be something inside my own template but ...

### Steps to Reproduce

print options
print report

### Expected behaviour

i would expect the PDF to be generated as before

### Deployment Method

Bare metal

### Version Information

<img width="705" height="383" alt="Image" src="https://github.com/user-attachments/assets/4387c609-578a-4356-bdc7-c912526cf5bf" />

the copy version info does not seem to work

### Try to reproduce on the demo site

I did not try to reproduce

### Is the bug reproducible on the demo site?

Not reproducible

### Relevant log output

```shell

```

SchrodingersGat:
@mabroens you can enable error reports for failed report printing:

https://docs.inventree.org/en/stable/settings/global/#reporting

Tunr on the `Log Report Errors` setting, then try printing again. And post the error message here

mabroens:
great that helped.

```
["Error rendering report: float() argument must be a string or a real number, not 'Money'"]

Traceback (most recent call last):

File "/opt/inventree/src/backend/InvenTree/report/models.py", line 575, in print

report = self.render(instance, request, None)

File "/opt/inventree/src/backend/InvenTree/report/models.py", line 274, in render

html = self.render_as_string(instance, request, context, **kwargs)

File "/opt/inventree/src/backend/InvenTree/report/models.py", line 261, in render_as_string

return render_to_string(self.template_name, context, request)

File "/opt/inventree/env/lib/python3.10/site-packages/django/template/loader.py", line 62, in render_to_string

return template.render(context, request)

File "/opt/inventree/env/lib/python3.10/site-packages/django/template/backends/django.py", line 61, in render

return self.template.render(context)

File "/opt/inventree/env/lib/python3.10/site-packages/django/template/base.py", line 175, in render

return self._render(context)

File "/opt/inventree/env/lib/python3.10/site-packages/django/template/base.py", line 167, in _render

return self.nodelist.render(context)

[...]

PR #10604, opened 2025-10-17
Report maths tweaks

SchrodingersGat:
Adjust the maths functions exposed to reports, to handle data types other than just floating point values.

- Now data types such as `Money` and `Decimal` can be provided and operated on
- String values are first converted to floats
- Function will raise an error if the operation cannot be performed on the provided types
- Closes https://github.com/inventree/InvenTree/issues/10603


matmair:
Great addition!

Issue #10585, opened 2025-10-14
[FR] Allow overriding both HTTP and HTTPS ports (if port 80/443 are in use)


tehkillerbee:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

After an automatic update caused inventree unable to serve the web page and giving an error instead, I decided it was time to update to the latest configuration, following the documentation listed here, as I was expecting it would fix the problem.
https://docs.inventree.org/en/stable/start/docker_install/

When I last installed inventree, I had to make a few adjustments to the `docker-container.yaml` as my host machine (NAS) already uses port 80 (and 443). While the HTTP WEB port can be set from .env, the HTTPS port cannot:
```
        ports:
#            - ${INVENTREE_WEB_PORT:-80}:80
#            - 443:443
            - ${INVENTREE_WEB_PORT:-8080}:80
            - ${INVENTREE_HTTPS_PORT:-8443}:443

```
This fixed the bind issue, but when trying to login, I now get the INVE-E7 error. 
```
INVE-E7: The visited path `http://server<abcd>.local:8080` does not match the SITE_URL `http://server<abcd>.local`. The INVENTREE_SITE_URL is set via `env` config method - deployment method `DOC`
```

IIRC, a similar error was given after the automatic update caused by this mismatch.

I tried setting `INVENTREE_SITE_LAX_PROTOCOL=True` but it made no change.

I also tried adding the port to SITE_URL, but then I get no error at all (Unable to connect)

If you can point me in the right direction, I'll gladly provide you with a PR to add a way to override these ports.

### Suggested solution

Since I am effectively stuck with no way to override these port, it would be nice if there was a better way to set override them. I can't be the only one hosting inventree in this way

### Describe alternatives you've considered

Changing the default ports used by my NAS to something else than 80 and 443.

### Do you want to develop this?

- [x] I want to develop this.

matmair:
Are you sure you are on latest? We should only be  checking hostname/paths, not ports when using lax comparison method.

tehkillerbee:
I checked and noticed a potential error in my configuration.
```
# InvenTree version tag (e.g. 'stable' / 'latest' / 'x.x.x')
INVENTREE_TAG=latest
```
I forgot 'latest' instead of latest.  So I assume this was the reason why the INVENTREE_SITE_LAX_PROTOCOL didn't work. I will check again ASAP.

[...]

PR #10619, opened 2025-10-19
Fixes for SITE_URL validity checks

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/10585

- Allow lax protocol checking to accept difference in port
- Refactor site URL validity checks
- Fix variable shadowing issue which caused conflicts with internal and external container ports
- Allows the HTTPS port for the proxy to be set externally
- Additional unit testing for site URL checks

### DIFF
--- a/src/backend/InvenTree/InvenTree/middleware.py
+++ b/src/backend/InvenTree/InvenTree/middleware.py
@@ -245,7 +245,7 @@ class InvenTreeHostSettingsMiddleware(MiddlewareMixin):
         site_url_match = (
             (
                 # Exact match on domain
-                is_same_domain(referer.netloc, urlsplit(settings.SITE_URL).netloc)
+                is_same_domain(referer.netloc, site_url.netloc)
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -1556,10 +1556,10 @@ def version(c):
 Backup      {get_backup_dir(error=False) or NOT_SPECIFIED}
 
 Versions:
-Python      {python_version()}
-Django      {InvenTreeVersion.inventreeDjangoVersion()}
 InvenTree   {InvenTreeVersion.inventreeVersion()}
 API         {InvenTreeVersion.inventreeApiVersion()}
+Python      {python_version()}
+Django      {InvenTreeVersion.inventreeDjangoVersion()}
```
matmair, line 1562: Was this change intended to be here?

SchrodingersGat: Yes, it's not strictly related but I think it's better to have our version first

matmair (approved):

PR #10628, opened 2025-10-19
feat(backend): extend schema intro

matmair:
This adds a OpenAPI vendor extension that helps with analyzing endpoints - would be very helpful for https://github.com/inventree/project-adr/issues/4
There is no immediate effect on deployments, as all the new code is disabled by the default setitngs.


matmair:
@SchrodingersGat @wolflu05 this is more meant for maintainers then normal users as it enables us to write jq or spectral rules to lint the API. Any thoughts on additional things that might be helpful? By default nothing changes

src/backend/InvenTree/config_template.yaml
```diff
@@ -38,6 +38,10 @@ debug: False
 debug_querycount: False
 debug_shell: False
 
+# Schema generation options
```
SchrodingersGat, line 43: As this is the default value, please comment it out in the `config_template.yaml` file - that way, if we ever change the default, we won't have users who are "locked in" to the old default

SchrodingersGat:
@matmair how is this intended to be used?

matmair:
> @matmair how is this intended to be used?

With this addition, a dev can set the config key, export the schema and then use any json/yaml tool to filter through the whole API to find issues like https://github.com/inventree/InvenTree/issues/8539 to ensure that the API works in a more uniform way.

Big picture, the goal is to create a set of rules that get run against the API spec to ensure all endpoints are working similar (ish).

matmair:
@SchrodingersGat I added short docs on the added functionality is that enough or should I expand more?

SchrodingersGat (approved):

SchrodingersGat:
I think what you have added is sufficient!

PR #10629, opened 2025-10-19
refactor(backend): add enums for ordering fields

matmair:
Makes enum easier to introspects as this is right now completely untyped

### DIFF
--- a/src/backend/InvenTree/InvenTree/mixins.py
+++ b/src/backend/InvenTree/InvenTree/mixins.py
@@ -223,6 +223,20 @@ class OutputOptionsMixin:
         if getattr(cls, 'output_options', None) is not None:
             schema_for_view_output_options(cls)
 
+    def __init__(self) -> None:
+        """Initialize the mixin. Check that the serializer is compatible."""
+        super().__init__()
+
+        # Check that the serializer was defined
+        if (
+            hasattr(self, 'serializer_class')
+            and isinstance(self.serializer_class, type)
+            and (not issubclass(self.serializer_class, FilterableSerializerMixin))
+        ):
+            raise Exception(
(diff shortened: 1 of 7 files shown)


matmair:
@1337joe any thoughts on this? This section was originally added by you.

src/backend/InvenTree/InvenTree/schema.py
```diff
@@ -105,9 +105,11 @@ def get_operation(
             parameters = operation.get('parameters', [])
             for parameter in parameters:
                 if parameter['name'] == 'ordering':
-                    parameter['description'] = (
-                        f'{parameter["description"]} Possible fields: {", ".join(ordering_fields)}.'
-                    )
+                    schema_order = []
+                    for field in ordering_fields:
+                        schema_order.append(field)
+                        schema_order.append('-' + field)
```
1337joe, line 111: Adding to the list twice gives duplicate entries.

```suggestion
```

1337joe (approved):
While the code generator I use for Java doesn't actually enforce the enum constraint on string fields annotated this way, I prefer this over my original code for the improved readability of the schema and api-doc.

src/backend/InvenTree/InvenTree/schema.py
matmair: That is intended, you can use a dash followed by the fieldname for a revere ordering

1337joe: Ah, I've used that but seeing it as an enum value looked out of place. My mistake.

matmair:

SchrodingersGat (approved):

SchrodingersGat:
@matmair thanks for the addition

PR #10632, opened 2025-10-20
[API] Bug fix for FilterableSerializerMixin

SchrodingersGat:
- Handle case where data is being exported
- Follow-up to https://github.com/inventree/InvenTree/pull/10498
- Adds unit test to ensure we catch this error next time

### DIFF
--- a/src/backend/InvenTree/part/test_api.py
+++ b/src/backend/InvenTree/part/test_api.py
@@ -3339,3 +3339,27 @@ class PartTestTemplateTest(PartAPITestBase):
         response = self.patch(url, {'choices': 'a,b,c,d,e,f,f'}, expected_code=400)
 
         self.assertIn('Choices must be unique', str(response.data['choices']))
+
+
+class PartParameterTests(PartAPITestBase):
+    """Unit test for PartParameter API endpoints."""
+
+    def test_export_data(self):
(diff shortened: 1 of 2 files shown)


src/backend/InvenTree/part/test_api.py
```diff
@@ -3339,3 +3339,27 @@ def test_choices(self):
...
+            data={
+                'export': True,
+                'export_plugin': 'inventree-exporter',
+                'part_detail': True,
+                'template_detail': True,
+            },
+            expected_code=200,
+        )
+
+        fields = response.data['actions']['GET'].keys()
```
matmair, line 3362: Would it not be easier to find/understand if this test for `data_exporter` behavior would be in either `data_exporter` or `InvenTree` in a generic form? A refactor of this specific section of Business API might remove coverage from a core mechanism

matmair (approved):

src/backend/InvenTree/part/test_api.py
SchrodingersGat: Do you mean the unit test should be in the `data_exporter` directory?

matmair: I think we should probably add a decoupled test in data_exporter

PR #10645, opened 2025-10-21
Tweak for currency_code_default

SchrodingersGat:
- Did not run full CI suite last time
- Follow up to https://github.com/inventree/InvenTree/pull/10641

PR #10646, opened 2025-10-21
[bug] Remove 'project_code' from optional fields

SchrodingersGat:
- enable_filter overrides field options
- Field does not auto-describe as a related field any more
- Breaks front-end forms
- Includes and closes https://github.com/inventree/InvenTree/pull/10645

This PR is a follow-up to https://github.com/inventree/InvenTree/pull/10498, now that I have narrowed in on the actual bug here.

The `project_code` field in the `BuildList` API endpoint was placed behind an `enable_filter` method. However this "disguises" the actual field type - changing it from a `RelatedField` to an `IntegerField`. This means that the field introspection does *not* work when making an `OPTIONS` request.

@matmair sorry for the noise on this, I think this would have been caught if we'd tagged a `full_run` on the original PR - as the frontend code is where this would be caught.

### Before Change

image

### After Change (Previous Field Definition)

image

### DIFF
--- a/src/backend/InvenTree/build/test_api.py
+++ b/src/backend/InvenTree/build/test_api.py
@@ -1240,7 +1240,6 @@ class BuildListTest(BuildAPITest):
             self.url,
             [
                 'part_detail',
-                ('project_code_detail', 'project_code'),
                 'project_code_detail',
                 ('user_detail', 'responsible_detail'),
                 ('user_detail', 'issued_by_detail'),


matmair:
We could (in theory) also add a filterable related field to solve this

matmair:
We could (in theory) also add a filterable related field to solve this

SchrodingersGat:
@matmair maybe a good addition for a future update - however I think this fix is acceptable for now. There's no harm in leave the `project_code` pk there by default

PR #10651, opened 2025-10-22
Add missing nullable annotation to optional field

1337joe:
#10549/v412 added an optional parameter but the field returned is not nullable, so any query returned without the optional field requested fails to validate against the schema. This adds the allow_null annotation to the new field.


matmair (approved):

SchrodingersGat:
@1337joe please address conflicts here

SchrodingersGat:
@1337joe thanks again for the ongoing fixes and improvements!

PR #10654, opened 2025-10-23
[UI] Implement "checked_by" for SalesOrderShipment

SchrodingersGat:
The `checked_by` flag has existed in the API for a while but not ever been utilized.

This PR adds the following UI features:

- Display "checked_by" status for a SalesOrderShipment
- Add API and table filtering
- Adds UI actions to mark a shipment as "checked"
- Adds option to prevent sending shipments which have not been checked
- Additional documentation and unit testing

### DIFF
--- a/src/backend/InvenTree/InvenTree/api_version.py
+++ b/src/backend/InvenTree/InvenTree/api_version.py
@@ -1,13 +1,18 @@
 """InvenTree API version information."""
 
 # InvenTree API version
-INVENTREE_API_VERSION = 416
+INVENTREE_API_VERSION = 417
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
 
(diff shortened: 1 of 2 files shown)


src/backend/InvenTree/part/serializers.py
```diff
@@ -1042,9 +1042,7 @@ def get_starred(self, part) -> bool:
     )
 
     price_breaks = enable_filter(
-        PartSalePriceSerializer(
-            source='salepricebreaks', many=True, read_only=True, allow_null=True
-        ),
```
1337joe, line 1047: Looks like a merge problem: this just reverted #10651.

If nobody gets to it first I'll put in another PR tomorrow.

SchrodingersGat: @1337joe I'll fix that now, that was unintended, probably due to the multiple parallel commits. Sorry about that!

Issue #8607, opened 2024-12-01
Refactor pricing mechanics and move to plugins


matmair:
## Problem statment
The current state of pricing is mostly coded into InvenTree's code base. This makes part pricing more complicated than necessary for hobby users and lacks the nuance needed for business users.

## Vision
Plugins should be the primary way to handle prices. The default system should provide a simple pricing setup that only considers BoM part costs.

## Breakdown
General goals:
1. Pricing calculations should be passed through plugins
a. Plugins should be able to add or subtract values and add metadata that can be added to reports, the UI etc.
b. Plugins should be aware of currencies and time
2. Move pricing calculations for parts into a plugin that can be disabled
a. Move settings for internal pricing to plugin
b. Move variant pricing overrides to the plugin
c. Move supplier pricing to the plugin (https://github.com/inventree/InvenTree/issues/7765, https://github.com/inventree/InvenTree/issues/3928, https://github.com/inventree/InvenTree/issues/7603)
3. Add pricing calculations via plugins to orders
a. Pass through metadata and part parameters
b. Enable pricing customisable types via states (for example shipping, packaging, handling fees, taxes, discounts, allowance)
4. Add cost to builds (https://github.com/inventree/InvenTree/issues/7949, https://github.com/inventree/InvenTree/issues/7279)
a. BoM cost
b. General Handling cost
c. Per build order cost
d. Per part cost
e. Possibly depending on worker steps (could be very usable to asses currently sunken cost per build)
5. Costs on tests?
6. Add permissions to ensure that detailed breakdowns etc can be restricted
7. Possibly add the option for plugins to require manual checks of prices

Plugin design ideas:
- Make pricing composable
- Make pricing runs repeatable (and time-fixed)
- Fix plugin order
- Store pricing runs and configurations in database

Open questions:
- Data flow design
- Python primitives
- Plugin API
- UI flow and integration

> [!WARNING]
This is a draft and would be a major amount of work. This will probably require a major release and intense testing and UI/UX design. Do not expect this to be done fast.

Issue #4022, opened 2022-12-06
[Feature] Permissions system for pricing


SchrodingersGat:
Add permission "role" to control who can view / edit / delete pricing data

- Internal pricing
- Sale pricing
- Supplier pricing

Purchase order and sale order pricing will still be controlled based on the SalesOrder and PurchaseOrder permission roles.

matmair:
Probably a part of a bigger permissions refactor once the main things for React are implemented

SchrodingersGat:
Fresh!

SchrodingersGat:
Not stale

PR #10669, opened 2025-10-25
[WIP] Pricing plugin

SchrodingersGat:
This PR implements a plugin system for pricing integration.

- The default plugin provides a base pricing structure which matches the previous code for backwards compatibility.
- Custom plugins can be implemented to override or extend part (or all) of the pricing functionality
- This is an initial implementation that can be extended in the future
- The scope of this PR is just to implement the plugin structure for pricing

### Breaking Change

This PR introduces a breaking change - the existing global settings for adjusting pricing behavior have been moved to the (newly introduced) default pricing plugin. This means that if you have adjusted the default values of the pricing configuration settings, these settings will need to be re-adjusted against the new plugin.

### Tasks

- [ ] Documentation for new plugin mixin
- [ ] Documentation for custom pricing functions via plugins
- [ ] Move pricing configuration settings to the default plugin
- [ ] Remove duplicated pricing settings from the global settings list
- [ ] Move all pricing functions to a plugin-driven code structure
- [ ] Document the default pricing plugin
- [ ] Document each setting for the default pricing plugin
- [ ] Reimplement functionality for calculating cost of "stock on hand"
- [ ] Migrate `update_bom_cost`
- [x] Migrate `update_purchase_cost`
- [x] Migrate `update_internal_cost`
- [x] Migrate `update_supplier_cost`
- [x] Migrate `update_variant_cost`
- [ ] Migrate `update_overall_cost`
- [ ] Migrate `update_sale_cost`
- [ ] Add plugin hooks for calculating the quantity purchase pricing for a part
- [ ] Add plugin hooks for calculating the quantity sale pricing for a part

### Talking Points

1. Can we (or should we) support having multiple pricing plugins operating, allowing a custom plugin to implement only a part of the functionality?

2. If not 1) then how to we support a custom plugin which only wants to make a "small" change. Does the plugin need to re-implement *all* functionality?


matmair:
Interesting approach; Regarding your questions:


> 1. Can we (or should we) support having multiple pricing plugins operating, allowing a custom plugin to implement only a part of the functionality?

For now, I think only having 1 plugin active will make things simpler



> 2\. If not 1) then how to we support a custom plugin which only wants to make a "small" change. Does the plugin need to re-implement _all_ functionality?

Maybe - this is mainly a thought experiment - developers could use inheritance? If the default plugin is split up enough in their function that should be reasonable

On a more general note: I think this is a good step, but still not a stable footing that satisfied the request we are getting. Looking at https://github.com/inventree/InvenTree/issues/8607 . We definitely should think about changing calculations and changing the data structure we use to save prices and price calculations. We ignore time for example right now - time can be very important for price and cross-currency pricing

SchrodingersGat:
Do you have any suggestions for specific changes to the data structure?

matmair:
I have not looked deeply into it as I am currently working on https://github.com/inventree/InvenTree/issues/4022, but I thought about it a bunch of the RFQ request. 

Currently, in my mind, the following fields would make sense in the DB; not all needed in the API:
- Calculation meta: plugin, plugin version, calculation date
- if the value was calculated or set manually
- money value in decimal and currency (already there)
- maybe some correlation to the calculation run
- maybe some references to the data that was used (currency exchange rates come to mind)

These would ensure that faulty calculations can be identified and re-run and that we can try to get calculation stability

PR #10671, opened 2025-10-25
Remove junk file

SchrodingersGat:
Introduced by CI script in https://github.com/inventree/InvenTree/pull/10660#pullrequestreview-3380213188


SchrodingersGat:
@matmair looks like a glitch or error in the CI script, something to look out for on next translation update.

matmair:
Certainly something worth looking into; this seems like nothing a translation pipeline should introduce

PR #10720, opened 2025-10-30
[UI] Plugin actions

SchrodingersGat:
Support custom spotlight actions in the UI.

### Tasks

- [x] Playwright tests


matmair:
We already have something called action in the backend; can we change the key to spotlightaction?

matmair (approved):

PR #10722, opened 2025-10-30
Fix selection list items not displaying values

### LINKED ISSUE #10701: Selection list don't display items

### Describe the bug*

The selection list GUI don't display the items values, just create the lines.

Image

### Steps to Reproduce

1. Go to Admin Center > part-parameters
2. Drill down Selection Lists
3. edit one selection list

### Expected behaviour

Display values of selection list items.

### Deployment Method

Docker

### Version Information

1.0.8

I did not try to reproduce

Not reproducible

### Relevant log output

```shell

```

akhilsingh-git:
Fixes #10701

The issue was that the table row key was using 'props.item.pk', but the API returns 'id' instead. This caused React to not properly render the rows with their values.

Changed the key to use 'props.item.id' with a fallback to 'props.idx' for new rows that don't have an ID yet.

### DIFF
--- a/src/frontend/src/forms/selectionListFields.tsx
+++ b/src/frontend/src/forms/selectionListFields.tsx
@@ -1,11 +1,9 @@
+import type { ApiFormFieldSet, ApiFormFieldType } from '@lib/types/Forms';
 import { t } from '@lingui/core/macro';
 import { Table } from '@mantine/core';
 import { useMemo } from 'react';
-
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
If you run `pre-commit install` in your repo, these style issues will be automatically fixed on a git commit

SchrodingersGat:
@akhilsingh-git thanks for the fix!

Issue #10732, opened 2025-11-01
install.sh: line 74: syntax error near unexpected token `;;'


tlf30:
### Deployment Method

- [x] Installer
- [ ] Docker Development
- [ ] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

Hello,
I did a clean install of Ubuntu Server 20.04.6 LTS (As advised in the documentation). 
When I try to run the installer, I get the following error:
```
trevor@akinv:~$ wget -qO install.sh https://get.inventree.org && bash install.sh                                                                                                                                                             install.sh: line 74: syntax error near unexpected token `;;'
install.sh: line 74: `          ;;'
```

### Steps to Reproduce

1. Go to `https://docs.inventree.org/en/stable/start/installer/`
2. See the requirements as `Supported OSs are Debian 11 and Ubuntu 20.04 LTS.`
3. Install the latest Ubuntu 20.04 LTS (20.04.6)
4. From the same document run the install command: `wget -qO install.sh https://get.inventree.org && bash install.sh`

### Relevant log output

```bash
login as: trevor
trevor@10.0.0.24's password:
Welcome to Ubuntu 20.04.6 LTS (GNU/Linux 5.4.0-144-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  System information as of Sat 01 Nov 2025 05:26:06 PM UTC

  System load:  0.62               Processes:              239
  Usage of /:   2.5% of 241.08GB   Users logged in:        0
  Memory usage: 7%                 IPv4 address for ens34: 10.0.0.24
  Swap usage:   0%


Expanded Security Maintenance for Infrastructure is not enabled.

0 updates can be applied immediately.

Enable ESM Infra to receive additional future security updates.
See https://ubuntu.com/esm or run: sudo pro status

Failed to connect to https://changelogs.ubuntu.com/meta-release-lts. Check your                                                                                                                                                              Internet connection or proxy settings


Last login: Sat Nov  1 17:23:26 2025
/usr/bin/xauth:  file /home/trevor/.Xauthority does not exist
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

[...]

PR #10744, opened 2025-11-02
fix: typo ins installer stops script from working

matmair:
Fixes https://github.com/inventree/InvenTree/issues/10732


matmair:
@SchrodingersGat this needs to be backported to `stable`

SchrodingersGat (approved):

PR #10751, opened 2025-11-03
fix(dbbackup): cannot use custom backup_storage

lehoangnam040:
- after #10140 the `STORAGES` with dbbackup is override
- If I want to use custom backup_storage it will automatically fallback to use local storage
- the fixes here to allow uses `dbbackup` from `InvenTree.backup`


SchrodingersGat:
@lehoangnam040 thanks for the patch

PR #10752, opened 2025-11-03
Auth Improvements

SchrodingersGat:
- Reject auth requests from user accounts which are inactive
- Additional detail in MFA required response

Note: This PR does not fix any of the issues listed in #10754


matmair:
Adding a test for the new route would be good IMO

Issue #10754, opened 2025-11-03
MFA Issues


SchrodingersGat:
Multiple issues related to MFA login.

### Enforced MFA Login

If the `LOGIN_ENFORCE_MFA` option is selected, the user is not correctly notified, or redirected to the MFA page.

1. Login as admin user (without MFA)
2. Go to the system settings page
3. Enable the `LOGIN_ENFORCE_MFA` setting
4. Experience broken UI

The user is not logged out, however any successive API requests fail silently with `401`:

<img width="1451" height="368" alt="Image" src="https://github.com/user-attachments/assets/b19af25f-172e-4681-b83a-090b1fcb34cd" />

However, the UI still gives appearance that the user is correctly logged in

### Media File Access

- Ref: https://github.com/inventree/inventree-app/issues/705#issuecomment-3478480925

Issues accessing media files from the app, when `LOGIN_ENFORCE_MFA` is active. Although the `/api/` requests work OK on the app, any requests to `/media/` fail - due to not meeting the strict MFA check.

### Login Screen

A user without MFA can still log in - they are *not* presented with an MFA scanning option if they do not have any MFA configured. A user without MFA configured should be redirected to the MFA configuaration screen, rather than being logged in

### Assign TOTP Codes

There seem to be issues assigning codes via TOTP in the current version. The UI just throws a generic 401 error.

<img width="1538" height="429" alt="Image" src="https://github.com/user-attachments/assets/27bc4fd9-0a19-4d44-b835-45a550d32d4b" />

This is with a "fresh" TOTP code via Google authenticator app - it should be working just fine

From the console logs

```
Unauthorized: /api/auth/v1/account/authenticators/totp
2025-11-03 10:36:17,494 WARNING Unauthorized: /api/auth/v1/account/authenticators/totp
```

### Web Index Access

The 'index' page cannot be accessed if `LOGIN_ENFORCE_MFA` is enabled:

<img width="766" height="328" alt="Image" src="https://github.com/user-attachments/assets/ddd9828c-57c8-48e9-9033-8c504d01c1ea" />

### INVENTREE_MFA_ENABLED setting

This currently has no effect - whether enabled or disabled

SchrodingersGat:
@matmair I've been spending some time digging into these issues - there are a few key issues with our frontend code which are preventing the MFA workflows from working properly! Will document further over the next few days

matmair:
@SchrodingersGat I can not reproduce 4; on which instance did you test that?

[...]

PR #10767, opened 2025-11-04
Allow additional MFA configuration options

SchrodingersGat:
- Add `INVENTREE_MFA_PASSKEY_LOGIN_ENABLED` setting
- Add `INVENTREE_MFA_WEBAUTHN_ALLOW_INSECURE setting

Part of working towards general MFA improvements

### DIFF
--- a/src/backend/InvenTree/InvenTree/settings.py
+++ b/src/backend/InvenTree/InvenTree/settings.py
@@ -1378,18 +1378,27 @@ ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE = True
 
 HEADLESS_ONLY = True
 HEADLESS_CLIENTS = 'browser'
-MFA_ENABLED = get_boolean_setting(
-    'INVENTREE_MFA_ENABLED', 'mfa_enabled', True
-)  # TODO re-implement
-MFA_SUPPORTED_TYPES = get_setting(
-    'INVENTREE_MFA_SUPPORTED_TYPES',
-    'mfa_supported_types',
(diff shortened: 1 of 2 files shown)


src/backend/InvenTree/InvenTree/settings.py
```diff
@@ -1378,18 +1378,27 @@
...
+    )
+    if MFA_ENABLED
+    else []
 )
+
 MFA_TRUST_ENABLED = True
-MFA_PASSKEY_LOGIN_ENABLED = True
-MFA_WEBAUTHN_ALLOW_INSECURE_ORIGIN = DEBUG
+
+MFA_PASSKEY_LOGIN_ENABLED = get_boolean_setting(
```
matmair, line 1395: This might leave users without a way to login; you can login with inky a passkey in theory;
Adding a way to disable it after it was enabled means we need to mitigate that

src/backend/InvenTree/InvenTree/settings.py
```diff
@@ -1378,18 +1378,27 @@
...
+    'INVENTREE_MFA_PASSKEY_LOGIN_ENABLED', 'mfa_passkey_login_enabled', True
+)
+
+MFA_WEBAUTHN_ALLOW_INSECURE_ORIGIN = get_boolean_setting(
```
matmair, line 1399: There are a lot of things modern features that do not work in an insecure context anymore - even if we allow it. I would recommend to not add a way to enable this

SchrodingersGat:
@matmair OK given that there are issues with adding configurability on both these settings, I"ll close this out.

Issue #10765, opened 2025-11-04
[FR] Sorting Parts by Category


xlfds:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

We have some parts for which there are many items in the bill of materials. For example, some of these are wires, screws and PCBs.
We would like to sort these lines by part category.

### Suggested solution

Add a new column to the BO/Required Parts table and the Part/Bill of Materials table. Then allow sorting by this column.

### Describe alternatives you've considered

Sorting by name helps partially.

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
@xlfds a reasonable request I think. Would you be interested in implementing this?

wolflu05:
Do you mean adding row dragging to the individual lines on the frontend and then update the additional "order" column in the background, or wouldn't ordering by the part caregory.lft value be enough?

xlfds:
I mean, add a sortable 'Part Category' column.

<img width="2642" height="632" alt="Image" src="https://github.com/user-attachments/assets/80008599-b399-4dc5-a448-481b859549f4" />

SchrodingersGat:
@xlfds this is a nice addition that I would immediately use - so, feature incoming in https://github.com/inventree/InvenTree/pull/10772

PR #10772, opened 2025-11-04
[UI] BOM part category

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/10765
- Adds "category" column to BOM table
- Adds "category" column to build lines table
- Adds required API filters and detail attributes

image


matmair:
@SchrodingersGat would a filter also be helpful?
Edit: I see it already is there in the API

PR #10774, opened 2025-11-05
improve docker dx with empty migration set

matmair:
In docker deploys, try running migrations if the database is empty and not testing


src/backend/InvenTree/InvenTree/tasks.py
```diff
@@ -661,6 +661,12 @@ def get_migration_plan():
     return plan
 
 
+def get_migriation_count():
```
SchrodingersGat: Please fix spelling here: `get_migration_count`

src/backend/InvenTree/InvenTree/apps.py
```diff
@@ -321,6 +321,17 @@ def ensure_migrations_done(self=None):
             return
 
         if not InvenTree.tasks.check_for_migrations():
-            logger.error('INVE-W8: Database Migrations required')
-            sys.exit(1)
+            # Detect if this an empty database - if so, start with a fresh migration
+            if (
+                settings.DOCKER
+                and not InvenTree.ready.isInTestMode()
```
SchrodingersGat, line 327: Should we also check that migrations are not already running?

matmair: yes, added now

SchrodingersGat:
Thanks, this will be a good improvement to the workflow

wolflu05:
What does this change exactly in the deployment process? Do we need to document this somewhere?

PR #10775, opened 2025-11-05
[UI] MFA Refactor

SchrodingersGat:
- Fixes various bugs / issues with MFA auth flows in the UI
- Comprehensive checks of edge cases, with user feedback / notifications
- Adds playwright testing for TOTP code generation

### DIFF
--- a/src/frontend/src/pages/Index/Settings/AccountSettings/MFASettings.tsx
+++ b/src/frontend/src/pages/Index/Settings/AccountSettings/MFASettings.tsx
@@ -11,6 +11,7 @@ import {
   Group,
   Loader,
   Modal,
+  Paper,
   PasswordInput,
   SimpleGrid,
   Stack,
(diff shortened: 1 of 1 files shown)


src/frontend/src/pages/Index/Settings/AccountSettings/MFASettings.tsx
```diff
@@ -0,0 +1,996 @@
...
+            {unusedCodes.length > 0 ? (
+              unusedCodes.map((code: string) => (
+                <Group
+                  p={3}
+                  justify='space-between'
+                  key={`mfa-recovery-code-${code}`}
+                >
+                  <Text>{code}</Text>
+                  <CopyButton value={code} />
+                </Group>
```
matmair: The individual copy buttons are a bit misleading here; should users one by one copy these codes to secure offline storage? Normally one should just grap all of them as they each only work once

matmair (commented):
LGTM in general;
- i still get shown recovery odes after a while with no top enabled
- with this we now do not need useReauth/useConfirm anymore so it should be removed

I have not had a way to test the reauth/mfareauth flows: have you tested that in practice?

src/frontend/src/pages/Index/Settings/AccountSettings/MFASettings.tsx
SchrodingersGat: Good point, I have made a single copy button now.

SchrodingersGat:
>  i still get shown recovery odes after a while with no top enabled

This might need some further investigation, I can't seem to reproduce this

>  with this we now do not need useReauth/useConfirm anymore so it should be removed

Done!

> I have not had a way to test the reauth/mfareauth flows: have you tested that in practice?

Yes, I have tested extensively with real TOTP and webauthn, and re-authentication

PR #10776, opened 2025-11-05
Remove custom token middleware

SchrodingersGat:
- Token auth handled by allauth now
- Ref: https://github.com/inventree/InvenTree/pull/10775

This fixes the auth issues in the app when MFA tokens are enforced. The custom token check middleware interferes with allauth middleware layer. I have tested with the app, this code works well.

- Closes https://github.com/inventree/inventree-app/issues/705

### DIFF
--- a/src/backend/InvenTree/InvenTree/middleware.py
+++ b/src/backend/InvenTree/InvenTree/middleware.py
@@ -6,8 +6,8 @@ from urllib.parse import urlsplit
 from django.conf import settings
 from django.contrib.auth.middleware import PersistentRemoteUserMiddleware
 from django.http import HttpRequest, HttpResponse, JsonResponse
-from django.shortcuts import redirect, render
-from django.urls import resolve, reverse_lazy
+from django.shortcuts import render
+from django.urls import reverse_lazy
 from django.utils.deprecation import MiddlewareMixin
 from django.utils.http import is_same_domain
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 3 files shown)


matmair:
Have you checked that user actions get correctly correlated to the correct user when tokens are used and that token validity is checked (ie. expired token do not still work)? Both was not the case the last time we tried this

SchrodingersGat:
> Have you checked that user actions get correctly correlated to the correct user when tokens are used and that token validity is checked (ie. expired token do not still work)? Both was not the case the last time we tried this

If we do not have this as part of the test suite already, I'll add it as part of this PR

src/backend/InvenTree/InvenTree/middleware.py
```diff
@@ -51,101 +50,6 @@ def get_token_from_request(request):
...
-                path = request.path_info
-
-                if path not in urls and not any(
-                    path.startswith(p) for p in paths_ignore
-                ):
-                    # Save the 'next' parameter to pass through to the login view
-
-                    return redirect(
-                        f'{reverse_lazy("account_login")}?next={request.path}'
-                    )
```
matmair: Will this be implemented somewhere else?

src/backend/InvenTree/InvenTree/middleware.py
```diff
@@ -51,101 +50,6 @@ def get_token_from_request(request):
...
-
-        # anymail requests are handled by the anymail library
-        if request.path_info.startswith('/anymail/'):
-            return self.get_response(request)
-
-        # Is the function exempt from auth requirements?
-        path_func = resolve(request.path).func
-
-        if getattr(path_func, 'auth_exempt', False) is True:
-            return self.get_response(request)
```
matmair: @SchrodingersGat how will this be handled in the future?

matmair:
I have had time to test this now; without the mixin things like plugins are not auth protected anymore; the API has it's own checks so it does not cause issues there but plugins, media and all other paths just do not check auth anymore

SchrodingersGat:
@matmair clearly not fit for purpose then. Will have to find a different way to address the issues outlined in https://github.com/inventree/inventree-app/issues/705

PR #10796, opened 2025-11-10
fix: MFA enforce flows / interactions

### LINKED ISSUE #11122: "You must enable two-factor authentication before doing anything else.", while trying to enable 2fa

### Describe the bug*

If I tick the box that forces 2fa, it forces 2fa, makes sense, right?
I ticked that as the only admin user in my instance, then went to the user section to try enable 2fa. But, it says I need to enable 2fa to render the table... The table that allows me to sign up for 2fa...
Video of the bug: https://youtu.be/Xp4vES-Pg5I

### Steps to Reproduce

1. Go into settings
2. Enable 2fa for all users
3. Go into user settings
4. Try to enable 2fa

### Expected behaviour

That the 2fa logic does not kick me out while trying to enable 2fa

### Deployment Method

Docker

### Version Information

I cannot do this. It logs me out.

I did not try to reproduce

Not reproducible

### Relevant log output

```shell

```

matmair:
Improve MFA experience when MFA enforcement gets enabled.
- All users are logged out when mfa enforcement is changed
- There is an explanation os-level alert before that happens
- The mandatory MFA registration now works on new logins
- Fix single-form login with password managers

Closes https://github.com/inventree/InvenTree/issues/10754
Closes #11122

### DIFF
--- a/src/backend/InvenTree/InvenTree/api_version.py
+++ b/src/backend/InvenTree/InvenTree/api_version.py
@@ -1,10 +1,13 @@
 """InvenTree API version information."""
 
 # InvenTree API version
-INVENTREE_API_VERSION = 439
+INVENTREE_API_VERSION = 440
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
+
+v440 -> 2026-01-15 : https://github.com/inventree/InvenTree/pull/10796
+    - Adds confirm and confirm_text to all settings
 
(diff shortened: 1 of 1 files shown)


matmair:
@SchrodingersGat ready for review and maybe also a trial. I have spent a lot of time testing this, but maybe I have overseen some edge case.

SchrodingersGat (approved):

SchrodingersGat:
@matmair nice work, looks good from my testing. Thanks for fixing this!

PR #10814, opened 2025-11-13
Fix for string form fields

SchrodingersGat:
- replace null values with empty strings
- Ensure that strings are not replaced with null values sometimes


wolflu05:
We have to carefully consider this change, I'm not really sure if this does not break something.

SchrodingersGat:
@wolflu05 I have added some extra checks here, it will only overwrite a null value if the serializer does not allow null values

PR #10730, opened 2025-11-01
feat(backend)!: bump to dj 5.2 lts / py 3.11

matmair:
Bump to Django to 5.2 LTS, Python 3.11 will give us support till 2027-10 (PEP 664)

This removes Debian 11 and Ubuntu 20.04 as those do not support the minimum python version

Addresses https://github.com/inventree/InvenTree/security/dependabot/239
Addresses https://github.com/inventree/InvenTree/security/dependabot/218

PR #10855, opened 2025-11-18
[bug] Media url fix

SchrodingersGat:
- Fixes bug introduced in https://github.com/inventree/InvenTree/pull/10730/files
- The `getMediaUrl` function was not handled correctly
- The `fully_qualified_url` method on the `Attachment` model was passing the wrong value to `getMediaUrl`

### DIFF
--- a/src/frontend/netlify.toml
+++ b/src/frontend/netlify.toml
@@ -7,6 +7,7 @@
 
 [build.environment]
   VITE_DEMO = "true"
+  PYTHON_VERSION = "3.11"
 
 # Send requests to subpath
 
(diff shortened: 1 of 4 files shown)


matmair:
Have you validated this against a s3 / s3-like backend?

SchrodingersGat:
> Have you validated this against a s3 / s3-like backend?

No, although the `TypeError` check is in-line with the type hinting in the function definition, whereas it was not before. I don't have ability to test against S3 - do you want to if you think it is worth it?

matmair:
found time to test, seems to work with s3

PR #10873, opened 2025-11-20
[plugin] Add printing delay to sample machines

SchrodingersGat:
- Useful for testing
- Default is zero

### DIFF
--- a/src/backend/InvenTree/InvenTree/api_version.py
+++ b/src/backend/InvenTree/InvenTree/api_version.py
@@ -1,10 +1,13 @@
 """InvenTree API version information."""
 
 # InvenTree API version
-INVENTREE_API_VERSION = 425
+INVENTREE_API_VERSION = 426
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
+
+v426 -> 2025-11-19 : https://github.com/inventree/InvenTree/pull/10867
+    - Adds optional "price_breaks" filter to the SupplierPart API endpoint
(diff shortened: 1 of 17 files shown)


src/backend/InvenTree/machine/machine_type.py
```diff
@@ -357,14 +357,19 @@ def reset_errors(self):
...
+        cache: bool = False,
+        backup_value: Any = None,
     ):
         """Return the 'value' of the setting associated with this machine.
 
         Arguments:
             key: The 'name' of the setting value to be retrieved
             config_type_str: Either "M" (machine scoped settings) or "D" (driver scoped settings)
             cache: Whether to use RAM cached value (default = False)
+            backup_value: A backup value to return if the setting is not found
```
wolflu05: Where is the advantage of introducing another method of setting a default? Can't we just set this as default for the setting and make it required?

src/backend/InvenTree/plugin/samples/machines/sample_printer.py
```diff
@@ -44,7 +53,14 @@ def print_label(
         **kwargs,
     ) -> None:
         """Send the label to the printer."""
+        print_delay = int(machine.get_setting('DELAY', 'D', backup_value=0) or 0)
```
wolflu05: Is that int cast and or 0 really necessary? I mean in theory it should also convert it to an int because that's the type of the setting?

SchrodingersGat:
@wolflu05 thanks for the feedback, I have simplified it a lot now

PR #10900, opened 2025-11-24
[Docker] Allow gunicorn webserver bind address to be adjusted

kquinsland:
Very minor fix, discussed here: https://github.com/inventree/InvenTree/discussions/10893#discussioncomment-15055253

I could make an argument for a new small section (`Web Server Settings`) next to the `Email Settings` on `docs/docs/start/config.md` but figured I'd ask, first.

### DIFF
--- a/contrib/container/Dockerfile
+++ b/contrib/container/Dockerfile
@@ -149,7 +149,7 @@ COPY --from=builder_stage ${INVENTREE_BACKEND_DIR}/InvenTree/web/static/web ${IN
 COPY --from=builder_stage /root/.local /root/.local
 
 # Launch the production server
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@kquinsland thanks for the fix here. I can confirm that this still works with the default values set.

If you want to add some documentation about how to configure the docker image for IPv6 support, that would be great!

kquinsland01:
> If you want to add some documentation about how to configure the docker image for IPv6 support, that would be great!

Can do! I'll try to get that PR'd before mid-week. I have not spent the time to dig through the app code to understand if the psql and redis/valkey clients even "know" to use IPv6 so at least for now, the scope of docs is "WebServer" only.

PR #10913, opened 2025-11-26
[UI] refactor "inactive alerts" panel

SchrodingersGat:
Adjusts panel which displays a list of alerts which are *not* active.

@matmair unless you have any strong objections to this? I had not been looking at it but I have received multiple questions about whether the admin needs to worry about these "inactive" alerts.

### DIFF
--- a/src/frontend/src/components/nav/Alerts.tsx
+++ b/src/frontend/src/components/nav/Alerts.tsx
@@ -119,35 +119,35 @@ export function getAlerts(
       key: 'debug',
       title: t`Debug Mode`,
       code: 'INVE-W4',
-      message: t`The server is running in production mode.`,
+      message: t`The server is running in debug mode.`,
       condition: server?.debug_mode || false
     },
     {
       key: 'worker',
       title: t`Background Worker`,
       code: 'INVE-W5',
-      message: t`The background worker process is running.`,
+      message: t`The background worker process is not running.`,
       condition: !server?.worker_running
     },
     {
       key: 'restart',
       title: t`Server Restart`,
       code: 'INVE-W6',
-      message: t`Server restart not required.`,
+      message: t`The server requires a restart to apply changes.`,
       condition: globalSettings.isSet('SERVER_RESTART_REQUIRED')
     },
     {
       key: 'email',
       title: t`Email settings`,
       code: 'INVE-W7',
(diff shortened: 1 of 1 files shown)


matmair:
Maybe we can improve the phrasing but in my opinion this panel is important so quickly check which things are infact configured correctly;
I think removing this would be a mistake

SchrodingersGat:
@matmair what about this?

image

matmair (approved):

wolflu05:
Why don't we add a system status page to the admin center and then add a list of all alerts there. Then we can remove the terminology of active and inactive and add a green check mark before all satisfied checks and a alert triangle before warnings and a red x before all Critical warnings. And then we only show the warnings and errors as a popup and add a link to the all checks page.

Because inactive implies to me that this check is currently not performed because it is deactivated, but from your explanation I have understood that they actually should mean that the "check" is satisfied.

Also having them all in one column under each other is much more readable then having to check in a grid.

SchrodingersGat:
@wolflu05 that's a much better idea. Here's the new look:

image

matmair:
So we are duplicating the amount of strings we need to translate per status just doubles?

SchrodingersGat:
image

@matmair simplified, no need for duplicate translations now.

matmair:
Not sure if this is more helpful than it was before, the messaging is now seeming more confusing - at least to me.

src/frontend/src/components/nav/Alerts.tsx
```diff
@@ -112,45 +114,45 @@ export function getAlerts(
...
     Number.parseInt(globalSettings.getSetting('_PENDING_MIGRATIONS')) ?? 0;
 
-  const allalerts: ExtendedAlertInfo[] = [
+  const allAlerts: ExtendedAlertInfo[] = [
     {
       key: 'debug',
       title: t`Debug Mode`,
       code: 'INVE-W4',
-      message: t`The server is running in debug mode.`,
+      message: t`The server is running in production mode.`,
```
matmair: These changes lead to the actually triggering alerts being very confusing - is this intended?

SchrodingersGat: Ah, yes good point. I've reverted the messages to their original values now

Issue #10908, opened 2025-11-25
Bulk stock adjustment does not ignore lines with quantity = 0


78kfb:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

According to the intended logic of stock adjustments, in bulk Add / Remove / Move operations
any lines with quantity = 0 should be ignored (no changes applied for these lines).

However, in the bulk stock operation form, lines with quantity = 0 are still processed and
the stock item quantity is set to 0. After that, the result depends on the "Delete on deplete"
setting:

- If "Delete on deplete" is enabled, the stock item is deleted from the location
  (it looks like the item disappears from the warehouse).
- If "Delete on deplete" is disabled, the stock item remains, but its quantity becomes 0.

### Steps to Reproduce

Create several parts (2 or more) and place them on a location with a positive quantity (e.g. 10 units each).

Open the bulk stock operation dialog (Add / Remove / Move – not stocktake/count).

Add multiple lines for these parts/this location:

Line 1: set a positive quantity (e.g. 5).

Line 2: leave quantity = 0 (or explicitly enter 0).

Apply the operation.

### Expected behaviour

In bulk Add / Remove / Move operations, any lines where quantity = 0
  should be completely ignored (no change to stock quantity, no deletion),
  in line with the original StockAdjust logic where 0 means "do nothing"
  for these operation types.

This behaviour is dangerous in practice: a simple mistake (leaving 0 in the form)
  can lead to items disappearing from the UI or being silently zeroed out.

### Deployment Method

None

### Version Information

- InvenTree version: 1.1.5 or older
- Installation method: Installer, docker
- Database backend: default, no chages
- Browser: (Chrome / Firefox / etc.)

InvenTree-Version: 1.1.5
Django Version: 4.2.26
Commit Hash: c09fb48
Commit Date: 2025-11-22
Commit Branch: null
Database: django.db.backends.postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.14.0-36-generic-x86_64-with-glibc2.41
Installer: DOC
[...]

PR #10914, opened 2025-11-26
[bug] Stock adjust

SchrodingersGat:
- Fixes UI bug which caused incorrect stock adjustments if initial form value was not adjusted
- Specify explicit default value when initializing form data
- Adds additional unit testing to prevent regresssion
- Fixes https://github.com/inventree/InvenTree/issues/10908#issuecomment-3577868025


SchrodingersGat:
python test is broken, not associated with this fix

Issue #10806, opened 2025-11-11
Django Warning Message


SchrodingersGat:
After update to django 5 (https://github.com/inventree/InvenTree/pull/10730) we now see the following warning output in the logs:

```
/home/inventree/dev/venv/lib/python3.11/site-packages/django/db/backends/utils.py:98: RuntimeWarning: Accessing the database during app initialization is discouraged. To fix this warning, avoid executing queries in AppConfig.ready() or when your app modules are imported.
  warnings.warn(self.APPS_NOT_READY_WARNING_MSG, category=RuntimeWarning)
```

I believe this is because we are hitting the database to check the state of the loaded plugins, during app loading. Django specifically warns against this now.

@matmair any thoughts on how we can better handle this?

matmair:
I will look into it

SchrodingersGat:
@matmair any thoughts on this one? It is difficult as there are multiple places where we hook into the `ready` function(s) for each app. Any requests to the database in these methods will trigger the warning

matmair:
It is difficult, I am not really seeing an official way to handle this.

PR #10938, opened 2025-12-01
App ready warning

SchrodingersGat:
- Adds function decorator to optionally suppress django warnings (in certain cases)
- Closes https://github.com/inventree/InvenTree/issues/10806

### DIFF
--- a/src/backend/InvenTree/InvenTree/api_version.py
+++ b/src/backend/InvenTree/InvenTree/api_version.py
@@ -1,10 +1,14 @@
 """InvenTree API version information."""
 
 # InvenTree API version
-INVENTREE_API_VERSION = 428
+INVENTREE_API_VERSION = 429
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
+
+v429 -> 2025-12-04 : https://github.com/inventree/InvenTree/pull/10938
+    - Adjust default values for currency codes in the API schema
+    - Note that this does not change any functional behavior, only the schema documentation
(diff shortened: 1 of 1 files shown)


src/backend/InvenTree/InvenTree/helpers.py
```diff
@@ -1170,3 +1171,21 @@ def plugins_info(*args, **kwargs):
...
+    Ref: https://github.com/inventree/InvenTree/issues/10806
+    """
+
+    def wrapper(*args, **kwargs):
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                'ignore',
+                message='Accessing the database during app initialization is discouraged',
+                category=RuntimeWarning,
+            )
```
matmair: Are we sure this warning is not valid? When I looked into it, it seemed like we could be causing real issues with the way we use the `ready` functions.

SchrodingersGat: Yes, it is a "valid" warning, however I do not see a way to cleanly avoid this warning entirely. We need to load the plugin registry, which definitely hits the database, and can even cause an entire reload of the apps structure.

So, the approach here is to provide a decorator around functions that we know will throw this warning, but we have determined are necessary to run in `ready` code - and cannot be placed elsewhere.

SchrodingersGat: Note that the `currency_codes` function has been refactored to not hit the database until after apps have been loaded. This is obviously preferable wherever possible

matmair (approved):

PR #10952, opened 2025-12-03
Fixed typo in shebang interpreter directive

ttftw:


SchrodingersGat:
@ttftw thanks for the fix!

PR #10954, opened 2025-12-04
Update version checks:

SchrodingersGat:
- Add error code for old python version
- Fix min python version in docs
- Various spelling fixes in docs

### DIFF
--- a/src/backend/InvenTree/InvenTree/version.py
+++ b/src/backend/InvenTree/InvenTree/version.py
@@ -128,8 +128,8 @@ def inventreeDocsVersion() -> str:
     """
     if isInvenTreeDevelopmentVersion():
         return 'latest'
-    else:
-        INVENTREE_SW_VERSION
+
+    return INVENTREE_SW_VERSION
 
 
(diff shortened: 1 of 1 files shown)


src/backend/InvenTree/InvenTree/version.py
```diff
@@ -110,39 +115,40 @@ def inventreeVersionTuple(version=None):
...
     """Return the version string matching the latest documentation.
 
     Development -> "latest"
     Release -> "major.minor.sub" e.g. "0.5.2"
     """
     if isInvenTreeDevelopmentVersion():
         return 'latest'
-    return INVENTREE_SW_VERSION  # pragma: no cover
+    else:
+        return 'stable'
```
matmair: @SchrodingersGat are you sure this will work as intended?
If someone installs 1.0 and we release bit, the docs will point at eg. 1.2 instead of 1.0

SchrodingersGat: You're right, I'll revert that

PR #10966, opened 2025-12-06
Update email docs

SchrodingersGat:
Tweak docs on email setup


SchrodingersGat:
I've opted to improve the docs instead, I think you're right and that's a better approach

PR #10983, opened 2025-12-08
feat: add on-demand price calculation

matmair:
This re-introduces the option to calculate the price for a line item


SchrodingersGat:
@matmair is this going to be implemented soon or should we push off to the next release cycle?

matmair:
TBH i think the quick fix i pushed that enables inline calc should be enough

matmair:
This was addressed enough with https://github.com/inventree/InvenTree/pull/11296

Issue #10996, opened 2025-12-11
[db] taggit field overhead


SchrodingersGat:
Ref: https://github.com/afkiwers/inventree_kicad/issues/137

While trying to understand what was causing the kicad plugin API endpoints to be slow, I discovered that the [django-taggit](https://django-taggit.readthedocs.io/en/latest/) "tags" field that we use introduces a significant overhead to the API calls.

Profiling an API call with a large number of parts (> 1000) shows that most of the time is spent in the `tags_for` method:

https://github.com/jazzband/django-taggit/blob/4fd99fdee21f8d1a087523a091cc5c3fbaee2fdd/taggit/models.py#L200

And, this function gets called for *every* item in the queryset - *even if you prefetch the field*.

With some simple debugging:

```python
    @classmethod
    def tags_for(cls, model, instance=None, **extra_filters):
        tag_relname = cls.tag_relname()
        model = model._meta.concrete_model
        kwargs = {
            "%s__content_type__app_label" % tag_relname: model._meta.app_label,
            "%s__content_type__model" % tag_relname: model._meta.model_name,
        }
        if instance is not None:
            kwargs["%s__object_id" % tag_relname] = instance.pk
        if extra_filters:
            kwargs.update(extra_filters)

        print("tags_for:", kwargs)  # Debug statement

        return cls.tag_model().objects.filter(**kwargs).distinct()
```

And then:

```
http://localhost:8000/api/part/?format=json&limit=500
```

```
tags_for: {'taggit_taggeditem_items__content_type__app_label': 'part', 'taggit_taggeditem_items__content_type__model': 'part', 'taggit_taggeditem_items__object_id': 50}
tags_for: {'taggit_taggeditem_items__content_type__app_label': 'part', 'taggit_taggeditem_items__content_type__model': 'part', 'taggit_taggeditem_items__object_id': 51}
tags_for: {'taggit_taggeditem_items__content_type__app_label': 'part', 'taggit_taggeditem_items__content_type__model': 'part', 'taggit_taggeditem_items__object_id': 54}
tags_for: {'taggit_taggeditem_items__content_type__app_label': 'part', 'taggit_taggeditem_items__content_type__model': 'part',
...
...
(x 500)
```

This issue is apparent for the KiCad plugin - but also for any API endpoint on the core InvenTree codebase which points to a model with a `TaggableManager`. The manager enforces this behaviour in the `get_queryset` method:

[...]

PR #10997, opened 2025-12-11
Add helper funcs for debugging

SchrodingersGat:
- Ref: https://github.com/inventree/InvenTree/issues/10996
- Adds some helper functions for common profiling tasks
- Not to be used in production code


SchrodingersGat:
@matmair I've opted to raise an exception if the file is imported outside of debug mode

PR #11005, opened 2025-12-13
Rearrange python package installs in are metal setup 

SergeoLacruz:
the python bindings must be installed before before invoke install. Otherwise it fails with 

```
Installing plugin packages from '/home/michael/src/config/plugins.txt'
Unknown command: 'collectplugins'
Type 'manage.py help' for usage.
ERROR: InvenTree command failed: 'python3 manage.py collectplugins'
- Refer to the error messages in the log above for more information
```
At least on Ubuntu 24

### DIFF
--- a/docs/docs/start/install.md
+++ b/docs/docs/start/install.md
@@ -131,21 +131,6 @@ If all packages have been installed run:
 invoke install
 ```
 
-#### Install Python Bindings
-
-Depending on your database the python bindings must also be installed (into your virtual environment). 
-For PostgreSQL install:
-
-```
-pip3 install psycopg pgcli
-```
-
-For MySQL install:
-
-```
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@SergeoLacruz thanks for the addition

CapnKernel:
This gave me a conflict when I merged it into the install instructions I'm working on.

> the python bindings must be installed before before invoke install

My solution is to run `invoke install --skip-plugins` instead of `invoke install`.  That way, the db isn't touched, and db setup instructions can be done a bit later in the instructions, as they were before.

I hope we can avoid duplication of effort!

(Draft version of my revised instructions here)

https://github.com/CapnKernel/InvenTree/blob/feature/bare-metal-install-improvements/docs/docs/start/install.md

PR #11012, opened 2025-12-14
[refactor] Optional prefetch

SchrodingersGat:
## Problem Description

We have multiple places in the code where we call `enable_filter` - to dynamically add (or remove) nested related fields from an API serializer.

To prevent **1 + N** database fetches, we must call `prefetch_related` on all fields which are to be included - even if they are *not* included. So, potentially we are prefetching querysets which never even get used.

This is a problem as there is a non-zero cost associated with prefetching - see https://github.com/inventree/InvenTree/issues/10996

Ideally, we should only prefetch related fields when they are actually *needed* - that way we do not pay the price of prefetching when it is not needed

## Solution

This PR adds the option to specify a list of prefetch fields to the `enable_filter` hook. 

- If the field is added to the queryset, the required fields will be prefetched
- If the field is *not* added, then the prefetches do not fire

This PR introduces the new functionality, but does not refactor all instances, to reduce scope.

## Performance Comparison

- Comparing between "master" and this PR
- Testing with and without parameter annotations
- Tested via external API calls
- Averaged across 5 runs
- Times are in milliseconds

| API URL | master | PR | 
| --- | --- | --- |
| `/api/part/?limit=500` | 1521  | 1454 |
| `/api/part/?limit=500&parameters=0` | 1608  | 1434  |
| `/api/part/?limit=500&parameters=1` | 1816  |  1861  |
| `/api/company/?limit=500` | 1000  |  251  |
| `/api/company/?limit=500&parameters=0` | 932 |  251  |
| `/api/company/?limit=500&parameters=1` | 1230 |  534  |

## Follow-up PRs

- [ ] Make "tags" field filterable - and prefetch accordingly
- [ ] Refactor all instances of `enable_filter` calls
- [ ] Performance profiling for all API endpoints

### DIFF
--- a/src/backend/InvenTree/build/serializers.py
+++ b/src/backend/InvenTree/build/serializers.py
@@ -1241,9 +1241,7 @@ class BuildItemSerializer(
             'stock_item__tags',
             'stock_item__part',
             'stock_item__supplier_part',
-            'stock_item__supplier',
-            'stock_item__manufacturer_part',
-            'stock_item__manufacturer_part__manufacturer',
+            'stock_item__supplier_part__manufacturer_part',
         ],
     )
 


src/backend/InvenTree/common/filters.py
```diff
@@ -310,3 +313,82 @@ def order_by_parameter(
...
+            allow_null=True,
+        ),
+        default,
+        filter_name=filter_name,
+        prefetch_fields=['project_code'],
+    )
+
+
+def enable_parameters_filter(
+    source: Optional[str] = None, filter_name: str = 'parameters', default: bool = False
```
matmair: Do we need these options or can we just build the serializer once and return a reference here?

src/backend/InvenTree/common/filters.py
```diff
@@ -310,3 +313,82 @@ def order_by_parameter(
...
+            source='project_code', many=False, read_only=True, allow_null=True
+        ),
+        default,
+        filter_name=filter_name,
+        prefetch_fields=['project_code'],
+    )
+
+
+def enable_project_label_filter(
+    filter_name: str = 'project_code_detail', default: bool = True
```
matmair: Do we need the options?

src/backend/InvenTree/common/filters.py
```diff
@@ -310,3 +313,82 @@ def order_by_parameter(
         f'{prefix}parameter_value_numeric',
         f'{prefix}parameter_value',
     )
+
+
+def enable_project_code_filter(
+    filter_name: str = 'project_code_detail', default: bool = True
```
matmair: Do we need the options here?

src/backend/InvenTree/company/serializers.py
```diff
@@ -399,12 +391,7 @@ def __init__(self, *args, **kwargs):
         # Check if 'available' quantity was supplied
         self.has_available_quantity = 'available' in kwargs.get('data', {})
 
-        # TODO INVE-T1 support complex filters
         brief = kwargs.pop('brief', False)
-        detail_default = not brief
-        part_detail = kwargs.pop('part_detail', detail_default)
-        supplier_detail = kwargs.pop('supplier_detail', detail_default)
-        manufacturer_detail = kwargs.pop('manufacturer_detail', detail_default)
```
matmair, line 407: This is a breaking change in the API, is that intended?

src/backend/InvenTree/company/serializers.py
```diff
@@ -413,16 +400,6 @@ def __init__(self, *args, **kwargs):
...
-
-        if supplier_detail is not True:
-            self.fields.pop('supplier_detail', None)
-
-        if manufacturer_detail is not True:
-            self.fields.pop('manufacturer_detail', None)
-            self.fields.pop('manufacturer_part_detail', None)
-
         if brief or prettify is not True:
             self.fields.pop('pretty_name', None)
```
matmair: If we remove the brief function for the other details we could use the default `enable_filter` for `pretty_name` too

matmair (commented):
Looks like a good improvement, using common filters is great;

I am not sure if we even need the provided flexibility for the common filters; did you have something specific in mind when providing it? Having these things static might promote a more uniform API

src/backend/InvenTree/common/filters.py
SchrodingersGat: I'll simplify it, thanks

src/backend/InvenTree/common/filters.py
SchrodingersGat: Good call, I will simplify

src/backend/InvenTree/common/filters.py
SchrodingersGat: Good call, I will simplify

src/backend/InvenTree/company/serializers.py
SchrodingersGat: The default behaviour should be the same, but the API schema will need to be updated as these are now filterable

src/backend/InvenTree/company/test_api.py
```diff
@@ -682,6 +710,42 @@ def test_active(self):
         for result in response.data:
             self.assertEqual(result['supplier'], company.pk)
 
+    def test_filterable_fields(self):
+        """Test inclusion/exclusion of optional API fields."""
+        fields = {
```
matmair, line 715: @SchrodingersGat you could also use the `run_output_test` helper for this

matmair (approved):

src/backend/InvenTree/InvenTree/serializers.py
```diff
@@ -113,6 +124,34 @@ def __init__(self, *args, **kwargs):
...
+        # Gather up the set of simple 'prefetch' fields and functions
+        prefetch_fields = set()
+
+        filterable_fields = [
+            field
+            for field in self.fields.values()
+            if getattr(field, 'is_filterable', None)
+        ]
+
+        for field in filterable_fields:
```
matmair, line 145: The serializer already contains `filter_targets`, that gets build by the `gather_filters` function. Could we use that here instead of iterating over all fields again?

Issue #11002, opened 2025-12-13
[dev] API Query Profiling / Improvements


SchrodingersGat:
## Overview

While investigating various profiling options for https://github.com/inventree/InvenTree/issues/10996 I have found that there are some places where we need to seriously consider the (in)efficiencies of our API. 

Many of our API endpoints (although some more than others) need a lot of optimization (to reduce the total query response time). There are a number of shortcomings of our current design which I have identified. Some will be easy(ish) to fix, some may require some larger architectural considerations.

## Related Issues

- https://github.com/inventree/InvenTree/issues/10996

## Related Pull Requests

- https://github.com/inventree/InvenTree/pull/11006
- https://github.com/inventree/InvenTree/pull/11007
- https://github.com/inventree/InvenTree/pull/11009
- https://github.com/inventree/InvenTree/pull/11010
- https://github.com/inventree/InvenTree/pull/11012
- https://github.com/inventree/InvenTree/pull/11021
- https://github.com/inventree/InvenTree/pull/11023
- https://github.com/inventree/InvenTree/pull/11073

### Part API

Consider the following API request: `http://localhost:8000/api/part/?format=json&limit=500`

This has a round-trip time of ~900ms (and increases with higher limit numbers).

The biggest culprit here is our immense queryset aggregation:

<img width="987" height="680" alt="Image" src="https://github.com/user-attachments/assets/d5a59f6a-e7ce-4a6e-b9e4-ae3059229e55" />

Which expands to the SQL in the attached file: [part_query.sql](https://github.com/user-attachments/files/24138785/part_query.sql)

There are a number of other identified bottlenecks in this (and other) API endpoints.

## Problem Scope

There are a number of things we need to address:

### Complex Annotations

We have many database queries which have complex annotations - this is to avoid the `1+N` problem, which has historically been the cause of extremely large query times (> 30 seconds).

However the annotations (especially related to "order requirements" on the `Part` model) are now so complex that just running the generated SQL is the largest part of the API request.

We need to address these, somehow - although I'm not sure how we do this without either:

- Re-introducing `1+N` issues
- Losing existing functionality in the API

### Annotate Before Filter

There are a number of points in the code where we annotate *before* filtering - which means that the database calls are slow because they annotate across all rows in the table - even ones which will be later removed via filtering.

### Table Indexes

There are a number of places where we should add custom indexes to the tables to aid in lookup. We do not currently have any custom indexes and there are a lot of cross-table references that are used very often.

### taggit issues

[...]

PR #11017, opened 2025-12-15
feat(backend): add performance tests

matmair:
Adds pytest based benchmarking tests that submit results to codspeed.io, which seems to have good usage in the python tooling ecosystem (astral, pydantic)  

See #11002

### DIFF
--- a/src/backend/InvenTree/part/test_api.py
+++ b/src/backend/InvenTree/part/test_api.py
@@ -3357,7 +3357,7 @@ class PartApiPerformanceTest(PartAPITestBase, InvenTreeAPIPerformanceTestCase):
 
     @pytest.mark.django_db
     @pytest.mark.benchmark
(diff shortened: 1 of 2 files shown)


matmair:
@SchrodingersGat I have just requested to add codspeed to the InvenTree/InvenTree repo; I do not think we need it on more repos for now

matmair:
@SchrodingersGat seems to work now; which endpoints should we cover?

SchrodingersGat:
I'd suggest we start with the `/part/` and `/stock/` API endpoints, they are likely the "heaviest" due to largest number of items

matmair:
No detailed traces for now, running 3.14 seems to be an issue

matmair:
@SchrodingersGat ready for review and merge; for now it is a bit limited but getting 3.14 running should provide a lot of advantages in the near future regarding insights.

PR #11021, opened 2025-12-16
[API] Tags filters

SchrodingersGat:
Refactor "tags" field

- Off by default
- Only prefetch when requested (expensive)
- Ref: https://github.com/inventree/InvenTree/pull/11012
- Ref: https://github.com/inventree/InvenTree/issues/11002
- Closes https://github.com/inventree/InvenTree/issues/10996

### DIFF
--- a/src/backend/InvenTree/InvenTree/serializers.py
+++ b/src/backend/InvenTree/InvenTree/serializers.py
@@ -213,8 +213,9 @@ class FilterableSerializerMixin:
 
         # Skip filtering for a write requests - all fields should be present for data creation
         if request := self.context.get('request', None):
-            if str(request.method).lower() in ['post', 'put', 'patch']:
-                return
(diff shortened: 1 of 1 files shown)


matmair (approved):

src/backend/InvenTree/part/test_api.py
```diff
@@ -1735,10 +1732,6 @@ def test_part_operations(self):
         # 2021-06-22 this test is to check that the "duplicate part" checks don't do strange things
         response = self.patch(url, {'name': 'a new better name'})
 
-        # Try to remove a tag
```
matmair, line 1738: Can we leave that test in to ensure tagging still works,

SchrodingersGat: Yep, I've added it back in now - it required a tweak to ensure that all fields are available on data creation (even if they are default OFF)

PR #11024, opened 2025-12-16
[API] Metadata performance fix

SchrodingersGat:
The DRF `OPTIONS` endpoint is expensive to compute - and rarely changes for a given model type.

We use the `OPTIONS` endpoint extensively in the user interface, to construct forms, build out tables, etc. Any speed improvement we can extract here is great.

This PR implements a caching strategy to significantly improve UI responsiveness - as each table sends an OPTIONS request whenever it is loaded.

### New Features

- `get_api_url` is expensive to calculate due to the call to `reverse` - can take a few hundred milliseconds in some cases
- Cache the computed URL value for a given model type improves request speed
- For most cases the entire `OPTIONS` call can be cached
- Cache is based on user, URL, query parameters
- Prevent auto-prefetching of related field data for an `OPTIONS` request (big speed up!)

### DIFF
--- a/src/backend/InvenTree/InvenTree/serializers.py
+++ b/src/backend/InvenTree/InvenTree/serializers.py
@@ -134,9 +134,10 @@ class FilterableSerializerMixin:
             The modified queryset with prefetching applied.
         """
         if request := getattr(self, 'request', None):
-            if request.method.lower() == 'options':
-                # Don't prefetch for an OPTIONS request
-                return queryset
+            if method := request.method:
+                if method.lower() == 'options':
+                    # Don't prefetch for an OPTIONS request
+                    return queryset
 
             if getattr(request, '_dummy_metadata_request', False):
                 # Don't prefetch for a metadata request


wolflu05:
> I like the idea; at least the machine APIs are also dynamic, so they would need to be excluded

What exactly do you mean here?

matmair:
@wolflu05 at least from what I remember some of the fields returned differ depending on the machine driver; am I misremembering?

wolflu05:
Not really sure, I cannot remember that. Do you have an example?

SchrodingersGat:
On further inspection this will likely not work - any translated strings in the metadata (strongly required) will be rendered in their base language (english) due to where the caching operation takes place in the rendering chain.

I do not think it will be easy to bypass this, without being overly complex and risky.

I'll close this out for now, might revisit at some point as there is some significant performance overhead to be gained here.

Typical numbers during testing were ~150ms for non-cached, ~50ms for cached.

PR #11030, opened 2025-12-16
deps(backend): bump allauth

matmair:
Bumps allauth to address some provider specific shortcomings.

https://github.com/inventree/InvenTree/security/dependabot/277


SchrodingersGat:
Should we add a unit test to ensure that a token associated with an inactive user is rejected?

matmair:
We already have tests for that, this only effected allauth APIs in our software as we have a custom auth middleware that checks these things for the core api.

PR #11035, opened 2025-12-17
refactor(backend): reduce API surface by unifying metadata endpoints

matmair:
Replaces individual metadata endpoints with a generic endpoint and a bunch of permanent redirects. Obviously a breaking change.
This:
- reduces the amount of duplicated endpoints doing the same things and the required tests too
- removes the necessity to declare new endpoints for every metadata-supporting model
- is plugin friendly as models defined in those can also use the new endpoint

Requires a change to the client libary - https://github.com/inventree/inventree-python/pull/285

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -116,7 +116,7 @@ jobs:
           update: true
       - name: Check types
         run: |
-          ty check --python ${Python_ROOT_DIR}/bin/python3
+          ty check --python ${Python_ROOT_DIR}/bin/python3 src/backend
 
   mkdocs:
     name: Style [Documentation]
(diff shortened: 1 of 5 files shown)


matmair:
@SchrodingersGat this is ready for review, the needed change for the API client is already proposed https://github.com/inventree/inventree-python/pull/285

You can see the passing test with that branch in https://github.com/inventree/InvenTree/actions/runs/20733404036/job/59525935718

SchrodingersGat:
@matmair this does look a lot cleaner.

To confirm what you are doing here - all metadata is accessed via `/api/metadata/pk/` - but the existing URLs all continue to work via redirect?

matmair:
@SchrodingersGat it sends a permanent redirect. If the client follows the redirect it still works. Most clients do that for GET but not for POST/PATCH, it is often configurable. That is largely why I marked the PR as breaking.

matmair:
Metadata is accessed via `/api/metadata/<model>/<pk>/`

SchrodingersGat:
Happy with this to merge?

matmair:
Yes

PR #11044, opened 2025-12-19
Improvements to the documentation instructions

CapnKernel:
Some refinements to the README for the docs.

PR #11045, opened 2025-12-19
Move docs/README.md into docs/docs/ so mkdocs can include it

CapnKernel:
In discussions about previous changes I'd made to the README.md for docs, we noted that the page wasn't currently reachable from the main docs page, and perhaps we could do that.  Well here it is.

With these changes, the `README.md` file previously in the top level docs dir has been moved one level deeper to `docs/docs/documentation-howto.md`.  I've moved it deeper so mkdocs can see it.  (Is there another way that doesn't involve moving it down?)   Also, we can't call the moved file `docs/docs/README.md`, because mkdocs says it clashes with `docs/docs/index.md`.

There are three commits.

1. Already sent in #11044.
2. Move `docs/README.md` to `docs/docs/documentation-howto.md` with no changes.
3. Changed examples to stop mkdocs interpreting `{{ image() }}` and `{{ icon() }}`, and added `documentation-howto.md` to `mkdocs.yaml`.

I suggest reviewing them as commits, rather than the end-to-end changes.


SchrodingersGat:
@CapnKernel please address merge conflicts here.

At the top of this page, the wording needs some tweaking now that it is included into the documentation itself:

image

Maybe instead of "this repository hosts..." it should be something like "This page describes..."

SchrodingersGat:
@CapnKernel are you still interested in this? Merge conflicts and changes needed

matmair:
@SchrodingersGat what do we want to do about this? do we have interest in merging this?

PR #11048, opened 2025-12-19
If a tag of "auto" is given, frontend-download gets the version from installed src

CapnKernel:
This will make downloading the matching front-end files easier when following the bare metal install instructions.


wolflu05:
Not sure if the inventree version from the Python version file is enough to determine the correct frontend version. This version is the same for several commits on master until a version bump occurs. What do you mean @SchrodingersGat ?

CapnKernel:
Thanks for your comment @wolflu05 

Context: My [rewritten instructions](https://inventree--11049.org.readthedocs.build/en/11049/start/install/#download-front-end-files) say this:

## Download Front End Files

A front end app runs in the user's browser page.  Let's download the files for the app:
```
mkdir -p ~/src/src/backend/InvenTree/web/static
invoke frontend-download --tag=auto
invoke static
```

Let me check my thinking: Backend code is associated with a particular version, and there are releases of the compiled front end code on github.  Even if there have been several commits to a particular version on the back end, as long as the version doesn't change, the front end files should still work.

I don't understand anything deeper than this.  Clarifying comments welcome!

matmair:
Have you tested this before submitting your PR? Try running this on a dev build like https://packager.io/gh/inventree/InvenTree/builds/4155

PR #11059, opened 2025-12-22
feat(backend): unify dependencies in pyproject

matmair:
This changes the management flow to completely depend on uv / modern python tooling standards and removes the individual listings for one central one.
Makes SBOM management easier and enables QOL things like environment syncing.
The tooling is a bit fast this way too as more of the resolution can be cached, but it is not really noticeable on modern ARM chips.


matmair:
this is out of scope because of current deploy methods

Issue #10740, opened 2025-11-02
INV-E1 error after upgrading from 1.0.1 to 1.1.0 with docker


christianvw:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

I tried upgrading my docker compose setup of InvenTree from 1.0.1 to 1.1.0 but ended up with the following on the web ui:

> If you see this text there might be an issue with your update.
> See [INVE-E1](https://docs.inventree.org/en/stable/settings/error_codes/#inve-e1) in the docs for help.

The logs from the container when it applied the migrations of the update looked fine (applying migrations...). Just one line was yellow when it was in maintenance mode and then forced to get out of it.

My `docker-compose.yml`:

```lml
---
services:
  prod-invtree-app:
    container_name: prod-invtree-app
    image: inventree/inventree:1.0.1
    volumes:
      - ${INVTREE_HOME}/data:/home/inventree/data
      - ./bg.jpg:/home/inventree/data/static/img/inventree_splash.jpg
    restart: unless-stopped
    expose:
      - '8000'
    environment:
      INVENTREE_SITE_URL: "https://inv.example.org"
      INVENTREE_USE_X_FORWARDED_HOST: 'True'
      INVENTREE_USE_X_FORWARDED_PROTO: 'True'
      INVENTREE_USE_X_FORWARDED_PORT: 'True'
      INVENTREE_ALLOWED_HOSTS: "inv.example.org"
      INVENTREE_TRUSTED_ORIGINS: "https://inv.example.org"
      INVENTREE_AUTO_UPDATE: "False"
      INVENTREE_PLUGINS_ENABLED: "True"
      INVENTREE_ADMIN_USER: "${ADMIN_USER_USERNAME}"
      INVENTREE_ADMIN_PASSWORD: "${ADMIN_USER_PASSWORD}"
      INVENTREE_ADMIN_EMAIL: "it@example.org"
      INVENTREE_DB_ENGINE: "postgresql"
      INVENTREE_DB_NAME: "invtree"
      INVENTREE_DB_HOST: "prod-invtree-postgres"
      INVENTREE_DB_PORT: "5432"
      INVENTREE_DB_USER: "${DB_USER}"
      INVENTREE_DB_PASSWORD: "${DB_PASSWORD}"
      INVENTREE_CACHE_ENABLED: "True"
      INVENTREE_CACHE_HOST: "prod-invtree-redis"
[...]

PR #11067, opened 2025-12-26
feat(backend): better for logging to detect issues with static files discovery

matmair:
This adds more information to the frontend directory to debug issues reported in #10740 ; it is not a fix for it but should help find it

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -392,7 +392,7 @@ jobs:
 
   performance:
     name: Tests - Performance
-    runs-on: codspeed-macro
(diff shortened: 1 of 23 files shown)


src/backend/InvenTree/InvenTree/version.py
```diff
@@ -22,6 +24,7 @@
 
 logger = logging.getLogger('inventree')
 
+warning_txt = 'INVE-W3: Could not detect git information.'
```
SchrodingersGat: Maybe this constant could be a little more descriptive e.g. `git_warning_text`

matmair: done

PR #11073, opened 2026-01-01
[API] Filter refactor

SchrodingersGat:
This PR is a major refactor of the `enable_filter` functionality which allows optional fields to be added (or removed) dynamically to API serializers.

### Problem Description

The *old* code was written in such a way that the optional serializer class was instantiated on definition.

Consider the following code

```python
    category_detail = enable_filter(
        CategorySerializer(
            source='category', many=False, read_only=True, allow_null=True
        ),
        prefetch_fields=['category'],
    )
```

The `CategorySerializer` class is instantiated (which is quite expensive) - and then this is also chained down for any multi-level child serializers.

In many cases, the fields are later removed, so instantiating them early is a waste of resources.

Also, due to the (wasteful?) way that DRF deep-copies all the fields (multiple times throughout the lifespan of the serializer) this can be very prohibitive.

Some particularly bad API endpoints which had deep nested serializers, saw hundreds of thousands of serializer objects created, and then later deleted.

### Solution

The PR introduces an `OptionalField` approach, which lazily evaluates the optional fields only after we have decided that they should definitely be included in the serializer:

```python
    category_detail = OptionalField(
        serializer_class=CategorySerializer,
        serializer_kwargs={
            'source': 'category',
            'many': False,
            'read_only': True,
            'allow_null': True,
        },
        prefetch_fields=['category'],
    )
```

### Justification

The new API endpoints are significantly faster, we have been introducing a huge amount of wasted overhead (for years now in the codebase) due to the unnecessary serializer evaluation

## Benchmarks

Benchmarking shows that serializers which are deeply nested have the most benefit from this PR. Both GET and OPTIONS requests are improved substantially by deferring serializer instantiation.

### Methodology

- Perform external requests via python API bindings
- Average across 100 requests per test
- All times specified in ms

### GET

 - limit=50
 - offset=0

| URL | Master | PR |
| --- | --- | --- |
| /api/part/ | 151 | 89|
| /api/part/category/ | 43 | 43 |
| /api/stock/ | 173 | 74 |
| /api/stock/location/ | 64 | 42 |
| /api/company/ | 47 | 38 |
| /api/build/ | 86 | 47 |
| /api/build/line/ | 646 | 351 |
| /api/build/item/ | 276 | 41 |
| /api/order/so/ | 75 | 50 |
| /api/order/so/shipment/ | 90 | 79 |
| /api/order/po/ | 89 | 50 |
| /api/order/po-line/ | 187 | 62 |
| /api/user/roles/ | 32 | 37 |
| /api/parameter/ | 42 | 36 |
| /api/parameter/template/ | 34 | 33 |

### OPTIONS

| URL | Master | PR |
| --- | --- | --- |
| /api/part/ | 85 | 71 |
| /api/part/category/ | 43 | 33 |
| /api/stock/location/ | 42 | 37 |
| /api/company/ | 43 | 36 |
| /api/build/ | 72 | 61 |
| /api/build/line/ | 266 | 44 |
| /api/build/item/ | 210 | 31 |
| /api/order/so/ | 67 | 60 |
| /api/order/so/shipment/ | 70 | 50 |
| /api/order/po/ | 66 | 57 |
| /api/order/po-line/ | 185 | 60 |
| /api/user/roles/ | 28 | 30 |
| /api/parameter/ | 36 | 35 |
| /api/parameter/template/ | 30 | 33 |

### Search

| URL | Master | PR |
| --- | --- | --- |
| `/api/search/` | 381 | 294 |


matmair:
I might be overlooking something, but are we not completely loosing static introspection with this change? I am not sure if there is much actual runtime performance to be gained by this, have you benchmarked it?

src/backend/InvenTree/InvenTree/serializers.py
```diff
@@ -57,60 +58,32 @@ def __init__(self, *args, **kwargs):
...
-        filter_by_query (bool): If True, also look for filter parameters in the request query parameters.
-        prefetch_fields (list of str, optional): List of related fields to prefetch when this field is included. This can be used to optimize database queries.
+@dataclass
+class OptionalField:
+    """DataClass used to optionally enable a serializer field.
 
-    Returns:
-        The decorated serializer field, marked as filterable.
+    This is used in conjunction with the `FilterableSerializerMixin` to allow
+    dynamic inclusion or exclusion of serializer fields at runtime.
```
matmair, line 42: This docstrings seem very slim compared to the replaced function

src/backend/InvenTree/InvenTree/serializers.py
```diff
@@ -120,135 +93,178 @@ def __init__(self, *args, **kwargs):
...
-            # Optionally also look in query parameters
-            # Note that we only do this for a top-level serializer, to avoid issues with nested serializers
+        # Skip filtering for a write request - all fields should be present for data creation
+        if method := getattr(self.request, 'method', None):
             if (
-                is_top_level
-                and val is None
-                and self.filter_on_query
-                and v.get('filter_by_query', True)
+                str(method).lower() in ['post', 'put', 'patch']
```
matmair: this could be a negative check against `permissions.SAFE_METHODS`

SchrodingersGat:
> I am not sure if there is much actual runtime performance to be gained by this, have you benchmarked it?

Benchmarks added to the top comment.

> I might be overlooking something, but are we not completely loosing static introspection with this change?

We should not lose any introspection as the fields are still created - they are just defered until we determine they are actually needed. This is most important in the case of deeply nested serializer fields which will never be exposed to the final serializer tree.

The `OPTIONS` endpoints still show all the optional fields, and I believe the API docs should be the same. LMK if you spot any items which are no longer included.

src/backend/InvenTree/InvenTree/api_version.py
```diff
@@ -1,11 +1,14 @@
...
 
 # InvenTree API version
-INVENTREE_API_VERSION = 435
+INVENTREE_API_VERSION = 436
 """Increment this API version number whenever there is a significant change to the API that any clients need to know about."""
 
 INVENTREE_API_TEXT = """
 
+v436 -> 2026-01-03 : https://github.com/inventree/InvenTree/pull/11073
+    - Add OptionalField class for cleaner handling of optional fields in serializers
```
matmair, line 10: There seem to be other API changes too, should they also be mentioned?

src/backend/InvenTree/InvenTree/mixins.py
```diff
@@ -238,7 +238,11 @@ def __init__(self) -> None:
             )
 
     def get_serializer(self, *args, **kwargs):
-        """Return serializer instance with output options applied."""
+        """Return serializer instance with output options applied.
+
+        Note that the serializer is cached after first evaluation,
+        as 'get_serializer' may be called multiple times per request.
```
matmair: Is this still true?

matmair:
> We should not lose any introspection as the fields are still created - they are just defered until we determine they are actually needed. This is most important in the case of deeply nested serializer fields which will never be exposed to the final serializer tree.

At least in my setups the python LSP seems to not work with this new setup. The kwargs are not validated anymore during startup and you can not rename fields on the serializer automatically because it is not inspectable throught the LSP server anymore. Type checking also seems to be lost on the kwargs.

src/backend/InvenTree/InvenTree/mixins.py
SchrodingersGat: Good catch, have removed it now.

src/backend/InvenTree/InvenTree/serializers.py
SchrodingersGat: Yes, I will add more documentation here

src/backend/InvenTree/InvenTree/serializers.py
SchrodingersGat: Done

SchrodingersGat:
> At least in my setups the python LSP seems to not work with this new setup. The kwargs are not validated anymore during startup and you can not rename fields on the serializer automatically because it is not inspectable throught the LSP server anymore. Type checking also seems to be lost on the kwargs.

can you please provide some specific examples of differences, perhaps we can add some additional unit tests? The speed improvements make the UX much snappier, so I'm keen to move this forward but want to make sure you are still happy with it.

SchrodingersGat:
As an interesting observation - the speed "penalty" which we are incurring by instantiating optional fields and then later removing them has been in the code base from the beginning, it was not introduced by the `enable_filter` refactor.

matmair:
> can you please provide some specific examples of differences, perhaps we can add some additional unit tests?

I am not sure if we really should test against any specific python LSP implementation; this is a fundamental design decision / issue that causes this - not a specific LSP implementation detail

Due to only passing a formless dict into this new function, we lose any type inference from the serializer class. If you insist, I can paste a screenshot from VS Code showing no refactor options etc. but showing something not existing is not adding much context.



> you are still happy with it.

This diverges from how Django Rest Framework works even more than we already do, I am not sure anyone outside of core devs will grasp how this works in a few weeks. It breaks a lot of development UX and makes the request/response cycle even more difficult to understand. I am not sure it is worth it



> As an interesting observation - the speed "penalty" which we are incurring by instantiating optional fields and then later removing them has been in the code base from the beginning, it was not introduced by the `enable_filter` refactor.

A serializer being initialized is part of the basic design of DRF. I purposefully did not touch it when I added enable_filter.


If we really want to change the fundamental behavior of DRF we could refactor the whole mechanism similar to https://www.django-rest-framework.org/api-guide/serializers/#dynamically-modifying-fields or https://github.com/khaledsukkar2/drf-shapeless-serializers

SchrodingersGat:
I think that this implementation is pretty similar to the concepts in `drf-shapeless-serializers` (which I had not known of before). 

> It breaks a lot of development UX and makes the request/response cycle even more difficult to understand.

I do not think that this is any more difficult to understand (from a developer perspective) than the `enable_filter` approach. The `enable_filter` approach was a great refactor of our existing "add-then-remove" approach for nested serializers which we have had for years. However it was not until I started digging into [API profiling](https://github.com/inventree/InvenTree/issues/11002) that I discovered how much overhead this approach can add.

> I am not sure it is worth it

Deeply nested serializers add a lot of overhead when responding to an API call. Consider building a table for `/api/build/item` in the user interface.

- First we issue an `OPTIONS` request to learn the structure of the table data
- Then we fetch the actual data with a `GET` request

In current master, these two requests are > 200ms each, which means that this table in particular feels very slow. Additionally, it reduces the number of requests being processed overall, which means that the UI is less snappy in general.

After this optimization, we are down to ~40ms + ~30ms in comparative testing. That's ~70ms vs ~450ms to load the table. 

This is a significant speed up (that is noticable in the UI) which I would like to retain. Maybe there's a more "DRF-ey" approach?

matmair:
While I understand the possible performance improvements, it should be at least noted that this makes it considerable harder to contribute. Even seasoned Django devs will not know this pattern and it will be in most important API endpoints.

One can not jump through the arguments with IntelliSense anymore, one loses type checking on the arguments on the serializers. In my eyes, that is a hard hit as it places more burden on the devs to know what they are doing.

I still was not able to reproduce the benchmarked improvements that drastically in my setups; are you using a live environment or one of the test datasets?

SchrodingersGat:
> are you using a live environment or one of the test datasets?

This is with (in both test cases)

- devcontainer development setup
- DEBUG=True

And also reproduced with both the django built-in server and gunicorn

SchrodingersGat:
@matmair any idea why the codespeed performance checks do not have any baseline data from master to compare against?

matmair:
No, looking into it now

SchrodingersGat:
I have also found [drf-flex-fields](https://github.com/rsinger86/drf-flex-fields), similar in approach to [drf-shapeless-serializeres](https://github.com/khaledsukkar2/drf-shapeless-serializers).

However from my reading of both the nested fields are not available as class attributes until after the top level serializer class is instantiated - meaning that any type hinting / introspection would also not be available?

Unless you have any experience with these libraries that shows working type hinting?

SchrodingersGat:
> No, looking into it now

image

SchrodingersGat:
@codespeed-hq run

PR #11080, opened 2026-01-05
feat (backend): Add more performance tests

matmair:
Using the full django setup for performance testing is expensive, doing it via the python API is much faster as seen in https://github.com/inventree/InvenTree/pull/11073#issue-3774414851 so the listed tests from that issue are added here for future CI tests.

Ref #11002


matmair:
Note: with walltime we would get syscalls included https://github.com/inventree/InvenTree/pull/11081 and make times more reflective of end user experience

SchrodingersGat:
Let's see what we can learn from this!

PR #11081, opened 2026-01-05
refactor(backend): use walltime for performance testing

matmair:
This PR switches the profiling to use [walltime](https://codspeed.io/docs/instruments/walltime) with bare metal runners by codspeed.

This should provide a lot more insights into underlying performance issue causes as it provides detailed flamegraphs and diffs across runs.


matmair:
This is blocked by a missing setting on the inventree GitHub org right now

matmair:
@SchrodingersGat ready for review and merge, this makes the times measured much more similar to timings you measured in recent PRs.

SchrodingersGat:
Certainly some room for improvement.on a lot of Jose endpoints!

PR #11092, opened 2026-01-07
Fix Schema Nullables

1337joe:
I finally caught back up to recent schema changes. Here's a small batch of added `allow_null` annotations for nullable `read_only` fields.

I also removed the default value of '' because that's not a valid selection from the choices list (which was resulting in code generated from the schema failing to compile). As far as I've spotted the changes to parameter/parameter template don't result in changes to the UI.

### DIFF
--- a/src/backend/InvenTree/build/models.py
+++ b/src/backend/InvenTree/build/models.py
@@ -1743,11 +1743,10 @@ class BuildItem(InvenTree.models.InvenTreeMetadataModel):
 
     def save(self, *args, **kwargs):
         """Custom save method for the BuildItem model."""
-        self.clean()
-
+        self.clean(raise_error=False)
         super().save()
 
-    def clean(self):
+    def clean(self, raise_error: bool = True):
         """Check validity of this BuildItem instance.
 
         The following checks are performed:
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
> WARN: diff exceeds the 1MB limit, truncating output...

The API diff comparison is failing here

1337joe:
> > WARN: diff exceeds the 1MB limit, truncating output...
> 
> The API diff comparison is failing here

No idea why: the entire diff fits on one screen for me when I run it locally. I just rebased to make sure I'm up to date with master and I'll look into it more if that fails again.

matmair:
the diff seams small on my local instance as well; looking into it

1337joe:
I found [a tool](https://www.oasdiff.com/diff-calculator) that gives a browser-viewable version of the generated change log and it is indeed massive (in part because changes to parameters makes it report changes to everywhere parameters are available). I should have time this afternoon to figure out if this is valid or if I need to split my changes to make the diff easier to review.

1337joe:
Long story short: changing the schema for parameters (such as allowing `updated_by_detail` to be nullable) results in reported changes for every api that has a detail field that includes parameters.

I rebased to pick up the new api_version changes and reduced the scope of my changes to try to fit within the tool for easier review.

PR #11095, opened 2026-01-07
ci(gc): add missing oidc permissions

fargito:
Hello, CodSpeed team member here.

New performance tests were added in #11080. However, contrary to 'Tests - Performance' job, OIDC is not used to authenticate the new job with CodSpeed. We don't currently gracefully handle the case where one job is using OIDC and the other one is not.

More details on how CodSpeed uses OIDC to authenticate workflows [here](https://codspeed.io/docs/integrations/ci/github-actions/configuration\#oidc-recommended).

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -286,6 +286,9 @@ jobs:
 
     needs: ["pre-commit", "paths-filter"]
     if: needs.paths-filter.outputs.server == 'true' || needs.paths-filter.outputs.force == 'true'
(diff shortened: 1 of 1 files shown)


matmair (approved):

matmair:
Thank you for the fix @fargito, the error was not really clear on what the issue was.
@SchrodingersGat IMO this does not need a full CI run as it only changes permission values and should solve https://github.com/inventree/InvenTree/pull/11073#issuecomment-3718227894

PR #11099, opened 2026-01-08
Switch away from codspeed-macro to ubuntu-24.04

matmair:
The current usage limits are not feasible for our CI/CD model

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -392,7 +392,7 @@ jobs:
 
   performance:
     name: Tests - Performance
-    runs-on: codspeed-macro
+    runs-on: ubuntu-24.04 # codspeed-macro
 
     needs: ["pre-commit", "paths-filter"]
     if: needs.paths-filter.outputs.server == 'true' || needs.paths-filter.outputs.force == 'true'


adriencaccia:
Hey @matmair, co-founder of @CodSpeedHQ here.

Just FYI, using `ubuntu-latest` will most likely lead to highest variance on walltime benchmarks. If after some time you see that the results get too flaky, we can [provide additional free macro runners since InvenTree is an open-source project](https://codspeed.io/docs/features/macro-runners#additional-usage).

Have fun using CodSpeed!

matmair:
@adriencaccia thank you for offering.

Just to be clear on the kind of usage we have: Last month we had at least 2,186 minutes usage on the faster of the 2 checks. At least from a quick test it seems like the GitHub runners are faster than the codspeed-macro runners with our current setup.
Based on simple extrapolation, we would have at least 7.000 minutes of usage per month in a typical month. Are you sure you want/can provide that kind of resources?

adriencaccia:
> @adriencaccia thank you for offering.
> 
> Just to be clear on the kind of usage we have: Last month we had at least 2,186 minutes usage on the faster of the 2 checks. At least from a quick test it seems like the GitHub runners are faster than the codspeed-macro runners with our current setup. Based on simple extrapolation, we would have at least 7.000 minutes of usage per month in a typical month. Are you sure you want/can provide that kind of resources?

This is completely feasible yes. Just let us know by mail if/when you want to use them, and we will continue discussion there 😉

Issue #10977, opened 2025-12-07
[FR] Docker - create volume for inventree-cache (redis)


B-z-F:
### Please verify that this feature request has NOT been suggested before.

- [x] I checked and didn't find a similar feature request

### Problem statement

Each start of inventree creates temporary volume with *dump.rdb*.

### Suggested solution

The volumes declaration for inventree-cache (redis) is missing in *docker-compose.yml*. Redis is enabled by default (INVENTREE_CACHE_ENABLED=True in *.env*), even though *docker-compose.yml* says otherwise (redis as the cache manager (optional, disabled by default)). */data* from redis should be in volume.

### Describe alternatives you've considered

.

### Do you want to develop this?

- [ ] I want to develop this.

SchrodingersGat:
@B-z-F good catch - do you want to submit a PR to add this?

PR #11100, opened 2026-01-08
Save redis data in docker volume to prevent creating new volume on each container start

B-z-F:
Fixes inventree/InvenTree#10977


SchrodingersGat:
@B-z-F thanks for the contribution :)

PR #11104, opened 2026-01-09
feat(backend): Add dedicated health endpoint

matmair:
This can be used in monitoring etc. and is less expensive than the /api/ route. No auth, only an indicator.
OK if the system loaded and (if they are enabled) plugins are loaded


matmair:
@SchrodingersGat should we add this to the docker image as a healthcheck cmd with curl?

SchrodingersGat:
> @SchrodingersGat should we add this to the docker image as a healthcheck cmd with curl?

That seems sensible.

We should also add some unit tests to this PR for the new endpoint

SchrodingersGat (approved):

Issue #10911, opened 2025-11-25
SSO Authentication fails with 401


vsviridov:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

I've configured SSO using Zitadel/Open ID Connect using the config file, and now when coming back from the IDP i get redirected to the login page with a 401 error on `/api/auth/v1/auth/session` endpoint. There's no logs even at "TRACE" level and I can't figure out at which stage it fails.

### Steps to Reproduce

Add openid configuration to the `config.yaml`
```yaml
      social_backends:
        - 'allauth.socialaccount.providers.openid_connect'
      social_providers:
        openid_connect:
          APPS:
            - provider_id: Zitadel
              name: Zitadel
              client_id: '-----'
              settings:
                fetch_userinfo: false
                oauth_pkce_enabled: true
                server_url: 'https://auth.------.com'   
```
Enabled relevant SSO Flags

<img width="1629" height="430" alt="Image" src="https://github.com/user-attachments/assets/316d5cb7-7b43-4715-a240-fc26abb1bd64" />

### Expected behaviour

Should be authenticated

### Deployment Method

None

### Version Information

InvenTree-Version: 1.1.5
Django Version: 4.2.26
Commit Hash: c09fb48
Commit Date: 2025-11-22
Commit Branch: null
Database: django.db.backends.postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.17.2-1-pve-x86_64-with-glibc2.41
Installer: DOC
[...]

PR #11112, opened 2026-01-09
Fix OIDC logins not displaying

bunkmate5127:
Only openid is installed into django, but oidc is openid_connect. With any luck this should solve #10911 (though I have not tested/confirmed).


matmair:
The extra you reference does not exsist, pelase do not waste our time with untested PRs.

PR #11120, opened 2026-01-11
chore(ci): Re-enable codspeed runner

matmair:
Reverses #11099 as we are getting much more resources by codspeed so we should not run into issues anymore


SchrodingersGat:
Any idea why the performance has degraded so much on this change?

matmair:
@SchrodingersGat in my testing the codspeed runners seem to be less performant than the machines by GitHub.

SchrodingersGat:
@matmair some conflicts have snuck in here

matmair:
@SchrodingersGat conflicts resolved

Issue #8909, opened 2025-01-17
StockItems are missing further information for differentiation in "Allocate Stock" form


simonkuehling:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

When allocating StockItems to a Sales Order and I have multiple StockItems of the same part (and worse case with the same quantity), there is currently no way to distinguish between them.

Examples:
- In my case I have a part for tracking service work through build orders (engineering hours, or 3D printing service for example) and use the batch code of the build outputs to identify for which sales order they are. Without seeing the batch code information in the allocation-drop-down, i cannot tell which one to select for this sales order
- In another case, I have multiple StockItems of the same part, again with the same quantity - one of them has already been moved to the packaging table for shipment. In the drop-down I cannot tell which StockItem to allocate though, because I am missing the location information.
- When stocking items with a best-before date, I want to be able to ensure a first-in-first-out principle to always ship the oldest stock first. The expiration date is not visible in the drop-down right now.

Screenshot:

![Image](https://github.com/user-attachments/assets/7066c348-5d76-4f13-9aa0-a6ea50f9b59f)

### Steps to Reproduce

1. Open a Sales Order > Line Items > Allocate Stock
2. Open the ```Stock Item``` drop-down

### Expected behaviour

have more information about available stock items in the allocation-drop-down, including:
- location
- batch code
- expiry date

(Another solution that could potentially work as a work-around is to openly show a unique identifier for all Stock Items, like a ```Stock Item ID``` - similar to the IPN for parts - to confidently identify stock items and tell multiple apart from each other.)

### Deployment Method

- [x] Docker
- [ ] Package
- [ ] Bare metal
- [ ] Other - added info in Steps to Reproduce

### Version Information

InvenTree-Version: 0.18.0 dev
Django Version: 4.2.17
Commit Hash: c7e9607
Commit Date: 2025-01-15
Commit Branch: null
Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.1.0-rpi7-rpi-v8-aarch64-with
Installer: DOC
[...]

PR #11137, opened 2026-01-14
[ui] Show stock item ID in `RenderStockItem`

wlcx:
Show the stock item ID in when rendering stock items. 

This is primarily so that stock items can be positively identified when allocating stock to a build order: currently two stock items in the same location with the same quantities are rendered identically, which easily leads to stock being allocated from the wrong stock item. This issue is also addressed with the new "barcode in forms" functionality in 1.2.0.

It's only really useful if stock items are being tracked individually, with barcoded labels, but this is how inventree encourages stock to be managed (and is the main thing causing us pain in our deployment) so I figured it was probably generally useful to just always show. Let me know if you think it should be behind a config option though.

Here's how it looks in the allocate stock dialog:
image


matmair:
Not sure if this makes sense as the default; growing to many options is also not good;

will have to think about this before giving an option; thx for the contribution

wlcx:
Yeah, understood on both points. I don't know if there's a simpler option to achieve this though, and from having used InvenTree quite extensively in production now it's the main shortcoming in an otherwise very smooth flow from PO -> stock -> BO.

Otherwise (forthcoming barcode scanning in forms aside) all other options rely on unwieldy changes to workflow (splitting stock items and assigning them to a dedicated location before assigning or having separately identifiable locations for every different stock item)

wlcx:
I think #8909 describes the same thing but focussing on batch code - but this is of course not always present or applicable for all parts/stock items.

SchrodingersGat:
Some menu entries render with a "mouse over" element - we could add the ID here instead?

The ID/PK I would consider "clutter" - however there are certainly times when you might need to know it. So a [HoverCard element](https://mantine.dev/core/hover-card/) with the ID value (as well as any other "additional items")?

matmair:
@SchrodingersGat not sure if a hover element is ergonomic enough;

## Future note; not ready for dev / adoption through this PR

Should we make the highlighted values configurable from the backend?

Maybe with a structured response in the options call so that frontend code is not made more and more complicated. Could mash together well Profiles as described in WIP [ADR-0002](https://github.com/inventree/project-adr/blob/1964c162c7d3590287ebe82229c30ffc4714c416/docs/adr/0002-permissions.md#profiles) to provide a lot of configuration without blowing up the code base too much.

matmair:
@wlcx please add a setting for this for now (default off) so we can merge this

SchrodingersGat:
@wlcx please adjust as requested :)

wolflu05:
Maybe we can implement a generic approach here so that users can e.g. enable "advanced mode" and see the pk for all related model selects

SchrodingersGat:
Closing as lack of response - generic implementation incoming in https://github.com/inventree/InvenTree/pull/12038

PR #11147, opened 2026-01-16
[ui] Show generated serial numbers as placeholder when receving stock

### LINKED ISSUE #11146: [FR] Make serial number generation opt-in when receiving stock from a PO

### Problem statement

When receiving trackable stock items from a purchase order, serial numbers are generated and populate the serial number field by default (when the "assign batch code and serial numbers" button is clicked"). This is frequently not what you want if you are recieving serialized stock from other manufacturers, as there's no guarantee serials will be consecutive, and can be a source of errors. `*`

This differs from the behavior when adding a stock item, where the generated serial is shown as a suggestion, with an "accept suggestion" button being present at the far right of the field.

`*` I think there's a broader topic here of "internally serialized" parts (i.e. the serial numbers are allocated by "us" at the point of stock item creation) vs "externally serialized" parts (i.e. the serial numbers are allocated by someone else already). The serial number generation logic is only really useful for the former. Perhaps this should be another issue?

### Suggested solution

Update the UI to use the "accept suggestion" UX as used in the stock item create dialog, which make the UI more consistent overall and helps reduce unintentional errors.

### Examples of other systems

_No response_

- [x] I want to develop this.

wlcx:
Update the "receive stock" dialog on the purchase orders page to show generated serial numbers for trackable parts as a placeholder that can be accepted, rather than a concrete value. This is in line with the behavior in the "add stock item" dialog.

I've tested this locally and it behaves as I'd expect, also when receiving multiple stock items at once. But I'm not that knowledgable about react so please do check to see if there's some edge case I've missed!

Looks like this:
image

Resolves #11146

### DIFF
--- a/src/frontend/src/forms/PurchaseOrderForms.tsx
+++ b/src/frontend/src/forms/PurchaseOrderForms.tsx
@@ -361,12 +361,7 @@ function LineItemFormRow({
 
   // Serial number generator
   const serialNumberGenerator = useSerialNumberGenerator({
-    isEnabled: () => batchOpen && trackable,
-    onGenerate: (value: any) => {
-      if (value) {
-        props.changeFn(props.idx, 'serial_numbers', value);
-      }
-    }
+    isEnabled: () => batchOpen && trackable
   });
 
   const [packagingOpen, packagingHandlers] = useDisclosure(false, {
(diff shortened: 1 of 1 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@wlcx great addition, thanks!

Regarding the style issues - you can setup your dev environment to auto-format the code on commit by running:

```bash
invoke setup-dev
```

wlcx:
Yep sorry, I don't like having stuff run on commit but of course that necessitates you to remember to run it manually 😅 

Looks to be passing now.

SchrodingersGat:
@wlcx thanks for the contribution :)

PR #11164, opened 2026-01-19
fix(auth): Use correct data for SSO group sync

p-fruck:
https://docs.allauth.org/en/latest/release-notes/recent.html#id18
Since django-allauth 65.11.0 (InvenTree v1.1.X), the extra_data is divided into separate userinfo and id_token objects.

I just upgraded to Inventree v1.1.8 from v1.0.X yesterday, hence I didn't notice this for a longer time. I would love to see a hotfix release since group sync is broken for the entire v1.1.X release line if I am not mistaken. Broken means that new groups are not synchronized and if group removal is disabled, all groups are deleted from the user on each signin.

### DIFF
--- a/src/backend/InvenTree/InvenTree/sso.py
+++ b/src/backend/InvenTree/InvenTree/sso.py
@@ -96,7 +96,9 @@ def ensure_sso_groups(sender, sociallogin: SocialLogin, **kwargs):
     group_map = json.loads(get_global_setting('SSO_GROUP_MAP'))
     # map SSO groups to InvenTree groups
     group_names = []
-    for sso_group in sociallogin.account.extra_data.get(group_key, []):
+    for sso_group in sociallogin.account.extra_data.get('userinfo', {}).get(
+        group_key, []
+    ):
         if mapped_name := group_map.get(sso_group):
             group_names.append(mapped_name)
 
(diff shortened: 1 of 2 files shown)


p-fruck:
Also, fixed an old SSO-related typo in 14303ea

matmair:
@p-fruck thank
You for the contribution! Please remove the translation file changes, we handle those in a separate process.

do you think it would be possible to check this in ci so it does not break again? Last time I looked into doing that I could not find an easy way

p-fruck:
@matmair I removed the translation commit again. I also thought about this. Last time I worked with testing oidc, I utilized the following project: https://github.com/Soluto/oidc-server-mock

We could create a github actions pipeline that uses the oidc-server-mock as service container and perform authentication against this container. This would be a full e2e test. However, I don't know how to configure the sso provider. Until now, I always used manual configuration via django admin ui. I could write some python code to add the config using the django ORM...

I think its doable but will take some time 

Are you aware of a simpler approach?

matmair:
I am not aware of a simpler approach; let’s keep this PR focused on fixing the issue and getting a release issued and do a quick follow up with the testing PR; this might be time to add a proper API/UI for application setup

SchrodingersGat (approved):

p-fruck:
@matmair I'd love to see a declarative way of configuring the SSO stuff :)

Thank you for the quick review and release, faster than any commercial support :rocket:

matmair:
Declarative configuration should already work via config file / Environment variables with the `INVENTREE_SOCIAL_PROVIDERS` variable.
API access can probably be addressed while doing #4715

PR #11179, opened 2026-01-21
Stock Tracking - Add Old Status to Deltas

jacobfelknor:
I'm often interested in finding stock status transitions. For example, I'd like to generate a report of stock items who transitioned from a custom "Inspection" state to the "OK" state within a given time period.

Rather than attempt a non-straightforward way of determining this transition from a set of tracking events, simply add an `old_status` field to the `deltas`. Now, I can just do something like

```python
yesterday = datetime.today() - timedelta(days=1)

inspection_custom_status = common.models.InvenTreeCustomUserStateModel.objects.get(
    name="INSPECTION", reference_status="StockStatus"
)

StockItemTracking.objects.filter(
    date__date=yesterday,
    deltas__old_status=inspection_custom_status.key,
    deltas__status=StockStatus.OK.value,
)
```

Additionally, match the stock tracking behavior between custom state changes in the "Edit" form and the "Update Status" form. Before, "Edit" would show the logical key's status pill in the tracking history, but "Update Status" showed the custom status pill. We want the "Edit" form to also show the custom status pill


src/backend/InvenTree/stock/models.py
```diff
@@ -800,7 +800,17 @@ def save(self, *args, **kwargs):
 
                 # Status changed?
                 if old.status != self.status:
-                    deltas['status'] = self.status
+                    # Custom status changed?
+                    # Matches custom status tracking behavior of StockChangeStatusSerializer
+                    if old.get_custom_status() != self.get_custom_status():
```
matmair: Q: can we put the 4 calls to `get_custom_status` into 2 vars at the begining to make it easier to read?

src/backend/InvenTree/stock/serializers.py
```diff
@@ -463,12 +463,16 @@ def update(self, instance, validated_data):
...
-            if not instance.compare_status(status_code):
-                instance.set_status(status_code)
-                instance.save()
+            # avoid a second .save() call and perform both status updates at once (to support `old_status` in tracking event)
+            # by setting the values in validated_data as computed by set_status()
+            instance.set_status(status_code)
+            validated_data['status'] = instance.status
+            validated_data['status_custom_key'] = (
+                status_code  # for compatibility with custom "leader/follower" concept in super().update()
+            )
```
matmair, line 473: C: not sure if this 100% safe; I worry if there are any other interaction on update() afterwards it will be hard to diagnose

src/backend/InvenTree/stock/test_api.py
```diff
@@ -1796,6 +1801,35 @@ def test_set_status(self):
             tracking = item.tracking_info.last()
             self.assertEqual(tracking.tracking_type, StockHistoryCode.EDITED.value)
 
+        # Test tracking with custom status
+        # *from* standard *to* custom
+        INSPECT_CUSTOM_STATUS = common.models.InvenTreeCustomUserStateModel.objects.get(
```
matmair: Q: please make this a separate test function within this class to ensure easier distinction in the performance tracking

matmair (commented):
LGTM overall; fallbacks should also work this way
what I am unclear about: There is no UI change as mentioned in the PR - did I miss that or is that to follow?

src/backend/InvenTree/stock/serializers.py
jacobfelknor: To be honest, I'm not sure either. I got (perhaps too) creative to avoid the second `.save()` call, because it appeared to me during debugging that the change to `status` and to `status_custom_key` happened in two different contexts, leaving me unable to properly document in the tracking event what the old status was.

There may be a better way of doing this, a fresh perspective may see it and I'd gladly take input.

jacobfelknor:
> LGTM overall; fallbacks should also work this way what I am unclear about: There is no UI change as mentioned in the PR - did I miss that or is that to follow?

~I did not include any UI changes. I don't feel strongly about having in the UI since I have a plugin generating a report for me. I can see the utility in adding it though if you'd like~

@matmair  My statement from before was misleading. While I didn't make any UI edits, I _did_ change what you saw in the UI in a particular scenario.

Before, when using the Edit form to modify the status to a custom status, its corresponding logical key was show in the tracking history. For example, if a `Inspection` custom status pointed to `Quarantined`, it was `Quarantined` that was shown in the tracking. This was in contrast to when you used the "Change Status" form, where the custom status (in my example, `Inspection`) was shown in the history.

I much prefer seeing the custom status in the tracking history, so modified how the tracking entry was created from the Edit form to use the custom status when it exists, so that the tracking history UI reflects this custom status.

Does that make sense? 

What I _didn't_ add was a representation of the `deltas['old_status']` in the UI. The UI entry still only shows the status it was updated to. This is the case that I claim not to feel strongly about in my original statement

src/backend/InvenTree/stock/test_api.py
jacobfelknor: resolved in b597fcdf912f640be1e915d54b458a204defa324

src/backend/InvenTree/stock/models.py
jacobfelknor: resolved in e1da45bde90fa820780d82ce135fd56cbe36274c

jacobfelknor:
Failing test is because I added a `InvenTreeCustomUserStateModel` to the fixtures, but `CustomStockItemStatusTest.setup()` creates some itself and doesn't expect any to already exist.

I'll move the custom state out of the fixture and create in in `StockItemTest.setup()` to match the current pattern

matmair (approved):
LGTM with that explanation;
maybe tracking the old logical state might be wise too if a custom status is removed.

SchrodingersGat (approved):

SchrodingersGat:
@jacobfelknor do you want to display this information in the UI too, or just an API change?

jacobfelknor:
> @jacobfelknor do you want to display this information in the UI too, or just an API change?

Yeah, I'm convinced. Initially I thought it appeared too distracting in the table, but I asked around and it sounds like my users would like to see it.  I've added this

> LGTM with that explanation; maybe tracking the old logical state might be wise too if a custom status is removed.

This got me thinking.... I went down a bit of a rabbit hole here. I decided to add `status_logical` and `old_status_logical` to the deltas, so that lets me fallback to to the logical state if the custom state is removed. I added a `fallbackStatus` arg to `StatusRenderer` to accomplish this. I do like this behavior. I noticed though, that this wasn't the behavior on the Stock Detail page, which would still show a generic "number" status in this case, so I decided to add a `fallBackStatus` to each instance of a `StatusRenderer` for a stock item that I could find. 

Of course, this isn't unique to stock items, and custom statuses should probably fall back to their logical status for display on other models too. This started to feel like too big of a change to include in this PR, so I stopped at only stock item cases I found.

Curious your thoughts. I'm inclined to leave the rest of the model types to a different PR.

Also okay with reverting my fallback commit (6f906383c6ac53732aba4d1681cf15fa969547f3) all together if I started to feature creep too much. It feels like a nice-to-have, but I don't necessarily need it.

src/backend/InvenTree/stock/serializers.py
```diff
@@ -1072,9 +1076,20 @@ def save(self):
...
+            if item.get_custom_status():
+                deltas['old_status'] = item.get_custom_status()
+            else:
+                deltas['old_status'] = item.status
+
             item.set_status(status, custom_values=custom_status_codes)
             item.save(add_note=False)
 
+            # after save, can track new status_logical
+            deltas['status_logical'] = item.status
```
matmair, line 1091: Is this actually saving this way?

matmair (approved):
LGTM; I like the fallback mechanism;

in theory the name of the fallback field is directly related to the name of the status field; not sure if doing that deduction automatically would make usage easier; for now I think this is good enough

src/backend/InvenTree/stock/serializers.py
matmair: a short search did not find a better way; still feels somewhat iffy but with full test coverage it should be fine

src/backend/InvenTree/stock/serializers.py
jacobfelknor: Yes, because the `deltas` are included in the `StockItemTracking` bulk create at the end, independent of the stockitem save event

jacobfelknor: Although, I am modifying a dictionary in a loop and passing that to the `StockItemTracking` instance. If the instance isn't making a "copy" of the data passed, I could suffer from a shared reference bug here.

I'm not able to look at this immediately, but that is a valid concern. Using a fresh `deltas` instance for each loop iteration would be safer

jacobfelknor:
Let's hold off on this for a moment. I'm now realizing that for this to be complete, I need to add `old_status` to the deltas of other places you can change status, like in `StockItem.move()`, `StockItem.stocktake()`, etc

I'll also address my concern in https://github.com/inventree/InvenTree/pull/11179#discussion_r2718166035

src/backend/InvenTree/stock/serializers.py
jacobfelknor: fixed in c62395b0ee28926c784652e9439030703b77a837

jacobfelknor:
> Let's hold off on this for a moment. I'm now realizing that for this to be complete, I need to add `old_status` to the deltas of other places you can change status, like in `StockItem.move()`, `StockItem.stocktake()`, etc

addressed in 384af7d33bf060f2a2636f6e665819e831e10c1e

matmair:
This looks good; I have merged in current master to see if enough coverage is there (whole mechanism should be covered and tested with assertations)

SchrodingersGat (approved):

SchrodingersGat:
@jacobfelknor thanks for the contribution :)

SchrodingersGat:
@jacobfelknor see spelling error fixed in https://github.com/inventree/InvenTree/pull/11230

jacobfelknor:
Good catch... Thanks

PR #11190, opened 2026-01-22
fix (backend): finer grained registration control for Single Sign On

### LINKED ISSUE #10281: Cannot disable user registration

### Describe the bug*

In the InvenTree admin settings, we have disabled registration:

Image

For the traditional Django UI, the registration form is successfully disabled:

Image

However, the new platform UI still allows registration:

Image

As the registration form does work, this is not only a visual frontend issue, but also an issue on the backend.

We do have SSO and SSO Registration enabled, which might play a role in this issue

### Steps to Reproduce

1. Disable registration
2. Enable SSO
3. Enable SSO registration
4. Log out
5. Go to new platform UI
6. Use the registration form

### Expected behaviour

Registration should not be possible.

### Deployment Method

Docker

### Version Information

# Version Information:
InvenTree-Version: 0.17.4
Django Version: 4.2.17
Commit Hash: de431b8
Commit Date: 2025-01-28

Database: postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.1.0-38-amd64-x86_64-with
Installer: DOC

Active plugins: False

I did not try to reproduce

Not reproducible

### Relevant log output

```shell

```

### LINKED ISSUE #11162: Bugs with Single Sign On

### Describe the bug*

Two things I noticed when setting up SSO:
1. When SSO-self registration is enabled, regular account registration is also enabled (#10281)
2. "Group on signup" is not getting applied to users created via SSO

### Deployment Method

Helm chart

### Version Information

Version 1.1.8 (and also 1.1.7).

```
inventree=# select * from common_inventreesetting where key like 'LOGIN_%';
 id  |              key              | value 
-----+-------------------------------+-------
  20 | LOGIN_ENABLE_REG              | False
  21 | LOGIN_ENABLE_PWD_FORGOT       | False
  64 | LOGIN_ENABLE_SSO              | True
  19 | LOGIN_ENABLE_SSO_REG          | True
  39 | LOGIN_ENFORCE_MFA             | False
  41 | LOGIN_MAIL_REQUIRED           | False
  45 | LOGIN_SIGNUP_MAIL_RESTRICTION | 
  47 | LOGIN_SIGNUP_MAIL_TWICE       | False
  79 | LOGIN_SIGNUP_SSO_AUTO         | True
  83 | LOGIN_SIGNUP_PWD_TWICE        | True
 142 | LOGIN_ENABLE_SSO_GROUP_SYNC   | True
(11 rows)

inventree=# select * from common_inventreesetting where key like 'SIGNUP_%';
 id |     key      | value 
----+--------------+-------
 93 | SIGNUP_GROUP | 1
(1 row)
```

matmair:
This corrects the behavior to match more closely what admins expect.
Fixes #11162
Fixes #10281

### DIFF
--- a/src/backend/InvenTree/InvenTree/auth_overrides.py
+++ b/src/backend/InvenTree/InvenTree/auth_overrides.py
@@ -86,7 +86,7 @@ class CustomSignupForm(SignupForm):
         return cleaned_data
 
 
-type RegistrationKeys = Literal['LOGIN_ENABLE_REG', 'LOGIN_ENABLE_SSO_REG']
+RegistrationKeys = Literal['LOGIN_ENABLE_REG', 'LOGIN_ENABLE_SSO_REG']
 
 
 def registration_enabled(setting_name: RegistrationKeys = 'LOGIN_ENABLE_REG'):
--- a/src/backend/InvenTree/InvenTree/auth_overrides.py
+++ b/src/backend/InvenTree/InvenTree/auth_overrides.py
@@ -1,5 +1,7 @@
 """Overrides for allauth and adjacent packages to enforce InvenTree specific auth settings and restirctions."""
 
+from typing import Literal
+
 from django import forms
 from django.conf import settings
 from django.contrib.auth.models import Group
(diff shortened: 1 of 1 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@matmair nice, glad we finally have this sorted

matmair:
let us wait a few days for user feedback before we celebrate; this login things seem to be always lurking somewhere

PR #11208, opened 2026-01-27
feat(frontend): disable_theme_storage

matmair:
Adds option to disable theme / widget / language loading; this should help the demo not change theme every reload.

This is only used to disable applying the setting, not the API calls to set or load the principal setting so that we can still get tracing data for these endpoints from the demo site.


SchrodingersGat (approved):

matmair:
@SchrodingersGat ready for review, merge and application of the new key `disable_theme_storage` on the demo

SchrodingersGat:
Thanks, I've updated the demo server which should update within 24 hours

PR #11213, opened 2026-01-27
Split "Assign Batch Code" and "Assign Serials" in PO Receive

jacobfelknor:
Split batch and serial input toggles on PO receive.

Depending on the project, our practices change whether we want to assign a batch code, serialize, or both at receive time. This lets us individually toggle these form fields rather than require receiving to clear the generated serial numbers in cases we don't want to serialize immediately.


matmair (approved):
LGTM

SchrodingersGat (approved):

SchrodingersGat:
@jacobfelknor thanks for the addition

PR #11221, opened 2026-01-29
feat: connect and migrate inventree DB in paw health DB

asad-behram:


matmair:
@asad-behram you might want to change these credentials as the whole world and all the bots scraping the PRs here now know your database credentials

PR #11238, opened 2026-02-01
chore(backend): bump weasyprint

matmair:


matmair:
Does not work currently due to dependency problems, closing for now

PR #11242, opened 2026-02-02
[API] BOM active filter

SchrodingersGat:
Adds an API filter for BOM items based on the 'active' status of the linked 'sub_part'


matmair (approved):
LGTM; does this have API performance impacts?

SchrodingersGat:
Should the API performance checker not be running here?

matmair:
If there are performance changes on the default callpaths (that we measure): yes. But I am not sure our CI has enough datasets to find issues with this filter

PR #11244, opened 2026-02-02
feat(frontend): Add better frontend tracing

matmair:
This PR gives us a generic way to associated API calls with a single event in the frontend. This is helpful to more easily find bottlenecks like serial API calls that do not depend on each other, API calls that are always grouped and access the same data and much more.

Changeset is kept minimal to get some practice with the process around the gathered analytics before we maybe extend OTEL to the frontend - which is much more involved data and security wise.


SchrodingersGat:
Should we be concerned about the extra (albeit small) overhead associated with the extra API call to end the trace? Have you profiled this in action?

matmair:
I have profiled it on my test setup and the duration for the whole interaction is at about 50-65 ms as it does not really do much intensive action

matmair:
I have profiled it on my test setup and the duration for the whole interaction is at about 50-65 ms as it does not really do much intensive action

SchrodingersGat:
@matmair happy with this if you can fix conflicts, so we can get to the 1.2.0

matmair:
I have adressed the conflict

PR #11245, opened 2026-02-02
refactor(frontend): load server and auth info in paralell

matmair:
first discovery of https://github.com/inventree/InvenTree/pull/11244


SchrodingersGat:
@matmair what is the benefit here?

matmair:
The browser is done faster with fetching all required information so the login window disappears a bit faster.

PR #11264, opened 2026-02-06
BO Allocation Table MultiSelect with Change Role

jacobfelknor:
The multiselect check boxes on the Build Order Allocations table is currently tied to the BO delete role. This PR changes that to the BO modify role.

My justification is that users with change role already can consume allocated stock, but this must currently be done on a one-by-one basis unless they also have the delete role.


matmair (approved):
LGTM

SchrodingersGat:
@jacobfelknor thanks for the fix!

PR #11266, opened 2026-02-08
Add missing part/stocktake nullable annotations for schema

1337joe:
I tested querying the updated types from schema v448 and got nulls in a couple of fields that the schema thought weren't nullable. This adds the missing annotations.


SchrodingersGat:
@1337joe thanks for the updates

PR #11277, opened 2026-02-10
[bug] Fix table ordering

SchrodingersGat:
Many places in the UI where table ordering does not work, or works improperly.

This PR introduces a number of changes for this:

### TODO

- [x] Generic "backend" tests for API ordering fields
- [x] Generic "frontend" tests for table ordering options


matmair (commented):
LGTM

PR #11281, opened 2026-02-10
Transfer Order

jacobfelknor:
Add a Transfer Order to InvenTree which is based on other order types, such as a Sales Order.

### Done:

* On create, we select a source and destination location
* Add line items, selecting the part objects you need to transfer
* Allocate stock to each line item
    * like a Build Order, each allocation defaults to the source location of the order, but may be overridden in the allocate form
* Upon completion of the order, all allocated stock is transferred to the destination location
    * you may optionally accept incomplete allocations in the complete order form
*  When completed, the "allocated stock" tab and table changes to "transferred stock"
* There is a sample default report included
* settings for `TRANSFERORDER_ENABLED`, `TRANSFERORDER_REFERENCE_PATTERN`, and `TRANSFERORDER_REQUIRE_RESPONSIBLE`
* Transfer order rule set, so permissions can be assigned

### Differences between Transfer Order and other Orders
* no "extra lines"
* no shipment tracking, like a Sales Order
* no customer

### Items I'd still like to explore

- [x] Adding a "consume" option which consumes stock against an order rather than moving it. 
    - this is admittedly a bit strange, but we have scenarios, especially against consumables, where we're interested in queuing up stock to release from inventory for review, then want record (a report) of it occurring.  
    - Still undecided whether its better to just manually remove the stock after the transfer takes place. The fact that it all exists in the "transferred stock" panel means it wouldn't be too hard to delete it post-transfer, but explicitly adding a "consume" option would make the intention more clear

- [ ] Immutable tracking on the transfer.
    - right now, the "transferred stock" panel changes over time due to this being a direct link to the stock item.
    - preferably, this panel can show forever which stock was transferred, independent of further operations
    - this behavior isn't unique to TOs. Purchase Orders also exhibit this behavior on received stock
    - must do it in a way that retains the user's ability to navigate to that stock item if it still exists.
    - _**the fact that the report can operate on the original line items, not the transferred stock itself, makes this less of a priority than I had originally envisioned.**_
        - however, using line items means we still loose track of some information, such as which batch code was transferred 

- [ ] Similarly to BO's "consume stock" feature, it may be nice to execute the Transfer Order in smaller increments. We could add a "transfer now" feature to allocated stock, doing partial transfers before the entire order is complete

- [x] add a Transfer Order Allocations table to the stock item Allocations panel

- [x] unit testing

- [x] documentation

### General questions

- Should transfer orders update the "quantity required" number on parts if we want to transfer more than what is available?
    - I hadn't considered this before, because my assumption was that TOs were only intended to operate on existing stock. 
    - If the TO is used like a request (hey, engineering needs 5 of these by this date), then hooking this in the quantity required may be helpful.
- Similarly, do allocations to TOs reduce the available stock for an item until the order is completed, upon which it restores this qty to available
    - `IN_STOCK_FILTER` would need to account for transfer orders in this case

### Testing
- No unit tests exist for this yet. Of course, these will need to be added


src/backend/InvenTree/order/migrations/0118_transferorder.py
matmair, line 1: once this is moving to be ready for a merge these migrations should all be redone into one

src/backend/InvenTree/order/api.py
```diff
@@ -2158,6 +2674,84 @@ def item_link(self, item):
             ),
         ]),
     ),
+    # API endpoints for transfer orders
+    path(
+        'to/',
```
matmair: I think we should strive for a more human-readable API so transfer-order would be better here IMO

src/backend/InvenTree/order/api.py
```diff
@@ -2158,6 +2674,84 @@ def item_link(self, item):
...
+                name='api-to-allocation-detail',
+            ),
+            path(
+                '', TransferOrderAllocationList.as_view(), name='api-to-allocation-list'
+            ),
+        ]),
+    ),
     # API endpoint for subscribing to ICS calendar of purchase/sales/return orders
     re_path(
         r'^calendar/(?P<ordertype>purchase-order|sales-order|return-order)/calendar.ics',
```
matmair: Will this be supported by transfer-orders too?

src/backend/InvenTree/order/models.py
```diff
@@ -371,11 +378,12 @@ def clean(self):
...
 
         # Check that the referenced 'contact' matches the correct 'company'
-        if self.company and self.contact:
-            if self.contact.company != self.company:
-                raise ValidationError({
-                    'contact': _('Contact does not match selected company')
-                })
+        if hasattr(self, 'company'):
+            if self.company and self.contact:
+                if self.contact.company != self.company:
```
matmair: can we chain these to make them more readable?

src/backend/InvenTree/order/models.py
```diff
@@ -385,11 +393,12 @@ def clean(self):
...
 
         # Check that the referenced 'address' matches the correct 'company'
-        if self.company and self.address:
-            if self.address.company != self.company:
-                raise ValidationError({
-                    'address': _('Address does not match selected company')
-                })
+        if hasattr(self, 'company'):
+            if self.company and self.address:
+                if self.address.company != self.company:
```
matmair: same as above

src/backend/InvenTree/part/models.py
```diff
@@ -1783,15 +1784,58 @@ def sales_order_allocation_count(self, **kwargs):
...
-        """Return the total quantity of stock allocated for this part, against both build orders and sales orders."""
+        """Return the total quantity of stock allocated for this part, against build orders, sales orders, and transfer orders."""
         if self.id is None:
             # If this instance has not been saved, foreign-key lookups will fail
             return 0
 
         return sum([
             self.build_order_allocation_count(**kwargs),
             self.sales_order_allocation_count(**kwargs),
+            self.transfer_order_allocation_count(**kwargs),
```
matmair: are we sure we want to include this here?

matmair (commented):
Very good start; looking at this we might want to change how a few parts of the code base work, this seems much to complicated for someone new to wrap their head around. Thanks for doing so still

src/backend/InvenTree/order/api.py
jacobfelknor: I hadn't realized it existed, but probably

src/backend/InvenTree/part/models.py
jacobfelknor: I'm not sure. This starts to creep in my question about whether we want to consider stock allocated to a Transfer Order as "available" or not, which also impacts the "required quantity"

jacobfelknor:
I'll be on vacation for the next 10 days, so won't be working on this until I'm back.

Before spending any more serious time on it when I return, I'd like to hear from your perspective what should be my priorities before this can be merged. I've detailed some of my thoughts in the PR description.

The basic functionality present is probably enough for me to start using it on my instance - maybe behind a feature flag - but I know there's a lot of housekeeping left to do before it could be considered "official."

SchrodingersGat:
### Documentation

@jacobfelknor this will need to also have comprehensive docs added, under the "stock" master docs location :)

src/backend/InvenTree/part/models.py
SchrodingersGat: It is an interesting question. I think that stock which is "in transfer" is not necessarily "unavailable" - merely moving from one location to another?

Alternatively if they are "unavailable" until the transfer order is complete, this also could make sense because it precludes them being assigned elsewhere (until the transfer is completed)

SchrodingersGat (commented):
@jacobfelknor this is a really good implementation, great effort!

src/backend/InvenTree/order/api.py
jacobfelknor: I tend to agree. I matched the pattern already present, such as `so/` for Sales Orders, `ro/` for Return Orders, etc. I don't think I should change _all_ API paths, so would you rather me match existing patterns or have Transfer Order diverge from the pattern but start more human-readable?

matmair: Currently starting to write API docs in https://github.com/inventree/project-adr/issues/4 that make things like this clearer; I think starting to diverge would be good. One less thing that we will need to break

src/backend/InvenTree/part/models.py
jacobfelknor: Here's an argument against including "reduced availability" logic for Transfer Orders.....

The spot it hurts most is actually not at the part allocation counts, referenced here, but at the stock item allocation count in `StockItemSerializer.annotate_queryset()`. The part level allocation counts at least are "released" once the order is no longer "pending", as defined by `TransferOrderStatusGroups` and `Part.transfer_order_allocations()` 

However, if I don't clear allocations, the available stock remains reduced even after the order completes from the point of view of the stock item.

```python
   @staticmethod
    def annotate_queryset(queryset):
        ....
        # Annotate the queryset with the total allocated to sales orders
        queryset = queryset.annotate(
            allocated=Coalesce(
                SubquerySum('sales_order_allocations__quantity'), Decimal(0)
            )
            # THIS IS THE PROBLEMATIC LINE!
            # this count isn't 'released' after the TO completes, 
            # forever reducing available qty for a stock item that has been through a TO
            + Coalesce(SubquerySum('transfer_order_allocations__quantity'), Decimal(0))
            + Coalesce(SubquerySum('allocations__quantity'), Decimal(0))
        )
```

And since in the current implementation, the `TransferOrderAllocation` is the only thing attaching a `StockItem` to a `TransferOrder`, if I clear allocations on complete, I lose the reference for which stock was transferred from this order.

In lieu of a more sophisticated immutable tracking on the transfer order - which I'm beginning to think is more scope than I'm willing to commit to in this PR - I think excluding TO allocations from any availability calculations is best, and then revisit when/if the immutable tracking exists.

src/backend/InvenTree/order/migrations/0118_transferorder.py
jacobfelknor: squashed in 15d4674357908af33af679f70a2f484b960d647f

jacobfelknor:
Alright, I think I've matched expectations for `TransferOrder` unit tests at this point.

I'll start on documentation next

jacobfelknor:
I've fixed a couple tests that were previously failing in actions

- `common.tests.SettingsTest.test_defaults`
- `users.tests.RuleSetModelTest.test_scope_names`
- `generic.states.tests.ApiTests.test_all_states`

I'm not sure that I understand the barcode tests well enough to determine if I caused these failures or if they're unrelated

jacobfelknor:
I *think* I finally tracked down the barcode test issue. I was missing a `TransferOrder.barcode_model_type_code()` implementation.

matmair:
> I _think_ I finally tracked down the barcode test issue. I was missing a `TransferOrder.barcode_model_type_code()` implementation.

do we need to improve that part of the code base to raise better warnings/errors?

jacobfelknor:
Potentially, but it ended up being pretty clear once I committed some real time to it and stepped through with the debugger. I hadn't revisited this until today

jacobfelknor:
Is the failing API Schema Documentation something I need to address before this is ready to merge? I'm not familiar with exactly what is happening here, but it looks like it checks for breaking changes in the API. Do my new transfer order API paths qualify as a "breaking" change, or should this be expected to pass?

Beyond that, let me know if there's anything I else I need to do in order for this to be ready.

jacobfelknor:
Ah - I think [I see what I need to do](https://docs.inventree.org/en/stable/develop/contributing/#api-versioning)

[Pushed](https://github.com/inventree/InvenTree/pull/11281/commits/41be1a895cc270217824abd6926a67b0f638e99b)

matmair:
There are a bunch of security warnings / code smells - please take a look https://github.com/inventree/InvenTree/pull/11281/checks?check_run_id=68568902493

the api job warning can be ignored by you, I will review that by hand

Not sure why the InvenTree-python test is failing. That will need adressing.

overall great work. We will try to get this over the finish line in an appropriate time

jacobfelknor:
Ironically, the `inventree-python` test was passing prior to [41be1a8](https://github.com/inventree/InvenTree/pull/11281/commits/41be1a895cc270217824abd6926a67b0f638e99b)

matmair:
That seems unrelated @jacobfelknor
This is in a good state, I am very happy with this

matmair:
Having a bit more coverage on the following things would be nice:
- TransferOrderSerialAllocationSerializer / the related api action
- the filters on TransferOrderFilter / TransferOrderAllocationFilter / TransferOrderLineItemFilter

src/frontend/src/forms/TransferOrderForms.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+          if (quantity != props.item.quantity) {
+            props.changeFn(props.idx, 'quantity', quantity);
+          }
+        }
+      }
+    };
+  }, [sourceLocation, record, props]);
+
+  console.log(sourceLocation);
```
SchrodingersGat: Please remove this debug line

src/frontend/src/forms/TransferOrderForms.tsx
```diff
@@ -0,0 +1,330 @@
...
+            <TransferOrderAllocateLineRow
+              key={`table-row-${row.idx}-${record.pk}`}
+              props={row}
+              record={record}
+              sourceLocation={sourceLocation}
+            />
+          );
+        }
+      }
+      // transfer: {
```
SchrodingersGat: If this field is no longer required, please remove the commented-out lines

src/frontend/src/pages/part/PartDetail.tsx
```diff
@@ -905,6 +907,17 @@ export default function PartDetail() {
         hidden: !part.assembly || !user.hasViewRole(UserRoles.build),
         content: part.pk ? <BuildOrderTable partId={part.pk} /> : <Skeleton />
       },
+      {
+        name: 'transfer_orders',
+        label: t`Transfer Orders`,
+        icon: <IconTransfer />,
+        hidden: part.virtual || !globalSettings.isSet('TRANSFERORDER_ENABLED'),
```
SchrodingersGat: Should also check if the user has the required view permissions on the `TransferOrder` ruleset

SchrodingersGat (changes requested):
Fantastic and thorough implementation!

SchrodingersGat:
@jacobfelknor I'll not have time this week to review , however it remains a priority. Can you please fix conflicts and see if you can address the outstanding web UI tests?

jacobfelknor:
Yep, no worries

SchrodingersGat:
@jacobfelknor sorry for the commit noise - I have fixed the database migrations and some other conflicts

SchrodingersGat:
Some additions I'd like to see here

### Demo Data

Add some sample transfer orders (and assigned stock) to the [demo dataset](https://github.com/inventree/demo-dataset/) - this will be used for running frontend tests via playwright

### Frontend Tests

Add some [playwright tests](https://docs.inventree.org/en/stable/develop/react-frontend/#testing) to cover the new UI elements

jacobfelknor:
Good call. I will do my best to get these addressed this week

jacobfelknor:
Added a couple playwright tests. I probably need a few more to be considered complete. Goal is to do this tomorrow if I have the time.

Using demo dataset at https://github.com/jacobfelknor/demo-dataset/tree/transfer_order

SchrodingersGat:
Nice, the tests are looking good! Please address the conflicts here which have snuck in

SchrodingersGat (approved):

SchrodingersGat:
@jacobfelknor I've fixed a bunch of conflict issues here - I think there are just a couple of the new playwright tests which are not passing. If you can get to these soon, that would be great! Otherwise, we might start to see more conflicts again which adds difficulty

jacobfelknor:
On it

jacobfelknor:
I think I've fixed an issue with the transfer order test.... I was able to see a case every now and then where the allocation form hadn't finished its API request to populate the stock item dropdown before it tried to proceed. Hopefully 479a1c5 helps this.

I wasn't able to reproduce the Build Order output test that also failed previously

jacobfelknor:
I also was assuming the test would only be ran *once* per pipeline, but I'm not sure that is true. 1afbd0bb147ce971d9ae80a99d10f92b19c29279 duplicates the transfer order I complete before running so that it can be ran many times in a row without issues

jacobfelknor:
I'm stuck.... I can't reproduce these failures locally

SchrodingersGat:
> I also was assuming the test would only be ran once per pipeline, but I'm not sure that is true

The order of tests can also not be guaranteed. It's a good idea to try and design the tests to be idempotent, as far as possible

SchrodingersGat:
@jacobfelknor all passing now!

@matmair I'm happy with this, are you happy for this to be merged?

jacobfelknor:
Thanks for the assist 😁

matmair:
Great effort, this is ready for merge!

There surely will be follow ups once users get access and discover new ways to use this but I think this is pretty well rounded as-is.
Thank you for your perseverance @jacobfelknor !

matmair (approved):

SchrodingersGat:
@matmair due to the large schema diff, the "schema push" action has failed - can you manually fix this?

matmair:
on it @SchrodingersGat

PR #11294, opened 2026-02-11
[1.1.x] bump django due to... all the releases vulns (#11265)

matmair:
# Backport

This will backport the following commits from `master` to `1.1.x`:
 - [bump django due to... all the releases vulns (#11265)](https://github.com/inventree/InvenTree/pull/11265)

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

PR #11295, opened 2026-02-11
chore(backend): fix stable batch vulns

matmair:
Fixes a few high vulns on the stable image


matmair:
@SchrodingersGat we should release ASAP after this and https://github.com/inventree/InvenTree/pull/11294 is merged; some serious django vulns in there

PR #11296, opened 2026-02-11
feat(frontend): add warning to SO line calcs

matmair:
Quick frontend-only addressing of what #10983 will do.

This adds a warning if the currently entered unit price is not matching the calculated unit price based on sales price price-breaks.


SchrodingersGat (approved):

SchrodingersGat:
Nice addition :)

PR #11299, opened 2026-02-12
Bump version number to 1.2.0

SchrodingersGat:


matmair:
Should we also change the CHANGELOG.md with this?

SchrodingersGat:
@matmair good point, have fixed that now

PR #11303, opened 2026-02-12
Supplier Part Updates

SchrodingersGat:
Currently the "primary" supplier part for a given "part" is specified on the Part model itself, which is contrary to the data pattern we use for other tables, which keeps them "self contained".

This PR does the following:

- Add a `primary` field to the `SupplierPart` model
- Migrate `default_supplier` info from `Part` model to the new field
- Remove the `default_supplier` field from the `Part` model

And also:

- Updated UI / API code

This means that we get better filtering for "primary" supplier parts, and more self-contained database tables

### Important Note

The `default_supplier` field is not currently used for any business logic in the application

### TODO

- [x] Unit tests for "primary" field
- [x] Playwright testing updates


matmair (approved):
LGTM

PR #11309, opened 2026-02-12
Update base image in Dockerfile

matmair:
Updated base image to the latest version as of 2026-02-12.


matmair:
@SchrodingersGat we have a bunch of high vulns in our container image because the base is neraly 6 months old. this needs a backport

PR #11336, opened 2026-02-15
refactor(backend): switch release pipeline to new pkgr version

matmair:
This switches the release mechanism for packages to https://go.packager.io/u/inventree/InvenTree
We will run in parallel for a while during the 1.2.x release window before we recommend the new mechanism


matmair:
ready for review and merge

PR #11340, opened 2026-02-16
[setup] invoke command updates

SchrodingersGat:
- wait for db before migrating data
- improve task state reporting
- early return from isGeneratingSchema (significant speed-up)

This fix ensures that the database is ready before running the migrate command.

Ref: https://github.com/inventree/InvenTree/issues/11338


src/backend/InvenTree/InvenTree/backends.py
```diff
@@ -84,10 +84,11 @@ def set_value(self, value: bool, retries: int = 5, minutes: int = 5):
 
             r -= 1
 
-            if r == 0:
-                logger.warning(
-                    'Failed to set maintenance mode state after %s retries', retries
-                )
+            # Disable this warning message (for now) as it is confusing users with no upside
+            # if r == 0:
+            #     logger.warning(
```
matmair, line 89: Why is this disabled?

matmair (approved):

src/backend/InvenTree/InvenTree/backends.py
SchrodingersGat: We are getting a lot of users reporting the following warning:

`Failed to set maintenance mode state after 5 retries`

And the users are reporting it as an error during the update process.

PR #11357, opened 2026-02-17
Add --no-drop option for dbrestore

SchrodingersGat:
Allow `--no-drop` option to be passed through to the `dbrestore` command


SchrodingersGat:
Closing, this option does not work correctly with django-dbbackup

PR #11365, opened 2026-02-18
Test fixes 2026 02

fanzart-dev:
This PR contains some test updates made while working through recent failing cases locally.

Please review and advise if any of these adjustments are no longer needed.


matmair:
The overall changes look interesting and reasonable but I would appreciate some added end user documentation for the tax and tenant models/functionality.
This change set is also pretty big; splitting it up into multiple PRs might speed up review times.

PR #11369, opened 2026-02-18
Update to /api/barcode/po-recieve/ error messages

Idea-Junkie:
# Includes

Updates to barcode error messages, letting users know if a barcode was found with a plugin, and what is missing from that information.

# Example: PO found, Supplier part Not found
<kbd>
Screenshot 2026-02-18 120850
</kbd>

See more in #11361 

## No Supplier Associated to Plugin
Also added "no_supplier_plugin _error" to notify the user if plugins do not have suppliers set in settings 

<kbd>
Screenshot 2026-02-18 143623
</kbd>

### This is set to only pop up if there is an error, if a match is found it skips this step

# List of Tasks for this to be a Completed PR
- [x] Documentation Update
- [x] Error Message implementation
- [ ] Add Unit Tests
- [x] Pre-Commit Style Verification


src/backend/InvenTree/plugin/base/barcodes/api.py
```diff
@@ -546,13 +549,35 @@ def handle_barcode(self, barcode: str, request, **kwargs):
...
-            if result is None:
-                continue
-
+            
+            # No_Match Determines if it found a exact match for all the required fields from scan_recieve_item
+            if result['No_Match'] is True:
+
+                # Supplier does not have associated Supplier ID
+                if result['supplier'] is None:
+                    no_supplier_plugin_error.append(current_plugin.name)
```
matmair: .slug should be used instead

src/backend/InvenTree/plugin/base/barcodes/api.py
```diff
@@ -546,13 +549,35 @@ def handle_barcode(self, barcode: str, request, **kwargs):
...
+                    continue
+                
+                # No PO(Purchase Order) or Supplier Part Found
+                if result['PO'] is None and result['supplier_part'] is None:
+                    continue
+
+                # PO(Purchase Order) exists and is found but Supplier part DNE 
+                if result['PO'] != None and result['supplier_part'] is None:
+                    # Adds what is causing error for barcode scan
+                    response[current_plugin.name + " Debug" ] = result
```
matmair: This should be a nested dict with slug as keys and the result as value for more reliable parsing

Idea-Junkie:
@matmair I will be making the documentation edits on here as well which is why its a PR draft. 

I believe this was a good first issue for me to get my feet wet, and I hope to be more of assistance on some other issues.

any comments on how I went about this? Should I work on it and then make a PR draft request and include the conversation in there instead of how I did it with the discussion? or is the discussion a better method?

src/backend/InvenTree/plugin/base/barcodes/api.py
```diff
@@ -546,13 +549,35 @@ def handle_barcode(self, barcode: str, request, **kwargs):
...
+            # No_Match Determines if it found a exact match for all the required fields from scan_recieve_item
+            if result['No_Match'] is True:
+
+                # Supplier does not have associated Supplier ID
+                if result['supplier'] is None:
+                    no_supplier_plugin_error.append(current_plugin.name)
+                    continue
+                
+                # No PO(Purchase Order) or Supplier Part Found
+                if result['PO'] is None and result['supplier_part'] is None:
```
matmair: these dict accesses are done multiple times; extract them to variables to lower likelyhood of problematic refactors

src/backend/InvenTree/plugin/base/barcodes/api.py
```diff
@@ -546,13 +549,35 @@ def handle_barcode(self, barcode: str, request, **kwargs):
...
+                # PO(Purchase Order) exists and is found but Supplier part DNE 
+                if result['PO'] != None and result['supplier_part'] is None:
+                    # Adds what is causing error for barcode scan
+                    response[current_plugin.name + " Debug" ] = result
+                    plugin_error = _('Purchase order Found\rNo supplier Part Match')
+
+                # Supplier Part is Found but PO(Purchase Order) DNE
+                elif result['PO'] is None and result['supplier_part'] != None:
+                    # Adds what is causing error for barcode scan
+                    response[current_plugin.name + " Debug" ] = result
```
matmair: Same as above

src/backend/InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -344,32 +344,48 @@ def scan_receive_item(
         # Extract supplier information
         supplier = supplier or self.get_supplier(cache=True)
 
+        # Construct Debug Response 
+        # Will hold info on what is missing from Barcode
+        Debugresponse = {}
```
matmair: variables should be starting lower case, this looks like a class

src/backend/InvenTree/plugin/base/barcodes/api.py
```diff
@@ -569,13 +594,19 @@ def handle_barcode(self, barcode: str, request, **kwargs):
...
+        elif no_supplier_plugin_error:
+            response['no_supplier_plugin_error'] = no_supplier_plugin_error
 
         # A plugin has not been found!
         if plugin is None:
             response['error'] = _('No plugin match for supplier barcode')
 
+        # A plugin was found, with a Error
+        elif plugin_error:
+            response['error'] = _(plugin_error)
```
matmair: Q: Why are already translated strings passed in for translation agian?

src/backend/InvenTree/plugin/base/barcodes/mixins.py
matmair: blocking: Please type this

src/backend/InvenTree/plugin/base/barcodes/mixins.py
```diff
@@ -344,32 +344,48 @@ def scan_receive_item(
...
-        if not line_item:
-            line_items = purchase_order.lines.filter(part=supplier_part)
-            if line_items.count() == 1:
-                line_item = line_items.first()
+        if not line_item and purchase_order != None:
+            try:
+                line_items = purchase_order.lines.filter(part=supplier_part)
+                if line_items.count() == 1:
+                    line_item = line_items.first()
+            except Exception:
```
matmair: This is a very wide exception; please make it more specific to the case you want to catch

matmair (changes requested):
Looks good in general; a few smaller things to address

I will allow CI now to see if there are style issues etc.

matmair:
> any comments on how I went about this? Should I work on it and then make a PR draft request and include the conversation in there instead of how I did it with the discussion? or is the discussion a better method?

making generalized rules is hard; if it is a concrete change or widening to/of system behavior, a feature request with "I want to dev this" selected is the way to go. That allows us more concrete roadmap management and improves visibility to the change and that someone is working on it.

Small things can just be done and PR'ed (with sufficient PR description). If the thing will take you longer than 1 week to dev I would certainly open an issue or discussion to ensure you do not waste time on duplicate / unwanted things.

src/backend/InvenTree/plugin/base/barcodes/api.py
Idea-Junkie: Ah, I was originally doing it with a bare string value but forgot to get rid of that

src/backend/InvenTree/plugin/base/barcodes/mixins.py
Idea-Junkie: getting rid of this as this was added during testing but I see no need for it now

Idea-Junkie:
Updated the requests you made, let me know if there is anything else I need to add or change, as the workflow changed a little bit with the changes your requested.

Also updated documentation (was only like two notes really)

Also while Im at it, is there a way to make the images on the documentation smaller? Its always overblown for me on my desktop because its a vertical picture instead of horizontal, also tried to preview it in VScode but it just showed how the text would work.

docs/docs/app/barcode.md
```diff
@@ -65,7 +65,9 @@ Transfer the currently selected stock location into another location. Scanning a
 
 #### Scan Received Parts
 
-Receive incoming purchase order items into the selected location. Scanning a *new* barcode which is associated with an item in an incoming purchase order will receive the item into the selected location.
+Receive incoming purchase order items into the selected location. Scanning a *new* barcode which is associated with an item in an incoming purchase order will receive the item into the selected location. 
+
+*Note: Both purchase order number and supplier SKU's are required to be found on the barcode for this function to find the associated line item. You cannot have one without the other.*
```
matmair: I would change the second sentence a bit.
```suggestion
*Note: Both purchase order number and supplier SKU's are required to be found on the barcode for this function to find the associated line item. Missing one will lead to an error.*
```

src/backend/InvenTree/plugin/base/barcodes/api.py
```diff
@@ -546,12 +557,56 @@ def handle_barcode(self, barcode: str, request, **kwargs):
...
+
+                # Supplier for PO or Supplier part in barcode was found
+                if supplier_found is True:
+                    # Adds info on for what was found in the barcode
+                    response[current_plugin.slug + "_debug"] = {
+                            'purchase_order':supplier_purchase_order,
+                            'no_match':no_match,
+                            'supplier':plugin_supplier,
+                            'supplier_part': supplier_part,
+                        }
```
matmair: As mentioned before: Please put all of these debug messages into one dict, not under multiple generated keys that can not be easily checked.

in the end the repsonse should be something likes this:
```
{
...
'supplier_matches': {
    'slug1': {
        'purchase_order': 123,
        'no_match': true,
        ...
    },
    'slug2': {
        'purchace_order': 456,
        'no_match': false,
    }
...
}
```

matmair (commented):
LGTM; adding some test coverage would be good too

src/backend/InvenTree/plugin/base/barcodes/api.py
Idea-Junkie: I am confused by this comment? it will only output the plugin debugresponse dict that found it as it breaks the for loop looking for more plugins to match the barcode info to. 

So it wont have multiple plugin debugresponses like it seems how youre describing it in the response as we decided that would be an issue, so I went away with that and stuck to only output the plugin it found that matches. 

or are you just talking about defining the dict to set variable instead of having the suppliers name in the dict?

like `plugin_debug` or something? this is pretty much what I am doing down below just without the `current_plugin.slug` which I can see would make accessing it a little bit more predictable in testing. 


````
...
'plugin_debug': {
    'purchase_order':  123,
    'no_match': true,
    ...
},
...
````

I guess your asking for this?
````
...
'plugin_debug': {
    'current_plugin.slug': {
        'purchase_order':  123,
        'no_match': true,
        ...
    },
},
...
````


Unless I'm misunderstanding what you are explaining. let me know 

# New format
<kbd>
Screenshot 2026-02-19 135029
</kbd>

**Comment by SchrodingersGat** [comment]
@Idea-Junkie looking good in general -are you going to add support for these messages to the app, too?

**Comment by Idea-Junkie** [comment]
@SchrodingersGat 

## Mobile Error Pop-Up
The original line was too long but with a carriage return added to it, it is readable on the app as well

<kbd>
image
</kbd>
is this what youre talking about?

**Comment by SchrodingersGat** [comment]
@Idea-Junkie yes, perfect!

File: src/backend/InvenTree/plugin/base/barcodes/api.py (hunk 9 above)
Comment (matmair) [reply src/backend/InvenTree/plugin/base/barcodes/api.py]: This is not about one single log item, this is about barcode log messages being made harder to parse
ok I going to try explain this with a scenario, maybe that is clearer:
1. admin enables barcode logging
2. users scan broken / not working barcodes
3. logs for different plugins get created
4. admin starts to analyze the logs by downloading them
5. admin wants to create stats about problematic pruchaseorders: We arrive at the problem

Aggregating the log lines with your format requires manual aggregating the fields "digikeyplugin_debug", "mysecondbarcode_debug", "hopefullyIrememerberthisplugin_debug" into one field to get usable stats across failing plugins
Aggregating the log lines with my suggestion: aggregate "supplier_matches"; always the same field; no data shape changing depending on the plugins / pluginslugs (which can and have changed)

Comment (Idea-Junkie) [reply src/backend/InvenTree/plugin/base/barcodes/api.py]: Ohhhh okay gotcha I think I understand now.

Comment (Idea-Junkie) [reply src/backend/InvenTree/plugin/base/barcodes/api.py]: Changed the name to supplier_matches now. To build off that and double-check I understand it, when in your representation you say `slug1`, in the context of the Json format can I just add a 'plugin_slug' variable to it without nesting it?  I know it mentions the plugins name earlier in the context of the response but maybe it would be useful? Check out the recent commit and see what your thoughts are.

Also I am a little new to tests for stuff like this, so from my understanding, in the tests file I would essential just replicate the scenarios that could cause the errors, but with the code directly?

Comment (matmair) [reply src/backend/InvenTree/plugin/base/barcodes/api.py]: I think this should be sufficient for now

File: src/backend/InvenTree/plugin/base/barcodes/api.py [hunk 10]
```diff
@@ -546,12 +559,58 @@ def handle_barcode(self, barcode: str, request, **kwargs):
...
+                # Supplier does not have associated Supplier ID
+                if plugin_supplier is None:
+                    no_supplier_plugin_error.append(plugin_slug)
+                    continue
+
+                # No Purchase Order or Supplier Part Found
+                if supplier_purchase_order is None and supplier_part is None:
+                    continue
+
+                # Purchase Order exists and is found but Supplier part DNE
```
matmair: This DNE causes style issues; please use the written out version in all 3 occurrences

src/backend/InvenTree/plugin/base/barcodes/api.py
Idea-Junkie: Is this good to be resolved now?

Idea-Junkie:
I believe this is ready for a commit now, Unless we think the tests would be necessary.

matmair:
Not really a fan of such changes without coverage, but I will defer to @SchrodingersGat for judgment if adding / extending them is required in this case

SchrodingersGat:
I would definitely want to see some unit tests here. This barcode scanning to receive parts has gone through a number of bug-fix cycles already and it's very hard to trace down errors "in the field" because of the interplay between the different systems.

So at the very least, some additional unit tests to cover the new modes of operation / expected error codes

Idea-Junkie:
Ok, I will take a look at adding more unit tests. Thank you guys for your guidance and help with this!

Idea-Junkie:
Would this apply to [`src/backend/InvenTree/plugin/builtin/suppliers/test_supplier_barcodes.py`? or [`src/backend/InvenTree/plugin/base/barcodes/test_barcode.py`](https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/plugin/base/barcodes/test_barcode.py)? 

EDIT: Found in https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/plugin/builtin/suppliers/test_supplier_barcodes.py

SchrodingersGat:
@Idea-Junkie where are you at with adding tests to this?

matmair (approved):

matmair:
> Found in https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/plugin/builtin/suppliers/test_supplier_barcodes.py

yes that would be a good target

SchrodingersGat:
@Idea-Junkie I'm going to bump this to the next release target - but please LMK if you are able to work on this again soon? It would be great to get it completed and merged in.

SchrodingersGat:
@Idea-Junkie any updates here?

SchrodingersGat:
@Idea-Junkie it would be great to get this one closed out

PR #11370, opened 2026-02-18
Change warning to info

SchrodingersGat:
- Generates a lot of scary logs during update process


src/backend/InvenTree/common/settings.py
```diff
@@ -47,9 +47,7 @@ def set_global_setting(key, value, change_user=None, create=True, **kwargs):
     from InvenTree.ready import canAppAccessDatabase
 
     if not canAppAccessDatabase(allow_shell=True, allow_test=True):
-        logger.warning(
-            f'Cannot set global setting "{key}" - database is not accessible'
-        )
+        logger.info(f'Cannot set global setting "{key}" - database is not accessible')
```
matmair, line 50: Can we maybe guard when this is raised better but keep the level? Not being able to write a global setting can be pretty bad in normal production for security

matmair (approved):

PR #11371, opened 2026-02-18
Batch update of records during migration

SchrodingersGat:
Reduce issues with data migration for very large datasets:

- Prefetch the queryset to significantly reduce database overhead
- Batch the "bulk update" operation to reduce memory and CPU usage

### DIFF
--- a/src/backend/InvenTree/stock/migrations/0118_auto_20260205_1218.py
+++ b/src/backend/InvenTree/stock/migrations/0118_auto_20260205_1218.py
@@ -10,6 +10,11 @@ def add_part_links(apps, schema_editor):
 
     history_entries = []
 
+    N = StockItemTracking.objects.count()
+
+    if N > 0:
+        print(f"\nUpdating {N} StockItemTracking entries with part links...")
+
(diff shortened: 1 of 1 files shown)


src/backend/InvenTree/stock/migrations/0118_auto_20260205_1218.py
```diff
@@ -10,24 +10,40 @@ def add_part_links(apps, schema_editor):
...
-    
+
+        # Process in batches to avoid issues with very large datasets
+        if len(history_entries) >= 100:
+            StockItemTracking.objects.bulk_update(history_entries, ['part'])
+            history_entries = []
+            print(".", end='', flush=True)
+
     if len(history_entries) > 0:
         StockItemTracking.objects.bulk_update(history_entries, ['part'])
```
matmair, line 46: I might be ignorant but are line 41 and 46 not the same?

SchrodingersGat: 45 is taking care of any entries which were "left over" at the end of the for loop

Issue #11235, opened 2026-02-01
[BUG] Fix restore of backups when plugins with modesl are used


matmair:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

If plugins with AppMixin run within an instance, we might not be able to restore all the backed up information as there is a direct connection between database and available plugins. This is not reflected in the current backup / restore setup.

### Steps to Reproduce

N/A

### Expected behaviour

do a multi-step process like proposed in https://github.com/inventree/InvenTree/issues/11214#issuecomment-3830173005

### Deployment Method

None

### Version Information

InvenTree-Version: 1.2.0 dev
Django Version: 5.2.10
Commit Hash: d00d472
Commit Date: 2026-01-30
Commit Branch: null
Database: django.db.backends.postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-5.15.0-100-generic-x86_64-with-glibc2.41
Installer: DOC
Active plugins: false

### Try to reproduce on the demo site

I did not try to reproduce

### Is the bug reproducible on the demo site?

Not reproducible

### Relevant log output

```shell

```

matmair:
also see https://github.com/inventree/InvenTree/issues/11214#issuecomment-3830173005

> I think we just have to ensure that the correct plugins are enabled before updating.
> 
> So we may need to ensure that the data is re-imported in a correct sequence:
> 
>     Run InvenTree table migrations
>     Import core InvenTree data
>     Load plugin configs (to ensure the plugins are enabled to allow generation of tables)
>     Re-run migrations (to create the plugin tables)
>     Import plugin data from the backup
>

PR #11372, opened 2026-02-18
feat(backend): ensure restore of backups only works in correct enviroments

### LINKED ISSUE #11214: [FR] ensure restore of backups only works in correct enviroments

### Problem statement

User might restore backups in too old environments. This can cause unreliability and data loss

### Suggested solution

Use https://github.com/Archmonger/django-dbbackup/pull/663 once it is merged and released to enrich backup metadata with data around api and software version as well as plugin state

Write docs

### Examples of other systems

_No response_

- [x] I want to develop this.

matmair:
This uses a new function I upstreamed in dbbackup to allow us checking various environment markers.
The names of variables are chosen in a manner to keep long-term restore compatibility in place as we might evolve this schema. It is no rocket science but should help with some common issues, especially if we get logs for debugging.

To-do:
- [x] add error codes?
- [x] docs
- [x] flag to allow a dangerous restore
- [x] e2e test?

Fixes #11214


SchrodingersGat:
Nice work, and great job getting the functionality merged upstream.

### dbbackup version

Have we already updated to the required version of django-dbbackup? There is no version bump in this PR?

### docs

It would be great to add a brief paragraph to the backup docs regarding compatibility

matmair:
> Have we already updated to the required version of django-dbbackup? There is no version bump in this PR?

yes, that was included in the latest bump (5.2.0 included the change)


> It would be great to add a brief paragraph to the backup docs regarding compatibility

This PR mostly prevents footguns, there is no functional change other than (by default) stopping restores from newer versions. I will add a short section on some basic restrictions. Real world functional decencies will probably only enter with #11235

SchrodingersGat:
@matmair fair enough - if you can submit some brief docs too that would be great, but happy to merge this now

Issue #11375, opened 2026-02-19
Sorting by Unit Price not worked


xlfds:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

Sorting by Unit Price not working in Stock -> StockItems table.
Sorting by Stock column not working too.

### Steps to Reproduce

1. Go to https://demo.inventree.org/web/stock/location/index/stock-items
2. Show Unit price column
3. Click to this column for sorting up
4. Click again for sorting down
5. We see this not working

### Expected behaviour

Order of rows will sorting by Unit Price.

### Deployment Method

None

### Version Information

InvenTree 1.3.0 dev
Python 3.11.14
Django 5.2.11
Commit f620630
Commit date 2026-02-18

### Try to reproduce on the demo site

I tried to reproduce

### Is the bug reproducible on the demo site?

Reproducible

### Relevant log output

```shell

```

SchrodingersGat:
@xlfds thanks for reporting - fix incoming in https://github.com/inventree/InvenTree/pull/11379

PR #11376, opened 2026-02-19
[UI] Bug fix for stock column ordering

SchrodingersGat:
- Cannot have a blank accessor


matmair:
see https://github.com/inventree/InvenTree/issues/11375?

SchrodingersGat:
@matmair thanks, I'll check that out too. This was discovered separately!

PR #11411, opened 2026-02-23
Fix auto pricing overwriting manual purchase price #10846

### LINKED ISSUE #10846: "Auto Pricing" overwrites manually entered Purchase Price

### Describe the bug*

**Note:** this is more or less a duplicate of #7600, but because that was targeting the previous UI I thought it might be better to raise a new issue.

When modifying the purchase price of a line item of a purchase order, the price entered by the user is overwritten by the Auto Pricing functionality. This leads the user to believe the price has been modified, where in fact it stayed the same.

### Steps to Reproduce

1. Open a purchase order with line items with historical pricing information, e.g. https://demo.inventree.org/web/purchasing/purchase-order/1/line-items
2. Edit a line item -> modify purchase price -> submit
3. Purchase price hasn't changed...

### Expected behaviour

When modifying the purchase price, the "Auto Pricing" slider should disable automatically.

It'd be nice if, when pressing enter, the form is submitted.

### Deployment Method

Other - added info in Steps to Reproduce

### Version Information

InvenTree-Version: 1.2.0 dev
Django Version: 5.2.8
Commit Hash: 770f7a2
Commit Date: 2025-11-16
Commit Branch: null
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-5.15.0-100-generic-x86_64-with-glibc2.41
Installer: DOC
Active plugins: false

I tried to reproduce

Reproducible

### Relevant log output

```shell

```

JustusRijke:
Fixes #10846 

Changes:

1. Disable the purchase price input when auto pricing is enabled.
1. Default auto pricing to false. When the user adds a line item and enables auto pricing, the form remembers this (so when adding another line item, auto pricing will still be enabled).
1. When editing a line item, auto pricing is always disabled by default to avoid unexpected price recalculation.

What seemed like a simple fix was harder than expected: auto_pricing is a transient serializer flag that defaults to true and gets included in GET responses. The frontend's edit form fetches initial data via GET and treats it as the source of truth, so any local state change gets overwritten by the fetched value on every sync cycle - hope I described this correctly. Setting the auto_pricing  flag to false is a bit of a workaround, but it gets the job done, with very little code changes. 

IMHO auto_pricing might be better modeled as an action (e.g., a "Calculate price" button) rather than a toggle. This would make intent explicit, avoid state sync issues, and better match the actual data model where auto_pricing is a one-shot request flag, not persistent state. That's a larger UX change, this fix addresses the immediate bug.


SchrodingersGat:
As you have changed the API schema, you will need to add an entry to `./src/backend/InvenTree/InvenTree/api_version.py`

JustusRijke:
> As you have changed the API schema, you will need to add an entry to `./src/backend/InvenTree/InvenTree/api_version.py`

Done. Having a hard time getting the tests to run reliably in my devcontainer, so falling back to the GitHub action runners for now

SchrodingersGat:
@JustusRijke thanks for the contribution!

Issue #11408, opened 2026-02-23
Notification locale translation


Tr4um4z:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*
I want to change the backend language to recive email notfication in italian.

i have set up lot of configs working in /etc/inventree/config.yaml but language: it-it do not work

`# Select default system language , or use the environment variable INVENTREE_LANGUAGE
language: it-it`

set a env variable in /etc/inventree/conf.d/other , do not work

`export INVENTREE_LANGUAGE="it-it"`

any suggestion?
Thanks in Advance

### Steps to Reproduce

Inventree in a LCX container under proxmox installed from https://community-scripts.github.io/ProxmoxVE/  script

### Expected behaviour

change language to IT

### Deployment Method

None

### Version Information

latest

### Try to reproduce on the demo site

I did not try to reproduce

### Is the bug reproducible on the demo site?

Not reproducible

### Relevant log output

```shell

```

SchrodingersGat:
You have the correct config variable - `INVENTREE_LANGUAGE`.

You can confirm that it is set correctly by going to `/api/` and you should see:

<img width="354" height="230" alt="Image" src="https://github.com/user-attachments/assets/0915aa08-5436-40a2-8319-6c81857994a5" />

So, that part is at least correct. Perhaps there are some other issues going on with the email notifications? Any email messages in particular you are looking at?

Tr4um4z:
![Image](https://github.com/user-attachments/assets/788ea677-bfcd-4218-84fe-877444b04683)

it's seems correct , i have all email (new user , change password , low stock notification) and web notification in english, web frontend in italian from user setting.

it's the backend that is not in italian

![Image](https://github.com/user-attachments/assets/5ba9d1b6-1a27-4c4c-a49b-409880b1d589)

Tr4um4z:
> You have the correct config variable - `INVENTREE_LANGUAGE`.
> 
> You can confirm that it is set correctly by going to `/api/` and you should see:
> 
[...]

PR #11412, opened 2026-02-23
[UI] Default locale

SchrodingersGat:
As part of the investigation for https://github.com/inventree/InvenTree/issues/11408 I discovered that the server default locale is not honored in the user interface.

This PR adds the following:

- Ensure that the UI language initially defaults to the server "default locale" setting
- Add a user profile option to select "default language"

### DIFF
--- a/src/frontend/src/contexts/LanguageContext.tsx
+++ b/src/frontend/src/contexts/LanguageContext.tsx
@@ -176,19 +176,17 @@ export function LanguageContext({
   return <I18nProvider i18n={i18n}>{children}</I18nProvider>;
 }
 
+// This function is used to determine the locale to activate based on the prioritization rules.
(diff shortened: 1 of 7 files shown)


src/frontend/src/contexts/LanguageContext.tsx
```diff
@@ -65,9 +65,30 @@ export function LanguageContext({
...
+    if (!!language) {
+      locale = language;
+    } else if (!!server.default_locale) {
+      locale = server.default_locale;
+    } else {
+      locale = defaultLocale;
+    }
+
+    if (locale != activeLocale) {
+      console.info('Changing locale from', activeLocale, 'to', locale);
```
matmair: Q: is this meant to be left in?

matmair (approved):
LGTM

PR #11437, opened 2026-02-27
Allow Zero Quantity BOM Item with System Setting

jacobfelknor:
Add a system setting that allows a zero-quantity BOM item.

We have a situation where we'd like to use setup quantity to define a quantity required per build, but should be independent of the BO build quantity (e.g. qty 4 of some part required to build 10 items, or 20 items, etc.)

Is this an acceptable way of doing such a thing? It seems to work in our local fork, but I'm worried about other side effects.


matmair (approved):
LGTM

SchrodingersGat:
You will need to add some documentation around this new setting - or the docs CI won't pass

docs/docs/settings/global.md
```diff
@@ -170,6 +170,7 @@ Configuration of label printing:
 {{ globalsetting("PART_SALABLE") }}
 {{ globalsetting("PART_VIRTUAL") }}
 {{ globalsetting("PART_COPY_BOM") }}
+{{ globalsetting("PART_BOM_ALLOW_ZERO_QUANTITY") }}
```
jacobfelknor, line 173: Is this all that is necessary?

SchrodingersGat: That should be sufficient, thanks

SchrodingersGat:
@jacobfelknor thanks for the contribution!

PR #11457, opened 2026-03-04
re-enable codspeed

matmair:
Updates the runner selection logic to only use `codspeed-macro` if the workflow is triggered by a push to the `master` branch or if the pull request is labeled with `performance-run`.


matmair:

PR #11459, opened 2026-03-04
Stock parameters

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/discussions/11423

Adds support for "parameters" against stock items

### Tasks

- [x] Update model
- [x] Update API / serializer
- [x] Add "parameters" panel to StockItem detail
- [x] Add "parameter view" for stock item table?
- [x] Support location relation filtering for stock parameters
- [x] Playwright testing
- [x] Update docs (link "parametric view" back to "parameters)

### DIFF
--- a/src/backend/InvenTree/stock/models.py
+++ b/src/backend/InvenTree/stock/models.py
@@ -400,6 +400,7 @@ class StockItemReportContext(report.mixins.BaseReportContext):
 
 class StockItem(
     InvenTree.models.PluginValidationMixin,
+    InvenTree.models.InvenTreeParameterMixin,
     InvenTree.models.InvenTreeAttachmentMixin,
     InvenTree.models.InvenTreeBarcodeMixin,
     InvenTree.models.InvenTreeNotesMixin,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#10887](https://github.com/inventree/InvenTree/pull/10887) adds the ability to auto-allocate tracked items against specific build outputs. Currently, this will only allocate items where the serial number of the tracked item matches the serial number of the build output, but in future this may be extended to allow for more flexible allocation rules.
 - [#11372](https://github.com/inventree/InvenTree/pull/11372) adds backup metadata setter and restore metadata validator functions to ensure common footguns are harder to trigger when using the backup and restore functionality.
 - [#11374](https://github.com/inventree/InvenTree/pull/11374) adds `updated_at` field on purchase, sales and return orders.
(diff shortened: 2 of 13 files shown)


src/backend/InvenTree/stock/models.py
```diff
@@ -400,6 +400,7 @@ class StockItemReportContext(report.mixins.BaseReportContext):
 
 class StockItem(
     InvenTree.models.PluginValidationMixin,
+    InvenTree.models.InvenTreeParameterMixin,
```
matmair, line 403: I think just applying the default business logic for parameters and parameter template <> stock item is not really intuitive.

As far as I can see this just enabled parameters but no mechanisms to handle the splitting / merging of stock items (as regularly happens) and what should happen to parameters in that case. That leaves holes compared to the default usage flow for stock items.

Was this considered, and should this limitation be documented or addressed?

lifeisafractal: I'll chime in since I started the discussion thread. FWIW, I didn't consider stock splitting/merging. I avoid using merging at all in my workflow due to it destroying history in the merge. Since merging is already destructive, I'd vote that the "winning" stock item retains its original parameters as with the rest of history and document that. That's just my $0.02.

Splitting is an interesting, and I'd say new, case. For my use cases I'd expect split items to inherit their parents' parameters. Now that raises the question is it a one-time copy-on-split, or do child stock items dynamically inherit their parents parameters? I honestly don't have a good opinion on this and it feels like this has nuance depending on use case. Either way, I'd expect stock items to inherit parent parameters via some method.

lifeisafractal: I thought more about this, and my vote is to do a "copy-on-split" of params into the split child stock item. new/changed params on the parent after that time don't reflect in the child. I think that is both most clear to document and implement. Again, just my $0.02.

woju: If I may, as the second requester: merging stock can be blocked altogether if there are custom parameters present that have different values, same as there are options to block for mismatched suppliers and states.

SchrodingersGat:
> As far as I can see this just enabled parameters but no mechanisms to handle the splitting / merging of stock items (as regularly happens) and what should happen to parameters in that case. That leaves holes compared to the default usage flow for stock items.

> Was this considered, and should this limitation be documented or addressed?

Good point, this should be considered further.

This has also highlighted that many other models lack proper management of parameters on duplication (orders, for example) - so I will have to look into that too.

SchrodingersGat:
> If I may, as the second requester: merging stock can be blocked altogether if there are custom parameters present that have different values, same as there are options to block for mismatched suppliers and states.

This will require significant further coding (and testing / validation) to ensure it is working correctly.

PR #11482, opened 2026-03-10
Fix order of hooks for stabilizing table columns

SchrodingersGat:
Ref: https://github.com/icflorescu/mantine-datatable/issues/759

Actually, still not fixed - see https://github.com/icflorescu/mantine-datatable/issues/759#issuecomment-4030837972


matmair (approved):

SchrodingersGat:
Adjustments as per comment in https://github.com/icflorescu/mantine-datatable/issues/759#issuecomment-4148942070

PR #11484, opened 2026-03-10
fix: wrap create_default_labels in transaction.atomic() savepoint to …

### LINKED ISSUE #11469: TransactionManagementError on stock operations when plugin registry reloads during HTTP request

### Describe the bug*

When the plugin registry detects a changed hash and reloads during a live HTTP request, Django re-runs all app ready() methods including report.ReportConfig. Inside create_default_labels() (report/apps.py line 163), LabelTemplate.objects.create() is called for label templates whose files already exist on disk. InvenTree's own check_existing_file() validator in report/models.py raises ValidationError.
Although a try/except ValidationError block is present, this is not sufficient on PostgreSQL. In PostgreSQL, any database error raised inside an atomic() block marks the entire transaction as aborted.  Even if the Python exception is caught. All subsequent queries in that same request then fail with TransactionManagementError, causing HTTP 500 on every stock write operation (add, count, transfer). Even InvenTree's own exception logger fails with the same error, preventing the error from being recorded normally.

### Steps to Reproduce

1. Have a plugin write a setting (set_setting()) during its initialization, this causes the plugin registry hash to change on every HTTP request

1. With the server running, perform any stock write operation: Stock > Add Stock, Stock Count submission, or Transfer

1. The registry detects a hash change mid-request and reloads

1. create_default_labels() attempts to INSERT label templates that already exist on disk

1. ValidationError is raised and caught in Python, but PostgreSQL transaction is now aborted

1. All subsequent DB queries in the same request fail with TransactionManagementError

1. The stock operation returns HTTP 500 and rolls back

### Expected behaviour

A duplicate template INSERT failure inside create_default_labels() should be fully isolated and should not affect any other database transaction running in the same request. Stock operations should complete successfully regardless of whether the plugin registry reloads during the request.

Wrap the LabelTemplate.objects.create() call in a nested transaction.atomic() savepoint. If the INSERT fails, only the savepoint rolls back — the outer request transaction remains intact.
Add import if not already present:
from django.db import transaction
Change in create_default_labels() (~line 163):
# Before:
try:
    report.models.LabelTemplate.objects.create(
        **template, template=self.file_from_template('label', filename)
    )
except ValidationError:
    logger.warning("Could not create label template: '%s'", template['name'])
 
# After:
try:
    with transaction.atomic():
        report.models.LabelTemplate.objects.create(
            **template, template=self.file_from_template('label', filename)
        )
except ValidationError:
    logger.warning("Could not create label template: '%s'", template['name'])
Note: The same pattern should be reviewed and applied to create_default_reports() if a similar structure exists there.

### Deployment Method

Bare metal

### Version Information

V1.2.3

This bug requires a specific trigger condition: a plugin that writes a setting unconditionally during initialization, causing the registry hash to change on every HTTP request. This condition would need to be deliberately induced on a demo site. The underlying latent bug in create_default_labels() can be confirmed by:
Enabling a PostgreSQL backend on the demo site
Manually calling registry.reload_plugins() inside an active atomic() block via the Django shell
Observing that any subsequent DB query raises TransactionManagementError

I did not try to reproduce

Not reproducible

### Relevant log output

```shell
Error event — 2026-03-08T02:16:16Z (request_id: fd36d6f5)
 
Step 1: create_default_labels() raises ValidationError
File "report/apps.py", line 163, in create_default_labels
    report.models.LabelTemplate.objects.create(
File "report/models.py", line 69, in rename_template
    instance.check_existing_file(path, raise_error=True)
File "report/models.py", line 127, in check_existing_file
    raise ValidationError({'template': ['Template file with this name already exists']})
 
Step 2: Outer transaction now aborted. Next query in apps.py fails:
File "InvenTree/apps.py", line 106, in remove_obsolete_tasks
    if obsolete_tasks.exists():
django.db.transaction.TransactionManagementError: An error occurred in the
current transaction. You can't execute queries until the end of the 'atomic' block.
 
Step 3: Even the exception logger fails with the same error:
File "InvenTree/exceptions.py", line 92, in log_error
    Error.objects.create(kind=kind, info=info or '', data=data or '', path=path)
django.db.transaction.TransactionManagementError: An error occurred in the
current transaction. You can't execute queries until the end of the 'atomic' block.
```

tigger2000ttfn:
…prevent PostgreSQL transaction abort

When the plugin registry reloads during an active HTTP request,  create_default_labels() raises ValidationError on duplicate template  inserts. On PostgreSQL this aborts the entire outer atomic() block,  causing TransactionManagementError on all stock operations in that request.

Fixes #11469


SchrodingersGat:
@tigger2000ttfn thanks for the contribution!

PR #11487, opened 2026-03-11
fix style issue from 11485

matmair:
see https://github.com/inventree/InvenTree/pull/11485

Dependabot PRs need to be reviewed; see https://github.com/inventree/InvenTree/pull/11445


matmair:

SchrodingersGat:
@matmair thanks - is there a way we can improve the dependabot PRs here? This issue keeps happening

matmair:
Not really, there are some mentions of better support for uv from the dependabot team but nothing released yet.
the issue is easy to spot in review: there should be no change to requirements.txt without the same also applied to requirements-3.14.txt and vice versa

PR #11492, opened 2026-03-11
fix: low-privilege user token creation

### LINKED ISSUE #11486: [bug] Users cannot create their own API tokens

### Describe the bug*

Any non-admin user does not have the required API role permissions to generate their own API tokens. This prevents users from generating API tokens unless they have "admin" (role permissions).

Image

### Steps to Reproduce

1. Login using non-admin account (e.g. user: `reader`, pass: `readonly`)
2. Navigate to "User Settings" / "Security"
3. Select "Access Tokens" -> "Add Token"
4. Get "Permission Denied" error

### Expected behaviour

Users should be able to control their own API tokens

### Relevant Code

from `./src/backend/InvenTree/InvenTree/metadata.py:InvenTreeMetadata:determine_metadata` (lines 133-140):

```python
            # Remove any HTTP methods that the user does not have permission for
            for method, permission in rolemap.items():
                result = check_user_permission(user, self.model, permission) or (
                    role_required and check_user_role(user, role_required, permission)
                )

                if method in actions and not result:
                    del actions[method]
```

Note here that the `users_apitoken` model requires the `RuleSetEnum.ADMIN` role permission:

```python
def get_ruleset_models() -> dict:
    """Return a dictionary of models associated with each ruleset.

    This function maps particular database models to each ruleset.
    """
    ruleset_models = {
        RuleSetEnum.ADMIN: [
            'auth_group',
            'auth_user',
            'auth_permission',
            'users_apitoken',
            ...
```

Which only certain groups of users should have.

matmair:
Switches frontend user token generation to the intended user-specific endpoint and fixes metadata permission discovery for IsAuthenticated permissioned endpoints

fixes #11486


SchrodingersGat (approved):

SchrodingersGat:
@matmair nice work and fix. Can you please add a quick playwright test to prevent a regression here?

matmair:

SchrodingersGat:
Thanks for the fix @matmair

PR #11500, opened 2026-03-12
[db] Backend setting improvements

SchrodingersGat:
Some improvements / refactoring for supported database backends.

### Documentation

- Fixes some spelling mistakes in the documentation - `KEEPALIVE_INTERNAL` should be `KEEPALIVE_INTERVAL`
- Add warnings about using sqlite in production

### Refactoring

- Pull database config code out of `settings.py` into own file - for clarity.

### SQLite improvements

- Adds default configuration values to improve sqlite database locking / concurrency issues
- Required to fix issues with large atomic processes (e.g. completing a shipment with a high number of items)


SchrodingersGat:
I have tested the code changes on a sqlite setup, which fails when trying to complete a shipment with a large number of allocated items (~500).

In such a case, the new default values for the database connection allow successful completion of the task - whereas it fails with the previous setup.

I have added a unit test here - https://github.com/inventree/InvenTree/pull/11503

**However** - I cannot get it to "fail" when running in the main test thread - it only seems to become an issue when running within django-q2 / background worker threads

matmair:
@SchrodingersGat I am not sure concurrent access from worker and server works 100% with any setting with sqlite as it is file based and there is always the option for a long blocking action either from a long task in a worker or the frontend.

matmair (approved):

PR #11512, opened 2026-03-13
Fix table for s3 backup in docs

Knirps1:
The default value for some options for s3 backup configuration are on a seperate line. This moves them to the correct position.


SchrodingersGat:
Thanks for the fix @Knirps1

PR #11520, opened 2026-03-13
refactor(frontend): use central user creds for testing

matmair:
this is helpful when testing against an instance with more secure passwords.


SchrodingersGat (approved):

SchrodingersGat:
Nice addition, thanks @matmair

PR #11536, opened 2026-03-16
fix(plugin): use app_name instead of plugin_path when deregistering models

nino-tan-smartee:
## Summary

- `_deactivate_mixin` uses `plugin_path` (full dotted module path) as the key into `apps.all_models` when removing plugin models during reload
- Django registers models under the `app_label` (short `app_name`), not the full `plugin_path`
- For plugins with nested module paths (e.g. `myplugin.myplugin`), `plugin_path != app_name`
- Since `apps.all_models` is a `defaultdict`, looking up `plugin_path` silently creates an empty `OrderedDict`, then `.pop(model)` raises `KeyError`
- This causes recurring `KeyError` crashes every plugin reload cycle (~1 minute) for any external plugin with a nested package structure

## Fix

- Use `app_name` (already computed at line 98) instead of `plugin_path` on line 118
- Add default `None` to `.pop()` for defensive safety
- Consistent with line 123 which already correctly uses `app_name`

## Reproduction

1. Create an external plugin with a nested module structure (e.g. `purchase_quotation/purchase_quotation/`)
2. Install and activate the plugin
3. Observe recurring `KeyError: '<model_name>'` in error reports every plugin reload cycle

## Test plan

- [ ] Verify external plugins with nested module paths no longer crash during reload
- [ ] Verify built-in plugins continue to work correctly (no regression)
- [ ] Verify plugin model cleanup still works (models are properly deregistered)

```python
from setuptools import find_packages, setup

setup(
    name="inventree-purchase-quotation",
    version="0.1.0",
    description="RFQ and Purchase Quotation management for InvenTree",
    author="....",
    packages=find_packages(),
    entry_points={
        "inventree_plugins": [
            "PurchaseQuotationPlugin = "
            "purchase_quotation.plugin:PurchaseQuotationPlugin",
        ],
    },
    install_requires=[
        ...
    ],
)

```

### DIFF
--- a/src/backend/InvenTree/plugin/base/integration/AppMixin.py
+++ b/src/backend/InvenTree/plugin/base/integration/AppMixin.py
@@ -113,9 +113,14 @@ class AppMixin:
                 break
 
             # unregister the models (yes, models are just kept in multilevel dicts)
+            # Note: Django registers models under the app_label (app_name),
+            # not the full dotted plugin_path. For plugins with nested module
+            # paths (e.g. "myplugin.myplugin"), plugin_path != app_name, so
+            # using plugin_path here would look up the wrong key in the
+            # defaultdict and raise KeyError on .pop().
             for model in models:
                 # remove model from general registry
-                apps.all_models[plugin_path].pop(model)
+                apps.all_models[app_name].pop(model, None)
 
             # clear the registry for that app
             # so that the import trick will work on reloading the same plugin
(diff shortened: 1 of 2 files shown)


matmair:
@nino-tan-smartee thank you for the PR; please add a unit test that ensures the fixed issues does not appear again

nino-tan-smartee:
> @nino-tan-smartee thank you for the PR; please add a unit test that ensures the fixed issues does not appear again

@matmair I added the unit test

matmair (approved):
Thank you for this PR; good stuff

SchrodingersGat:
@nino-tan-smartee please confirm when you have successfully completed your test plan!

nino-tan-smartee (commented):
Test plan confirmed — all items verified. Ready to merge.

SchrodingersGat:
@nino-tan-smartee thanks for confirming

PR #11537, opened 2026-03-16
chore(backend): Bump ty

matmair:
Bumps ty as there is better support for django types now


src/backend/InvenTree/plugin/plugin.py
```diff
@@ -92,16 +92,16 @@ class MetaBase:
     TITLE = None
 
     @mark_final
-    def get_meta_value(self, key: str, old_key: Optional[str] = None, __default=None):
+    def get_meta_value(self, key: str, old_key: Optional[str] = None, default=None):
```
matmair, line 95: @SchrodingersGat this is technically breaking, but I only think this was every used internally

SchrodingersGat (approved):

PR #11539, opened 2026-03-16
Focus+select the quantity input in the stock add/count/remove dialog

### LINKED ISSUE #11511: [FR]Set focus and select the quantity field in the Add/remove stock dialog

### Problem statement

When opening this dialog:

Image

In almost every case you want to enter a number and hit submit (maybe you add notes, but it clearly happens after you entered the number).

It would help the overall UX if the 0 number would be always selected and would have focus.

### Suggested solution

Set the focus and selection when this dialog opens.

Manually selecting the input :)

### Examples of other systems

_No response_

- [x] I want to develop this.

martonmiklos:
Initial PR for review.
The selectAndFocus attribute is only propagated to the Numeric input types.
Shall I implement it for all the input types?
Should I separate them?

Maybe the implementation should be done in two PRs: one for providing the select and focus "API" and the second for the stock dialogs?

Fixes #11511


src/frontend/src/components/forms/fields/NumberField.tsx
```diff
@@ -14,7 +14,9 @@ export default function NumberField({
   placeholderAutofill,
   placeholderWarningCompare,
   placeholderWarning,
-  onChange
+  onChange,
+  selectAndFocus = false,
+  apiTriggerValue
```
SchrodingersGat: What is the purpose of `apiTriggerValue` here?

SchrodingersGat:
@martonmiklos please address my question as above

SchrodingersGat:
@martonmiklos update?

martonmiklos:
> @martonmiklos update?

I will dedicate some time to InvenTree this weekend + Monday and will get this through.

src/frontend/src/components/forms/fields/NumberField.tsx
martonmiklos: It is a residue from a former attempt. Not needed anymore, removed it. Thanks for spotting it!

src/frontend/src/forms/StockForms.tsx
```diff
@@ -1299,7 +1306,17 @@ export function useAddStockItem(props: StockOperationProps) {
       <Alert color='blue'>
         {t`Increase the quantity of the selected stock items by a given amount.`}
       </Alert>
-    )
+    ),
+    onOpen: () => {
+      setTimeout(() => {
+        const input = document.querySelector<HTMLInputElement>(
+          '#number-field-quantity, input[aria-label="number-field-quantity"]'
```
matmair, line 1313: Not sure if that is the best way to solve it as i can see these label changing in the future but I also do not have a better proposal; thoughts @SchrodingersGat ?

SchrodingersGat: It seems like a bit of an anti-pattern to me - we should be able to set it with normal properties, not trying to introspect the DOM

SchrodingersGat:
@martonmiklos please review and update based on requested changes

PR #11540, opened 2026-03-16
Implement Latest parts dashboard widget

martonmiklos:


SchrodingersGat:
@martonmiklos nice idea - can you please provide a screenshot of the new widget?

martonmiklos:
> @martonmiklos nice idea - can you please provide a screenshot of the new widget?

Sure:
kép


There are 3 things about this which annoys me a bit:

If I add the width is not correct:
kép

After hitting F5 it got to a state as in the upper screenshot.

The another bit annoying thing is this warning:

kép

Maybe I would need to add the ID it to the tables filtering capabilities?

SchrodingersGat:
Ah ok, I thought the widget was actually going to display the set of latest parts, but that's fine :)

> If I add the width is not correct:

There are some outstanding state issues with the dashboard, I have not had any time to look at it - and it is way down on the pile. If you can work it out, please submit a patch!

> The another bit annoying thing is this warning:

What is annoying about this? It is intended to show that the current table filters are defined by the path parameters, and the custom table filtering icon is disabled in this case

martonmiklos:
> What is annoying about this? It is intended to show that the current table filters are defined by the path parameters, and the custom table filtering icon is disabled in this case

Well technically I understand this, however on my notebook screen the content for what I clicked on the dasboard starts only at the middle of the page:

kép

Additionally the sidemenu items does not lead to any meaningful/associatable places.
Maybe we should create a new view for the latest part where only the part list is being shown?
Same could be reused for the subscribed parts and other part related shortcut dashboard links.

wolflu05:
What about combining these two features, so that clicking on a link with query params adds these just as normal filters, and adding normal filters adds them to the url, so that sharing of table views works a lot smoother. In my opinion this warning alert is just distracting, and the 1 above the filter icon is just enough and well known from other solutions to visually indicate that there are filters active.

martonmiklos:
> What about combining these two features, so that clicking on a link with query params adds these just as normal filters, and adding normal filters adds them to the url, so that sharing of table views works a lot smoother. In my opinion this warning alert is just distracting, and the 1 above the filter icon is just enough and well known from other solutions to visually indicate that there are filters active.

I do agree, but let's fix things one by one (the same alert is shown in the already existing Subscribed parts widget).

wolflu05:
Yeah, definitely, these were just my thoughts about this topic how I imagine it at the end, not sure what others think about this.

martonmiklos:
@SchrodingersGat 
What's the plan with this one? I do not think that adding more feature this PR makes sense, I can improve the redirection with new PRs later.
Shall I fix the conflicts?

SchrodingersGat:
@martonmiklos yes please address conflicts here

SchrodingersGat (approved):

SchrodingersGat:
@martonmiklos nice work, thanks for the addition!

PR #11565, opened 2026-03-19
[plugin] Cache busting for plugin static files

SchrodingersGat:
Plugins can provide files which are copied into the "static" directory for serving. Often these plugins are generated using the plugin creator tool, which (until this point) have bundled javascript files using non-hashed names.

This means that if a new version of the plugin is installed, the file names will be the same, and caching (either at the browser level or proxy level) can prevent the new version of the file from being properly loaded.

This PR adds support for "cache busting" - by trying to find the "latest" version of the plugin and returning that version of the file to the frontend.

While his will require updates to any existing plugins to make use of - the default behaviour of "non hashed files" is retained, such that any existing plugins will still function.

### Functionality

1. Look for a vite manifest file
2. Look in this file for an entry mapping to the requested file
3. If available, return the hashed version of the file

### References

https://github.com/inventree/plugin-creator/pull/90

### TODO

- [x] Add unit tests with and without manifest files
- [x] Improve generic staticfiles support


src/backend/InvenTree/plugin/plugin.py
```diff
@@ -578,35 +582,122 @@ def define_package(self):
...
+            *args: Path components to the static file (e.g. 'js', 'admin.js')
+
+        Returns:
+            str: Path to the hashed version of the file if it exists, else None
+        """
+        storage = StaticFilesStorage()
+
+        # First, try to find a manifest file which maps original filenames to hashed filenames
+        # We only support vite manifest files - as generated by the plugin creator framework
+        manifest_file = os.path.join(*self.get_static_path(), '.vite', 'manifest.json')
```
matmair: I would prefer usage of path lib where possible

src/backend/InvenTree/plugin/plugin.py
```diff
@@ -578,35 +582,122 @@ def define_package(self):
...
+            with storage.open(manifest_file) as f:
+                manifest = json.load(f)
+        except json.JSONDecodeError:
+            logger.error(f"Failed to parse manifest file for plugin '{self.SLUG}'")
+            return None
+
+        # Find the entry associated with the requested file
+        # Remove the file extension, as the manifest may contain hashed files with different extensions (e.g. .js, .css)
+        filename = str(args[-1] or '').split('.')[0]
+        pattern = re.compile(rf'{re.escape(filename)}\.(js|jsx|tsx)')
```
matmair, line 619: This could be pre-compiled once during startup

src/backend/InvenTree/plugin/plugin.py
```diff
@@ -578,35 +582,122 @@ def define_package(self):
...
+
+        if check_exists:
+            storage = StaticFilesStorage()
+
+            if not storage.exists(full_path):
+                logger.error(
+                    f"Static file not found for plugin '{self.SLUG}': {full_path}"
+                )
+
+        url = os.path.join(settings.STATIC_URL, full_path)
```
matmair: Have you confirmed this works with S3?

SchrodingersGat: Not specifically, this behavior is not changed from current master though

matmair: I am not sure doing a path join on a url is working with pre-signed s3 urls

SchrodingersGat: Will pathlib handle this?

matmair: no, you need to get the url from the storage backend

SchrodingersGat: Ok, I have made this generic now. An improvement over previous behavoir

matmair:
Great changeset!

PR #11577, opened 2026-03-20
Support image upload from clipboard

martonmiklos:
Title says all. I tend to upload screenshotted images to 3D modelled parts what I needed to save to file so far. With this change we can upload it instantly from the clipboard.

### DIFF
--- a/src/frontend/src/components/details/DetailsImage.tsx
+++ b/src/frontend/src/components/details/DetailsImage.tsx
@@ -19,7 +19,7 @@ import {
 } from '@mantine/dropzone';
 import { useHover } from '@mantine/hooks';
 import { modals } from '@mantine/modals';
-import { useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 
 import { ActionButton } from '@lib/components/ActionButton';
 import type { UserRoles } from '@lib/enums/Roles';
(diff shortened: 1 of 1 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@martonmiklos nice work, this seems to work well. Some small requests below:

### Discoverability

Unless you know that this feature exists, it's very hidden. I'd suggest that in the "Upload Image" dialog, you can add a friendly `<Alert />` message above the "drag and drop" field - with a message something like:

`Drag and drop an image, or paste an image from the clipboard`

### Testing

It would be great to add a playwright test to ensure this feature does not regress in the future.

martonmiklos:
>  nice work, this seems to work well.

Well most of the credits goes for ChatGPT, I am still not into React to a level to implement things like this by myself. 

> 
Unless you know that this feature exists, it's very hidden. I'd suggest that in the "Upload Image" dialog, you can add a friendly <Alert /> message above the "drag and drop" field - with a message something like:

I agree. Instead of having an alert extending the existing message would be acceptable for you?
kép

> It would be great to add a playwright test to ensure this feature does not regress in the future.

Absolutely!

SchrodingersGat:
> Instead of having an alert extending the existing message would be acceptable for you?

That looks fine!

SchrodingersGat:
@martonmiklos thanks for the addition!

martonmiklos:
> @martonmiklos thanks for the addition!

I still owe you with the playwright test 😆

PR #11602, opened 2026-03-24
Restrict queryset for DataImportSession

SchrodingersGat:
- Only allow non-staff users to see their own sessions
- Follow-up to https://github.com/inventree/InvenTree/pull/11580

### Tasks

- [ ] Add API unit tests

### DIFF
--- a/src/backend/InvenTree/importer/api.py
+++ b/src/backend/InvenTree/importer/api.py
@@ -80,7 +80,7 @@ class DataImportSessionMixin:
         try:
             user = self.request.user
         except AttributeError:
-            return importer.models.DataImportSession.objects.none()
+            raise PermissionDenied('User information is not available')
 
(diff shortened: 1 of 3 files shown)


src/backend/InvenTree/importer/api.py
```diff
@@ -73,6 +73,22 @@ class DataImportSessionMixin:
...
     permission_classes = [InvenTree.permissions.DataImporterPermission]
 
+    def get_queryset(self):
+        """Return the set of DataImportSession objects that the user has permission to view."""
+        queryset = super().get_queryset()
+
+        try:
+            user = self.request.user
+        except AttributeError:
+            return importer.models.DataImportSession.objects.none()
```
matmair: Should we rather throw an error here to be more clear of what is happening?

matmair (approved):
LGTM

PR #11611, opened 2026-03-27
feat(backend): add inventree version to every response

matmair:
Having a header with the version on response objects helps with making client-side checks for correct parsing easier and can help reduce the number of required calls in stateless CLI usage as there is no need to call /api/ to determine API version.


SchrodingersGat (approved):

SchrodingersGat:
What about adding the API version also?

matmair:
Done @SchrodingersGat

SchrodingersGat (approved):

SchrodingersGat:
LGTM, nice addition thanks @matmair

PR #11612, opened 2026-03-27
fix(frontend): complete zh-Hans UI localization

amm10090:
## Summary
- fill in all missing `zh_Hans` frontend messages and bring the current catalog coverage to `2142 / 2142`
- normalize manufacturing and inventory terminology across the frontend catalog, including `Build`, `Stocktake`, and `Shipment`
- move remaining user-visible hardcoded frontend strings into Lingui, including language loading errors, host labels, stock location scan actions, and subscription notifications

## Verification
- `docker compose -f .devcontainer/docker-compose.yml run --rm --entrypoint /bin/bash inventree -lc 'cd /home/inventree/src/frontend && yarn compile && yarn build'`\n- confirmed `zh_Hans` frontend catalog coverage is `2142 / 2142`\n\n## Notes\n- this PR only includes frontend localization and source-string updates\n- Crowdin-managed localization branches were not edited directly\n- local devcontainer and environment setup changes were intentionally excluded


src/frontend/src/contexts/LanguageContext.tsx
```diff
@@ -165,8 +166,7 @@ export function LanguageContext({
   if (loadedState === 'error') {
     return (
       <Text>
-        An error occurred while loading translations, see browser console for
-        details.
+        {t`An error occurred while loading translations, see browser console for details.`}
```
matmair: This can be no text if loading translations fails. So defeats the point of having the text to translate this one

SchrodingersGat (changes requested):
The changes to the untranslated UI strings are great but please remove changes to the .po files - these should be handled via our crowdin service

src/frontend/src/contexts/LanguageContext.tsx
wolflu05: have you made sure that this actually works here, as this is outside of the LanguageContext and rendered when loading the translations has failed

amm10090:
Implemented in a3fc1b1cfd.

 review feedback:
- Removed all .po file changes from this PR (these should indeed come via Crowdin)
- Kept only frontend source string fixes in TS/TSX files
- Updated LanguageContext error fallback to plain text (no translation lookup in the translation-load-failure path)

Please take another look.

src/frontend/src/contexts/LanguageContext.tsx
amm10090: Good catch — you are right.

I updated this in a3fc1b1cfd: the error fallback in LanguageContext is now plain text (not wrapped with translation macros), so it does not depend on translation context during translation-load failure.

matmair (approved):
Looks good

SchrodingersGat (approved):

SchrodingersGat:
@amm10090 thanks for the contribution!

PR #11613, opened 2026-03-27
fix(i18n): compile backend translations during update

amm10090:
## Summary
- add a dedicated `invoke backend-trans` task to compile backend Django translations
- run backend translation compilation as part of `invoke update`
- keep backend `django.mo` files in sync during normal update / deployment flows

## Why
The backend already ships `zh_Hans` translation sources for settings metadata, but these translations do not take effect at runtime unless `compilemessages` is run manually. As a result, pages such as System Settings can still show English labels and descriptions even when the UI language is set to Simplified Chinese.

## Verification
- `python3 -m py_compile tasks.py`
- `docker exec inventree_devcontainer-inventree-1 bash -lc 'cd /home/inventree && . dev/venv/bin/activate && invoke backend-trans && test -f /home/inventree/src/backend/InvenTree/locale/zh_Hans/LC_MESSAGES/django.mo'`\n- verified compiled backend translations resolve `Base URL`, `Company name`, `Server Instance Name`, and `Update Check Interval` under `zh-hans`\n\n## Notes\n- this PR intentionally keeps the change scoped to the existing update workflow\n- no Crowdin-managed localization branches were edited directly


matmair:
I think this misses the point that running translation compilation is expensive on some platforms and that docker images for example come with pre-ran translations.
This should be either optional or decided based upon the deployment platform

tasks.py
```diff
@@ -922,6 +932,9 @@ def update(
     # Ensure required components are installed
     install(c, uv=uv)
 
+    # Compile backend translation files to ensure django.mo stays in sync
+    backend_trans(c)
```
matmair: This should not be run every time on ever platform, only when needed / optional. Similar to frontend compilation

amm10090:
Implemented the requested change in dabbba271f.

`invoke update` no longer runs backend translation compilation unconditionally.

Current behavior:
- Skip backend compilation by default in docker environments
- Use `--backend` to force backend compilation
- Use `--no-backend` to skip explicitly
- Keep `invoke backend-trans` as the explicit command when needed

Also updated the `update` task help/docstring to mark backend translation compilation as optional.

tasks.py
amm10090: Agreed, and addressed in dabbba271f.

I changed `invoke update` so backend translation compilation is now optional / platform-aware:
- default behavior: skip backend compilation in docker environments (`INVENTREE_DOCKER`)
- `--backend`: force compilation during update
- `--no-backend`: explicitly skip compilation
- `invoke backend-trans` remains available as an explicit standalone command

This mirrors the existing frontend update pattern and avoids paying compile cost on platforms where translations are already precompiled.

matmair:
@amm10090 it seems like you added changes that were not intended for the public/upstream. Please remove them

matmair:
I have removed the publishing actions from this PR; should be good to merge now

SchrodingersGat:
@matmair I think you need to re-approve for merge

SchrodingersGat:
@amm10090 thank you for the contribution!

PR #11617, opened 2026-03-27
refactor(backend): move various endpoints to modelviewsets

matmair:
This makes it easier to have uniform URL names and behaviors; also adds a bit more testing and cleans up some path docstrings.

Still to figure out:
- [x] bulk delete only on list endpoints
- [x] some wired API behaviors are hard to refactor - keep them or use this to clean them up? -> added todo for now

Todo when stable:
- [ ] document how to add API endpoints the correct way


matmair:
@SchrodingersGat would appreciate a look here. This
a) makes it harder (more explicit) to declare API endpoint that do not follow a uniform style
b) basically replaces the `mixin + list + detail view ... url defintion` pattern with `viewset + urlregistration` keeping code fragments together
c) enables us to move url definition patterns into one singe route generator instead of distributed url definitions

As seen this can be done in small steps and is still very similar to current patterns / allows reuse of current view mixins

SchrodingersGat:
@matmair I think that this is a good move overall. Would we be looking to move a greater portion of our endpoints to this setup? Or just a small number to start with?

matmair:
my goal would be to move all model/object based API endpoint to viewsets and keep view/function based endpoints only for stand-alone actions.

As that would be a large diff and there are small decisions to be taken regularly regarding abnormlaities in the current API I think batching this over a longer time is the way to go

SchrodingersGat:
@matmair I'd like to get 1.3.0 out soon (as there are some major fixes that people are waiting on) - so I'll bump this to 1.4.0

matmair:
one permission todo, then this is ready

matmair:
@SchrodingersGat this is ready for review and merge. I have identified and documented various inconsistent API details, those are kept for now. My plan would be to do a separate PR to change these (breaking) and discuss details there then to keep this one focused on the pure refactor

SchrodingersGat (approved):

SchrodingersGat:
Nice, this is a great refactor and much cleaner code. Please address the conflict and then I'm happy to merge

matmair:

PR #11630, opened 2026-03-30
Enhancements for recort import/export

SchrodingersGat:
This PR is the first step towards supporting sequential data import to allow migration of a setup which has plugins which have defined database models - https://github.com/inventree/InvenTree/issues/11235

- Adds "metadata" functionality to the `import_records` and `export_records` commands
- Allows for validation of imported data to check that it matches against the currently running setup
- Required for https://github.com/inventree/InvenTree/issues/11235
- Similar to checks implemented in https://github.com/inventree/InvenTree/pull/11372

### During Export

- Custom post-processing step adds metadata to the exported json file

### During Import

- Pre-processing step strips the metadata from the imported json file
- Compares the compatibility of the imported data file to the currently running setup

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -45,6 +45,51 @@ def is_debug_environment():
     )
 
 
+def get_django_version():
+    """Return the current Django version."""
+    from src.backend.InvenTree.InvenTree.version import (
+        inventreeDjangoVersion,  # type: ignore[import]
+    )
+
+    return inventreeDjangoVersion()
+
+
+def get_inventree_version():
+    """Return the current InvenTree version."""
+    from src.backend.InvenTree.InvenTree.version import (
+        inventreeVersion,  # type: ignore[import]
+    )
+
+    return inventreeVersion()
+
+
+def get_inventree_api_version():
+    """Return the current InvenTree API version."""
+    from src.backend.InvenTree.InvenTree.version import (
+        inventreeApiVersion,  # type: ignore[import]
+    )
+
+    return inventreeApiVersion()
+
+
+def get_commit_hash():
+    """Return the current git commit hash."""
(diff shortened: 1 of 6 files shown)


tasks.py
```diff
@@ -414,7 +472,23 @@ def manage(c, cmd, pty: bool = False, env=None):
...
     """
-    run(c, f'python3 manage.py {cmd}', manage_py_dir(), pty, env)
+    return run(
+        c, f'python3 manage.py {cmd}', manage_py_dir(), pty=pty, env=env, **kwargs
+    )
+
+
+def installed_apps(c) -> list[str]:
+    """Returns a list of all installed apps, including plugins."""
+    result = manage(c, 'list_apps', pty=False, hide=True)
```
matmair, line 482: Why not directly handoff the data? Going over the shell just adds failure states;
Via call_command

SchrodingersGat: Because `tasks.py` is not within the scope of the `manage.py` script, it does not operate within the correct django environment:

e.g.

```python
def installed_apps(c) -> list[str]:
    """Returns a list of all installed apps, including plugins."""
    
    from django.core.management import call_command

    output = call_command('list_apps', stdout=subprocess.PIPE).getvalue().decode('utf-8').strip()
```

throws:

```python
  File "/home/inventree/dev/venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 110, in call_command
    app_name = get_commands()[command_name]
               ~~~~~~~~~~~~~~^^^^^^^^^^^^^^
KeyError: 'list_apps'
```

matmair: Pretty sure this will break in unexpected ways depending on the local shell environment but we will see

PR #11631, opened 2026-03-30
Support physical units for BOM lines

### LINKED ISSUE #4922: [Feature] Support physical units for BOM lines
Now that we support physical units, this could be included for BOM items also.

If the internal unit for a wire is "metres", the BOM could specify inches / mm / etc

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/4922

This PR implements support for physical units in BOM items.

### Breaking Change

The "quantity" field on the BOMItem model is now calculated based on the provided "raw_amount" field - any plugins which are trying to write directly to the "quantity" field will experience errors.

Plugins should either write to the "raw_amount" field directly, or use the new "set_quantity" helper method on the BOMItem model.

### API Change

This PR introduces a new "raw_amount" field to the BomItem API - which is preferenced over the "quantity" field. However, if only the "quantity" field is provided, it will be used, to allow backwards compatibility with existing systems.

### Tasks

- [x] Data / schema migrations
- [x] Update serializers and API endpoints
- [x] Continue to allow write to "quantity" field (if "raw_amount" field not present)
- [x] Logic for validating raw amount
- [x] Update existing unit tests
- [x] Additional unit tests for physical unit / quantity validation
- [x] Additional playwright tests
- [x] Update documentation
- [x] Update CHANGELOG.md
- [x] Update "wireviz" plugin to allow specification in raw amount values - https://github.com/inventree/inventree-wireviz/pull/60
- [x] Perform data migration on demo dataset - https://github.com/inventree/demo-dataset/pull/111
- [x] Add non-numeric raw-amount values to demo dataset (for testing) - https://github.com/inventree/demo-dataset/pull/108


src/frontend/src/forms/BomForms.tsx
```diff
@@ -38,7 +38,10 @@ export function bomItemFields({
       },
       addCreateFields: newPartFields
     },
-    quantity: {},
+    raw_amount: {
+      label: t`Quantity`,
+      description: t`Raw quantity of the required part`
```
matmair: I think "raw" is an implementation detail the user does not need to know and would remove it here

src/frontend/tests/customization/customization.spec.ts
```diff
@@ -21,13 +19,11 @@ test('Customization - Splash', async ({ page }) => {
...
 
-test('Customization - Logo', async ({ page }) => {
-  await doLogin(page, {
-    user: noaccessuser
-  });
+// TODO: Implement this test
+// test('Customization - Logo', async ({ page }) => {
+//   await doLogin(page, {
+//     user: noaccessuser
+//   });
```
matmair: Q: Why not just remove it?

matmair (approved):
LGTM; I am not sure if showing "raw" quantity and quantity to the user is an implementation detail the user needs to see long-term

PR #11641, opened 2026-03-31
Shipment parameters

SchrodingersGat:
Add support for parameters against sales order shipments

### DIFF
--- a/.github/workflows/qc_checks.yaml
+++ b/.github/workflows/qc_checks.yaml
@@ -706,7 +706,7 @@ jobs:
           npm: true
           install: true
           update: true
-          apt-dependency: postgresql-client libpq-dev
+          apt-dependency: gettext postgresql-client libpq-dev
           pip-dependency: psycopg2
       - name: Set up test data
         run: |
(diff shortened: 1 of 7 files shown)


SchrodingersGat:
@matmair the API schema test here is again failing with the following message:

image

```
WARN: diff exceeds the 1MB limit, truncating output...
```

Can you remember what the fix for this was last time?

matmair:
That just means there is a lot of diff; manual review of the schema changes (the new schema is attached to the run) and ignore

matmair (approved):

matmair:
The diff is just to big to handle for GitHub actions which has a hard limit at 1mb; please don’t touch this part of the pipeline

SchrodingersGat:
@matmair I will revert this - was just checking if the format was the issue.

I am confused as to how the diff is so large with only this minimal change. I will have to verify it locally

matmair:
I have looked into the API changes; there seems like something unrelated is going on regarding the API.
The real diff is only a few lines long and looks ok

matmair:
@SchrodingersGat fix in https://github.com/inventree/InvenTree/pull/11649

SchrodingersGat:
@matmair thanks for the associated fix!

PR #11643, opened 2026-03-31
Enforce deterministic ordering for plugin settings

SchrodingersGat:
- Plugin settings display in a "random" order in the UI
- Unlike built-in settings we cannot enumerate them ahead of time
- Editing a setting can cause the order of displayed settings to change
- This PR enforces a standard enumeration order in the API (based on KEY property)

### DIFF
--- a/src/backend/InvenTree/machine/api.py
+++ b/src/backend/InvenTree/machine/api.py
@@ -97,6 +97,10 @@ class MachineSettingList(APIView):
             )
             all_settings.extend(list(settings_dict.values()))
 
+        # Sort settings by the 'key' field before returning,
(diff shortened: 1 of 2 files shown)


wolflu05:
Ideally I think the order should be the same as defined in the plugin. Since dicts are ordered in python but the order gets lots when serializing?? we could also think about adding an extra key order and sort them on the frontend? Or maybe just completly switch to an array response shape?

SchrodingersGat:
@wolflu05 good point, the settings could be returned in the order they are defined in the plugin itself

SchrodingersGat:
see https://github.com/inventree/InvenTree/pull/11646

PR #11646, opened 2026-03-31
Better ordering for plugin settings

SchrodingersGat:
- Followup to https://github.com/inventree/InvenTree/pull/11643
- Use order as provided by plugin


wolflu05:
LGTM, thanks for that quick fix

PR #11651, opened 2026-04-01
Reduce load on background worker

SchrodingersGat:
This PR introduces two changes to reduce the load on the background worker by removing duplicate entries:

- Do not save global settings if the value is unchanged
- Do not queue background tasks if there is already an identical entry

### Tasks

- [x] Add unit test for new de-duplication code

### DIFF
--- a/src/backend/InvenTree/InvenTree/tasks.py
+++ b/src/backend/InvenTree/InvenTree/tasks.py
@@ -159,8 +159,55 @@ def record_task_success(task_name: str):
     set_global_setting(f'_{task_name}_SUCCESS', datetime.now().isoformat(), None)
 
 
+def check_existing_task(taskname, group: str, *args, **kwargs) -> Optional[str]:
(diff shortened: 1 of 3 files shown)


src/backend/InvenTree/InvenTree/tasks.py
```diff
@@ -159,8 +159,56 @@ def record_task_success(task_name: str):
     set_global_setting(f'_{task_name}_SUCCESS', datetime.now().isoformat(), None)
 
 
+def check_existing_task(taskname, group: str, *args, **kwargs) -> Optional[str]:
```
matmair, line 162: how does this handle repeat events on the same item?

SchrodingersGat: The default action is to ignore duplicate events, but once the first event is complete, a new one can be added.

Also the `offload_task` function now has a `check_duplicates` argument which can be set to False to force a duplicate

PR #11652, opened 2026-04-01
Memoize parameters for useInstance

SchrodingersGat:
Should help to reduce re-renders


matmair (approved):
lgtm;
as an aside: can we test for to many re-renders in playwrigth?

SchrodingersGat:
> lgtm; as an aside: can we test for to many re-renders in playwrigth?

That's an interesting idea - I'm not sure if that's possible but can look into it.

PR #11655, opened 2026-04-01
refactor(backend): replace bleach with nh3 and bump weasy

matmair:
Most work done by AI: Github Copilot (claude sonnet 4.6)

weasyprint has some issues and we need to bump, but we can not upgrade due to complex dependencies between weasyprint and bleach (via some transitive deps); so our usage of bleach needs to be replaced by something - in this case nh3 which seems to be what most users of bleach switched to

Replaces https://github.com/inventree/InvenTree/pull/11415
Fixes https://github.com/inventree/InvenTree/security/dependabot/301


SchrodingersGat (approved):

SchrodingersGat:
@matmair nice work getting this sorted finally, as the tests are unchanged here, this looks good to me!

PR #11659, opened 2026-04-01
feat(frontend): improve comms around danger of staff users

matmair:
follow up to https://github.com/inventree/InvenTree/pull/11658
- renames "staff" to "administrator" to make it more clear what level of access is granted
- adds a warning code and permanent banner when using an elevated user image
- adapts various text to make comms around security boundaries clearer

todo:
- [x] is there a need for a permanent option to disable the banner?
- [x] should we change frontend refs and api fields from `is_staff` to `is_admin`


SchrodingersGat (approved):

matmair:
@SchrodingersGat what is your opinion on these questions?

1. is there a need for a permanent option to disable the banner? Should that be global or per user?
2. should we change frontend code refs and api fields from `is_staff` to `is_admin`

SchrodingersGat:
> 1. is there a need for a permanent option to disable the banner? Should that be global or per user?

We could at least make it dismissable by the user, even if only for the current session?
And I think a global setting to disable it entirely would be good, and easy to implement.

>   2. should we change frontend code refs and api fields from `is_staff` to `is_admin`

I think that it is sufficient to change the labels and messaging around this, without breaking the API fields

matmair:
@SchrodingersGat I have added a close button and a global customize option to hide the banner

matmair:
@SchrodingersGat ready for merge;
As this increments the backend a backport is probably not feasible

SchrodingersGat (approved):

SchrodingersGat:
@matmair nice work

PR #11674, opened 2026-04-04
Ensure configuration settings are documented

SchrodingersGat:
- Follow-up to https://github.com/inventree/InvenTree/pull/11673
- Ensure that configuration options are documented correctly
- Adds on to existing pre-build and post-build docs hooks

### Note

This does not automatically catch every config option which is available - because there are some that are dynamically loaded. However, it at least ensures that the primary configuration options (which are not gated during loading) are observable and included in the documentation.


tasks.py
```diff
@@ -2205,12 +2210,19 @@ def doc_schema(c):
...
+    c,
+    address='localhost:8080',
+    compile_schema: bool = False,
+    export_definitions: bool = True,
+):
     """Start a local mkdocs server to view the documentation."""
     # Extract settings definitions
-    export_definitions(c, basedir='docs')
+    if export_definitions:
+        export_definitions(c, basedir='docs')
```
matmair, line 2318: I think this will produce a collision

SchrodingersGat: Thanks, fixed

matmair (approved):
LGTM

PR #11677, opened 2026-04-06
Worker debugging

gunstr:
Ref #11665

Open questions:

- Should an option to start debugging with the sync=True setting be added and covered in docs update in this PR or is a manual change in settings.py enough for now?
- When I built the docs I noted that the API version in schema.md was also updated from 449 to 470. I see in the repo that the version is 449 but in the documentation it's 470. Should I just dropt that?

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -36,7 +36,9 @@
       "program": "${workspaceFolder}/src/backend/InvenTree/manage.py",
       "args": [
         "runserver",
-        "0.0.0.0:8000"
+        "0.0.0.0:8000",
+        // "--sync",// Synchronize worker tasks to foreground thread
+        // "--noreload", // disable auto-reload
       ],
       "django": true,
       "justMyCode": false
--- a/docs/docs/develop/devcontainer.md
+++ b/docs/docs/develop/devcontainer.md
@@ -60,7 +60,12 @@ If you only need a superuser, run the `superuser` task. It should prompt you for
 
 #### Run background workers
 
-If you need to process your queue with background workers, run the `worker` task. This is a foreground task which will execute in the terminal.
+If you need to process your queue with background workers, open a new terminal and run the `worker` task with `invoke worker`. This is a foreground task which will execute in the terminal.
+
+If you are developing functions that will be executed by background workers there are a few debugging options.
+
+- If the workers are started with the `worker` task you can add `print` or `logger` statements to the code and monitor the output in the terminal.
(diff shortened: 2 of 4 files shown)


SchrodingersGat:
>     * Should an option to start debugging with the sync=True setting be added and covered in docs update in this PR or is a manual change in settings.py enough for now?

It might be a good addition for developers to have a `--sync-worker` flag to the `invoke dev.server` command - which only applies in debug mode. The `invoke dev.server` command already runs the development server, so it's a good place for it.

>     * When I built the docs I noted that the API version in schema.md was also updated from 449 to 470. I see in the repo that the version is 449 but in the documentation it's 470. Should I just dropt that?

Do not commit these changes - they are updated as part of our GitHub CI process

gunstr:
> It might be a good addition for developers to have a --sync-worker flag to the invoke dev.server command - which only applies in debug mode. The invoke dev.server command already runs the development server, so it's a good place for it.

This was a bit unclear to me... As I see it the only reason to use `sync=True` is when the server is started from one of the entries in launch.json with debugpy. And what I can see invoke is not used to start a debugging session.  Or are there more things happening "under the hood" that I have not understood?

SchrodingersGat:
I mean, in `.vscode/launch.json` adding an argument here:

```json
  // Use IntelliSense to learn about possible attributes.
  // Hover to view descriptions of existing attributes.
  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
  "version": "0.2.0",
  "configurations": [
    {
      "name": "InvenTree Server",
      "type": "debugpy",
      "request": "launch",
      "program": "${workspaceFolder}/src/backend/InvenTree/manage.py",
      "args": [
        "runserver",
        "0.0.0.0:8000", // expose server in network (useful for testing with mobile app)
        //"--sync"// Sychronize worker tasks to foreground thread
        // "--noreload" // disable auto-reload
      ],
      "django": true,
      "justMyCode": true
    },
```

And then, in `settings.py` determine if we are running with this argument (and in DEBUG mode) and then set `sync=True`

gunstr:
Got it - I will try later today.

gunstr:
I have implemented the changes but I cannot commit them. Looking at the output I can find this:
```
codespell................................................................Failed
- hook id: codespell
- exit code: 65
```
I cannot figure out what is wrong. Any hints?

gunstr:
Found the spelling error... It was not marked in the editor, but reding the output log more carefullty it was mentioned there

SchrodingersGat:
@gunstr thanks for the addition!

PR #11696, opened 2026-04-08
[UI] Adjust "superuser warning" banner

SchrodingersGat:
- fix typo: evelevated -> elevated
- visual tweaks
- adds some playwright tests
- adds new settings to control display of banner(s)


src/frontend/src/components/nav/Header.tsx
```diff
@@ -130,12 +129,14 @@ export function Header() {
...
 
+  const [showSuperuserAlert, setShowSuperuserAlert] = useState<boolean>(true);
+
   const showElevated = useMemo(
     () =>
-      (user?.is_staff || user?.is_superuser || false) &&
-      !elevatedAlertClosed &&
-      !window.INVENTREE_SETTINGS.dangerous_hide_evelevated_alert,
-    [user, elevatedAlertClosed]
+      (user?.is_superuser || false) &&
```
matmair: This defeats half of the purpose of this. As we recently learned the hard way: the “staff” user flag is also dangerous and should be guarded. Not warning against it will lead to ppl not reading the docs and being unaware of the   risky thing they are doing by using a user with that flag

SchrodingersGat: I 100% agree for a superuser - but IMO operating as a staff user is expected.

Otherwise, does every "staff" user need a secondary account for their day-to-day operations?

matmair: given the current security boundaries for templates: yes

SchrodingersGat: I have added two separate settings to control this then. They can be hard-coded via environment variables - as with any other setting.

PR #11705, opened 2026-04-09
Selection lists updates

SchrodingersGat:
Some API / UI updates for the "selection lists" support:

- When adding a parameter based on a selectionlist, fetch the entries dynamically via the API,  rather than fetching as an entire block.
- Adds searching / ordering / filtering to the existing API endpoints
- Improved rendering, etc

### Tasks

- [x] Playwright tests for new features
- [x] Updated demo dataset - https://github.com/inventree/demo-dataset/pull/107
- [x] Update documentation


matmair (approved):

matmair:
@SchrodingersGat I have pushed some small fixes to make this build

src/backend/InvenTree/common/models.py
```diff
@@ -2320,11 +2320,39 @@ def get_api_url():
         """Return the API URL associated with the SelectionList model."""
         return reverse('api-selectionlist-list')
 
-    def get_choices(self):
-        """Return the choices for the selection list."""
-        choices = self.entries.filter(active=True)
+    def get_choices(self, active: Optional[bool] = True):
```
matmair, line 2323: Question: what is this used for?

matmair (approved):

src/backend/InvenTree/common/models.py
SchrodingersGat: Did not end up using this addition for the PR, but I think it is still a useful option

PR #11746, opened 2026-04-14
fix table in docs

ZhangzrJerry:
https://docs.inventree.org/en/latest/start/docker_install/#web-server-bind-address

image


matmair (approved):
Thank you!

PR #11761, opened 2026-04-17
feat(currency): add native support for BRL (Brazilian Real)

itallogravina:
- Add BRL currency to currency registry


matmair:
I see no need for this change that would cause huge schema diffs; supported currencies can be set via the configuration per instance https://docs.inventree.org/en/stable/settings/global/#pricing-and-currency

SchrodingersGat:
Agreed with @matmair - this is a user configurable setting, no need to adjust the defaults here @itallogravina

PR #11764, opened 2026-04-18
No remote image

SchrodingersGat:
Removes the "download image from URL" functionality.

This has historically been a hack because copying images from the clipboard didn't work. Now that clipboard images have been implemented, this kludge is no longer required. As it introduced a substantial security hole it may be best to just remove it entirely.

### Tasks

- [x] Update documentation
- [x] Remove API fields
- [x] Remove UI elements


matmair:
@SchrodingersGat this looks good; will there be a follow up?

SchrodingersGat:
@matmair replacement PR here - https://github.com/inventree/InvenTree/pull/11962

PR #11772, opened 2026-04-20
feat(frontend): warn if notes are dirty

### LINKED ISSUE #301: notes are not automatically saved

matmair:
This adds an alert if the user tries to navigate away after changing the content of a panel - for now only notes.

closes https://github.com/invenhost/InvenTree/issues/301

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- [#11772](https://github.com/inventree/InvenTree/pull/11772) the UI now warns if you navigate away from a note panel with unsaved changes
+
 ### Changed
 
 ### Removed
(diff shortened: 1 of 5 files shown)


wolflu05:
Nice addition, we may can use the [useBlocker](https://reactrouter.com/how-to/navigation-blocking) hook from react router here to block at the router level to ensure no other navigation on a different link occurs. We may can also block closing the tab when there is unsaved data

SchrodingersGat:
This is a great addition! The `useBlocker` hook looks interesting - do you think you will add that here @matmair ?

matmair:
@wolflu05 useBlocker does not work with our router model / react router version.

matmair:
i think i got this as far as I can without touching the router - which i am not intrested in

SchrodingersGat (approved):

wolflu05:
Oh, you're right. I haven't used React Router for a while. It seems that they have different modes now, which I was not aware of.

PR #11778, opened 2026-04-21
feat(frontend): add inline create modal to PurchaseOrderLineItem dialog

### LINKED ISSUE #299: inline creation of parts in dialogs

matmair:
Based on real-world user feedback from a HW producing company this reduces the amount of required context switches by adding inline addition dialogs.
image
In this specific example I have chosen to keep the new dialog as slim as possible as the goal is speed - not adding a new entry point to the existing dialog. Maybe a link to the full creation dialog might be interesting.
image
The newly created item is directly selected in the related field. This needs to be specifically enabled per form
 

fixes https://github.com/invenhost/InvenTree/issues/299

### DIFF
--- a/src/frontend/src/components/forms/fields/RelatedModelField.tsx
+++ b/src/frontend/src/components/forms/fields/RelatedModelField.tsx
@@ -561,9 +561,10 @@ function InlineCreateButton(
     () => calculateModalData(definition, form),
     [definition.filters, definition.addCreateFields, form]
   );
+  const model = useMemo(() => modelInfo?.label() ?? '', [modelInfo]);
 
   const create_modal = useCreateApiFormModal({
-    title: t`Create New ${modelInfo.label() ?? ''}`,
+    title: t`Create New ${model}`,
     url: apiUrl(modelInfo.api_endpoint),
     modelType: definition.model,
     initialData: relatedInitialData,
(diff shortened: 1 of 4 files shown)


src/frontend/src/components/forms/fields/RelatedModelField.tsx
```diff
@@ -525,3 +550,69 @@ export function RelatedModelField({
...
+  form: UseFormReturn<FieldValues, any, FieldValues>,
+  setValue: (value: number) => void
+): ReactNode {
+  const relatedInitialData = useMemo(
+    () => calculateModalData(definition, form),
+    [definition.filters, definition.addCreateFields, form]
+  );
+
+  const create_modal = useCreateApiFormModal({
+    title: t`Create New ${modelInfo.label() ?? ''}`,
```
SchrodingersGat: As this string has to be translated dynamically, and also perform string replacement, I would prefer to see a separate step where we extract the `modelInfo.label` to pass to the translation layer. We have previously had issues where complex inline string replacements have broken the crowdin translation integration.

e.g.

```typescript

const model = useMemo(() => modelInfo?.label() ?? '', [modelInfo]);

  const create_modal = useCreateApiFormModal({
    title: t`Create New ${model}`,
```

SchrodingersGat:
@matmair this looks great, we have been missing a structure for multi-level modals!

matmair:
@SchrodingersGat adressed

SchrodingersGat (approved):

SchrodingersGat:
@matmair nice work!

I think that (separately) it would be great to add a new playwright test for this, as we are implementing a brand new UI feature. But I'm happy to merge this functionality in now.

PR #11785, opened 2026-04-22
[UI] More sub-forms

SchrodingersGat:
- Follow-up to https://github.com/inventree/InvenTree/pull/11778
- Adds more inline forms


src/frontend/src/components/forms/fields/RelatedModelField.tsx
```diff
@@ -546,25 +543,42 @@ export function RelatedModelField({
             onScanSuccess={onBarcodeScan}
           />
         )}
+        {addButton && modelInfo && (
```
matmair: IMO it makes more sense to have this left-centered as we already have the option for other general actionbuttons on the right

matmair (approved):

src/frontend/src/components/forms/fields/RelatedModelField.tsx
SchrodingersGat: Sure, I have reverted that

PR #11790, opened 2026-04-23
Fix typecast for db settings

SchrodingersGat:
- Follow-up to https://github.com/inventree/InvenTree/pull/11674

### DIFF
--- a/src/backend/InvenTree/InvenTree/setting/db_backend.py
+++ b/src/backend/InvenTree/InvenTree/setting/db_backend.py
@@ -12,12 +12,16 @@ logger = structlog.get_logger('inventree')
 def get_db_backend():
     """Return the database backend configuration."""
     db_config = {
-        'ENGINE': get_setting('INVENTREE_DB_ENGINE', 'database.engine', None),
-        'NAME': get_setting('INVENTREE_DB_NAME', 'database.name', None),
-        'USER': get_setting('INVENTREE_DB_USER', 'database.user', None),
-        'PASSWORD': get_setting('INVENTREE_DB_PASSWORD', 'database.password', None),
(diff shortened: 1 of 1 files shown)


src/backend/InvenTree/InvenTree/setting/db_backend.py
```diff
@@ -12,12 +12,16 @@
...
+        'ENGINE': get_setting(
+            'INVENTREE_DB_ENGINE', 'database.engine', '', typecast=str
+        ),
+        'NAME': get_setting('INVENTREE_DB_NAME', 'database.name', '', typecast=str),
+        'USER': get_setting('INVENTREE_DB_USER', 'database.user', '', typecast=str),
+        'PASSWORD': get_setting(
+            'INVENTREE_DB_PASSWORD', 'database.password', '', typecast=str
+        ),
+        'HOST': get_setting('INVENTREE_DB_HOST', 'database.host', '', typecast=str),
+        'PORT': get_setting('INVENTREE_DB_PORT', 'database.port', '', typecast=str),
```
matmair, line 24: This seems to remove the default; and are you sure this should be a `str` not a `int`?

SchrodingersGat: These settings have been adjusted to closely match the typing and default values expected by django:

https://docs.djangoproject.com/en/6.0/ref/settings/#port

For the "port" setting in particular - the default value here *should* be blank, because an empty value translates to the default port for the specific DB backend. Using the postgres port as the default does not work for other db backends for example

SchrodingersGat: And using `None` as a default value causes issues if the value is not specified.

PR #11793, opened 2026-04-23
Check validate BOM

SchrodingersGat:
Adds a new optional setting to specify a group of users who have permission to validate BOMs


src/backend/InvenTree/part/models.py
```diff
@@ -2026,9 +2026,26 @@ def is_bom_valid(self) -> bool:
...
+        if not user:
+            return False
+
+        if validate_group_id := get_global_setting('PART_BOM_VALIDATE_GROUP'):
+            try:
+                validate_group = Group.objects.get(pk=validate_group_id)
+                if not user.groups.filter(pk=validate_group.pk).exists():
+                    can_validate = False
+            except (ValueError, Group.DoesNotExist):
+                pass
```
matmair, line 2042: This feels like a re-implementation of the django permission model. why not add a permission `can_validate` to the model? https://docs.djangoproject.com/en/6.0/topics/auth/customizing/#custom-permissions

SchrodingersGat: This would require an expansion of our existing group / role permission system, right?

matmair: Not necessarily, these permissions could just be added to the model and then used. It might require adding the new permission to the role map

SchrodingersGat:
Closed, replaced by https://github.com/inventree/InvenTree/pull/11825

PR #11800, opened 2026-04-24
Adjust order quantity based on pack size

### LINKED ISSUE #10946: Ordering from BO adds item quantity, not pack quantity, to a PO

### Describe the bug*

When I check the box in the required items in a BO and try to order them, the total quantity is added to a PO, not the pack quantity. I'm on 1.1.3. 

### Steps to Reproduce

1. Create a purchased part with a supplier part that has a pack quantity = 10
2. Create an assembly part that has the new purchased part with quantity 20 in its BOM
3. Create a BO to make 2x of the assembly part (so total quantity of the purchased part needed = 40)
4. Go to "required items" in the new BO, check the box of the purchased part, and select "Buy Stock"
5. The default quantity shown will be 40, not 4 (packs). 

### Expected behaviour

It should instead show the quantity of the supplier part, using the pack size, not the actual item quantity. Otherwise way too many packs will be ordered by accident. 

### Deployment Method

None

### Version Information

1.1.3

I did not try to reproduce

Not reproducible

### Relevant log output

```shell

```

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/10946


matmair (dismissed):
Does this depend on https://github.com/inventree/InvenTree/pull/11631 because I do not think quantity_raw is a valid field in the current api

SchrodingersGat:
> Does this depend on #11631 because I do not think quantity_raw is a valid field in the current api

This is just a local temporary variable to track the original quantity (without pack size). It is not sent anywhere

PR #11834, opened 2026-04-30
Fix: preserve existing widget sizes when adding a dashboard widget

### LINKED ISSUE #11829: Adding a dashboard widget resets every other widget to its minimum size

### Describe the bug*

When you add a new widget from the drawer, every existing widget on the dashboard snaps back to its minWidth / minHeight. Custom sizing is gone, you have to resize everything from scratch.

### Steps to Reproduce

1. Open the dashboard
2. Resize a few widgets to non-default sizes
3. Add any widget from the drawer
4. Existing widgets are now all at their minimum size

### Expected behaviour

Existing widgets keep their sizes and positions. The newly-added widget gets auto-placed by react-grid-layout (default behavior — placed at the bottom or first available slot).

**Cause:**
src/frontend/src/components/dashboard/DashboardLayout.tsx, addWidget, line 104:
typescript_layouts[key] = updateLayoutForWidget(_layouts[key], widgets, true);
The true is overrideSize. When that flag is on, updateLayoutForWidget clobbers w and h on every layout entry to the widget's minimum. So adding one widget rewrites the size of all the others.
**Tested Fix:**
Change true to false:
diff- _layouts[key] = updateLayoutForWidget(_layouts[key], widgets, true);
+ _layouts[key] = updateLayoutForWidget(_layouts[key], widgets, false);
The new widget still gets a layout entry react-grid-layout auto-place it because there's no entry for it yet. Existing widgets keep whatever sizes the user set. Tested the change and it works. Happy to PR if useful.

### Deployment Method

Bare metal

### Version Information

1.3.0

I did not try to reproduce

Not reproducible

### Relevant log output

```shell

```

tigger2000ttfn:
Fixes #11829

## Problem
In `DashboardLayout.tsx`, `addWidget` called `updateLayoutForWidget` with `overrideSize=true` on every existing layout entry. With that flag set, `updateLayoutForWidget` forces `w` and `h` to the widget's `minWidth` / `minHeight`. So adding any widget silently snapped every other widget on the dashboard back to its minimum size, wiping the user's custom layout.

## Fix
Pass `overrideSize=false` instead. Existing widgets keep their layout entries (and therefore their user-set sizes). The newly added widget has no entry yet, so react-grid-layout auto-places it at default size, which is the desired behaviour.

This matches the pattern already used in `onLayoutChange`, which also passes `overrideSize=false`.

## Test
Adds a Playwright regression test in `pui_dashboard.spec.ts` that
1. Adds a widget
2. Resizes it (drag bottom-right resize handle)
3. Verifies it grew
4. Adds a second widget
5. Asserts the first widget retained its resized height

### DIFF
--- a/src/frontend/src/components/dashboard/DashboardLayout.tsx
+++ b/src/frontend/src/components/dashboard/DashboardLayout.tsx
@@ -97,11 +97,14 @@ export default function DashboardLayout() {
         setWidgets([...widgets, newWidget]);
       }
 
-      // Update the layouts to include the new widget (and enforce initial size)
+      // Update the layouts to include the new widget.
+      // Pass overrideSize=false so existing widgets keep their user-set
+      // dimensions; only the newly added widget will receive default sizing
+      // via react-grid-layout's auto-placement.
       const _layouts: any = { ...layouts };
 
       Object.keys(_layouts).forEach((key) => {
-        _layouts[key] = updateLayoutForWidget(_layouts[key], widgets, true);
+        _layouts[key] = updateLayoutForWidget(_layouts[key], widgets, false);
       });
 
       setLayouts(_layouts);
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
@tigger2000ttfn thanks for digging into this one and submitting a fix :)

PR #11880, opened 2026-05-05
Lock notes on orders if they are completed

### LINKED ISSUE #11879: Editing/saving a note on a completed PO fails with no user feedback

### Describe the bug*

If you try to save a note on a completed PO, the operation fails with no user-visible feedback. The underlying HTTP PATCH to `/api/order/po/<id>` returns 400 with the error `This order is locked and cannot be modified`, but the UI does not surface this to the user, who will likely not notice that their note didn't get saved.

I suspect this may be a more specific symptom of missing blanket UI error handling on this operation, but there are some specific UX considerations for this particular branch so I've kept this issue focussed on that. It may well be that the least-effort resolution would be to add/fix error toasts on note save operations failing.

### Steps to Reproduce

1. complete a PO
2. navigate to the notes panel, enter a note, click the save button
3. Navigate away from, then back to the notes panel

### Expected behaviour

Either:
- PO notes should be possible to edit after a PO has been completed - after all this is a not inconceivable thing for user to want to do
- The notes editor should be disabled in the UI when the PO is completed to make it obvious _before_ users spend time typing a note 
- (less ideal, but should be caught by general error handling on this code path which seems like it must be missing) a clear error should be shown to the user to let them know that the note could not be saved due to the PO being "locked".

### Deployment Method

None

### Version Information

InvenTree-Version: 1.2.7
Django Version: 5.2.11
Commit Hash: 01bf600
Commit Date: 2026-04-08
Commit Branch: null
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.6.50-x86_64-with-glibc2.41
Installer: DOC
Active plugins: [{"name":"InvenTreeBarcode","slug":"inventreebarcode","version":"2.1.0"},{"name":"BOM Exporter","slug":"bom-exporter","version":"1.1.0"},{"name":"InvenTree Exporter","slug":"inventree-exporter","version":"1.0.0"},{"name":"Parameter Exporter","slug":"parameter-exporter","version":"2.0.0"},{"name":"Part Stocktake Exporter","slug":"inventree-stocktake-exporter","version":"1.0.0"},{"name":"InvenTreeEmailNotifications","slug":"inventree-email-notification","version":"1.0.0"},{"name":"InvenTreeUINotifications","slug":"inventree-ui-notification","version":"1.0.0"},{"name":"InvenTreeCurrencyExchange","slug":"inventreecurrencyexchange","version":"1.0.0"},{"name":"InvenTreeMachines","slug":"inventree-machines","version":"1.0.0"},{"name":"PartNotificationsPlugin","slug":"partnotificationsplugin","version":"1.0.0"},{"name":"InvenTreeLabel","slug":"inventreelabel","version":"1.1.0"},{"name":"InvenTreeLabelMachine","slug":"inventreelabelmachine","version":"1.0.0"},{"name":"DigiKeyBarcodePlugin","slug":"digikeyplugin","version":"1.0.1"},{"name":"LCSCBarcodePlugin","slug":"lcscplugin","version":"1.0.1"},{"name":"MouserBarcodePlugin","slug":"mouserplugin","version":"1.0.1"},{"name":"TMEBarcodePlugin","slug":"tmeplugin","version":"1.0.1"},{"name":"Autolinker","slug":"autolinker","version":null},{"name":"Farnell Plugin","slug":"farnell-plugin","version":"1.0.0"},{"name":"ZebraPrinterDriverPlugin","slug":"zebra-printer-driver-plugin","version":null},{"name":"RS Barcode Plugin","slug":"rs-barcode-plugin","version":"1.0.0"},{"name":"Wireviz","slug":"wireviz","version":"1.1.1"}]

I tried to reproduce

Reproducible

### Relevant log output

```shell
API response from server, with HTTP 400: {"reference":["This order is locked and cannot be modified"]}
```

matmair:
Fixes #11879

### DIFF
--- a/src/frontend/src/pages/build/BuildDetail.tsx
+++ b/src/frontend/src/pages/build/BuildDetail.tsx
@@ -569,7 +569,9 @@ export default function BuildDetail() {
       NotesPanel({
         model_type: ModelType.build,
         model_id: build.pk,
-        has_note: !!build.notes
+        has_note: !!build.notes,
+        // TODO @matmair - change API to include a "locked" attribute that we can check here
+        editable: build.status == 40 ? false : undefined
       })
     ];
   }, [
--- a/src/frontend/src/pages/build/BuildDetail.tsx
+++ b/src/frontend/src/pages/build/BuildDetail.tsx
@@ -569,9 +569,7 @@ export default function BuildDetail() {
       NotesPanel({
         model_type: ModelType.build,
         model_id: build.pk,
-        has_note: !!build.notes,
-        // TODO @matmair - change API to include a "locked" attribute that we can check here
-        editable: build.status == 40 ? false : undefined
+        has_note: !!build.notes
       })
     ];
   }, [
--- a/src/frontend/src/pages/purchasing/PurchaseOrderDetail.tsx
+++ b/src/frontend/src/pages/purchasing/PurchaseOrderDetail.tsx
@@ -410,7 +410,9 @@ export default function PurchaseOrderDetail() {
       NotesPanel({
         model_type: ModelType.purchaseorder,
         model_id: order.pk,
-        has_note: !!order.notes
+        has_note: !!order.notes,
+        // TODO @matmair - change API to include a "locked" attribute that we can check here
+        editable: order.status == 30 ? false : undefined
       })
     ];
   }, [order, id, user]);
--- a/src/frontend/src/pages/purchasing/PurchaseOrderDetail.tsx
+++ b/src/frontend/src/pages/purchasing/PurchaseOrderDetail.tsx
@@ -412,7 +412,11 @@ export default function PurchaseOrderDetail() {
         model_id: order.pk,
         has_note: !!order.notes,
         // TODO @matmair - change API to include a "locked" attribute that we can check here
-        editable: order.status == 30 ? false : undefined
+        editable:
+          order.status == 30 &&
+          !globalSettings.isSet('PURCHASEORDER_EDIT_COMPLETED_ORDERS')
+            ? false
+            : undefined
       })
     ];
   }, [order, id, user]);
--- a/src/frontend/src/pages/purchasing/PurchaseOrderDetail.tsx
+++ b/src/frontend/src/pages/purchasing/PurchaseOrderDetail.tsx
@@ -413,7 +413,7 @@ export default function PurchaseOrderDetail() {
         has_note: !!order.notes,
         // TODO @matmair - change API to include a "locked" attribute that we can check here
         editable:
-          order.status == 30 &&
+          order.status == poStatus.COMPLETE &&
           !globalSettings.isSet('PURCHASEORDER_EDIT_COMPLETED_ORDERS')
             ? false
             : undefined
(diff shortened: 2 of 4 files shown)


src/frontend/src/pages/purchasing/PurchaseOrderDetail.tsx
```diff
@@ -410,7 +410,9 @@ export default function PurchaseOrderDetail() {
       NotesPanel({
         model_type: ModelType.purchaseorder,
         model_id: order.pk,
-        has_note: !!order.notes
+        has_note: !!order.notes,
+        // TODO @matmair - change API to include a "locked" attribute that we can check here
+        editable: order.status == 30 ? false : undefined
```
SchrodingersGat: This should check the "edit completed orders" global setting too

src/frontend/src/pages/sales/SalesOrderDetail.tsx
```diff
@@ -448,7 +448,9 @@ export default function SalesOrderDetail() {
       NotesPanel({
         model_type: ModelType.salesorder,
         model_id: order.pk,
-        has_note: !!order.notes
+        has_note: !!order.notes,
+        // TODO @matmair - change API to include a "locked" attribute that we can check here
+        editable: order.status == 30 ? false : undefined
```
SchrodingersGat: Check the global setting

SchrodingersGat (approved):

src/frontend/src/pages/purchasing/PurchaseOrderDetail.tsx
matmair: done

src/frontend/src/pages/build/BuildDetail.tsx
```diff
@@ -569,7 +569,9 @@ export default function BuildDetail() {
       NotesPanel({
         model_type: ModelType.build,
         model_id: build.pk,
-        has_note: !!build.notes
+        has_note: !!build.notes,
+        // TODO @matmair - change API to include a "locked" attribute that we can check here
+        editable: build.status == 40 ? false : undefined
```
SchrodingersGat: Should we add a new global setting to check for ability to edit completed build orders?

SchrodingersGat (changes requested):

src/frontend/src/pages/build/BuildDetail.tsx
matmair: I think a broader re-thiniking of locks, including better APIs and a more general flow might be good. I would not change it in this PR as I would like to backport this to fix @wlcx problem soon

SchrodingersGat: In that case, enforcing a new lock on the build order after completion is a breaking change - if you are not wanting to add a new setting, can we limit the changes here only to the orders which get "locked" (purchase / sales / return).

src/frontend/src/pages/sales/SalesOrderDetail.tsx
```diff
@@ -448,7 +448,13 @@ export default function SalesOrderDetail() {
       NotesPanel({
         model_type: ModelType.salesorder,
         model_id: order.pk,
-        has_note: !!order.notes
+        has_note: !!order.notes,
+        // TODO @matmair - change API to include a "locked" attribute that we can check here
+        editable:
+          order.status == 30 &&
```
SchrodingersGat: Rather than a hard-coded numerical value - you could check against the named status codes like we do elsewhere in the UI:

https://github.com/inventree/InvenTree/blob/81fbf52c60cd6261e4d51907f63fad162fab2fc4/src/frontend/src/pages/purchasing/PurchaseOrderDetail.tsx#L101

SchrodingersGat (approved):

PR #11905, opened 2026-05-09
chore(backend): bump dev tooling, switch to prek

matmair:
prek is a drop in replacement for pre-commit but faster; on my setup this 2.5x full repeated runs.
Also reduces our dependency footprint - which I always appreciate


matmair:

SchrodingersGat (approved):
Nice!

SchrodingersGat:
Conflicts will need to be addressed after django bump

matmair:
@SchrodingersGat done

PR #11920, opened 2026-05-11
[feature] Rename attachments

### LINKED ISSUE #5047: [FR] Allow multiple files drag and drop into attachment & allow rename files

### Problem statement

Currently  we need to drag and drop files 1 by 1, it is also impossible to rename them.

### Suggested solution

Allow multiple drag and drop. 

It would also be great to allow renaming files.

-

### Examples of other systems

-

- [ ] I want to develop this.

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/5047
- Allow rename of attachment files after upload

### Tasks

- [x] Prevent change of file extension (security risks?)
- [x] Docs update
- [x] Unit tests


src/backend/InvenTree/common/api.py
```diff
@@ -793,6 +793,25 @@ def destroy(self, request, *args, **kwargs):
         return super().destroy(request, *args, **kwargs)
 
 
+@extend_schema(responses={201: common.serializers.AttachmentSerializer})
+class AttachmentRename(AttachmentMixin, CreateAPI):
```
matmair: Great idea; would it not make sense to use a POST action and return code 200 for this as it is a change?

SchrodingersGat:
@matmair I have simplified this to make use of the existing API endpoint - LMK what you think

src/backend/InvenTree/common/models.py
```diff
@@ -1993,6 +1995,54 @@ def __str__(self):
...
+
+        if not new_path.is_relative_to(old_path.parent):  # pragma: no cover
+            raise ValidationError(_('Invalid filename'))
+
+        new_path = new_path.as_posix()
+
+        if default_storage.exists(new_path):
+            raise ValidationError(_('A file with this name already exists'))
+
+        # Create a new file with the new name, and delete the old file
```
matmair, line 2054: Can we ensure this is an atomic action?

matmair (approved):
Lgtm; a further improvement could be to use build-in validations of filenames instead of hard coding a regex

PR #11924, opened 2026-05-11
bump urlib3

matmair:
replaces https://github.com/inventree/InvenTree/pull/11923 https://github.com/inventree/InvenTree/pull/11922 https://github.com/inventree/InvenTree/pull/11921


matmair:
AFAIK we are not impacted by the high severity vulns addressed in the release but security scanners won't care

PR #11897, opened 2026-05-08
frontend: more exposing for plugins

matmair:
Smaller attempt of https://github.com/inventree/InvenTree/pull/11858

exposes the following in the context:
- useInstance
- renderRemoteInstance
- EditApiForm

exports:
- DetailDrawerComponent
- useLocalLibState

PR #11928, opened 2026-05-11
feat(frontend): custom model render registration

matmair:
This adds the facilities to register custom model renders; while this is functional similar to providing `modelRenderer` for each field definition it enables plugins to be written more similar to core code and makes providing smaller, reusable plugins that just provide data classes and basic UI blocks easier.

The registered renderer acts as a fallback by default but can be prioritized by setting custom_model on the field definition.

follow up to https://github.com/inventree/InvenTree/pull/11897 / #11917
works towards https://github.com/inventree/org/issues/50
closes https://github.com/inventree/org/issues/56


SchrodingersGat (approved):

SchrodingersGat:
@matmair nice idea, thanks for the addition!

PR #11934, opened 2026-05-12
[UI] Adjust week start date

SchrodingersGat:
Allow user to select the week starting day for calendar views


matmair (commented):
Looks good; should we maybe also add a system setting for this?

matmair (approved):

SchrodingersGat:
I'd like to see *either* a global or user setting - which would you prefer?

matmair:
If it is either I think a global setting would be more helpful; I do not think this will differ much between users on the same system.

SchrodingersGat:
@matmair I have changed this to use a global setting now

PR #11941, opened 2026-05-13
[bug] Fix for report generation context

### LINKED ISSUE #11940: Get current logged in user full name and email in report plugin
### Discussed in https://github.com/inventree/InvenTree/discussions/11933

<div type='discussions-op-text'>

<sup>Originally posted by **NerosKi** May 12, 2026</sup>
I have a report plugin to generate some extra data, but would also like to put on the report (quote in my case) who generated it. Can I get this information? Currently I take information who created the sales order but I rather have who actually generated the report pdf.

### Bug Description

- User who is printing the report is not captured in the report context.
- This is because the report is now offloaded to the background worker
- In this case, the "request" object is lost

Thus, the user information needs to be captured and passed to the background worker

SchrodingersGat:
### Summary
- Closes https://github.com/inventree/InvenTree/issues/11940
- Fixes a long-standing issue where user information was unavailable to report and label context
- This bug has been present ever since we offloaded report generation to the background worker

### Changes

- Fixes bug, ensures user information  is correctly retrieved by the background worker
- Removes existing references to "request" - which is *NOT* transferred across to the background worker anyway
- Substantial refactoring of the report generation pipeline code
- Adds a large amount of new test coverage to report generation
- Improved typing / code cleanup as we go

### Breaking Change

Any custom report plugins which implement any of the following methods will need to assess the impact of the changes to the plugin API:

```python
    def add_report_context(self, report_instance, model_instance, user, context):
        """Add extra context to the provided report instance.

        By default, this method does nothing.

        Args:
            report_instance: The report instance to add context to
            model_instance: The model instance which initiated the report generation
            user: The user to associate with the generated report
            context: The context dictionary to add to
        """

    def add_label_context(self, label_instance, model_instance, user, context):
        """Add extra context to the provided label instance.

        By default, this method does nothing.

        Args:
            label_instance: The label instance to add context to
            model_instance: The model instance which initiated the label generation
            user: The user to associate with the generated label
            context: The context dictionary to add to
        """

    def report_callback(self, template, instance, report, user, **kwargs):
        """Callback function called after a report is generated.

        Arguments:
            template: The ReportTemplate model
            instance: The instance of the target model
            report: The generated report object
            user: The user to associate with the generated report

        The default implementation does nothing.
        """
```

*However* - up to this point, the "request" object was being passed through instead of the user, and that has (for years) been `None`.

### Backporting

This will *not* be backported to `1.3.x` branch due to the change in the plugin API

### DIFF
--- a/src/backend/InvenTree/plugin/base/integration/ReportMixin.py
+++ b/src/backend/InvenTree/plugin/base/integration/ReportMixin.py
@@ -24,7 +24,7 @@ class ReportMixin:
         super().__init__()
         self.add_mixin(PluginMixinEnum.REPORT, True, __class__)
 
-    def add_report_context(self, report_instance, model_instance, request, context):
+    def add_report_context(self, report_instance, model_instance, user, context):
         """Add extra context to the provided report instance.
 
         By default, this method does nothing.
@@ -32,11 +32,11 @@ class ReportMixin:
         Args:
             report_instance: The report instance to add context to
             model_instance: The model instance which initiated the report generation
-            request: The request object which initiated the report generation
+            user: The user to associate with the generated report
             context: The context dictionary to add to
         """
 
-    def add_label_context(self, label_instance, model_instance, request, context):
+    def add_label_context(self, label_instance, model_instance, user, context):
         """Add extra context to the provided label instance.
 
         By default, this method does nothing.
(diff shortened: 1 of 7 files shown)


wolflu05:
~~Maybe its also a good idea to add the user info to the report context type, so that external tooling (e.g. docs, plugins) can correctly discover the available user in the context~~?

~~https://github.com/inventree/InvenTree/blob/3d2577b411aea048b2e918a6a3ad2096e90b5545/src/backend/InvenTree/report/mixins.py#L19-L20~~

EDIT: ok, found the point, its already there, but it was not provided previously. The correct class is:

https://github.com/inventree/InvenTree/blob/3d2577b411aea048b2e918a6a3ad2096e90b5545/src/backend/InvenTree/report/models.py#L140-L161

PR #11950, opened 2026-05-15
Add MariaDB link to database section in README

robertsilen:
InvenTree already documents MySQL and MariaDB installation in the docs and includes MariaDB-specific database options in the backend. The README tech stack only listed MySQL. This adds MariaDB as a supported database alongside the existing entries.


matmair:
@robertsilen why does the link point to a commercial offering instead of the foundation? .com vs .org?

robertsilen:
> @robertsilen why does the link point to a commercial offering instead of the foundation? .com vs .org?

@matmair Good question! It should be the most relevant option for the context. I updated it to .org now.

SchrodingersGat (approved):

PR #11951, opened 2026-05-15
feat(backend): add throtteling by default

matmair:
Part of SAST / AI-driven vuln discovery was how trivial some API points are to DoS attack (unknowingly even when using an MCP); while these limits are too lax for small instances they should keep the impact on most instances minimal while providing a good first barrier.

As I already mentioned in the threat model: a WAF is much better solution to this but most instances on the internet seem to not use that.


SchrodingersGat:
Any testing we can apply to this?

SchrodingersGat:
I also wonder whether these limits are sufficient for loading 2-3 pages simultaneously? A page load can throw a few requests to the server, this should be tested thoroughly with some abusive page loads.

matmair:
I will look into testing

matmair:
@SchrodingersGat I move to include this in 1.5.0 - oob comms to follow

SchrodingersGat (approved):

SchrodingersGat:
@matmair some throttling tests here would be good to. Are you going to update this PR with OOB docs / tests - or submit as a separate PR?

matmair:
Re:docs: There already are a changelog entry and config docs. As there are default values no action by users is required so I am not sure what more documentation is required

re:tests: in my local testing triggering throttling bled through test classes, which is why I added the check if running in debug. I will add a simple overwrite of that check so that we can have a single test in which throttling is applied

PR #11955, opened 2026-05-16
Feature: Auto-select items tab for final stock locations (#11766)

### LINKED ISSUE #11766: [FR]Final stock location

### Problem statement

When navigating through stock locations (sometimes back and forth) for looking a semi specific part (i.e. I am looking for a screw with a specific length and I need to peek through all the philips-, hex-, torxscrew etc. bins), it would be useful if the final locations (where no sub locations present) would list the contained parts automatically when clicking onto their links. 
 

### Suggested solution

I was thinking about adding a flag to the stock locations  which would disable the creation of the sub locations, (this would hide the "Stock locations" menu from here automatically:

Image

and would redirect the sublocations endpoint automatically to the stock-items

Changing the default behavior is not really a good choice, as it is possible that I navigate to the empty Sub locations  list to add a new one for e.g.

### Examples of other systems

_No response_

- [x] I want to develop this.

adityakrmishra:
Fixes #11766.

As discussed in the issue, I added a new user setting `DISPLAY_ITEMS_FOR_FINAL_LOCATION` (default True). 

If a stock location or category has 0 sublocations, the UI now automatically defaults to the 'Stock Items' or 'Parts' tab. The empty sublocations tab is still fully accessible if a user needs to manually click it to add something. 

Tested locally and working as expected. Let me know if you need any adjustments!


matmair:
1) this targets the wrong branch; should be `master`
2) imo this changes default behavior and should be false by default

matmair (approved):
lgtm

matmair (approved):
Lokks good @SchrodingersGat ready for review & merge

matmair:
@adityakrmishra if you install prek (or pre-commit) in your enviroment you get auto-formatting followng the applicable coding standards on each commit.
You can either do that manually or run `invoke setup-dev`. Should reduce the amount of CI-feedback required siginificantly.

src/backend/InvenTree/common/setting/user.py
```diff
@@ -268,4 +268,12 @@ def label_printer_options():
         'description': _('Save the last used printing machines for a user'),
         'default': '',
     },
+    'DISPLAY_ITEMS_FOR_FINAL_LOCATION': {
+        'name': _('Display Items for Final Location'),
```
SchrodingersGat: My only suggestion here is that "Location" is not correct wording - as this can apply to both "location" and "category" instances.

SchrodingersGat:
Happy with this overall, but perhaps we can tweak the wording for the user setting?

adityakrmishra:
Great catch! I've renamed the setting key to DISPLAY_ITEMS_FINAL_LEVEL and updated the display text and documentation to be completely agnostic so it accurately reflects that it handles both locations and categories.

SchrodingersGat (approved):

SchrodingersGat:
@adityakrmishra nice work, thanks!

SchrodingersGat:
Thanks for the contribution!

adityakrmishra:
Thank you so much for the merge and all the guidance!
This makes my second successful PR here, which means a lot to me. Being able to point to these real-world contributions is going to be amazing proof of my hard work as I start building my software career
I'm going to take a little screen-time break to recharge, but I definitely plan to come back and tackle more issues in the future. Thanks again for the opportunity!

PR #11956, opened 2026-05-16
[UI] Panel indicator badges

SchrodingersGat:
Extends the functionality of panel indicator badges, making it easier to define if a particular panel should display an indicator badge.

This PR extends the concept to:

- parameters
- attachments

image

image


src/frontend/lib/types/Panel.tsx
```diff
@@ -8,7 +8,7 @@ export type PanelType = {
   label: string;
   controls?: ReactNode;
   icon?: ReactNode;
-  notification_dot?: 'info' | 'warning' | 'danger' | null;
```
matmair, line 11: I am already using the warning dot in a plugin, i think this is valuable to keep

src/frontend/src/components/panels/AttachmentPanel.tsx
```diff
@@ -17,6 +19,21 @@ export default function AttachmentPanel({
     name: 'attachments',
     label: t`Attachments`,
     icon: <IconPaperclip />,
+    indicator: async () => {
+      if (!model_type || !model_id) {
+        return false;
+      }
+
+      return api
+        .get(apiUrl(ApiEndpoints.attachment_list), {
```
matmair, line 28: This PR adds multiple API call; can we maybe add all of these queries to a single api endpoint that just shows if a highlight is needed?

src/frontend/src/components/panels/ParametersPanel.tsx
```diff
@@ -21,6 +24,21 @@ export default function ParametersPanel({
     label: t`Parameters`,
     icon: <IconListDetails />,
     hidden: hidden ?? false,
+    indicator: async () => {
+      if (!model_type || !model_id) {
+        return false;
+      }
+
+      return api
+        .get(apiUrl(ApiEndpoints.parameter_list), {
```
matmair, line 33: See above comment re:api call volume

src/frontend/lib/types/Panel.tsx
SchrodingersGat: @matmair ok, I'll ensure that the interface does not change here

src/frontend/src/components/panels/AttachmentPanel.tsx
SchrodingersGat: @matmair not quite following - are you suggesting adding a new API endpoint? Instead of querying the attachment list (for e.g.)

SchrodingersGat:
@matmair I have adjusted this so that the existing interface is still supported.

src/frontend/src/components/panels/AttachmentPanel.tsx
matmair: If we add a number of indicators I think it would make sense to have a single API endpoint include either the indication or have a new endpoint instead of calling various endpoints that require; if we keep it at 2 extra calls we can probably ignore this tbh

matmair (approved):

src/frontend/src/components/panels/AttachmentPanel.tsx
SchrodingersGat: @matmair I'll keep it in mind - perhaps a generic API endpoint which returns the number of generic foreign key links against the specified model instance and returns something like:

```json
{
  "attachments": 4,
  "parameters": 2
}
```

PR #11971, opened 2026-05-20
[Refactor] Notes

### LINKED ISSUE #7400: Notes Improvements
- Follow on from https://github.com/inventree/InvenTree/pull/7284

### Ideas

- Allow image resizing (requires that we accept a subset of HTML tags on the server)
- Consider moving "notes" to own API endpoint (like "metadata") - as potentially quite data heavy

SchrodingersGat:
- This PR represents a major refactor / rethink of how we handle notes.
- Closes https://github.com/inventree/InvenTree/issues/7400

### Example Screenshots

| Description | Screenshot |
| ---- | ---- |
| Display Mode | image |
| Edit Mode | image |
| Rendered to report | image |

### Generic Notes Table

The major item here is to move all notes into a new table, which has generic FK links to other items. This brings it into line with how we handle attachments, parameters, etc. 

### Multiple Notes

We can now specify multiple, separate notes per item. This is useful for keeping different types of notes separate - e.g. shipping instructions vs inspection instructions (for e.g.)

### Render to report

Rendering notes to reports is now greatly improved. In particular, images are correctly handled now - and can even be resized based on the specified size in the HTML content

### Editor Updates

Also introduces the possibility of a more intuitive notes editor.

~At this stage, I am trialing [BlockNote](https://www.blocknotejs.org/) - which provides navite mantine support, and provides a "notion style" editing interface.~

I am using the [Rich Text Editor](https://mantine.dev/x/tiptap/) shipped as part of the Mantine framework.

**Pros:** 

- Intuitive, "modern" editing interface
- Supports more content layout features
- Resizable images

**Cons:**

- Requires HTML formatting
- Current notes use markdown

### Tasks

- [x] Display note information (title / description / last updated by)
- [x] Allow edit notes (if user has permission)
- [x] Allow delete notes (if user has permission)
- [x] Client side sanitize of notes
- [x] Server side sanitize of notes
- [x] Support direct image upload (drag and drop / copy-paste)
- [x] Render notes in PDF reports
- [x] Ensure inline images are rendered correctly in PDF reports
- [x] Documentation
- [x] Data migration for existing notes
- [x] Data migration for existing notes image
- [x] Handle legacy "NotesImage" concept - how to detect and embed these images?
- [x] prevent navigation away when notes are dirty
- [x] Allow note selection from query params
- [x] Update demo dataset with notes / images / etc (*BEFORE* migration)
- [x] Apply data migration to demo dataset
- [x] Unit tests (backend)
- [x] Duplicate notes from existing instance - also copies embedded images
- [x] Playwright tests
- [x] Unit tests for data migrations
- [x] Remove "notes" field from any models which use the new approach
- [x] Delete the old "notes" mixin class
- [x] Support dark mode
- [x] Support localization
- [x] Remove old dependencies for simplemde
- [x] Delete old "delete_old_notes_images" task
- [x] Mark "delete_old_notes_images" task as outdated
- [x] Ensure users cannot upload notes against models they do not have permission for
- [x] Ensure users cannot upload note images against models they do not have permission for
- [x] Allow creation of note "templates" (edit via admin interface)
- [x] Create new notes from template note
- [x] Add documentation for note templates
- [x] Add some note templates to the demo dataset - https://github.com/inventree/demo-dataset/pull/122
- [ ] Update app support (display multiple notes) - https://github.com/inventree/inventree-app/issues/859
- [x] Add CHANGELOG entry

### Ideas for future work

- Add plugin to link directly to attachments
- Add plugin to link to other pages (e.g. parts / orders / etc)
- Add plugin to render admonitions
- Export notes?
- Allow notes to be "scoped" to a certain group permission


SchrodingersGat:
@matmair @wolflu05 before I go too far with this I'd be interested in your feedback around two major points:

**Data Structure Refactor**: Are we on board with the idea of supporting multiple notes per item?

**Editor**: Do you have any objections to moving to a new editor, and move to HTML notes? We will obviously have to tackle any data sanitizing required here

matmair:
I think the general idea is very good; a few points/ideas:
- I would like to see some tests in the final PR to ensure we do not leak data due to this migration
- if we already build a new model we might as well add `created`, `updated` and `created_by`, `updated_by` fields
- does a note owner make sense?
- should we add user customizable types?
- should we addnote states?
- having a special "primary" note that still renders to the api seems like a good idea; i have seen a bunch of ppl use notes fields for all kind of plugins; this change would be pretty breaking

I like blocknote as a choice, the OpenProject and a bunch of other EU sovereign gov cloud products use it.
How we save/provide notes data in the future is still a bit of a question; there is markdown support [from what I can tell](https://www.blocknotejs.org/examples/interoperability/converting-blocks-to-md). And if we are already switching data model why not use the native document format and just save a html (or markdown) representation on the side? That seems safer and better for future customizations of the editor capabilities.

Things i am unsure about:
- what does rendering to reports look like? are existing reports breaking?
- possible security problems?

wolflu05:
I also like the general idea, the current notes editor seems very limited. But I would appreciate if there would be some easy migration path from markdown based notes to the new format, and a method to render the new notes format to reports.

SchrodingersGat:
> But I would appreciate if there would be some easy migration path from markdown based notes to the new format

All markdown features should be forward-compatible. This will be fully tested. I don't want to lose any of my notes, either.

> and a method to render the new notes format to reports.

100% will be supported and tested too.

The back-end storage of the notes will be editor-agnostic HTML. So rendering to reports will be easier (ideally) than markdown.

SchrodingersGat:
@matmair @wolflu05 I have switched to using the mantine rich text editor - https://mantine.dev/x/tiptap/

The **blocknotejs** editor is not as fully featured as I would like, and even a bunch of standard markdown does not work out of the box.

docs/docs/concepts/notes.md
```diff
@@ -0,0 +1,98 @@
+---
+title: Notes
+---
+
+## Notes
+
+*Notes* allow free-form rich-text content to be written and stored against a specific object within InvenTree. Notes can be used to record observations, instructions, historical context, or any other information associated with a model instance.
+
+!!! note "Business Logic"
+    Notes are not used for any core business logic within InvenTree. They are intended to provide supplementary documentation and context for objects, which can be useful for reference, communication, or reporting purposes.
```
matmair:
```suggestion
    Notes are not to be used for any core business logic within InvenTree. They are intended to provide supplementary documentation and context for objects, which can be useful for reference, communication, or reporting purposes. Plugins should not use them for storage and opt for object metadata or custom models instead.
```

docs/docs/concepts/terminology.md
```diff
@@ -13,7 +13,7 @@ Evolves around manufacturing of parts out of other parts. It keeps track of stoc
 
 ### Part Library Management *(PLM)*
 Keeps track of BOMs, part variants, possible substitutions, versions, IPNs and further part parameters.
-PLM can also mean product lifecycle management – those systems manage all stages from design through manufacturing up to customer support and recycling.
+PLM can also mean product lifecycle management - those systems manage all stages from design through manufacturing up to customer support and recycling.
```
matmair: Question: Is there any meaning to his change?

src/backend/InvenTree/common/api.py
```diff
@@ -1455,7 +1522,7 @@ def create(self, request, *args, **kwargs):
     # Webhooks
     path('webhook/<slug:endpoint>/', WebhookView.as_view(), name='api-webhook'),
     # Uploaded images for notes
-    path('notes-image-upload/', NotesImageList.as_view(), name='api-notes-image-list'),
+    path('notes-image/', NotesImageList.as_view(), name='api-notes-image-list'),
```
matmair: Suggestion: this should be moved down to the other notes endpoints

src/backend/InvenTree/common/models.py
```diff
@@ -3011,6 +2976,225 @@ def description(self):
         return self.template.description
 
 
+class Note(
```
matmair, line 3063: Suggestion: Having a stable order identifier on the model might be interesting

src/backend/InvenTree/common/models.py
```diff
@@ -3011,6 +2976,225 @@ def description(self):
...
+            ).exclude(pk=self.pk).update(primary=False)
+
+        self.cleanup_images()
+
+    def delete(self):
+        """Perform custom delete checks before deleting a Parameter instance."""
+        self.check_delete()
+        super().delete()
+
+    def clean(self):
```
matmair, line 3181: If possible I would like to see security relevant cleaning data outside of a model class and tested independently so that missing coverage is more visible in the reports

src/backend/InvenTree/InvenTree/helpers.py
```diff
@@ -939,33 +939,46 @@ def remove_non_printable_characters(value: str, remove_newline=True) -> str:
...
         extensions=extensions,
         extension_configs=extension_configs,
         output_format='html',
     )
 
-    # nh3 sanitizer settings
+    return html
+
+
+def clean_markdown(value: str) -> str:
```
matmair: Question: Where is this used?

src/frontend/package.json
```diff
@@ -40,6 +40,9 @@
         "compile": "lingui compile --typescript"
     },
     "dependencies": {
+        "@blocknote/core": "^0.51.0",
+        "@blocknote/mantine": "^0.51.0",
+        "@blocknote/react": "^0.51.0",
```
matmair: Question: Is blocknote still required?

src/frontend/package.json
```diff
@@ -138,6 +150,11 @@
         "vite-plugin-externals": "^0.6.2",
         "vite-plugin-istanbul": "^8.0.0"
     },
+    "overrides": {
+        "@blocknote/core": {
+            "prosemirror-model": ">=1.25.4"
+        }
+    },
```
matmair, line 154: Question: is this still required?

matmair (approved):
Great effort; looks very close already!
Maybe we can merge this in the `next-breaking` branch for now?

matmair:
Can we remove django-markdownify with this? would remove a few dependencies. I have made a small demo branch for this: https://github.com/matmair/InvenTree/pull/676

src/frontend/package.json
SchrodingersGat: No, I have removed them now!

src/backend/InvenTree/InvenTree/helpers.py
SchrodingersGat: Nowhere any more, I have removed it now.

src/backend/InvenTree/common/models.py
SchrodingersGat: @matmair can you clarify this point? I'm not sure what you mean here

src/backend/InvenTree/common/models.py
SchrodingersGat: Can you propose what that would look like?

docs/docs/concepts/terminology.md
SchrodingersGat: Not sure - must have crept in from a global find-and-replace. I have reverted the change now.

src/backend/InvenTree/common/models.py
matmair: Having a field that keeps the order of the notes for that object. That could be used to set a stable order of notes in the UI. Right now the order depends on the creation order/pk if I interpret it right.

wolflu05:
Although this is a big breaking change in the InvenTree api as not only the endpoints change but also how notes are stored, i think this is a huge step forward. In general it looks good, curious what @matmair says. Unfortunately I won't have the time to do a full review.

PR #11982, opened 2026-05-21
Change order custom status via api

### LINKED ISSUE #11927: PurchaseOrder PATCH accepts status_custom_key, but does not persist valid custom status

### Describe the bug*

`PATCH /api/order/po/{id}/` appears to accept `status_custom_key`, but a valid custom status is not persisted.

### Steps to Reproduce

1. Confirm custom PO status exists via API: `GET /api/generic/status/custom/?reference_status=PurchaseOrderStatus`
```json
[
  {
    "key": 110,
    "name": "NEEDS_APPROVAL",
    "label": "Needs Approval",
    "logical_key": 10,
    "reference_status": "PurchaseOrderStatus"
  }
]
```
2. Confirm target PurchaseOrder is has correct logical status.
```json
{
  "pk": 282,
  "status": 10,
  "status_text": "Pending",
  "status_custom_key": 10
}
```
3. PATCH with valid custom key for the current logicial status
```json
curl --request PATCH \
  --url https://<host>/api/order/po/282/ \
  --header 'Authorization: Token <redacted>' \
  --header 'Content-Type: application/json' \
  --data '{
    "status_custom_key": 110
  }'
```

### Expected behaviour

`status_custom_key` should persist.

```json
{
  "status": 10,
  "status_custom_key": 110
}
```

### Deployment Method

None

### Version Information

Inventree version 1.3.2
API version 477
Python 3.14.3
Django 5.2.13
Commit Hash a60e85b

I tried to reproduce

Reproducible

### Relevant log output

```shell

```

johnluetke:
Fixes #11927

As discussed in the issue, an order's custom status key can be updated via the API. 

Validation has been added to ensure that the custom_status_key value is aligned to the current logical state of the order.


SchrodingersGat (approved):

SchrodingersGat:
@johnluetke nice clean implementation :)

matmair (approved):
Looks good;
there are a few formatting things, those could be fixed with pre-commit or prek; I will submit them shortly

src/backend/InvenTree/order/serializers.py
```diff
@@ -118,11 +118,19 @@ class AbstractOrderSerializer(
...
-    status_text = serializers.CharField(source='get_status_display', read_only=True)
+    status_text = serializers.CharField(read_only=True)
 
     # status field cannot be set directly
     status = serializers.IntegerField(read_only=True, label=_('Order Status'))
 
+    # can be set directly, but must be valid for the current order status
+    status_custom_key = serializers.IntegerField(
+        label=_('Custom Status Key'),
+        help_text=_('Update order status to a custom value for this logical value'),
```
matmair, line 133: With this set we loose dynamic choice enumeration in the schema description; @SchrodingersGat is that acceptable or should we patch the schema generation mechanism

johnluetke: I've been trying to familiarize myself with how this works, but don't quite understand how it broke. It seems like the schema now thinks that the `status` and `custom_status_key` fields are more tightly coupled now? 

If you can point me in the right direction, I can see if my solution can be implemented in a way that doesn't change this.

SchrodingersGat: @matmair what would you propose as an alternative?

matmair: Deleting the description or patching our custom schema generator to still add the choices if a description is set on a serializer.

SchrodingersGat:
@johnluetke you will need to add a new entry into `InvenTree/api_version.py` because the API surface has changed.

src/backend/InvenTree/order/models.py
```diff
@@ -589,6 +589,21 @@ def order_address(self):
...
 
+    @property
+    def status_text(self):
+        """Return the text representation of the current status. This will consider any custom status."""
+        if self.get_custom_status() is not None:
+            from generic.states.custom import (
+                get_logical_value as get_custom_state_logical_value,
+            )
+
+            custom_status = get_custom_state_logical_value(
```
SchrodingersGat, line 601: This is going to result in a [N + 1 query problem](https://www.baeldung.com/cs/orm-n-plus-one-select-problem) - each retrieved item will have multiple database hits.

In fact, this problem already exists in the codebase. I'm currently working on a patch for this which I will amend to your branch

SchrodingersGat:
@johnluetke I have addressed the mentioned query issue, which has unfortunately introduced some conflicts here, but they should be minor.

Please review the changes I made in https://github.com/inventree/InvenTree/pull/12055 - you may have to adjust your approach here slightly

matmair:
@johnluetke we are slowly closing in on the 1.4 release - can you please try to address in the conflict in the next few weeks?

src/backend/InvenTree/order/models.py
johnluetke: After rebasing, I saw you fix for N+1 come in

johnluetke:
> @johnluetke we are slowly closing in on the 1.4 release - can you please try to address in the conflict in the next few weeks?

Done. Sorry it took so long.

src/backend/InvenTree/order/models.py
matmair: looking at the run results I do not think you have solved the n+1 issue

you can run the test `order.test_api.PurchaseOrderTest.test_po_custom_status_query_count` locally via this cli command `invoke dev.test --runtest order.test_api.PurchaseOrderTest.test_po_custom_status_query_count`

johnluetke: Yes i was notified of the test failures a few minutes ago.

SchrodingersGat:
@matmair are you happy with this?

src/backend/InvenTree/order/models.py
```diff
@@ -590,6 +590,21 @@ def order_address(self):
         """Return the Address associated with this order."""
         return self.address or self.company.primary_address
 
+    @property
+    def status_text(self):
+        """Return the text representation of the current status. This will consider any custom status."""
```
SchrodingersGat, line 595: I am concerned that this approach will result in the same `N+1` issue as the one I fixed in the other PR. Have you done any specific testing here for large query sets?

SchrodingersGat:
Conflicts fixed, I am looking into whether the new updates cause an "N + 1" issue

SchrodingersGat:
@johnluetke thanks for the contribution 😁

matmair:
This broke a few frontend status rendering functions; fixed in https://github.com/inventree/InvenTree/pull/12184/commits/c179abd5ee51d54a94fcf4a5c18839734c0577a9

PR #11991, opened 2026-05-23
Add CLAUDE.md file

SchrodingersGat:
Adds claude file for integration with claude code

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,148 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+InvenTree is an open-source inventory management system built on:
+- **Backend**: Python/Django + Django REST Framework, with async tasks via django-q2
+- **Frontend**: React 19 + TypeScript, Mantine UI, TanStack Query, Zustand, React Router, Lingui i18n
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,148 @@
...
+# Testing
+invoke test                            # Run all backend tests
+invoke test --runtest=order.test_api   # Run a specific test module
+invoke test --runtest=order.tests.TestClass.test_method  # Run single test
+invoke test --coverage                 # Run with coverage report
+invoke test --keepdb                   # Keep test DB between runs
+
+# Linting (backend uses ruff)
+ruff check src/backend/InvenTree       # Lint
+ruff format src/backend/InvenTree      # Format
```
matmair, line 36: why not just call prek?

matmair (commented):
this seems to be replicating info from the contributions file and also like it could very easily drift; can we not just link to the contributions documentation?

SchrodingersGat:
@matmair I think you're right, this doesn't seem to add much beyond just pointing claude to the existing files.

matmair:
A quick search suggests that only Anthropic stuff uses this Claude.md format;
The other AI tooling seem to broadly use a `Agents.md` file that links out to context files and add a simple `Claude.md` file with the contents of `@AGENTS.md`

PR #12013, opened 2026-05-26
Plugin validation tweak

SchrodingersGat:
Prevent plugin-based validation during read-only management operations.

In particular, prevent plugins from running custom validation during:

- Data import / export
- Data migrations
- Database dump / restore


matmair:
Data import is a point in time where I can imagine validation being helpful

matmair (approved):

SchrodingersGat:
@matmair here data "import" means from an external data file - from the command line, not the API.

It becomes quite complex to manage because we cannot control the order of data being imported, it's somewhat "all or nothing"

matmair:
@SchrodingersGat i think so; the alternative is that you have broken records in the database and have to correct them the next time a (possibly unrelated) validation runs over the record

Issue #6951, opened 2024-04-04
Transfer parts from stock to another without merging


SchrodingersGat:
### Discussed in https://github.com/inventree/InvenTree/discussions/6825

<div type='discussions-op-text'>

<sup>Originally posted by **MrShippeR** March 23, 2024</sup>
Hello comunity,
I would like ask you how use system right. I have 2 stock locations:

- New items on transport to Central stock
- Central stock

### Idea
My way of using Inventree is: I Add items at PC web interface to "items on transport" and someone transport it with car to "Central stock". 
When parts arrive, I want someone get Android app and check in delivered parts (count, undamaged...). To assign them to shelf where the part belogs, where is QR code for item and some parts are already there. Operator then scan QR on shelf and it should increase count of existing item. I should see that increase in Item history.

### Reality
When someone change location of item from "items on transport" to "Central stock" there are 2 separate items and I must Merge them to one. And when I use Merge button, Item history is out/non complete/starting from blank.

### Question
Is there method how do this without merging? Idealy by scanning QRs?</div>

SchrodingersGat:
Related issue: https://github.com/inventree/inventree-app/issues/480

SchrodingersGat:
@MrShippeR if you would like the development team to assign priority to this you can fund the issue using the link above

Velenka1:
Hi, sorry to bother. I am unclear on the history shown here but will this issue will be addressed? We have just run into this on our system and for us it is a bit silly to have multiple stock items for a single part at the same location.

While merging does work, it is a hassle, especially where the stock history is concerned (which is important for us).

We also observe this problem occurring upon receiving line items from a purchase order.

SchrodingersGat:
@Velenka1 there has been no movement on this - no-body has stepped up to implement this and the core dev team are focused on larger efforts. Would this be something that you would be keen to implement?

neilbeukes:
I'm also facing this issue and would like to get a solution to this. 

I'm willing to pick this up. 

The Idea is to have an additional option here (where the star is) that you can enable and it will "merge" it into the first existing stock it finds of the same part in the new location. 

<img width="2285" height="241" alt="Image" src="https://github.com/user-attachments/assets/57d013e0-5e6e-465b-9174-b13f921288c3" />

Sorry about the mock design quality (insert graphic design is my passion meme here)

matmair:
I am sure we would review a PR; the request is pretty old so it should be verified that this was not implemented already

[...]

PR #12022, opened 2026-05-27
Allow stock transfer to merge into existing stock (optional) #6951

neilbeukes:
As requested in [#6951](https://github.com/inventree/InvenTree/issues/6951), I am providing an option when transferring stock between locations, to optionally merge into the first available stock of the same type (of part). 

## New parts added. 
- Option on transfer modal to enable merge after transfer
- Global settings line to set the default state of the option on the modal
- Tracking entries to signal transfer of stock
- Tests for new transfer-merge logic.

This is will change existing behavior as follows. 

## If merge is toggled
- Look for first matching stock item in destination location.
  - If matching stock found, perform a standard merge between transferred stock and existing matched stock.
  - if no matching stock found then follow existing transfer logic.
-  History of destination stock remains the same, and get another tracking event signalling stock transferred and from where.

## If merged is not toggled
- Existing behaviour stays intact i.e transferred stock will become its "own" stock in the destination location.

## Some images:

Global settings
image

Modal with merge enabled on transfer
image

Middle event is an example of a merged transfer
image


neilbeukes:
Oh I see some builds failing, ill take a look at those. I'm assuming I need to also update docs with added functionality.

matmair:
1) you need to bump the version in api_version.py and add a entry in the docstring below
2) you need to add a doc section for STOCK_MERGE_ON_TRANSFER - see global.md

src/backend/InvenTree/stock/models.py
```diff
@@ -2275,6 +2307,13 @@ def merge_stock_items(self, other_items, raise_error=False, **kwargs):
                 self.parent = None
                 self.save()
 
+            if other.location:
+                location_note = _('Transferred from %(location)s') % {
+                    'location': other.location.pathstring
+                }
+
+                notes = f'{notes}\n{location_note}' if notes else location_note
```
matmair: This feels like it will be very hard to trace back and also like a data loss

matmair (approved):

src/backend/InvenTree/stock/models.py
SchrodingersGat: Any newlines are stripped out from char fields too

neilbeukes: You guys are right, thanks for pointing it out. Stripping the existing tracking event and replacing it with a lesser detailed event makes no sense. Im mostly reusing the existing event now. 
image

SchrodingersGat:
@neilbeukes looking good! Only issue is that the code is not meeting style guidelines:

image


If you run `pre-commit install` in your local repo, it will automatically format the code for you on each git commit.

neilbeukes:
> @neilbeukes looking good! Only issue is that the code is not meeting style guidelines:
> image
> 
> If you run `pre-commit install` in your local repo, it will automatically format the code for you on each git commit.

Hi @SchrodingersGat. I would expect this to automatically run on commit as per the docs. 

Is there a way I can run that manually? I dont see any scripts in package.json or as a invoke-able task. 

Thanks,

SchrodingersGat:
> Hi @SchrodingersGat. I would expect this to automatically run on commit as per the docs.

It will, *after* you install the pre-commit hooks: https://docs.inventree.org/en/stable/develop/contributing/#setup-devtools

After this, any git commit will automatically format the code.

> Is there a way I can run that manually? I dont see any scripts in package.json or as a invoke-able task.

Yes, you can replicate what the [pre commit file](https://github.com/inventree/InvenTree/blob/master/.pre-commit-config.yaml) is doing

neilbeukes:
> > Hi @SchrodingersGat. I would expect this to automatically run on commit as per the docs.
> 
> It will, _after_ you install the pre-commit hooks: https://docs.inventree.org/en/stable/develop/contributing/#setup-devtools
> 
> After this, any git commit will automatically format the code.
> 
> > Is there a way I can run that manually? I dont see any scripts in package.json or as a invoke-able task.
> 
> Yes, you can replicate what the [pre commit file](https://github.com/inventree/InvenTree/blob/master/.pre-commit-config.yaml) is doing

I see, 

I did setup dev tools and have had all the pre-checks running everytime I commit but it doesnt seem to pick up those formatting. 

image

Ill look into the more this weekend, maybe I can see why its not running as intended.

matmair (approved):

neilbeukes:
Thanks @matmair

SchrodingersGat:
There are still outstanding conflicts and errors here - please address soon or we will have to bump to the next release cycle

neilbeukes:
How do I keep my api_version conflict free if a new push happens every day. :rofl:

matmair:
I have a shortcut but we should find a generic solution probably ;.)

neilbeukes:
@SchrodingersGat Hi, 

I see this was moved over to Milestone 1.5. 

I was hoping for it to be released in 1.4. Any issues with the PR? 

The builds and tests all pass, only 1 failed when Matt merged in master.

SchrodingersGat:
@neilbeukes thanks for the contribution, sorry that it missed the 1.4.0 window!

smithcv:
I'm trying to validate the merge.  When merging items, it looks like Inventree is doing a good job of preserving as many of the records as possible, I have not been able to break that in my testing.  I know that the link to the purchase orders breaks for one of the two items being merged.  The linking back to the purchase orders gets a bit weird here, but maybe that is worth spinning off a new item to address those.

SchrodingersGat:
@smithcv if you can propose a fix for that shortcoming, we would certainly welcome a PR!

PR #12029, opened 2026-05-28
WIP: docker use distroless

matmair:
this would reduce the potential vuln footprint a lot;
not ready for merge or prod

### DIFF
--- a/contrib/container/Dockerfile
+++ b/contrib/container/Dockerfile
@@ -9,8 +9,9 @@
 # - Runs InvenTree web server under django development server
 # - Monitors source files for any changes, and live-reloads server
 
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
How does this change reduce our vulnerability?

matmair:
It reduces the amount of included software to the absolute minimum required

wolflu05:
I am not sure if that's a great decision regarding the history of InvenTree docker images. Don't we fall into the same rabbit hole again then? I mean then we may forgot some deps for printing or anything else someone relies on.

matmair:
This low priority; closing for now
Switching prod to be leaner Debian base is still a goal to reduce patching efforts

PR #12035, opened 2026-05-30
Beta

jbbailey917:


SchrodingersGat:
we will not be accepting any PRs which:

- Have zero description, context, or justification
- The contributor clearly has not put any effort into reading our contribution guide
- Over 50K lines changed in a first contribution
- Reproduces copies of our configuration files for no discernible reason

jbbailey917:
@SchrodingersGat Sorry, was meant to target my own local branch!

matmair:
I would recommend looking into the plugin and parameter system, your changes are mostly duplicating features that already are there or would be a few hundred lines at most in a plugin. We release quite often in a year, almost always with some security fixes or improvements. having a large local patch set is probably going to cause a lot of work in the long run for you. That is why we added plugins.

jbbailey917:
@matmair I appreciate the comment. I went ahead and refactored my work and will continue using the plugin framework. This was something I was worried about originally but it kind of got away from me - thanks for the inspiration.

matmair:
You are welcome; no one benefits from you having to spend more time on updates

PR #12037, opened 2026-05-30
[UI] Fix navigation tabs

### LINKED ISSUE #11256: "Stock" item in toolbar/menus not visible when no permissions for "Stock Items"

### Describe the bug*

"Stock" toolbar is not visible with following group permissions:
Image

### Steps to Reproduce

1. Add a group with only view permissions for "Stock Locations"
2. Make sure that the "View" and all other permissions for "Stock Items" is dropped.
3. Observe that "Stock" is missing from main toolbar (and other locations) for users of this group

One can still manually browse the "{SITE}/web/stock/location/index/sublocations" URL, so in backend things seem to be ok.

### Expected behaviour

"Stock" would be still visible.

### Deployment Method

Docker

### Version Information

1.1.11.

I did not try to reproduce

Not reproducible

### Relevant log output

```shell

```

SchrodingersGat:
- Better fidelity of user permissions
- Closes https://github.com/inventree/InvenTree/issues/11256

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -115,7 +115,7 @@
         "@lingui/babel-plugin-lingui-macro": "^5.9.2",
         "@lingui/cli": "^5.9.2",
         "@lingui/macro": "^5.9.2",
-        "@playwright/test": "^1.58.2",
+        "@playwright/test": "^1.16.0",
         "@types/node": "^25.5.0",
         "@types/qrcode": "^1.5.5",
         "@types/react": "^19.2.14",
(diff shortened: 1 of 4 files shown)


src/frontend/package.json
matmair, line 1: Was this meant to be left here,

matmair (approved):

src/frontend/package.json
SchrodingersGat: Yes - see https://github.com/microsoft/playwright/issues/40998

PR #12056, opened 2026-05-31
[change] Delivery address

### LINKED ISSUE #9303: [FR] Delivery Address on Purchase Orders

SchrodingersGat:
This PR expands the functionality of the "Address" model, allowing it to be used to represent *internal* addresses (not just addresses associated with external companies).

This is an alternative approach to the one suggested in https://github.com/inventree/InvenTree/pull/9768

### Relates Issues

- Closes https://github.com/inventree/InvenTree/issues/9303
- Replaces https://github.com/inventree/InvenTree/pull/9768
- Update demo dataset - https://github.com/inventree/demo-dataset/pull/117

### Breaking Change

This is a "slightly" breaking change - in that the Address model no longer is definitely linked to a Company instance. Any external code which expects an Address to always link to a Company model will need to be updated

### Changes

- Address.company is now nullable; addresses with no linked company are "internal addresses" representing your own sites
- Internal addresses are managed from the Admin Center (staff only) under a new Addresses panel
- Purchase Orders, Return Orders, and Transfer Orders now use an internal address for their address field; the picker is filtered to internal addresses only
- Sales Orders are unchanged — they continue to use a customer-linked address
- If no explicit address is set on a PO/RO/TO, the primary internal address is used as the fallback (mirrors the existing company primary-address fallback on Sales Orders)
- New ?internal=true/false filter on the address list API endpoint
- order_address property updated for all order types to reflect the internal/external split
- Unit tests added for address field validation on all four order types (PO, SO, RO, TO) via the API, and for the order_address fallback behaviour
- Documentation updated: new Internal Addresses section in the company concepts page, including fallback address behaviour for both internal and external orders


matmair:
this will probably break a few plugins; can we move the change in the company link to 2.0?

src/backend/InvenTree/company/api.py
```diff
@@ -105,27 +106,67 @@ class ContactDetail(RetrieveUpdateDestroyAPI):
...
+
+    internal = rest_filters.BooleanFilter(
+        label=_('Internal Address'), field_name='company', lookup_expr='isnull'
+    )
+
+
+def _require_staff_for_internal_address(request, address_company) -> None:
+    """Raise PermissionDenied if the address is internal and the user is not staff."""
+    if address_company is None and not request.user.is_staff:
+        raise PermissionDenied(_('Only staff users can modify internal addresses'))
```
matmair, line 126: this forces ppl to use staff accounts - which are well-known sources for security issues; imo this should be a seperate permissions

matmair:
Having looked deeper at it: this is a huge breaking change; I do not really understand why we are breaking every existing PO, RO, TO effectively - the value-add of this feature is blurry to me at best.

SchrodingersGat:
So currently, PO / RO / TO - all specify addresses incorrectly (IMO):

## Purchase Order

- Must specify an address linked to the *supplier* - NOT the delivery address of the org using InvenTree

## Return Order

- Must specify an address linked to the *customer* - NOT the return address of the org using InvenTree

## Transfer Order

- Not linked to a company at all
- However the "address" field here *must* be linked to a Company

----------------------------------

I think that in each of these cases having an "internal" address (not linked to any Company) would make more sense, right?

wolflu05:
Sorry I have no real opinion on this, since I am not really using the addresses, but this may be a big breaking change for some other plugin developers. Just an idea, why cant we create an "Internal Company" and add adresses there? Then instead of company name in system settings we can select the internal company?

matmair:
I think this solution with an `internal` flag that fundamentally changes validation and breaking the meaning of existing date entries is not a slight change; it changes how whole integration flows would need to work.

IMO a suggestion to solve the stated problems should not change the meaning of existing data fields without even warning the users.
For POs this would mean adding a new field `delivery address` and renaming the label and helper text of the current field to `supplier adress`. The new field could map to a helper property `target_field` that is exposed by any type of order for keeping a similar python api on all orders.
For ROs the same but names `return address`(new) and `customer address`.

For TOs either limit it to addresses that are additionally flagged as `internal` or add a system setting that enables admins to select if they want non-internal addresses to be allowed for TOs. TOs were not shipped in a stable so changing stuff there does not worry me.

Additionally, I think it would make sense to add a way to have one or more "internal" companies (marked with a flag or setting) - packaging/shipping slips still need company information and this would keep the python API interfaces stable

SchrodingersGat:
Clearly we will need to think carefully about how this is implemented. I'm happy to support "internal" companies (and thus linked addresses) if we all feel that's a cleaner approach

PR #12059, opened 2026-05-31
Fix: Apply default location correctly in Return Stock form (#11568)

### LINKED ISSUE #11568: "Move to default location" button does not apply changes when returning items from Customer

### Describe the bug*

Hi everyone! We are back with an issue we found when moving items back from customers and we think it would be helpful to solve.

To clarify, we usually use customers as projects, to assigning items for a period of time and moving them back when it ends. 

When we want to take back the items, we move them to their default locations. However, the "default location" button in the form only shows the preview with the name of the location, but it doesn't apply the changes when you press it. This behaviour only occurs when you try to return items from a customer, when transfering items from another location, it works correctly.

Image

Image

Image

Image

Image

Image

Chrome console: 
Image

### Steps to Reproduce

1. Go to any stock item with default location and which is salable.
2. Assign it to a customer.
3. Go to "Return stock".
4. Press the "Move to default location" button.
5. Press the "Submit" button in the modal.

### Expected behaviour

We think that it's behaviour would be the same when you used this button transfering items from one location to another.

### Deployment Method

None

### Version Information

Demo InvenTree: 1.3.0 dev

I tried to reproduce

Reproducible

### Relevant log output

```shell

```

adityakrmishra:
Fixes #11568. Resolved the state binding issue where the default location was not applied during customer stock returns.

### DIFF
--- a/src/frontend/src/forms/StockForms.tsx
+++ b/src/frontend/src/forms/StockForms.tsx
@@ -39,11 +39,11 @@ import { useFormContext } from 'react-hook-form';
 import { useNavigate } from 'react-router-dom';
 import { api } from '../App';
 import RemoveRowButton from '../components/buttons/RemoveRowButton';
+import { StandaloneField } from '../components/forms/StandaloneField';
 import {
   TableFieldExtraRow,
   type TableFieldRowProps
 } from '../components/forms/fields/TableField';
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
@adityakrmishra thanks for the fix!

SchrodingersGat (approved):

PR #12066, opened 2026-06-02
fix(forntend): generate UI coverage again

SchrodingersGat:
### Changes

- Split frontend tests into separate workflow file (due to increased complexity)
- Split firefox and chromium into separate workflows

#### Firefox Build

- Compiles frontend code first
- Runs playwright tests against compiled code (on port :8000)
- Does not run coverage

#### Chromium Build

- Runs playwright tests against dev server (on port :5173)
- Runs coverage
- Is split into multiple shards (single shard takes over 1 hour for all our tests)

---------------------------------

By running separate workflows for firefox and chromium we gain the following advantages:

- Reduce run time (due to parallel workflows)
- Test dev (5173) and prod (8000) UI code


SchrodingersGat:
It would appear that changing from using compiled UI code to the vite live-load is causing the playwright tests to timeout entirely..

src/frontend/playwright.config.ts
```diff
@@ -26,44 +23,40 @@ const MAX_RETRIES: number = 3;
...
 
-const BASE_URL: string = IS_CI
-  ? 'http://localhost:8000'
-  : 'http://localhost:5173';
+const BASE_URL: string =
+  process.env.PLAYWRIGHT_BASE_URL || 'http://localhost:5173';
+
+// If running in "production" mode, we can use multiple workers to speed up the tests
+const MAX_WORKERS: number = BASE_URL.endsWith('8000') ? 3 : 1;
+const MAX_RETRIES: number = IS_CI ? 2 : 5;
```
matmair: 5 retries seems exsessive

.github/workflows/frontend.yaml
```diff
@@ -0,0 +1,383 @@
...
+          if [[ '${{ github.event_name }}' == 'push' && "$GITHUB_REF" == 'refs/heads/master' ]]; then
+            is_main_push=true
+          fi
+          if [[ '${{ github.repository }}' == 'inventree/InvenTree' && ( "$is_main_push" == 'true' || "$PERFORMANCE" == 'true' ) ]]; then
+            echo "runner=codspeed-macro" >> "$GITHUB_OUTPUT"
+            echo "submit-performance=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "runner=ubuntu-24.04" >> "$GITHUB_OUTPUT"
+            echo "submit-performance=false" >> "$GITHUB_OUTPUT"
+          fi
```
matmair: pretty sure this is not needed for frontend testing

.github/workflows/frontend.yaml
```diff
@@ -0,0 +1,383 @@
...
+      server: ${{ steps.filter.outputs.server }}
+      migrations: ${{ steps.filter.outputs.migrations }}
+      frontend: ${{ steps.filter.outputs.frontend }}
+      api: ${{ steps.filter.outputs.api }}
+      force: ${{ steps.force.outputs.force }}
+      cicd: ${{ steps.filter.outputs.cicd }}
+      requirements: ${{ steps.filter.outputs.requirements }}
+      runner-perf: ${{ steps.runner-perf.outputs.runner }}
+      performance: ${{ steps.performance.outputs.force-performance }}
+      submit-performance: ${{ steps.runner-perf.outputs.submit-performance }}
```
matmair: I think we only need `frontend` and `force` path filters / outputs

matmair (approved):

SchrodingersGat:
@matmair please review again when you have time - it seems to actually be working now:

https://github.com/inventree/InvenTree/actions/runs/27020005409/job/79751496165?pr=12066

matmair (approved):

matmair:
Nice; I like how it cuts down on wait time significantly

matmair:
IMO this could also be included on 1.4

SchrodingersGat:
Let's see how the coverage upload goes now

PR #12093, opened 2026-06-05
Locked plugin settings

### LINKED ISSUE #11769: [FR] Support for configuring plugins from config.yaml

### Problem statement

Currently plugins have to be configured manually through the admin web interface, which makes automatic deployment of Inventree instances somewhat annoying.

### Suggested solution

An extra `plugin_settings` field in config.yaml with a format something like:

```yaml
plugin_settings:
  PluginName1:
    SettingName1: true
    SettingName2: "asdf"
  PluginName2:
    SettingName3: false
```

On startup, these settings would get applied, and settings defined this way would ideally be greyed out in the admin interface.
If a setting can't be applied (plugin not installed, value is invalid, etc. a warning notification is emitted.

A way of applying these settings via API would also work, but feels way clumsier than just allowing it to be done via the already existing config file.

### Examples of other systems

_No response_

- [x] I want to develop this.

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/11769
- Allow plugin settings to be locked down via configuration, much like global settings

### Tasks

- [x] Documentation


matmair (approved):
This looks clean enough, I think this will be fine

matmair (approved):

PR #12103, opened 2026-06-06
[UI] Table column filters

SchrodingersGat:
Extend table filtering to allow direct filtering from each column. It does not add "new" filters - instead hooks into the already defined filterset

- Display which columns are currently filtered
- Expose existing filter elements directly to columns
- Allow add / edit / remove filters directly

### Screenshot

image

### Tasks

- [x] Update documentation
- [x] Add playwright tests
- [x] Update CHANGELOG


matmair (approved):
LGTM; great change for fast usability
I am wondering how one could make this new addition easier to debug / include for plugin developers

PR #12104, opened 2026-06-06
Added bulk edit and bulk delete to PO line items

noobxcamper:


SchrodingersGat:
@noobxcamper did you intent to submit this here without any details / context / explanation / justification? There can be no consideration of merging without these things.

Secondly The code submitted here does not conform with existing code design patterns in the codebase. If you want to proceed with this (after addressing the context issues above) there will be some required changes

PR #12107, opened 2026-06-07
Bug fix for SalesOrderStatusGroups

### LINKED ISSUE #12087: [FR] Filter for orders that are not Completed or Cancelled

### Problem statement

In the SO table you can select a filter called "Outstanding". If you set that filter to YES you will see SO's not in Status `Completed`, `Shipped`, or `Cancelled`.

Image

Image

We do not consider the order to be fully closed until the status is actually set to `Complete` (or `Cancelled`). For orders being Shipped there might still be some paperwork and follow up to do until they are moved to the `Complete` status.

I think this definition of "Outstanding" has been like that all the time why I suppose it's not really feasible to redefine it, but would it be possible to add a new filter for "Closed" SO's (i.e. status `Completed`or `Cancelled`)? By setting that to "NO" it would then list all orders where there are still work to be done.

### Suggested solution

1. Remove status ´Shipped` from the definition of outstanding
2. Add a new filter for `Closed` orders - status `Completed` or `Cancelled`
3. Add the possibility to select more than one status when filtering on status
4. Any other relevant solution to customize the filtering

Depending on the complexity of the selected solution I would be happy to give it a try to implement this. Option 1 and possibly option 2 I think I know how to do...

So far we have not used the SO filter at all because of this and instead visually used the Status column to identify open orders. But when the volume of orders is growing I see an obvious risk that Open orders are being missed if they are being delayed and are listed far down in the table or even on following pages.

### Examples of other systems

_No response_

- [ ] I want to develop this.

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/12087

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking Changes
 
-- [#9604](https://github.com/inventree/InvenTree/pull/9604) - refactors user API endpoint to be less ambiguous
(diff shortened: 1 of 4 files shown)


gunstr:
Thanks a lot 👍

PR #12117, opened 2026-06-08
[UI] Preview Drawer

### LINKED ISSUE #7364: Open detail pages in drawer when interacting on a table for faster navigation
That issue leads me to another idea, what about opening the parts detail page, but only in a drawer with about 90% width so that one can easily step through parts, check things and quickly continue with the next part. (We could do that for every tables detail page). But if we open the url directly (so we're not coming from a table), then we open the part in 100% without loading and showing the table first

_Originally posted by @wolflu05 in https://github.com/inventree/InvenTree/issues/7271#issuecomment-2120497054_

SchrodingersGat:
Implements a global "preview drawer" which displays a "quick view" of a particular item, rather than navigating to that item.

Closes https://github.com/inventree/InvenTree/issues/7364

### Tasks

- [x] Implementation of component
- [x] Add preview render functions for all internal models
- [x] Expose to plugins
- [x] Documentation (and screenshots)
- [x] Playwright tests
- [x] Update CHANGELOG


neilbeukes:
This is a great idea and would help with the UX experience a lot. If you need any specific help here, give me a shout @SchrodingersGat

SchrodingersGat:
@neilbeukes cheers, thanks for the offer! The next part of the process is deciding *what* we will display here, for the various model types. We will have to define components for each type of model to preview. Perhaps you could offer some design support in what should be displayed?

wolflu05:
Why dont we just reuse the entire part detail page, but just in the drawer instead of reinventing the wheel a second time for the same model?

wolflu05:
Initially I imagined it like having this whole page, but accessible from within a drawer, so that we can quickly check parameters, the avialable stock, ...

image

SchrodingersGat:
@matmair @wolflu05 ready for review. One open question: Do we make this default enabled or disabled via user preferences?

src/frontend/src/components/previews/GlobalPreviewDrawer.tsx
```diff
@@ -0,0 +1,34 @@
...
+    state.isSet('ENABLE_PREVIEW_PANEL')
+  );
+  const isOpen = usePreviewDrawerState((state) => state.isOpen);
+  const modelType = usePreviewDrawerState((state) => state.modelType);
+  const id = usePreviewDrawerState((state) => state.id);
+  const instance = usePreviewDrawerState((state) => state.instance);
+  const filters = usePreviewDrawerState((state) => state.filters);
+  const preview = usePreviewDrawerState((state) => state.preview);
+  const targetUrl = usePreviewDrawerState((state) => state.targetUrl);
+  const closePreview = usePreviewDrawerState((state) => state.closePreview);
```
matmair: this seems like something that should be one call ot usePreviewDrawer

src/frontend/src/components/previews/PreviewType.tsx
```diff
@@ -0,0 +1,102 @@
...
+    case ModelType.salesordershipment:
+      return SalesOrderShipmentPreviewComponent({ instance, modelId });
+    case ModelType.transferorder:
+      return TransferOrderPreviewComponent({ instance, modelId });
+    case ModelType.stocklocation:
+      return StockLocationPreviewComponent({ instance, modelId });
+    case ModelType.partcategory:
+      return PartCategoryPreviewComponent({ instance, modelId });
+    default:
+      return null;
```
matmair: could this be a lookup and lazy loaded? we seem to add more and more points where one needs to add a new model

src/frontend/src/components/tables/InvenTreeTable.tsx
```diff
@@ -700,6 +705,18 @@ export function InvenTreeTableInternal<T extends Record<string, any>>({
...
+
+  // Callback to display "preview" view for a row (if available)
+  const showRowPreview = useCallback(
+    (pk: string | number) => {
+      if (tableProps.modelType && pk) {
+        previewDrawer.openPreview(tableProps.modelType, Number(pk));
+      }
+    },
+    [tableProps.modelType]
+  );
```
matmair, line 718: This seems to be repeated a lot; would it make sense to move this to InvenTreeTable?

matmair (approved):
Looks great;
I see some potential to make this easier to adopt in plugins / forks with a bit of restructuring

matmair:
> @matmair @wolflu05 ready for review. One open question: Do we make this default enabled or disabled via user preferences?

I think this should be default off for now; we can switch it on with 2.0

src/frontend/src/components/previews/PreviewType.tsx
SchrodingersGat: I've refactored this into `ModelInformationDict` now

SchrodingersGat:
@neilbeukes if you would like to assist, a thorough review of the UX in this branch would be greatly appreciated

neilbeukes:
> @neilbeukes if you would like to assist, a thorough review of the UX in this branch would be greatly appreciated

Sorry man, 

I was away on business trip for like 2 weeks. Only got back to my personal machine now.

PR #12128, opened 2026-06-08
feat(frontend): add hotkey registration function and hotkey helper modal

### LINKED ISSUE #358: Hotkey discovery not working

matmair:
This adds a new (wrapper) hook to register hotkeys that helps to make hotkeys easier to discover. Note: We do not use hotkeys to their full potential right now;

grafik
This dialog is placed in the spotlight as an action and is dynamic (ie. plugins can use it too) based on which hotkeys are actually mounted.

Based on real world user feedback
closes https://github.com/invenhost/InvenTree/issues/358

### DIFF
--- a/src/frontend/CHANGELOG.md
+++ b/src/frontend/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 This file contains historical changelog information for the InvenTree UI components library.
 
+### 1.4.6 - June 2026
+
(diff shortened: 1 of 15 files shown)


matmair:
this wrapper is necessary because event listeners can not be listed as far as I found out.
All hotkeys registered outside this helper are naturally not discoverable.

SchrodingersGat (approved):
Nice, this looks like a great functional addition

PR #12142, opened 2026-06-10
[report] Printing fixes

SchrodingersGat:
Implements a number of permissions fixes for printing labels and reports:

- Prevent users from printing reports against models they do not have "VIEW" permissions fro
- Prevent printing against disabled templates
- Enhanced unit tests


matmair (approved):

matmair:
@SchrodingersGat as this changes API behavior I think this should be considered breaking and not be back ported

PR #12151, opened 2026-06-10
Make plugin registry hash independent of plugin discovery order

### LINKED ISSUE #12150: Plugin registry hash depends on plugin discovery order, causing cross-process reload loops

### Describe the bug*

`calculate_plugin_hash()` in `plugin/registry.py` iterates `self.plugins.items()` in insertion order, which is the plugin discovery order of the local process. Two processes can hold the *same* registry state (same plugins, same versions, same active flags) in a *different* order and compute different hashes.

When that happens, every process that sees `_PLUGIN_REGISTRY_HASH` ≠ its own hash triggers a full registry reload and writes its own hash back — so two process groups with different discovery orders ping-pong the setting forever. On my instance this produced ~118 "Plugin registry hash has changed - reloading" events per hour, with `/api/` requests blocking 20-25s each while reloads ran, until health checks started timing out.

The subtle part is that discovery order is *stable* for a fixed site-packages state, so a freshly-restarted instance looks completely healthy. The order can shift when a plugin package is installed/reinstalled while the server is running: processes booted before the change keep their old in-memory order, anything booted after (respawned gunicorn workers, the background worker, a management shell) computes the new one. That's why this is easy to live with for months and then suddenly very loud, and why "restart everything" always appears to fix it.

Forensics from my instance (InvenTree 1.3.4, docker, 4 gunicorn workers + background worker, 16 active plugins), captured mid-loop: all 4 gunicorn workers agreed on hash `3ea062f3…` while a fresh `manage.py shell` in the same container computed `15d58d98…` — and dumping `slug|name|version|is_active` per plugin from both contexts showed identical tuples. Same set, different order. I also verified the proposed fix at runtime: a sorted-variant hash computes identically across every process context I could probe (server shells, background worker container).

### Steps to Reproduce

1. Run InvenTree with multiple gunicorn workers (e.g. `INVENTREE_GUNICORN_WORKERS=4`) and the background worker.
2. `pip install` a plugin package into the running container (i.e. without restarting the server processes), or otherwise change site-packages in a way that shifts entry-point discovery order.
3. Wait for any new process to boot with the new discovery order (worker respawn, shell, restarting just one container) while old processes keep serving.
4. Watch the logs: `Plugin registry hash has changed - reloading` repeats indefinitely while `/api/` latency degrades.

Timing of step 3 makes this awkward to reproduce on demand — on my instance the loop first appeared ~20h after the install. The mechanism is easier to demonstrate directly: compute `registry.calculate_plugin_hash()`, reverse `registry.plugins`, compute again. Same registry state, different hash.

### Expected behaviour

The registry hash represents the registry *state* (which plugins, which versions, active or not), independent of the iteration order of whichever process computes it.

### Deployment Method

- [x] Docker

### Version Information

InvenTree 1.3.4 (docker), observed mechanism unchanged on current master (`plugin/registry.py`, `calculate_plugin_hash`).

### Relevant log output

```text
INFO Plugin registry hash has changed - reloading
INFO Plugin registry hash has changed - reloading
INFO Plugin registry hash has changed - reloading
(~118/hour, /api/ blocking 20-25s per request while reloading)
```

I have a one-line fix (sort by slug before hashing) plus a regression test ready — PR incoming right after this issue.

Nasawa:
Fixes #12150

`calculate_plugin_hash()` iterated `self.plugins.items()` in insertion order, which is the plugin discovery order of the local process. Two processes holding the same registry state in a different order computed different hashes and ping-ponged `_PLUGIN_REGISTRY_HASH`, driving each other into endless registry reloads via `check_reload()` (full forensics in the issue).

This sorts by slug before hashing, so the hash represents the registry state rather than any process's iteration order.

Also adds a regression test: reverse `registry.plugins` and assert the hash is unchanged. The test fails against the old implementation and passes with the sort.

One behavioural note: existing deployments will compute a different hash the first time they run this code, causing exactly one reload wave on upgrade — same as any plugin set change today.

### DIFF
--- a/src/backend/InvenTree/plugin/registry.py
+++ b/src/backend/InvenTree/plugin/registry.py
@@ -1027,7 +1027,11 @@ class PluginsRegistry:
         data = md5()
 
         # Hash for all loaded plugins
-        for slug, plug in self.plugins.items():
+        # Note: Sort by slug, so the hash is independent of discovery order.
+        # Different processes can discover the same plugins in a different
+        # order, and the hash must represent the registry *state*, not the
+        # iteration order of any particular process.
+        for slug, plug in sorted(self.plugins.items()):
             data.update(str(slug).encode())
             data.update(str(plug.name).encode())
             data.update(str(plug.version).encode())
--- a/src/backend/InvenTree/plugin/registry.py
+++ b/src/backend/InvenTree/plugin/registry.py
@@ -1031,7 +1031,7 @@ class PluginsRegistry:
         # Different processes can discover the same plugins in a different
         # order, and the hash must represent the registry *state*, not the
         # iteration order of any particular process.
-        for slug, plug in sorted(self.plugins.items()):
+        for slug, plug in sorted(self.plugins.items(), key=lambda item: item[0]):
             data.update(str(slug).encode())
             data.update(str(plug.name).encode())
             data.update(str(plug.version).encode())
--- a/src/backend/InvenTree/plugin/test_plugin.py
+++ b/src/backend/InvenTree/plugin/test_plugin.py
@@ -538,6 +538,27 @@ class RegistryTests(TestQueryMixin, PluginRegistryMixin, TestCase):
             registry.registry_hash = 'abc'
             self.assertTrue(registry.check_reload())
 
+    def test_registry_hash_order_independence(self):
+        """Test that the registry hash does not depend on plugin iteration order.
+
+        Different processes (gunicorn workers, background worker, shell) can
+        discover the same set of plugins in a different order. If the hash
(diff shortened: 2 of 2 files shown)


matmair (approved):

SchrodingersGat:
@Nasawa thank you for looking into this and submitting a fix. An oversight that has gone unnoticed for quite some time I would imagine. 

Thanks for the contribution

wolflu05:
Just out of interest, do you think there is the same error with the machine registry hash?

matmair:
Could be; I think we should just factor out the hashing mechanism into a general implementation, in which we address these things centrally.

PR #12168, opened 2026-06-15
Permissions fix

SchrodingersGat:
Tighten permission scopes and roles around certain API endpoints

### DIFF
--- a/src/backend/InvenTree/common/api.py
+++ b/src/backend/InvenTree/common/api.py
@@ -1405,7 +1405,7 @@ class ObservabilityEndSerializer(serializers.Serializer):
 class ObservabilityEnd(CreateAPI):
     """Endpoint for observability tools."""
 
-    permission_classes = [AllowAnyOrReadScope]
+    permission_classes = [IsAuthenticated, AllowAnyOrReadScope]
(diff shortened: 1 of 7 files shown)


src/backend/InvenTree/common/api.py
matmair, line 1: @SchrodingersGat this was intended to be unauthenticared as frontend login tracing does not work without

src/backend/InvenTree/plugin/api.py
```diff
@@ -174,7 +174,10 @@ class PluginDetail(RetrieveDestroyAPI):
 
     queryset = PluginConfig.objects.all()
     serializer_class = PluginSerializers.PluginConfigSerializer
-    permission_classes = [InvenTree.permissions.IsSuperuserOrReadOnlyOrScope]
+    permission_classes = [
+        permissions.IsAuthenticated,
+        InvenTree.permissions.IsSuperuserOrReadOnlyOrScope,
+    ]
```
matmair, line 180: IMO it had value to have access to this data without login; i do not agree that that is a security issue.
All confidential info should be marked accordingly and will be filtered out

PR #12169, opened 2026-06-15
Data import permissions

SchrodingersGat:
Adjust API permissions for data import sessions to ensure that:

- Users can only access row data associated with import sessions they control
- Users can only access column data associated with import sessions they control
- Import session data is cleared once the import session is completed


matmair:
This has broken frontend tests in src/frontend/tests/pui_importing.spec.ts:5

PR #12180, opened 2026-06-16
[import] specify fk lookup field

SchrodingersGat:
- https://github.com/inventree/InvenTree/discussions/12170
- allow users to explicitly specify fk lookup fields on import


src/backend/InvenTree/importer/models.py
```diff
@@ -765,21 +841,35 @@ def lookup_related_field(self, field_name: str, value: str) -> Optional[int]:
...
         base_filters = (
             self.session.field_filters.get(field_name, {})
             if self.session.field_filters
             else {}
         )
 
-        # First priority is the PK (primary key) field
+        if lookup_field:
+            # A specific lookup field has been chosen by the user — query only that field
+            try:
```
matmair, line 852: Can we check lookupfield is an actual string and not a dictionary?

matmair (approved):

src/backend/InvenTree/importer/models.py
SchrodingersGat: On it

PR #12184, opened 2026-06-16
fix(backend): import session metadata

matmair:
Add storage for historic import results for reporting and display purposes

fixes breakage from https://github.com/inventree/InvenTree/pull/12169 and https://github.com/inventree/InvenTree/pull/11982


SchrodingersGat (approved):
Nice catch, I had not considered this as part of the previous change

matmair:
@SchrodingersGat this should be ready to merge and fix the last few frontend CI breaks

SchrodingersGat (approved):
Thanks for addressing these issues @matmair

PR #12187, opened 2026-06-17
Report Generation Updates

SchrodingersGat:
A number of improvements to report generation framework:

- Fix editor, allow user to drag the divider between editor and preview
- Correctly extract error information from server response
- Handle errors when report generation offloaded to the background worker
- Prevent multiple print attempts in case of error
- Additional e2e playwright testing


src/backend/InvenTree/report/models.py
```diff
@@ -578,13 +579,13 @@ def print(
...
                     msg = _('Template syntax error')
-                    output.mark_failure(msg)
+                    output.mark_failure(str(e) or msg)
                     raise ValidationError(f'{msg}: {e!s}')
                 except ValidationError as e:
                     output.mark_failure(str(e))
                     raise e
                 except Exception as e:
-                    msg = _('Error rendering report')
+                    msg = str(e) or _('Error rendering report')
```
matmair: Have you checked that we won't expose too much information to the user this way? The Exception is pretty wide

src/backend/InvenTree/report/models.py
```diff
@@ -617,13 +618,13 @@ def print(
...
                         msg = _('Template syntax error')
-                        output.mark_failure(error=_('Template syntax error'))
+                        output.mark_failure(error=str(e) or msg)
                         raise ValidationError(f'{msg}: {e!s}')
                     except ValidationError as e:
                         output.mark_failure(str(e))
                         raise e
                     except Exception as e:
-                        msg = _('Error rendering report')
+                        msg = str(e) or _('Error rendering report')
```
matmair: Same question as above

matmair (approved):

src/backend/InvenTree/report/models.py
SchrodingersGat: This error information is valuable when creating the template as it allows the template developer to determine what failed. Without this information, debugging broken templates is extremely difficult...

SchrodingersGat: If there is any specific "leaky information" you are worried about we can patch for that

matmair: Exception is just a verrrry broad surface: anything that raises an exception with sensitive info will have that returned straight to the users - ie. a git url that includes a PAT or basic auth for example.

Can we specifically trigger the most common exception types where we know they are save? This catchall is just an open-ended exfiltration without having to enabled DEBUG

SchrodingersGat: I have adjusted this now, any "catch all" exceptions leak no information - staff users can always view the error logs for more info

SchrodingersGat: This should now be consistent with previous behaviour - LMK if you are happy with this.

matmair (approved):

PR #12194, opened 2026-06-18
[bug] Fix double save

SchrodingersGat:
Fixes API bug where items are saved twice - where a single DB write would suffice

### DIFF
--- a/src/backend/InvenTree/stock/api.py
+++ b/src/backend/InvenTree/stock/api.py
@@ -1512,10 +1512,7 @@ class StockItemTestResultList(
         Also, check if an attachment was uploaded alongside the test result,
         and save it to the database if it were.
         """
-        # Capture the user information
-        test_result = serializer.save()
-        test_result.user = self.request.user
-        test_result.save()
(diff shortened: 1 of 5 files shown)


src/backend/InvenTree/stock/api.py
```diff
@@ -1676,14 +1673,7 @@ def create(self, request, *args, **kwargs):
         serializer = self.get_serializer(data=data)
         serializer.is_valid(raise_exception=True)
 
-        # Record the user who created this Part object
-        item = serializer.save()
-        item.user = request.user
-        item.system = False
-
-        # quantity field cannot be explicitly adjusted  here
-        item.quantity = item.item.quantity
```
matmair, line 1685: Is setting item.quantity not needed anymore?

matmair (approved):

src/backend/InvenTree/stock/api.py
SchrodingersGat: This endpoint does not actually support POST action at all - so I've now removed this code entirely

PR #12203, opened 2026-06-19
build(docker): compile backend translations during image build

### LINKED ISSUE #12188: Backend localization fails due to missing compilemessages in build stage

### Describe the bug*

I am writing to report a localization gap in the standard Docker deployment, which causes the application to fail to render localized text (such as field label and help_text served by the backend API) when switching to any language other than English.

Root Cause:
After digging into the git history, I found that commit d557d2dff modified the `invoke update` command and decoupled the localization generation logic from it. I assume the intention was to keep the container runtime environment clean and stateless, preventing the container from modifying source artifacts during startup (which makes total sense).

However, because `*.mo` files are excluded by .gitignore, removing this step from the runtime lifecycle without introducing it to the Docker image build stage (Dockerfile) leaves the localization pipeline uncompleted. In a production container environment, the backend currently has no way to serve compiled translation binaries out-of-the-box.

### Steps to Reproduce

1. Deploy InvenTree using the standard Docker / Docker-compose environment.
2. Switch the application language to any language other than English (e.g., German, French, or Chinese).
3. Inspect any dynamically rendered form and check its backend API response. (e.g., `OPTIONS /api/part/`)
4. Notice that fields fallback to English because the backend cannot load .mo translations.
5. Manually run `python manage.py compilemessages` inside the backend container and restart it, the translations instantly load correctly.

### Expected behaviour

To close this loop while respecting the philosophy of keeping runtime clean, the translation messages should be compiled during the Docker image build phase in the Dockerfile (e.g., RUN python manage.py compilemessages or via an equivalent invoke task during multi-stage build).

Before I open a Pull Request to fix this, I wanted to check in with the team to confirm your thoughts: Is this an overlooked gap from the previous refactoring, or is there an alternative preferred pattern for handling production i18n in InvenTree Docker images?

### Deployment Method

Docker

### Version Information

Inventree Version: `1.3.5`

I tried to reproduce

Reproducible

### Relevant log output

```shell

```

geekcj12:
# Summary

This PR compiles the Django backend translation files (.mo) during the production Docker image build, instead of relying on them being generated at runtime.

# Changed

- `contrib/container/Dockerfile`: Added a build step in the production stage that runs invoke int.backend-compilemessages, so compiled `.mo` files are baked into the image.
- `tasks.py`: Added a new int.backend-compilemessages invoke task that compiles the backend translations via django compilemessages without loading the full InvenTree settings (lightweight, build-friendly).
- `.github/workflows/docker.yaml`: Extended the Test Docker Image step to assert the compiled .mo artifacts exist for a few representative languages (de, fr, ru, zh_Hans), guarding the new build step.

# Testing approach

The check is added to the existing Test Docker Image step in docker.yaml, consistent with how other in-image artifacts (manage.py, requirements.txt, etc.) are already verified — a docker run ... test -f against each expected path.

It intentionally covers only a few representative languages rather than the full set. The compilation step runs django compilemessages, which iterates over every language in the locale/ directory in a single pass. If compilation fails for any reason (a malformed .po file, a missing gettext, an incorrect working directory), the command returns a non-zero exit code and the Docker build itself fails — so "all languages compiled" is already enforced by the build succeeding. The CI check therefore isn't trying to re-verify every language; its purpose is narrower and complementary:
- Confirm the artifacts landed at the expected path — that .mo files actually exist under locale/<lang>/LC_MESSAGES/, not just that the command exited cleanly.
- Guard against regressions — if someone later removes or relocates the compile step, or the output path changes, the check fails fast.

A small, diverse sample (de, fr, ru, zh_Hans — covering Latin, Cyrillic, and CJK scripts) is enough to catch these failure modes. Iterating over all ~40 languages would add noticeable runtime, make the step brittle against a single edge-case locale, and provide no additional signal that the build success doesn't already give us.

---

Fixes #12188

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -84,6 +84,11 @@ jobs:
           docker run --rm inventree-test test -f /home/inventree/gunicorn.conf.py
           docker run --rm inventree-test test -f /home/inventree/src/backend/requirements.txt
           docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/manage.py
+          # Check that backend translations were compiled into the image (a few representative languages)
+          docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/de/LC_MESSAGES/django.mo
+          docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/fr/LC_MESSAGES/django.mo
+          docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/ru/LC_MESSAGES/django.mo
+          docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/zh_Hans/LC_MESSAGES/django.mo
       - name: Build Docker Image
         # Build the development docker image (using docker-compose.yml)
         run: docker compose --project-directory . -f contrib/container/dev-docker-compose.yml build --no-cache
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -89,6 +89,11 @@ jobs:
           docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/fr/LC_MESSAGES/django.mo
           docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/ru/LC_MESSAGES/django.mo
           docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/zh_Hans/LC_MESSAGES/django.mo
+          # Check that frontend translations (lingui) were compiled into the static build output (a few representative languages)
+          docker run --rm inventree-test grep -q '"src/locales/de/messages.ts"' /home/inventree/src/backend/InvenTree/web/static/web/.vite/manifest.json
+          docker run --rm inventree-test grep -q '"src/locales/fr/messages.ts"' /home/inventree/src/backend/InvenTree/web/static/web/.vite/manifest.json
+          docker run --rm inventree-test grep -q '"src/locales/ru/messages.ts"' /home/inventree/src/backend/InvenTree/web/static/web/.vite/manifest.json
+          docker run --rm inventree-test grep -q '"src/locales/zh_Hans/messages.ts"' /home/inventree/src/backend/InvenTree/web/static/web/.vite/manifest.json
       - name: Build Docker Image
         # Build the development docker image (using docker-compose.yml)
         run: docker compose --project-directory . -f contrib/container/dev-docker-compose.yml build --no-cache
--- a/contrib/container/Dockerfile
+++ b/contrib/container/Dockerfile
@@ -145,6 +145,9 @@ ENV PATH=/root/.local/bin:$PATH
 COPY --from=builder_stage ${INVENTREE_BACKEND_DIR}/InvenTree/web/static/web ${INVENTREE_BACKEND_DIR}/InvenTree/web/static/web
 COPY --from=builder_stage /root/.local /root/.local
 
+# Compile Django backend translations during image build
+RUN bash -c "cd '${INVENTREE_HOME}' && invoke int.backend-compilemessages"
+
 # Launch the production server
 CMD ["sh", "-c", "exec gunicorn -c ./gunicorn.conf.py InvenTree.wsgi -b ${INVENTREE_WEB_ADDR}:${INVENTREE_WEB_PORT} --chdir ${INVENTREE_BACKEND_DIR}/InvenTree"]
 
--- a/tasks.py
+++ b/tasks.py
@@ -906,6 +906,23 @@ def backend_trans(c, verbose: bool = False):
     success('Backend translations compiled successfully')
 
 
+@task(help={'verbose': 'Print verbose output'})
+@state_logger('backend_compilemessages')
+def backend_compilemessages(c, verbose: bool = False):
+    """Compile backend Django translation files without loading InvenTree settings."""
+    info('Compiling backend translations...')
+
+    cmd = 'python3 -m django compilemessages'
+
+    if verbose:
+        cmd += ' -v 1'
+    else:
+        cmd += ' -v 0'
+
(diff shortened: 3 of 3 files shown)


tasks.py
```diff
@@ -906,6 +906,23 @@ def backend_trans(c, verbose: bool = False):
     success('Backend translations compiled successfully')
 
 
+@task(help={'verbose': 'Print verbose output'})
+@state_logger('backend_compilemessages')
+def backend_compilemessages(c, verbose: bool = False):
```
SchrodingersGat, line 911: This additional task is not needed here. Just run `invoke translate --no-frontend`

contrib/container/Dockerfile
```diff
@@ -145,6 +145,9 @@ ENV PATH=/root/.local/bin:$PATH
 COPY --from=builder_stage ${INVENTREE_BACKEND_DIR}/InvenTree/web/static/web ${INVENTREE_BACKEND_DIR}/InvenTree/web/static/web
 COPY --from=builder_stage /root/.local /root/.local
 
+# Compile Django backend translations during image build
+RUN bash -c "cd '${INVENTREE_HOME}' && invoke int.backend-compilemessages"
```
SchrodingersGat, line 149: Use `invoke translate` here instead

.github/workflows/docker.yaml
```diff
@@ -84,6 +84,11 @@ jobs:
           docker run --rm inventree-test test -f /home/inventree/gunicorn.conf.py
           docker run --rm inventree-test test -f /home/inventree/src/backend/requirements.txt
           docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/manage.py
+          # Check that backend translations were compiled into the image (a few representative languages)
+          docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/de/LC_MESSAGES/django.mo
+          docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/fr/LC_MESSAGES/django.mo
+          docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/ru/LC_MESSAGES/django.mo
+          docker run --rm inventree-test test -f /home/inventree/src/backend/InvenTree/locale/zh_Hans/LC_MESSAGES/django.mo
```
SchrodingersGat, line 91: This is great - we should also check that the frontend messages are present also

SchrodingersGat (changes requested):

tasks.py
geekcj12: Thanks for the review! I tried running `invoke translate --no-frontend` locally to build the image, but unfortunately, it doesn't work in this context. Before making any changes, I'd love to clarify the underlying bottleneck.

During the Docker production build stage, the database is not yet configured (the config.yaml is only available in the runtime data volume, and `INVENTREE_DB_*` env vars do not exist at build time).

The native translate task triggers `makemessages / compilemessages` via `manage.py`, which implicitly sets `DJANGO_SETTINGS_MODULE=InvenTree.settings`. When `settings.py` is loaded, it executes `get_db_backend()` at the module's top level. When required fields like ENGINE or NAME are missing, it raises a ValueError: Missing required database configuration key. As a result, the Docker build fails early during the initialization process.

This is exactly why I introduced the `int.backend-compilemessages` task. By running `python3 -m django compilemessages` directly, it bypasses loading the entire InvenTree settings module, completely decoupling the step from any database configurations.

Additionally, it fits the production image build philosophy better, at this stage, we only need to compile existing `.po` files to `.mo`. We do not need to extract source strings via `makemessages` or trigger other side effects.

Am I missing any build-time DB config? If not, I'd prefer to keep the lightweight custom task approach. If the task naming is a concern, I'm more than happy to rename it to match the project's convention!

.github/workflows/docker.yaml
geekcj12: Thanks for the feedback! I have already implemented and verified this verification logic in commit [1e534d6].

**How the frontend verification works:**

The frontend messages.ts source files only exist during the Docker builder stage and are excluded from the final production image. The production target only grabs the compiled Vite outputs.

In the production image, Vite bundles each `messages.ts` file into a hashed chunk and registers it inside `web/static/web/.vite/manifest.json`, using `src/locales/<lang>/messages.ts` as the lookup key.

 So the faithful analog is grepping that manifest for the same representative languages the backend uses (de, fr, ru, zh_Hans). Each command exits non-zero (failing the step) if a locale didn't make it into the build output.

tasks.py
SchrodingersGat: @geekcj12 ok fair point, in that case why not just run the manage.py python command directly in the django directory, rather than a separate wrapper?

e.g. something like

```docker
RUN bash -c "cd '${INVENTREE_HOME}/src/backend/InvenTree' && python ./manage.py compilemessages"
```

geekcj12: @SchrodingersGat That's a reasonable question! However, running the manage.py command directly inside the Django directory will unfortunately hit the exact same roadblock.

Here are the core technical considerations behind the current implementation:

1. manage.py Still Requires the Database: Any command invoked via python manage.py <command> implicitly sets up the Django application registries and fully loads InvenTree.settings. As mentioned earlier, loading settings.py triggers get_db_backend() at the module's top level, which immediately raises a ValueError during Docker build time due to the absence of the database configuration environment.
2. Following Existing Project Patterns: I intentionally designed this by mimicking the existing int.frontend-compile task pattern to introduce an internal, build-time task:
    - No manage.py Dependency: It runs `python3 -m django compilemessages` directly under the hood, which boots up the pure Django translation compiler without touching the InvenTree settings module, cleanly bypassing the database issue.
    - It explicitly marks this as a build-time/internal asset preparation step, preserving its user-facing semantics without breaking changes.
    - Maintainable Dockerfile: This keeps the Dockerfile clean and prevents scattering raw, complex shell commands inside it. All build-time automation entry points remain centralized inside tasks.py.

matmair (approved):

SchrodingersGat (approved):

SchrodingersGat:
@geekcj12 thanks for tracking down this issue and providing a fix!

PR #12206, opened 2026-06-19
New logo

SchrodingersGat:
Playing around with a new simplified version of the logo

### Example

| Old | New|
| --- | --- | 
| inventree | inventree-3 |
| image | image |
| image | image |

If anyone with actual graphic design chops has any ideas...

### Other Changes

Other repos to update with the change:

- [ ] website
- [ ] README
- [ ] Github repos
- [ ] docs
- [ ] update all png copies
- [ ] app
- [ ] crowdin


matmair:
Is there a reason behind switching the colors?

SchrodingersGat:
> Is there a reason behind switching the colors?

Increased contrast, basically

matmair:
In my very personal and not qualified opinion the new logo feels soulless and less technical than the old one. It is so far removed from a box or the original logo that is becomes hard to tell what the meaning is, especially of the brown elements.
Maybe I am alone with this opinion, but I do not see what is wrong with the current one. The switch before I understood, but for this I do not really get the benefits. Maybe those could be formulated? Why do we need a new logo?

SchrodingersGat:
I do not think it is that dissimilar, my main priority was to make it a bit more visually distinct. Especially when it is small, the combination of three separate colors and thin black lines are "lost". 

If we wanted to get "abstract" about it, the separate brown outer section represent the various plugins that "surround" the core functionality of InvenTree ;)

But, I just think it looks better :p

matmair:
maybe we post about this in https://opensourcedesign.net/ and see what professional eyes create with these objectives? To me the new proposal looses identity and looks generic

guest20:
I've not contributed to the project before, but I like the old icon better.

I think you can replace the infinitely thin black lines with darker/lighter surfaces when you're targeting really low pixel density situations... you can achieve a 3d cube-of-cube look with two tones of each colour.

The scales here are cheated, they're all the same image, but:

16: image 32: image 64: image 256: image

I did this in mspaint real quick, so apologies to anybody who's good at graphic design...

PR #12208, opened 2026-06-19
Report locale updates

SchrodingersGat:
- Adds support for specifying locale in various reporting helper functions
- Useful when the report should be generated to match a different locale
- This allows precise control of how currencies are rendered in reports
- Additional unit tests

The following reporting helper functions are now "locale aware":

- render_currency
- format_date
- format_number


src/backend/InvenTree/report/templatetags/report.py
```diff
@@ -871,101 +890,138 @@ def render_html_text(text: str, **kwargs):
 @register.simple_tag
 def format_number(
     number: int | float | Decimal,
-    decimal_places: Optional[int] = None,
     multiplier: Optional[int | float | Decimal] = None,
     integer: bool = False,
-    leading: int = 0,
```
matmair, line 877: Is this not breaking? I seems like it was replaced with `min_digits` for existing usages might have issues

SchrodingersGat: Good point, I have reverted this to use the same name

matmair (approved):

Issue #12209, opened 2026-06-19
Login with superuser account fails because it is treated as registration attempt


MartinHenselHLL:
### Deployment Method

- [ ] Installer
- [ ] Docker Development
- [x] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

I am trying inventree for the first time and deployed a completely new docker-compose install on Ubuntu-24 following the directions given in https://docs.inventree.org/en/stable/start/docker_install/#docker-installation

After the initial db setup via `invoke update` all containers come up as expected. However all login attempts using the superuser account created either in the `.env` file or via the `invoke superuser` fail. The inventree-server container reports `INVE-W12: Signup attempt blocked, because registration is disabled via setting LOGIN_ENABLE_REG.` which should not matter, when attempting a sign-in with an existing account?

### Steps to Reproduce

0. Follow docs https://docs.inventree.org/en/stable/start/docker_install/#docker-installation to deploy with docker-compose.
1. Run db initialization with `docker compose run --rm inventree-server invoke update`.
2. Create a superuser account by either
   a. Setting up the necessary variables in the `.env` file (`INVENTREE_ADMIN_USER`, `INVENTREE_ADMIN_PASSWORD`, `INVENTREE_ADMIN_EMAIL`).
   b. Running the superuser command `docker compose run inventree-server invoke superuser`.
3. Deploy the full container stack with `docker compose up -d`
4. Navigate to the instance landing page
5. Try to log in with the superuser credentails.
6. The frontend shows Error 403: `Check your input and try again.`
7. The backend logs can be found below.

Running the `docker compose run inventree-server invoke superuser` command again confirms, that the superuser account exists.

### Relevant log output

```bash
inventree-server  | 2026-06-19T11:23:46.378685Z [warning  ] INVE-W12: Signup attempt blocked, because registration is disabled via setting LOGIN_ENABLE_REG. [inventree] ip=##.###.###.## request_id=bee88131-5414-4ca9-ba92-31086f4aac7f user_id=None
[...]

PR #12216, opened 2026-06-20
Add meaningful message on CSRF failure

SchrodingersGat:
Ref: https://github.com/inventree/InvenTree/issues/12209

Provide an error message to the user if CSRF is not properly configured. This can be difficult to ascertain without otherwise digging into the server logs.

image

### DIFF
--- a/src/backend/InvenTree/InvenTree/middleware.py
+++ b/src/backend/InvenTree/InvenTree/middleware.py
@@ -106,6 +106,32 @@ apps_mfa_bypass = [
 """App namespaces that bypass MFA enforcement - normal security model is still enforced."""
 
 
+def csrf_failure(request, reason=''):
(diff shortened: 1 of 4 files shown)


src/backend/InvenTree/InvenTree/middleware.py
```diff
@@ -106,6 +106,32 @@ def ensure_slashes(path: str):
 """App namespaces that bypass MFA enforcement - normal security model is still enforced."""
 
 
+def csrf_failure(request, reason=''):
```
matmair, line 109: where is this called? I can not find a reference to this symbol

SchrodingersGat: Ah, yes forgot to push changes to settings.py - updated now

matmair (approved):

PR #12217, opened 2026-06-21
[UI] Tree select in form fields

### LINKED ISSUE #12120: [FR] improve tree releated model fields

### Problem statement

When selecting tree models (categories or locations) the context (parent locations) is very crucial. Currently this is very unvisible with large trees.

 

### Suggested solution

Mantine has added a [TreeSelect](https://mantine.dev/core/tree-select/#scrollable-dropdown) in v9 which looks really interesting for this usecase.

looking for another library

### Examples of other systems

_No response_

- [ ] I want to develop this.

SchrodingersGat:
- Implements a "tree select" for stock locations and part categories in UI forms:
- Closes https://github.com/inventree/InvenTree/issues/12120

image


SchrodingersGat:
@wolflu05 please review this carefully, it's quite a complex query / requery structure so I hope I've caught all the edge cases

src/frontend/src/components/forms/fields/ApiFormField.tsx
```diff
@@ -120,15 +124,70 @@ export function ApiFormField({
...
+          <TreeField
             controller={controller}
+            definition={fieldDefinition}
             fieldName={fieldName}
+            endpoint={ApiEndpoints.category_tree}
+            childIdentifier='subcategories'
+            model={ModelType.partcategory}
             navigate={navigate}
           />
         );
```
wolflu05: where are these field_types defined on the backend? I would prefere the option used below checking on the API url all within the related field.

src/frontend/lib/types/Forms.tsx
```diff
@@ -100,7 +100,9 @@ export type ApiFormFieldType = {
     | 'nested object'
     | 'dependent field'
     | 'table'
-    | 'tags';
+    | 'tags'
+    | 'location'
+    | 'category';
```
wolflu05: same as below

wolflu05 (commented):
This is looking really good. Thank you very much for implementing it. There are just a few small things I found:

1. when having a location selected and opening the tree picker, the selected value clears visually, but when closing it appears again. I think that should just show always
2. the full path does not show anymore; before this PR it was possible to show the full path of a location. After only the leaf location name is shown.
   Before:  Image
  After: Image
3. not sure how we can fix this better, but when opening from an already selected item, it only shows the locations in that location path. When I want to select a differet, e.g. a neighbor of the parent, I need to collapse and uncollapse its parent first: 

https://github.com/user-attachments/assets/90119e5b-e24c-4b6f-8cf3-77016584085d

src/frontend/src/components/forms/fields/ApiFormField.tsx
SchrodingersGat: These field types are a frontend-only "hack" to redirect to the special field type

SchrodingersGat:
@wolflu05 fix for 2) is in now - use pathstring for the display label

> not sure how we can fix this better, but when opening from an already selected item, it only shows the locations in that location path. When I want to select a differet, e.g. a neighbor of the parent, I need to collapse and uncollapse its parent first:

Correct, not sure how to address this better. It requires specific DB filtering on the backend to make it in any way efficient

> when having a location selected and opening the tree picker, the selected value clears visually, but when closing it appears again. I think that should just show always

Can you show a screenshare? Not sure exactly what you are saying here

src/frontend/src/components/forms/fields/ApiFormField.tsx
wolflu05: Where do they get produced? I did not find them.

wolflu05:
> Correct, not sure how to address this better. It requires specific DB filtering on the backend to make it in any way efficient

Yeah thats right. I think its fine to keep it like that.

> Can you show a screenshare? Not sure exactly what you are saying here

https://github.com/user-attachments/assets/f8020548-3c2a-418f-a7ef-567401a6fe01

You can see at the very beginning something is selected, then I opened the picker and it immediatly hides and just the "Select..." placeholder is shown. When closing again, it shows again.

You can actually also see a second issue in this recording. When opening the first time after the selection, it does not scroll the selected item into view

wolflu05:
One improvement would also be to vissually make "structural" locations show in a different way and make them non selectable.

SchrodingersGat:
> One improvement would also be to vissually make "structural" locations show in a different way and make them non selectable.

I like this idea - I have marked structural locations as italic / dimmed in the dropdowns

src/frontend/src/components/forms/fields/ApiFormField.tsx
SchrodingersGat: image

Is this what you are looking for?

wolflu05: This is just the typedef. I mean where an APIField with location as field_type is created. If I'm not mistaken, this must be on the backend, but I cannot find it in this PR.

SchrodingersGat: >  If I'm not mistaken, this must be on the backend, but I cannot find it in this PR.

It's entirely frontend driven

wolflu05: Ok, but I do not understand where this value is set on the frontend then. I must be overlooking something obvious

SchrodingersGat: I have just simplified the code so it should be a bit more obvious. There was a code path that got refactored and was not being used, so I have removed it now

wolflu05: Thanks, that makes sense now, that was what I was thinking and why I wrote the comments, but did not fully understand.

PR #12218, opened 2026-06-21
devops: upload Flakiness.io reports from PRs-from-forks

aslushnikov:
This is based upon https://docs.flakiness.io/ci/github-actions-open-source/


matmair:
Why would we need this info for forks?

aslushnikov:
@matmair this will upload test results for pull-requests-from-forks to the flakiness.io. 

This will give you the following:
- each PR will have a nice dashboard with testing results, like you have for dependabot PRs: https://flakiness.io/InvenTree/InvenTree/pull/12214
- all your trace artifacts will be easily accessible & viewable in case of certain test failures
- if there's a regression, we'll add a failing check on the PR

matmair:
Thank you for the PR. I do not think we are at a stage where we want to see this at the PR stage, master is fine for now.

Issue #11171, opened 2026-01-20
Doesn't remember darkmode setting


merspieler:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

When I log back into inventree, it always forgets that I have set dark mode.
I would expect the UI to remember this setting.

### Steps to Reproduce

* Set the UI to dark mode to dark mode
* clear browser data or log in from a different browser or device
* Observe the color move is again bright

### Expected behaviour

For the UI to remember my setting of dark mode

### Deployment Method

Docker

### Version Information

InvenTree-Version: 1.1.9
Django Version: 4.2.26
Commit Hash: 73c4f88
Commit Date: 2026-01-20
Commit Branch: null
Database: django.db.backends.postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.12.60-x86_64-with-glibc2.41
Installer: DOC
Active plugins: false

### Try to reproduce on the demo site

I tried to reproduce

### Is the bug reproducible on the demo site?

Reproducible

### Relevant log output

```shell

```

matmair:
The color schema is set per device and kept in the browsers local storage; if you clear browser data you loose your preference.

matmair:
Having the color mode follow you is very cumbersome when you switch a lot between office and workshop floor.

merspieler:
That's a problem with an ephemeral system which on purpose doesn't keep state.

matmair:
You call it a problem, I call it a ⭐️ design decision ⭐️ 

The color mode change button is on the login page on the lower left below the login box. You can select the desired one before logging in on each device and the selection will stay till you clear the browser. At that point you will have to log in again and the button will right there

matmair:
@SchrodingersGat what is your POV here?

SchrodingersGat:
If we tracked it within the user's profile, I'm sure that there would be others requesting that we revert to the previous behavior.

@merspieler would you be willing to submit a patch to implement this? Otherwise, I do not think that this will gather much attention

merspieler:
How much time do you think that needs? In theory is should be straight forward, but I've seen horrible code at work where such a small change meant touching almost 10 files and even more functions/classes, chasing references for a day.

matmair:
It would take me probably 30mins and then maybe another 5 if CI fails. No idea about your typescript and python/Django proficiency as your profile is private.

merspieler:
never touched typescript or django...
[...]

PR #12227, opened 2026-06-23
color scheme: Use the users prefered mode by default

reid-p:
It's likely if people want dark mode, they set it on their system / browser as well.
Use that hint to select the mode, if its not set in the store.
Probably Address #11171 in most cases.


SchrodingersGat (approved):

matmair (approved):

SchrodingersGat:
@reid-p thanks for the contribution!

PR #12233, opened 2026-06-23
extend barcode scans API

matmair:
This adds permission checks for the to be rendered/found model


matmair:

PR #12274, opened 2026-06-29
[UI] Table field refactor

### LINKED ISSUE #11425: Deleting rows to fast causes hiccup in the Allocate Stock for Build Order incomplete outputs form

## This is not ground breaking, but a really small edge case, so not urgent

I was allocating parts in bulk using the Allocate Stock function for incomplete outputs for a Build Order, however when I went to remove the two at the top rows, and accidently deleted the third as well, it refreshed and then took a second for all of the stock items to load again,

then when they did load it was off by one part because the top part reappeared

<kbd>
Image
</kbd>

### Steps to Reproduce

1. Create build order with incomplete ouptuts
2. Go to bulk allocate parts of that incomplete output, but do not allocate the first two but allocate the rest (preferably with 4+)
3. Delete the first three (its really quirky so if I do it slow it doesn't cause the issue)
4. watch as it tries to recreate the 3rd one deleted, and then move all the rest down by one

### Expected behaviour

delete the first to 3 items and then continue on as normal

### Deployment Method

Docker

### Version Information

InvenTree Version: 1.2.3
API Version: 453
Python Version: 3.11.14
Django Version: 5.2.11
Commit Hash: 12da3b0
Commit Date: 2026-02-22

I did not try to reproduce

Not reproducible

### Relevant log output

```shell

```

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/11425

### Breaking Change

Any custom plugins which use "table" field types in UI forms will need to adjust to match the new configuration. A unique ID value is required for each row.

### Description

This PR addresses a major design inefficiency in the way that "table fields" are rendered in the user interface. It fixes a number of structural problems:

- Removing any row causes all rows to be re-created
- Updating any row causes all rows to be re-created
- Each render takes multiple paint cycles

These issues were largely caused by:

- Using the array index to identify each row (replaced with unique ID value)
- Passing through too many props (now pass by ref)
- Not memoizing props properly (now added a better memo step)

### Before Refactor

https://github.com/user-attachments/assets/5890c135-5b67-4767-a997-42cfe47668d4

### After Refactor

https://github.com/user-attachments/assets/7572762d-5c3f-4918-b8c5-758480a699c1

### DIFF
--- a/src/frontend/src/components/forms/fields/TableField.tsx
+++ b/src/frontend/src/components/forms/fields/TableField.tsx
@@ -363,9 +363,9 @@ export const TableField = memo(
     return (
       previousProps.definition === nextProps.definition &&
       previousProps.fieldName === nextProps.fieldName &&
-      previousProps.value === nextProps.value &&
-      previousProps.onChange === nextProps.onChange &&
-      previousProps.error === nextProps.error
+      areShallowEqual(previousProps.value, nextProps.value) &&
+      areShallowEqual(previousProps.error, nextProps.error) &&
+      previousProps.onChange === nextProps.onChange
     );
   }
 );
--- a/src/frontend/src/components/forms/fields/TableField.tsx
+++ b/src/frontend/src/components/forms/fields/TableField.tsx
@@ -2,8 +2,14 @@ import { t } from '@lingui/core/macro';
 import { Trans } from '@lingui/react/macro';
 import { Alert, Container, Group, Stack, Table, Text } from '@mantine/core';
 import { IconExclamationCircle } from '@tabler/icons-react';
-import { type ReactNode, useCallback, useEffect, useMemo, useRef } from 'react';
-import type { FieldValues, UseControllerReturn } from 'react-hook-form';
+import {
+  type ReactNode,
+  memo,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef
+} from 'react';
 
 import { AddItemButton } from '@lib/components/AddItemButton';
 import { identifierString } from '@lib/functions/Conversion';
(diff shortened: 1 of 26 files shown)


src/frontend/src/components/forms/fields/TableField.tsx
```diff
@@ -125,15 +272,15 @@ export function TableField({
...
+      aria-label={`table-field-${fieldName}`}
       style={{ width: '100%' }}
     >
       <Table.Thead>
         <Table.Tr>
           {definition.headers?.map((header, index) => {
             return (
               <Table.Th
-                key={`table-header-${identifierString(header.title)}-${index}`}
+                key={`table-header-${identifierString(header.title)}`}
```
matmair: is this guaranteed to be unique?

src/frontend/src/forms/BuildForms.tsx
```diff
@@ -636,14 +646,31 @@ function BuildAllocateLineRow({
...
+            } else if (value.trim() !== '') {
+              const parsed = Number.parseFloat(value);
+              nextValue = Number.isFinite(parsed) ? parsed : '';
+            }
+
+            setQuantity(nextValue === '' ? 0 : nextValue);
+            props.changeFn(props.rowId, 'quantity', nextValue);
+          }}
           error={props.rowErrors?.quantity?.message}
         />
```
matmair: This might be a good candidate for a reusable component

src/frontend/src/functions/debug.tsx
```diff
@@ -0,0 +1,36 @@
+/** Various debugging helper functions for development */
+
+import { useEffect, useRef } from 'react';
+
+/**
+ * A custom hook that logs the previous and current props of a component whenever it updates.
+ */
+export function useWhyDidYouUpdate(name: string, props: any) {
+  const previousProps = useRef({});
```
matmair, line 9: Maybe this should be mentioned in the docs?

src/frontend/src/tables/sales/SalesOrderAllocationTable.tsx
```diff
@@ -285,7 +285,6 @@ export default function SalesOrderAllocationTable({
 
     return {
       items: stockItems,
-      model: ModelType.stockitem,
```
matmair, line 288: imo this could be helpful info in the future for plugins / customisation and should not cause re-renders

matmair (approved):
LGTM; this will require change on any plugins that use tables; right?

SchrodingersGat:
> LGTM; this will require change on any plugins that use tables; right?

Yes, marked as breaking now. And I'll update the CHANGELOG

src/frontend/src/components/forms/fields/TableField.tsx
SchrodingersGat: I have reverted this to ensure unique column IDs

src/frontend/src/forms/BuildForms.tsx
SchrodingersGat: Done

PR #12287, opened 2026-06-30
fix pin syntax

matmair:
corrects pinning syntax and a few instances of pins


SchrodingersGat:
@matmair why is github actions complaining about the new format?

matmair:
@SchrodingersGat the recently bumped version of zizmor intruders a new check for pins which helps finds wrong pin references

matmair:

SchrodingersGat (approved):

PR #12294, opened 2026-07-02
[API] Refactor API for duplicating objects

SchrodingersGat:
- Refactor duplicated code into a single "DuplicateOptionsSerializer"
- Consolidate functionality
- Add duplicate options for multiple new models
- Adding framework to allow copying of notes - https://github.com/inventree/InvenTree/pull/11971


matmair (commented):
I like the idea and this implementation.

Taking the idea a but further: would it make sense/be possible to just have a single endpoint for duplicating an object or are the endpoints to specialized for it to make sense?

SchrodingersGat:
I think that it's too specific to make it worth implementing a "generic" endpoint. At least this reduces duplication of code and consolidates the approach across all models

PR #12072, opened 2026-06-02
Feature: Backend scaffolding for Repair Orders (#12064)

adityakrmishra:
Addresses the backend requirements for #12064.

This PR introduces the initial Django architecture for the new Repair Order feature, integrating it into the `order` app as discussed.

### Additions:
* **Status Codes:** Added `RepairOrderStatus` enumerations.
* **Models:** Added `RepairOrder`, `RepairOrderLineItem`, and `RepairOrderAllocation` to `order/models.py`.
* **API Routing:** Implemented DRF serializers and added CRUD ViewSet endpoints to `order/api.py`.
* **Linting:** Validated via `ruff`.

*(Note: Migration files have not been generated in this commit to avoid environment conflicts. Please run `manage.py makemigrations order` when reviewing.)*

Ready for review before we move on to the Mantine UI scaffolding!

closes https://github.com/inventree/InvenTree/issues/12064

Issue #12064, opened 2026-06-01
[Feature] Repair orders


SchrodingersGat:
### Discussed in https://github.com/inventree/InvenTree/discussions/7694

> I run a repair service. From what I can see, Inventree is not able to create a repair order, and to track parts and a description of repair work done to a customer's unit. Is this a feature that would be likely implemented?

### Concept Requirements

- Determine scope of feature
- Required DB models
- Required UI elements
- Require API flows

adityakrmishra:
@SchrodingersGat @matmair I can pick this up. Since this requires core schema additions, I want to sync on the architecture before cutting a PR.

TL;DR: I'm visualizing this as a hybrid between a SalesOrder (customer-facing) and a BuildOrder (consuming internal stock).
Proposed DB Models (Django)
Integrated into the order app (or a new repair app):
RepairOrder: reference (auto-seq), customer (FK to Company), customer_reference, item_description, symptoms, status (integer choices: Pending, In Progress, Completed, etc.). Inherits from standard Attachment/Notes mixins.
RepairOrderLineItem & RepairOrderAllocation: To map and consume StockItems used during the repair.

API Routing
Standard DRF ModelViewSet CRUD for the models (e.g., /api/order/repair/).
Custom actions like /api/order/repair/<id>/issue/ to execute the permanent stock deduction once parts are installed.

UI (Mantine)
New RepairOrderTable index.
Detail view with panels for: Customer Details/Symptoms, Parts Allocation (similar to BO UI), and Markdown Notes for the tech's repair log.

Let me know if this scope aligns with what you had in mind. If this architecture LGTM, I'll start scaffolding the backend models and API as a v1

SchrodingersGat:
@adityakrmishra that sounds pretty thorough, if you would like to take this on that's great!

### Starting Points

- https://docs.inventree.org/en/stable/develop/
- https://docs.inventree.org/en/stable/develop/architecture/
- https://docs.inventree.org/en/stable/develop/devcontainer/
- https://docs.inventree.org/en/stable/develop/react-frontend/

And feel free to reach out with any questions!

Best thing to do would be to start a new PR so that we can provide any direct assistance there  👍

matmair:
there are a few important things missing here: business context on the function, what kind of actions might be required and the lifecycle of orders in this new concept.
The tech is the easy part that any llm can do

matmair:
[...]

PR #12303, opened 2026-07-05
feat(order): add RepairOrder order type (backend)

ameyypawar:
## Repair Orders — backend foundation (draft)

Opening as a **draft** to put a concrete backend proposal on the table for #12064 and get feedback on the data-model direction before building further. I'm aware of the discussion in #7694 and the in-progress work in #12072 — happy to consolidate; this is meant to move the conversation forward, not around anyone.

### Approach
A standalone `RepairOrder` modelled as a **customer-facing order**, closely mirroring `ReturnOrder` and wired end-to-end like the recent `TransferOrder` — reusing the order infrastructure (reference patterns, status state machine, permissions, reporting, barcodes) rather than reinventing it.

### Included
- `RepairOrder` + `RepairOrderLineItem` + `RepairOrderExtraLine` models; `RepairOrderStatus` codes
- Serializers + DRF API (`/api/order/repair-order/…`) incl. issue/hold/complete/cancel transitions
- Permissions (`repair_order` ruleset + OAuth2 scope), system settings, admin, API tests, fixture
- `api_version` bump + changelog entry

### Deliberately out of scope (for discussion)
- **Internal stock allocation/consumption** — how repair parts deduct from stock is the main open design question, so I left it out of this first pass rather than guess.
- Frontend (PUI) — backend only.

### Status
Passes locally on **both SQLite and PostgreSQL**: migrations apply cleanly, the API schema builds with only additive changes, and the order test suite (incl. new `RepairOrder` API tests) is green.

Feedback welcome on the model shape — especially stock-consumption, and standalone order type vs. extending `ReturnOrder`.

Refs #12064


SchrodingersGat:
This feature is already being implemented in https://github.com/inventree/InvenTree/issues/12064

Issue #12311, opened 2026-07-06
Conflicting migrations detected when upgrading 1.3.3 -> 1.4.1


JustusRijke:
### Deployment Method

- [x] Installer
- [ ] Docker Development
- [ ] Docker Production
- [ ] Bare metal Development
- [ ] Bare metal Production
- [ ] Digital Ocean image
- [ ] Other (please provide a link `Steps to Reproduce`

### Describe the problem*

Upgrade fails, see log below.

As suggested in the log output I executed `python manage.py makemigrations --merge`:

```
foobar@InvenTree-PT: /opt/inventree/src/backend/InvenTree$ python3 manage.py makemigrations --merge
INVE-W3: Could not detect git information.
INVE-W10 - Config for 'INVENTREE_CONFIG_FILE' not in recommended directory '/etc/inventree'.
2026-07-06 09:54:33,477 WARNING INVE-W10 - Config for 'INVENTREE_SECRET_KEY_FILE' not in recommended directory '/etc/inventree'.
Traceback (most recent call last):
  File "/opt/inventree/src/backend/InvenTree/manage.py", line 26, in <module>
    main()
  File "/opt/inventree/src/backend/InvenTree/manage.py", line 21, in main
    execute_from_command_line(sys.argv)
  File "/opt/inventree/env/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/opt/inventree/env/lib/python3.11/site-packages/django/core/management/__init__.py", line 382, in execute
    settings.INSTALLED_APPS
  File "/opt/inventree/env/lib/python3.11/site-packages/django/conf/__init__.py", line 81, in __getattr__
    self._setup(name)
  File "/opt/inventree/env/lib/python3.11/site-packages/django/conf/__init__.py", line 68, in _setup
    self._wrapped = Settings(settings_module)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/inventree/env/lib/python3.11/site-packages/django/conf/__init__.py", line 166, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #12320, opened 2026-07-07
Adjust default behavior of "migrate" command

SchrodingersGat:
- Does not create new migration files automatically
- Ref: https://github.com/inventree/InvenTree/issues/12311

The existing default behavior on "invoke update" is to create new migration files if the database schema has changed. This *may* lead to the situation outlined in #12311 where we end up with out-of-line migrations.

### Tasks

- [x] Add breaking changelog entry


SchrodingersGat:
@matmair LMK what you think of this - it changes the default behavior but in general, users should *NOT* be in a situation where they are generating new migration files

PR #12322, opened 2026-07-07
[refactor] File uploads

SchrodingersGat:
This PR refactors how file uploads are handled:

- Single file upload path renamer
- Keep files associated with their model type
- Optionally associate files with their underlying model ID

In addition to reducing duplicated code (and potential bugs) - this PR can be used in the future to implement an auth check of uploaded files - and reject if the user does not have the required permissions to access those files. We can use the file path to determine the model type (and potentially ID) and then check if the user has the correct roles/permissions.

### Tasks

- [x] Refactor existing `upload_to` endpoints
- [ ] Fix bugs
- [ ] Add unit testing for all potential upload points
- [ ] Data migrations for moving existing uploaded files

### DIFF
--- a/src/backend/InvenTree/common/media.py
+++ b/src/backend/InvenTree/common/media.py
@@ -6,7 +6,7 @@ from typing import Optional
 from django.core.exceptions import ValidationError
 
 
-def common_file_upload_handler(
+def rename_uploaded_file(
     filename: str,
     file_type: str,
     model_type: Optional[str] = None,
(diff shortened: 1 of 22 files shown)


src/backend/InvenTree/common/media.py
```diff
@@ -0,0 +1,68 @@
...
+
+    if not file_type:
+        raise ValidationError('File type must be specified.')
+
+    # First, remove any illegal characters from the filename.
+    # Keep '.' so valid file extensions are preserved.
+    illegal_chars = '\'"\\\\/`~#|!@#$%^&*()[]{}<>?;:+=,'
+
+    for c in illegal_chars:
+        filename = filename.replace(c, '')
```
matmair: Imo we should just throw an exception here

SchrodingersGat: This has been refactored to a more standard check now

PR #12332, opened 2026-07-08
Fix notes being dropped on part creation

ledvinap:
## Summary
 
  Fix `POST /api/part/` so the `notes` field is preserved during part creation. Previously, `NotesFieldMixin` stripped `notes` for any serializer used by a list-capable view, which also affected `ListCreateAPI` write  requests.
 
  ## Changes
 
  - Only hide `notes` for safe/read list requests
  - Keep `notes` available for create/update validation and persistence
  - Add regression coverage for creating a part with notes
  - Preserve existing behavior where part list responses omit `notes`
 
  ## Testing
 
  - Verified live API behavior on local dev server:
    - `POST /api/part/` with `notes`
    - `GET /api/part/{pk}/`
    - `GET /api/part/?limit=1`
  - Ran `python3 -m py_compile` for edited backend files

### DIFF
--- a/src/backend/InvenTree/InvenTree/serializers.py
+++ b/src/backend/InvenTree/InvenTree/serializers.py
@@ -822,9 +822,13 @@ class NotesFieldMixin:
         super().__init__(*args, **kwargs)
 
         if hasattr(self, 'context'):
+            request = self.context.get('request', None)
+            method = getattr(request, 'method', None)
+
             if view := self.context.get('view', None):
                 if (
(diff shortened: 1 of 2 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@ledvinap thanks for the fix!

This will be re-addressed again once we get https://github.com/inventree/InvenTree/pull/11971 merged, but for now this bug fix is a great catch

matmair (approved):

PR #12334, opened 2026-07-08
[UI] Page Load Improvements

SchrodingersGat:
This PR provides a significant speed up to initial page loading:

- Reduce duplicate API calls
- Defer loading of expensive components
- Improvements to suspended component delays

### Comparisons

| Test | Old | New |
| --- | --- | --- |
| page load time | ~2.5s | ~1.2s |
| screencap | https://github.com/user-attachments/assets/f4cecf40-14ae-43b9-8d9d-2b66ebb6f301 | https://github.com/user-attachments/assets/2014ac9a-12e6-410b-8205-dbed28647358 |

*Note: Most of the 1200ms can be nominally improved even further with caching - the python dev server does not perform any caching of static files*


src/frontend/src/pages/Auth/LoggedIn.tsx
```diff
@@ -9,10 +8,9 @@ import { Wrapper } from './Layout';
 export default function Logged_In() {
   const navigate = useNavigate();
   const location = useLocation();
-  const checkLoginStateDebounced = useDebouncedCallback(checkLoginState, 300);
 
   useEffect(() => {
-    checkLoginStateDebounced(navigate, location?.state);
+    checkLoginState(navigate, location?.state);
```
matmair, line 13: Have you validated that this does not increase api calls?

src/frontend/src/states/states.tsx
```diff
@@ -41,25 +41,62 @@ export interface ServerAPIProps {
...
+    setApiDefaults();
+    const traceId = setTraceId();
+    await Promise.all([
+      useServerApiState.getState().fetchServerApiState(),
+      useUserSettingsState.getState().fetchSettings(),
+      useGlobalSettingsState.getState().fetchSettings(),
+      useGlobalStatusState.getState().fetchStatus(),
+      useIconState.getState().fetchIcons()
+    ]);
+    removeTraceId(traceId);
```
matmair, line 84: Maybe I am blind but where are you starting the tracing session? It seems like it is not started with the actual fetch

matmair (approved):
Interesting improvements. I am always a bit sceptical of changes in the global state and locale mgmt, we have destroyed stuff inadvertently there before. But looks like it works.

2 general thoughts:
-  The whole login / state management mechanism gets more and more complex. I wonder if this could be handled by well tested third party components if we fully switch to OAuth2
- would it make sense to add some performance/size focused playwright tests?

src/frontend/src/states/states.tsx
matmair: Nevermind, I am blind

PR #12354, opened 2026-07-10
[UI] Show linked barcode string on stock detail page (#11745)

### LINKED ISSUE #11745: [FR] Show linked barcode string in user interface

### Problem statement

We recently updated our inventree instance from 0.16.1 to 1.3.0. In the earlier version we could easily see linked barcode string in the user interface on the details tab of a stock item. Since the update, we cannot find this string anywhere anymore. We used this for manual verification in our processes. 

### Suggested solution

Bring back the linked barcode string on stock item details.

fetch it from the database with a custom made app or service, a lot of overhead for the use case

### Examples of other systems

_No response_

- [ ] I want to develop this.

mahekp05:
## Description
Restores the linked (third-party) barcode string on the stock item detail page, which was visible pre-1.0 but lost after the UI rewrite.

Fixes #11745
## Changes
- Expose `barcode_data` as a read-only field on `StockItemSerializer`
- Display a "Linked Barcode" field (copyable) on the stock detail page, shown only when a barcode is linked

## Tests
- Backend API test asserting `barcode_data` is exposed and read-only
- Playwright test covering the link → display → unlink flow


SchrodingersGat:
Hi @mahekp05 thanks for the PR!

If we are going to fix this we should do it in the general case - not *just* for the StockItem detail.

There are many other pages which support custom barcode data, so these changes should be applied to those too, and also in a generic way

mahekp05:
> Hi @mahekp05 thanks for the PR!
> 
> If we are going to fix this we should do it in the general case - not _just_ for the StockItem detail.
> 
> There are many other pages which support custom barcode data, so these changes should be applied to those too, and also in a generic way

Thanks for the feedback! I've reworked this to apply generically across **all** models that support custom barcodes (`InvenTreeBarcodeMixin`), rather than special-casing StockItem.

## Approach

**Backend — one shared serializer mixin**
Added `BarcodeSerializerMixin` in `InvenTree/serializers.py`, which declares the read-only `barcode_data` field once. Each barcode-supporting serializer mixes it in and adds `'barcode_data'` to its `Meta.fields`. For the order types this is done once on the shared `AbstractOrderSerializer` / `order_fields()`, so PurchaseOrder, SalesOrder, ReturnOrder and
TransferOrder are all covered without duplication.

**Frontend — one shared helper**
Added `barcodeDataField(instance)` in `components/details/Details.tsx`, which returns the copyable "Linked Barcode" detail field (only shown when a barcode is linked). Each detail page spreads it into its details grid (consistently in the top-left panel). The original inline StockItem block was refactored to use this helper too, so nothing is special-cased.

## Coverage
The linked barcode string now displays on: **Part, StockItem, StockLocation, Build, SupplierPart, ManufacturerPart, PurchaseOrder, SalesOrder, ReturnOrder, TransferOrder, SalesOrderShipment**.

## Files changed

**Backend**
- `InvenTree/serializers.py` — new `BarcodeSerializerMixin`
- `stock/serializers.py`, `part/serializers.py`, `build/serializers.py`,
  `company/serializers.py`, `order/serializers.py` — mixin wired in + `barcode_data` field
- `InvenTree/api_version.py` — API v520 changelog
- `InvenTree/test_serializers.py` — new test asserting `barcode_data` is exposed & read-only
  across all barcode serializers

**Frontend**
- `components/details/Details.tsx` — new `barcodeDataField` helper
- 10 detail pages wired to use it (Stock, Part, Location, Build, Manufacturer/Supplier part,
  Purchase/Sales/Return/Transfer order)

## Tests
- Generic backend test (`BarcodeSerializerMixinTest`) covering all 11 serializers
- Existing StockItem end-to-end API test retained
- Playwright test covering the link → display → unlink flow

Open to feedback — happy to adjust the placement, naming, or approach if you'd prefer it done differently, or if any other changes are needed.

src/backend/InvenTree/InvenTree/serializers.py
```diff
@@ -834,6 +834,22 @@ def __init__(self, *args, **kwargs):
...
 
+class BarcodeSerializerMixin(serializers.Serializer):
+    """Serializer mixin for models which support custom (third-party) barcode assignment.
+
+    Exposes the read-only 'barcode_data' field (the linked barcode string) so that
+    it can be displayed in the user interface. Applies to any model inheriting from
+    InvenTree.models.InvenTreeBarcodeMixin (e.g. Part, StockItem, StockLocation).
+
+    Note: The serializer's Meta.fields must include 'barcode_data' for it to appear.
+    The field is read-only here; barcodes are assigned via the barcode link endpoints.
```
matmair, line 1028: can we add a check if the `Meta` is set correctly? Either in init or as a django system check?

src/frontend/src/components/details/Details.tsx
```diff
@@ -54,6 +54,23 @@ export type DetailsField = {
 type BadgeType = 'owner' | 'user' | 'group';
 type ValueFormatterReturn = string | number | null | React.ReactNode;
 
+/**
+ * Construct a reusable "Linked Barcode" detail field for any model which
+ * supports custom (third-party) barcode assignment (i.e. exposes 'barcode_data').
+ *
+ * The field is copyable and only rendered when a barcode is linked to the instance.
+ */
+export function barcodeDataField(instance: any): DetailsField {
```
matmair, line 63: @SchrodingersGat do we need to do a memo trick here?

matmair (approved):
LGTM in general

src/backend/InvenTree/InvenTree/serializers.py
mahekp05: Yes, I just added a check in BarcodeSerializerMixin.__init__ to make sure the serializer's model supports barcodes and that barcode_data is included in Meta.fields. If either is configured incorrectly, it raises ImproperlyConfigured. I also added tests for both cases.

I went with the init-time check so the configuration is validated whenever the serializer is instantiated

matmair:
@SchrodingersGat imo the approach looks good; I think this would be - in general - be ready for a merge

@mahekp05 a few merge conflicts have creeped in; please adress these

mahekp05:
@matmair @SchrodingersGat I just updated the branch with the latest `master` and resolved all conflicts.

`master` had refactored the detail pages into `DetailsPanel` components, so I moved `barcodeDataField(instance)` into each new panel and kept the shared helper + backend `BarcodeSerializerMixin`. Bumped API to v531.

Branch is now conflict-free.

Issue #11385, opened 2026-02-19
add part button disappear if grid style is changed to parametric view


saadyacu:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

In the parts windows, logged in as an admin, the add button disapprears and user will not be able to add parts if the view is switched from "Table View" to "Parametric View"

"Table View"
<img width="1685" height="211" alt="Image" src="https://github.com/user-attachments/assets/22954ccc-ae65-4339-9cf9-c6b17518a78a" />

"Parametric View"
<img width="1694" height="189" alt="Image" src="https://github.com/user-attachments/assets/c1afb4e0-552e-445c-b9eb-df2e46a9394a" />

### Steps to Reproduce

1. go to the parts windows
2. switch views 

### Expected behaviour

button "Add" should stay on based on user permission 

### Deployment Method

None

### Version Information

1.22

### Try to reproduce on the demo site

I did not try to reproduce

### Is the bug reproducible on the demo site?

Not reproducible

### Relevant log output

```shell

```

matmair:
There is not 1.22, provide the full version information as it is available for easy copy in the About InvenTree dialog via the big orange copy button.

SchrodingersGat:
@saadyacu yes the "add part" button is not available from that table. If you would like to see it there, I would happily review a PR.

hanielee:
Hi, I'd like to pick this up! Any guidance on where to look first would be appreciated.

SchrodingersGat:
### Dev Resources

- https://docs.inventree.org/en/stable/develop/
- https://docs.inventree.org/en/stable/develop/devcontainer/
- https://docs.inventree.org/en/stable/develop/react-frontend/

### Critical Code Points

- https://github.com/inventree/InvenTree/blob/master/src/frontend/src/tables/part/PartTable.tsx
- https://github.com/inventree/InvenTree/blob/master/src/frontend/src/tables/part/ParametricPartTable.tsx

PR #12392, opened 2026-07-14
[UI] Keep Add Parts button visible in Parametric View

hanielee:
## Problem
The "Add Parts" button disappears when switching to Parametric View on the Parts page, even for admin users. This isn't a permissions issue: `PartListTable` (Table View) and `ParametricPartTable` (Parametric View) are separate components, and only `PartListTable` builds a toolbar with an "Add Parts" action. The generic `ParametricDataTable` that Parametric View wraps only supports per-row actions (like "Add Parameter"), so the button was never in that view in the first place.

## Solution
Brought the "Add Parts" button from Table View into Parametric View so both views behave consistently. Added an optional `customActions` prop to `ParametricDataTable`, following the same pattern as the existing `customColumns` and `customFilters` props, and wired the same "Add Parts" dropdown (Create Part, Import from File, Import from Supplier) into Parametric View through that prop, gated by the same `hasAddRole(UserRoles.part)` permission check Table View already uses. However, since the dropdown was duplicated between the two tables, I extracted it into a shared `PartCreationMenu` component used by both.

Before:
issue-add-parts-button

After:
fix-issue-add-parts-button
- Closes `#11385`: #11385 

## Testing
Added Playwright tests confirming the button is visible for admins, hidden for read-only users, and that creating a part through the Parametric View works end to end.


SchrodingersGat (approved):

hanielee:
@SchrodingersGat Thank you for the approval on the draft, I've marked it as ready for review!

SchrodingersGat:
The combination of "create new part" / "import parts" is a repeated pattern here - I think it would be worth offloading this to a common component e.g. `PartCreationMenu` which can be used in both locations

hanielee:
@SchrodingersGat I've implemented the `PartCreationMenu` component that both the Table View and Parametric View now use. Thanks for mentioning this, I noticed the SonarCloud code duplication check flagging it while I was working on the PR and I wasn't sure if I should've created separate component, but it makes sense since it's cleaner this way.

SchrodingersGat (approved):

SchrodingersGat:
@hanielee please address merge conflicts here, sorry

hanielee:
@SchrodingersGat I noticed some of the checks are failing, but some seem unrelated to this PR (aside from the prek one which I'll look into). I'm wondering if there are other remaining blockers or anything else needing to be addressed? Thanks!

SchrodingersGat:
The failing frontend tests seem to be flaky tests elsewhere - I'll look into that. But please address the prek tests

hanielee:
> The failing frontend tests seem to be flaky tests elsewhere - I'll look into that. But please address the prek tests

@SchrodingersGat Thanks for confirming! I’ve addressed the prek test failure and pushed the changes.

SchrodingersGat:
@hanielee thanks for the contribution!

PR #12413, opened 2026-07-17
feat(ui): show variant stock indicator in parts table

### LINKED ISSUE #5959: [FR] highlight variant stock in parts views

### Problem statement

hi,

including the variant stock in the stock of the template part is in general a good idea but it causes a bit of confusion for some users on our side. If the part view is used, it is not clear if the displayed stock includes variants until the part is selected and the stock view is checked.

### Suggested solution

Adding an info symbol next to the stock like it is already done in the stock allocation view for build orders probably would clear things up already. It would be premium, if the actual variant stock would be displayed in the tooltip.

Obviously this is not really necessary as stock information could be directly viewed by going via the stock views and not the part views. However users will always have their weird ways of accessing information and one probably has to tailor into this...

### Examples of other systems

_No response_

- [ ] I want to develop this.

br413:
﻿## Summary
- Add variant and direct stock breakdown to the parts table hover card when `variant_stock > 0`
- Extract shared `renderPartStockCell` renderer and reuse it in Parametric View for consistency
- Matches the existing build allocation / BOM pattern using `TableHoverCard`

Fixes #5959

## Test plan
- [x] Create a template part with variants that have stock
- [x] Confirm parts table shows info icon next to stock count
- [x] Hover shows "Includes variant stock" and "Direct stock" breakdown
- [x] Confirm Parametric View shows the same hover behavior
- [x] Confirm parts without variant stock are unchanged

### DIFF
--- a/src/frontend/src/components/tables/PartStockCell.tsx
+++ b/src/frontend/src/components/tables/PartStockCell.tsx
@@ -1,0 +1,132 @@
+import { t } from '@lingui/core/macro';
+import { Group, Text } from '@mantine/core';
+import type { ReactNode } from 'react';
+import { formatDecimal } from '../../defaults/formatters';
+import { TableHoverCard } from './TableHoverCard';
+
+export function renderPartStockCell(record: any): ReactNode {
+  if (record.virtual) {
+    return (
+      <Text size='sm' c='dimmed' fs='italic'>
+        {t`Virtual part`}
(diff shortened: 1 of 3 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@br413 thanks for the contribution!

PR #12415, opened 2026-07-17
fix(ui): show initial stock fields when PART_CREATE_INITIAL enabled

br413:
﻿## Summary
- Fix Add Part form not showing Initial Stock fields when `PART_CREATE_INITIAL` is enabled in admin settings
- Replace incorrect `virtual != false` gate with `globalSettings.isSet('PART_CREATE_INITIAL')` in `PartForms.tsx`
- Add Playwright regression tests for enabled/disabled setting states

Fixes #12266

## Test plan
- [x] Enable `PART_CREATE_INITIAL` in settings — Initial Stock section appears in Add Part form
- [x] Create part with initial quantity and location — stock is created
- [x] Disable setting — Initial Stock section is hidden
- [x] Existing Nested Object Field / supplier tests unaffected

PR #12418, opened 2026-07-18
Fix Redis write failures on Windows Docker bind mounts

br413:
## Summary
- Restores the Redis `/data` volume mount from #11100 while keeping disk persistence disabled
- Builds on the persistence fix already merged via #12415 (`--save "" --appendonly no`)
- Prevents login and other cache-dependent requests from returning HTTP 500 when Redis cannot persist to disk on Windows Docker hosts

## Context
On Windows Docker Desktop, the `inventree-data/redis` bind mount can prevent Redis from writing `dump.rdb`, triggering `MISCONF` errors and blocking all write operations (including django-allauth login rate limiting).

#12415 merged the core fix (disable RDB/AOF persistence) but removed the external volume mount. This PR restores the #11100 volume layout for consistency, as discussed with @SchrodingersGat — persistence remains off, so no `dump.rdb` writes occur and the Windows permission issue is still resolved.

InvenTree uses Redis purely as a Django cache backend (`django_redis`); cache data is ephemeral and does not require durable storage.

## Test plan
- [x] Verified Redis accepts writes after change (`redis-cli SET testkey ok`)
- [x] Verified login succeeds at `http://localhost/web/login` with demo credentials
- [x] Confirmed existing Linux Docker behavior is unaffected


br413:
Hi @matmair — this fixes Redis `MISCONF` / HTTP 500 login failures on Windows Docker Desktop when the redis bind mount cannot write `dump.rdb`.

The change makes the cache service ephemeral (`--save ""` `--appendonly no`) and removes the data volume, which matches how Redis is used here (cache only, not durable storage).

Happy to adjust if you would prefer a platform-specific compose override instead.

SchrodingersGat:
This would prevent any persistent storage of redis data, correct? And thus revert the changes made in https://github.com/inventree/InvenTree/commit/ff9f465b8b1eea0e812c81464c69ba68e11e31d4

br413:
@SchrodingersGat — good question. The initial version did revert the volume mount from #11100; I've pushed an update that keeps it.

The fix is disabling disk persistence (`--save "" --appendonly no`), not removing the mount. That stops Redis from writing `dump.rdb` to `/data`, which is what triggers `MISCONF` on Windows bind mounts when permissions block writes.

The volume from #11100 stays in place for layout consistency. With persistence off, Redis no longer snapshots to disk, so the original #10977 concern (anonymous volumes accumulating `dump.rdb`) is also addressed without needing RDB writes.

InvenTree uses Redis purely as a Django cache backend (`django_redis`), so cache loss on container restart is expected and acceptable — unlike Postgres, nothing durable lives in Redis here.

br413:
Migration Test failure is pre-existing on `master` (`TestRemoveMpttFieldsMigration` / duplicate `status_custom_key` ORM bug from #12360). Docker Build Test passed.

br413:
Rebased onto latest `master` and resolved the merge conflict with #12415.

#12415 already landed the persistence disable (`--save "" --appendonly no`). This PR now only restores the `${INVENTREE_EXT_VOLUME}/redis:/data` volume mount from #11100, per the earlier discussion with @SchrodingersGat — persistence stays off, so Windows `MISCONF` errors remain fixed while the external volume layout is preserved.

Updated the PR description to reflect the narrower scope.

br413:
Hi @SchrodingersGat — rebased onto latest master and narrowed scope per your feedback: this PR now only restores the Redis volume mount from #11100 while keeping persistence disabled via #12415.

All CI checks are green on the latest push. Ready for another look when you have time.

SchrodingersGat:
@br413 thanks for the contribution

PR #12423, opened 2026-07-19
Fix creation date formatting on part detail page

aamirasmal15:
The Creation Date field on the part detail page is declared as a plain string, so it shows the raw ISO date from the API and ignores the user's date display format setting. The same field on the order and build pages is declared as 'date' and formats fine, so this just aligns the part page with those.

Seen on 1.4.2, still present on master.


SchrodingersGat (approved):

SchrodingersGat:
@aamirasmal15 thanks for the fix!

PR #12424, opened 2026-07-20
Fix 12399 update notification

### LINKED ISSUE #12399: Update notification still fails after #10435 due to NotificationEntry uid NULL

### Describe the bug*

This appears to be a follow-up to #10435.

Issue #10435 fixed the NULL target_object_id failure by using the UI notification plugin configuration as the notification target. However, the same update notification still fails later when trigger_notification() attempts to create a NotificationEntry with uid=None.

This also causes the notification to link to the UI notification plugin configuration and display "Plugin Detail - undefined" when clicked.

The scheduled `InvenTree.tasks.check_for_updates` task creates an `update_available` UI notification with no target object:

    trigger_notification(
        None,
        'update_available',
        ...
    )

The UI notification is successfully created, but the task then fails when recording the notification entry:

    django.db.utils.IntegrityError:
    null value in column "uid" of relation "common_notificationentry"
    violates not-null constraint

The failure occurs because `obj_ref_value` is `None` and is passed to:

    NotificationEntry.notify(category, obj_ref_value)

while the model declares:

    uid = models.IntegerField()

As a result, the scheduled task fails every day and creates another duplicate UI notification.

Additionally, `InvenTreeUINotifications.send_notification()` replaces the missing target with its own plugin configuration:

    if not target:
        target = self.plugin_config()

Therefore clicking the update notification attempts to open a plugin detail page and displays:

    Plugin Detail - undefined
    Plugin Inactive
    Plugin is not active

Image

Image

### Relevant traceback

    File "InvenTree/tasks.py", line 671, in check_for_updates
        trigger_notification(None, 'update_available', ...)

    File "common/notifications.py", line 218, in trigger_notification
        NotificationEntry.notify(category, obj_ref_value)

    File "common/models.py", in notify
        get_or_create(key=key, uid=uid)

    django.db.utils.IntegrityError:
    null value in column "uid" of relation "common_notificationentry"
    violates not-null constraint

### Additional investigation

The GitHub API already returns a valid release URL in the response from:

    https://api.github.com/repos/inventree/inventree/releases/latest

For the current release, the relevant fields are:

    tag_name: 1.4.1
    html_url: https://github.com/inventree/InvenTree/releases/tag/1.4.1

However, `check_for_updates()` only reads `tag_name`:

    tag = data.get('tag_name', None)

The `html_url` field is ignored and is not passed to the notification system.

The UI notification plugin also does not use a link from the notification context. It only stores:

    name=ctx.get('name')
    message=ctx.get('message')

The resulting notification API response currently contains:

    target.model_type = pluginconfig
    target.model_id = 10
    target.link = https://parts.embedded.homes/admin/plugin/pluginconfig/10/change/

This confirms that the incorrect link is generated by the backend serializer from the fallback PluginConfig target, rather than by the frontend.

The effective flow is:

    trigger_notification(None, 'update_available')
        -> target = self.plugin_config()
        -> NotificationMessage.target_object = PluginConfig
        -> NotificationMessageSerializer uses the staff admin URL fallback
        -> clicking the notification opens the plugin configuration page

A search of production call sites also shows that `update_available` is the only current call to `trigger_notification()` which passes `obj=None`. Other notification calls provide a real model instance.

Another detail is that `record_task_success('check_for_updates')` is called before notification delivery. Therefore the update check is recorded as successful, but the background task subsequently raises an IntegrityError while recording the NotificationEntry.

### Suggested direction

A global notification needs an explicit stable deduplication identifier instead of `None`. The notification also needs an explicit destination URL or dedicated handling for the `update_available` category instead of using the UI notification plugin configuration as its target.

### Steps to Reproduce

1. Run an InvenTree version older than the current stable release.
2. Enable the scheduled update check.
3. Wait for or manually invoke `InvenTree.tasks.check_for_updates`.
4. Check the worker log.
5. Open the generated `Update Available` notification in the UI.

### Actual behaviour

- The UI notification is created.
- Insertion into `common_notificationentry` fails because `uid` is NULL.
- The update check is recorded as successful before notification delivery, but the background task subsequently raises an IntegrityError while recording NotificationEntry.
- A new duplicate notification is created on every scheduled run.
- Clicking the notification opens `Plugin Detail - undefined`.

### Expected behaviour

- The update notification should be successfully deduplicated.
- The scheduled task should complete without an exception.
- Clicking the notification should open the relevant release or update page, or the notification should be intentionally non-clickable.

### Deployment Method

Docker

### Version information

    InvenTree: 1.4.1
    API: 511
    Python: 3.14.6
    Django: 5.2.15
    Deployment: official Docker image
    Database: PostgreSQL 17

### Relevant database schema

    common_notificationentry.uid integer NOT NULL

I did not try to reproduce

Not reproducible

### Relevant log output

```shell

```

Leopoldius:
## Summary

- add an explicit notification UID for update notifications without a model instance
- persist the GitHub release URL on UI notification messages
- prefer the stored notification link during serialization
- add a migration for the new optional notification link field
- add regression coverage for notification creation, deduplication, and serialized release links

Fixes #12399

## Testing

- `ruff check` on changed files
- `ruff format --check` on changed files
- `python src/backend/InvenTree/manage.py test InvenTree.test_tasks.InvenTreeTaskTests.test_task_check_for_updates --verbosity 2`
- `python src/backend/InvenTree/manage.py test InvenTree.test_tasks --verbosity 2`
  - 11 tests passed
  - existing unrelated `test_failed_task_notification` failure remains reproducible independently


Leopoldius:
Small Git history mishap on my side xD

 The branch was initially based on an outdated fork history, but it has now been rebased onto the current upstream/master. The PR should be clean now.

BR,
Leonid

Leopoldius:
A small follow-up after rebasing the branch onto the latest upstream/master:

- The migration was renamed from `0045_notificationmessage_link` to `0048_notificationmessage_link`.
- Its dependency was updated from `0044_notificationmessage_charfield_pk` to `0047_parametertemplate_unique` to avoid a conflicting migration leaf.
- The existing commit was amended and the branch was force-pushed to keep the PR history clean and limited to a single commit.

The commit SHA changed from `0fd43bf3f` to `0475f3e13`. The migration check and targeted test pass, and the Read the Docs build is now green.

BR,
Leonid

SchrodingersGat (approved):

SchrodingersGat:
@Leopoldius thanks for the contribution!

PR #12432, opened 2026-07-20
Read report/label templates in binary mode

### LINKED ISSUE #12431: Default report creation fails on Backblaze B2 with IncompleteBody
### Summary

Creating the default report and label templates fails on Backblaze B2 (and other S3 backends that validate the request Content-Length strictly). The upload returns `IncompleteBody: The request body was too small`, the create rolls back, and InvenTree retries it on every startup.

### Environment

- InvenTree 1.4.1, also present on current master
- Storage backend: S3 (`INVENTREE_STORAGE_TARGET=s3`) pointed at Backblaze B2
- django-storages `S3Storage` with boto3

### Cause

`ReportConfig.file_from_template` in `src/backend/InvenTree/report/apps.py` reads the template in text mode:

```python
return ContentFile(file.open('r').read(), os.path.basename(file_name))
```

`file.open('r').read()` returns a `str`, so `ContentFile.size` is the character count. For any template that contains a multi-byte UTF-8 character, that count is smaller than the encoded byte length. django-storages then sends the character count as the `Content-Length` header while the body is the encoded bytes. AWS S3 accepts the mismatch, but Backblaze B2 rejects it.

Only `inventree_transfer_order_report.html` is affected today, because it is the only default template with a non-ASCII character: 1458 characters, 1460 bytes.

### Steps to reproduce

With `default_storage` configured for an S3 backend on Backblaze B2:

```python
from django.core.files.base import ContentFile
from django.core.files.storage import default_storage
path = '.../report/templates/report/inventree_transfer_order_report.html'
content = open(path, 'r').read()   # str, 1458 chars
default_storage.save('test.html', ContentFile(content))   # raises IncompleteBody
```

Reading the same file with `open(path, 'rb')` uploads without error.

### Suggested fix

Read the template in binary mode so `ContentFile.size` matches the bytes that are uploaded.

impuls42:
### Fix

`file_from_template` reads default report and label templates in text mode, so the resulting `ContentFile` reports its size as a character count. For a template with multi-byte UTF-8 characters that count is smaller than the encoded byte length, and django-storages sends it as the `Content-Length`. Backblaze B2 and other strict S3 backends reject the upload with `IncompleteBody`. AWS S3 tolerates the mismatch, so the bug is not visible there.

`inventree_transfer_order_report.html` is the only default template with a non-ASCII character (1458 chars, 1460 bytes), so its upload fails during `create_default_reports`, the transaction rolls back, and it is retried on every startup.

Reading the file in binary mode makes `ContentFile.size` match the uploaded bytes.

Fixes #12431

### Testing

Reproduced on a live instance backed by Backblaze B2. Before the change, `create_default_reports` fails for the transfer-order template on every startup with `IncompleteBody`. After the change the template is created and the error is gone. Confirmed the mechanism in a shell for the same file: `ContentFile(open(path, 'r').read())` fails, `ContentFile(open(path, 'rb').read())` succeeds.

The mismatch only shows up against an S3 backend that enforces Content-Length, so it is not reproducible against the local storage used in CI.

### DIFF
--- a/src/backend/InvenTree/report/apps.py
+++ b/src/backend/InvenTree/report/apps.py
@@ -96,7 +96,7 @@ class ReportConfig(AppConfig):
             raise FileNotFoundError(
                 'Template file %s does not exist in %s', file_name, file
             )
-        return ContentFile(file.open('r').read(), os.path.basename(file_name))
+        return ContentFile(file.read_bytes(), os.path.basename(file_name))
 
     def create_default_labels(self):
         """Create default label templates."""
--- a/src/backend/InvenTree/report/tests.py
+++ b/src/backend/InvenTree/report/tests.py
@@ -1,7 +1,9 @@
 """Unit testing for the various report models."""
 
 import os
+import tempfile
 from io import StringIO
+from pathlib import Path
 from unittest.mock import patch
 
 from django.apps import apps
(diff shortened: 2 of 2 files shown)


impuls42:
Updated after a self-review:

- Switched to `Path.read_bytes()` instead of `open('rb').read()`, so the file handle is closed rather than relying on refcounting.
- Added a regression test in `report/tests.py` asserting that `file_from_template` sizes the `ContentFile` by the file's byte length. It fails on the old text-mode read (character count) and passes with this change, and it needs no S3 backend.

SchrodingersGat (approved):

SchrodingersGat:
@impuls42 thanks for the fix, this seems clean

SchrodingersGat:
This is the same issue as https://github.com/inventree/InvenTree/pull/12410

SchrodingersGat:
@impuls42 can you please add some testing against different text encodings as per the comments in https://github.com/inventree/InvenTree/pull/12410

impuls42:
Added a regression test (`test_file_from_template_is_encoding_agnostic`) that runs `file_from_template` over content in utf-8, utf-16 and latin-1 and asserts the ContentFile is byte-identical to the file on disk, with the size equal to the byte length. The utf cases include multi-byte characters where the byte length differs from the character count, which is the mismatch the old text-mode read produced. Reading the raw bytes never decodes, so non-English content can't be mis-sized or corrupted regardless of the file's encoding.

SchrodingersGat (approved):

SchrodingersGat:
@impuls42 thanks for the contribution!

PR #12507, opened 2026-07-30
feat(backend): switch to a more formal FSM approach

matmair:
This PR refactors the first few models to use django-fsm2 for handling state transitions in models. This approach also moves common behavior into a shared decorator like:
- refreshing state before evaluation
- keeping the whole transaction atomic
- triggering events after all state checks and actions are done

It also fixates a currently sometimes differing behavior: it disallows runs of transitions if there is no change in state.

Todo:
- [x] look at db efficiency
- [x] should we always raise validation errors instead of failing silently? (would be breaking)
- [x] this would be a great place to add specific transition permissions -> do we want to add that now / later? --> maybe later
- [x] add depreciation for the redundant `_can` checks
- [x] look into moving away from signals for transition handling


matmair:
@SchrodingersGat @wolflu05 @inventree/triage still have to figure out 2 db efficiency issues but I would appreciate a general look at this approach.

I also have these questions that I am unsure off and would appreciate your opinion on:
- [ ] should we always raise validation errors instead of failing silently? (would be breaking)
- [ ] this would be a great place to add specific transition permissions -> do we want to add that now / later?

SchrodingersGat:
I like this approach, the intent (and implementation) of the code is *much* cleaner now.

Question: Does this still allow plugins to override / intervene with transitions?

> should we always raise validation errors instead of failing silently? (would be breaking)

I think a validation error is a good thing - I have run into silent failures on transition and it's very hard to debug

> this would be a great place to add specific transition permissions -> do we want to add that now / later?

I think we should limit scope here, to prevent it from blowing out and never getting merged

matmair:
thank you for the fast response; I think this will make future use of state-transitions much more approachable. I think a lot of places in core and plugins could benefit from using this approach.

> Does this still allow plugins to override / intervene with transitions?

yes, that is fully preserved

matmair:

SchrodingersGat:
@matmair please address conflicts and then I am happy to merge

matmair:
merge conflict adressed @SchrodingersGat

SchrodingersGat:
Thanks for this improvement!

matmair:
Working on docs for plugins for this; have been building some very low LOC plugins with this

PR #12508, opened 2026-07-30
Fix admin search for PurchaseOrderLineItem

bhumin18:
## Summary

Expands `search_fields` on `PurchaseOrderLineItemAdmin` to allow searching by part name, part description, supplier SKU, order reference, and supplier name.

## Motivation

Previously only the optional `reference` field was searchable, making it impossible to find line items by part or order details in the Django Admin panel.

Searching for a part name like `Widget` returned 0 results, while the same search on `SalesOrderLineItemAdmin` works correctly because it has `part__name` in its `search_fields`.

## Changes

- `src/backend/InvenTree/order/admin.py`: Expanded `search_fields` on `PurchaseOrderLineItemAdmin`
- `src/backend/InvenTree/order/tests.py`: Added unit test to verify search_fields configuration

### DIFF
--- a/src/backend/InvenTree/order/admin.py
+++ b/src/backend/InvenTree/order/admin.py
@@ -94,7 +94,14 @@ class PurchaseOrderLineItemAdmin(admin.ModelAdmin):
 
     list_display = ('order', 'part', 'quantity', 'reference')
 
-    search_fields = ('reference',)
+    search_fields = [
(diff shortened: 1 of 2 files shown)


SchrodingersGat (approved):

SchrodingersGat:
@bhumin18 thanks for the contribution

PR #12510, opened 2026-07-30
Add Previous/Next navigation between detail views (#12397)

### LINKED ISSUE #12397: [FR] Add Next and Previous Icon to switch easily between 2 objects

### Problem statement

Hi,

### Suggested solution

Image

Best regards

### Examples of other systems

Image

- [ ] I want to develop this.

jdjioe5-cpu:
Resolves https://github.com/inventree/InvenTree/issues/12397

Adds a generic Previous/Next icon-button pair to entity detail pages so users can step between sibling instances without returning to the list view, matching the suggestion from the issue.

## What changes

* `NextPrevAction` — a small, dumb presentational component that renders two compact `ActionIcon` buttons (`IconChevronLeft` / `IconChevronRight`). It is intentionally model-agnostic: the caller resolves the destination `pk` and `onClick` and the component just disables the button when the target is undefined. It is exported from `lib/` so frontend plugins and other detail pages can reuse it.
* `useNextPrevSiblings` — a React-Query hook that queries the InvenTree REST API for the previous / next sibling using `pk__lt` / `pk__gt` filtering and the same `ordering` parameter the list view uses. Only two single-row queries are issued per detail page, so the cost is independent of list length.
* `PageDetail` now accepts an optional `nextPrev` prop and renders the action next to the existing actions group. The button is placed before the action icons so it stays close to the title area without crowding the rest of the toolbar.
* `PartDetail` is the first call site. It filters by the part's current `active` and `assembly` flags so the navigation stays within a coherent subset; this matches the spirit of the issue ("only for the filtered items") while staying conservative until a richer filter-context API is available.
* When a side has no sibling (e.g. the first or last item in the filtered set), the corresponding button is disabled instead of hidden, so the affordance is consistent across detail pages.

## Why this design

The issue explicitly asks for navigation scoped to the originating filter list. Rather than reimplement filter state on the client, the hook reuses the existing list API and only asks for one record per direction. This keeps the implementation small, keeps it correct under any list-ordering policy, and avoids duplicating the filter-parser logic.

The component is split out of `PartDetail` so future work can wire it into `StockItemDetail`, `PurchaseOrderDetail`, etc. without copy-pasting icon code. See `PageDetail` for the single integration point.

## Tests

* `tests/pui_part_next_prev.spec.ts` — Playwright regression that renders the detail page, asserts both buttons appear via stable `data-testid` hooks (`inventree-prev-item`, `inventree-next-item`), clicks `next` and `prev`, and verifies that the disabled-state fallback also works on single-item datasets.
* The new TypeScript code passes `tsc --noEmit` cleanly.

Refs https://github.com/inventree/InvenTree/issues/12397


matmair (changes requested):
See https://github.com/inventree/InvenTree/issues/12397#issuecomment-5140111349

matmair:

PR #12511, opened 2026-07-31
feat: Add generic previous/next navigation between detail views (closes #12397)

### LINKED ISSUE #12397: [FR] Add Next and Previous Icon to switch easily between 2 objects

chidii:
Closes #12397

 Design
Answers the two open questions from the issue thread:

- Filter/order fidelity: navigation context (endpoint + ordered pk list) is captured
  at row-click time in InvenTreeTable and stored in a small zustand slice
  (useLocalLibState, alongside the existing detailDrawerStack pattern). Prev/next on the
  detail page resolve as O(1) lookups into that exact list the user was viewing - not
  just "next by pk". No stored context (direct link, plugin embedding) falls back to an
  ordering-aware pk__gt/pk__lt query.
- List shrinks mid-navigation: if a resolved neighbor 404s, it's dropped from the
  stored context and the next available neighbor is fetched automatically.

 Generic for plugins
`nextPrev` is a first-class optional prop on `PageDetail` itself
(`{hasPrev, hasNext, onPrev, onNext}`), not per-page logic. Any page or plugin using
`PageDetail`/`Instance` gets it by calling `useNextPrevInstance` - no copy-pasted filter
code per detail page.

 Changes
- `hooks/UseNextPrevInstance.tsx` (new) - peer to `useInstance`, resolves prev/next pk
- `states/LocalLibState.tsx` - adds `listNavContexts` (excluded from persistence -
  ephemeral navigation state, not a user preference)
- `components/nav/PageDetail.tsx` - renders prev/next chevrons + alt+←/→ hotkeys when
  `nextPrev` is supplied
- `pages/part/PartDetail.tsx` - first integration

 Not yet done
- InvenTreeTable row-click wiring to populate `listNavContexts` (needs review against
  actual table component)
- Tests: hook-level (context hit/miss/stale-pk-recovery) + Playwright shrink-list case


matmair:
This approach leads to differing behaviors when a user shares the link to the page they currently have open - user 1 sees the nav, user 2 not; the filter, ordering etc need to be somewhere in the url path so that behavior is consistent and repeatable

I like the overall approach; the implementation should stire in a way where the logic is implemented in the generic components so that plugins do not need to re-implement all this logic. Similar to the approach that is taken with the InvenTreeTable component

matmair (changes requested):
See https://github.com/inventree/InvenTree/issues/12397#issuecomment-5140111349

matmair:

PR #12528, opened 2026-07-31
Add previous/next navigation to detail pages

### LINKED ISSUE #12397: [FR] Add Next and Previous Icon to switch easily between 2 objects

ejspeed-cmd:
## Summary

Adds previous/next navigation buttons to detail pages, allowing users to move through a filtered and ordered list without returning to the table view. 
- Closes #12397.

## Motivation

When reviewing many items from a filtered list (e.g. all in-stock items for a location), users currently have to navigate back to the list, find their place, and click into the next item. This adds prev/next chevron buttons to the detail page header that preserve the active filter and ordering context.

### Backend

- `stock/api.py`: Added `pk_gt` and `pk_lt` filter fields to `StockFilter`
- `part/api.py`: Added `pk_gt` and `pk_lt` filter fields to `PartFilter`

These allow the frontend to query for the adjacent item in the current ordered/filtered set.

### Frontend

- `hooks/UseNextPrev.tsx`: New hook `useNextPrevSiblings(pk)` that reads nav context from `_nav_*` URL params and issues `pk_gt`/`pk_lt` API queries to find the previous and next item PKs
- `components/nav/NextPrevAction.tsx`: New component rendering prev/next chevron buttons, disabled when no sibling exists
- `components/nav/PageDetail.tsx`: Accepts a `pk` prop; calls `useNextPrevSiblings` internally and renders `NextPrevAction` in the header actions area
- `components/tables/InvenTreeTable.tsx`: Row click handler now encodes the active endpoint, ordering and filters as `_nav_*` query params on the detail URL
- `components/panels/PanelGroup.tsx`: Panel navigation now preserves existing query params so nav context is not lost when switching tabs
- `pages/part/PartDetail.tsx`: Passes `pk` to `PageDetail`
- `pages/stock/StockDetail.tsx`: Passes `pk` to `PageDetail`

## Behaviour

- Buttons only appear when arriving via a table row click (i.e. `_nav_*` params are present in the URL)
- The URL is fully shareable: the filter/ordering context is encoded in query params
- Ordering matches the list view exactly
- Navigation preserves the current panel and nav context across clicks
- The feature is implemented entirely in generic components (`InvenTreeTable`, `PageDetail`) with no per-page logic beyond passing `pk`


src/frontend/src/locales/ar/messages.po
matmair, line 1: remove these as per contribution guidelines

src/frontend/src/components/tables/InvenTreeTable.tsx
```diff
@@ -747,11 +748,52 @@ export function InvenTreeTableInternal<T extends Record<string, any>>({
...
 
           if (!showPreviewPanel || eventModified(event as any)) {
-            navigateToLink(url, navigate, event);
+            // Build nav context from current table state so the detail page
+            // can resolve prev/next against the same filtered/ordered set.
+            if (url) {
+              const queryFilters = getTableFilters(false);
+              const ordering = queryFilters.ordering ?? '';
+
+              const DISPLAY_PARAMS = new Set([
```
matmair: this hardcodes a (partial) set of the /api/part/ endpoint - not a generic approach

ejspeed-cmd: done , removed the hardcoded parametes

ejspeed-cmd:
@matmair is there anything i can do to fix the failing test ?, i checked them and it doesn't seem related to my changes

ejspeed-cmd:
@matmair can you reapprove the ci ?

ejspeed-cmd:
The API schema diff failure is because  pk_gt/pk_lt are new filter fields that appear in the generated schema. I've left the api_version.py bump for maintainers as per the project convention

matmair:
Can you point out where it is documented that you should not bump the API version?
https://github.com/inventree/InvenTree/blob/master/docs/docs/develop/contributing.md#api-versioning

ejspeed-cmd:
> Can you point out where it is documented that you should not bump the API version? https://github.com/inventree/InvenTree/blob/master/docs/docs/develop/contributing.md#api-versioning

it's not - in fact it says the opposite, my bad :-( -i've pushed the commit to bump it

ejspeed-cmd:
@matmair Should prev/next navigation work when arriving via the preview panel sidebar link, and should
  the buttons appear without filter context (defaulting to pk ordering)?

matmair:
@ejspeed-cmd no

PR #12532, opened 2026-08-01
feat: add reusable previous/next detail navigation

### LINKED ISSUE #12397: [FR] Add Next and Previous Icon to switch easily between 2 objects

krysto9872:
## Summary

Implements the requested previous/next navigation for detail pages (ref #12397).

- Carries the current table page's model IDs through React Router state when opening a detail view.
- Renders accessible previous/next controls in `PageDetail`, including explicit targets for plugin pages.
- Restores the previously focused input by `id` or `name` after navigating between records.
- Keeps navigation bounded to the loaded table page and fails closed when context is missing or invalid.

## Validation

- `npm run build`
- `npm run lib`

No backend changes are required.
closes #12397


matmair:

PR #12533, opened 2026-08-01
Fix test result ordering by test timestamps

### LINKED ISSUE #12509: Test results not showing most recent test by date-time

### Describe the bug*

When uploading test results, instead of the most recent result by date-time being shown, the test result shown is just the one that was most recently recorded.

### Steps to Reproduce

1. Create a part, ensure that it has 'testable' toggled on.
2. Navigate to the part page > test templates > add a test template. (see image for my test template)
3. Create a stock instance of the part.
4. Open the stock page for that instance and navigate to 'test results'.
5. Add a passing test result. Ensure that the start/finish time are recent. (I chose today's date at 12:00:00-13:00:00)
6. Submit the test result 
Image
7. See that the test results shows this test as being in the state PASS.
Image
8. Add another test result. This time the test should be failing. Choose an older start/finish time. (I chose 1 week prior to today's date 12:00:00-13:00:00)
Image
9. Submit the test result.
10. See that now the test results show this test as being in the state FAIL, despite the actual time of testing being outdated.
Image

### Expected behaviour

The status and data of the most recent test result by recorded date-time are shown. (Not whenever the result happens to be inputted)
I understand that there are two times to consider, start and finish time. I think my team would prioritize finish time, but perhaps this could be a setting within InvenTree (order tests by start-time or finish time).

### Deployment Method

None

### Version Information

InvenTree-Version: 1.5.0 dev
API Version: 522
Python Version: 3.14.5
Django Version: 5.2.15

I'm not able to copy the full version information on my inventree instance.

I tried to reproduce

Not reproducible

### Relevant log output

```shell
The date-time fields are not available on the demo site. I was able to grab the API responses from POST requests when adding each test result.

Logging recent test:

pk    4430
stock_item    10636
result    true
value    "5" 
attachment    null
notes    "" 
test_station    "" 
started_datetime    "2026-07-30 00:00" 
finished_datetime    "2026-07-30 01:00" 
user    8
date    "2026-07-30 17:46" 

Logging older test:

pk    4431
stock_item    10636
result    false
value    "10" 
attachment    null
notes    "" 
test_station    "" 
started_datetime    "2026-07-23 00:00" 
finished_datetime    "2026-07-23 01:00" 
user    8
date    "2026-07-30 18:02"
```

jayasree-007:
## Summary

This PR updates the default ordering of `StockItemTestResult` to prioritize:

1. `finished_datetime`
2. `started_datetime`
3. `date`
4. `pk`

It also updates the frontend ordering in `StockItemTestResultTable` and `BuildOutputTable` so the displayed "latest" test result follows the same ordering logic.

Fixes #12509


SchrodingersGat (approved):

src/frontend/src/tables/build/BuildOutputTable.tsx
```diff
@@ -79,6 +79,45 @@ type TestResultOverview = {
...
 
+const compareTestResults = (a: any, b: any) => {
+  const finishedA = a.finished_datetime
+    ? new Date(a.finished_datetime).getTime()
+    : null;
+  const finishedB = b.finished_datetime
+    ? new Date(b.finished_datetime).getTime()
+    : null;
+
+  if (finishedA !== finishedB) {
```
SchrodingersGat: Perhaps this repeated comparison pattern could be refactored into a helper? It's repeated three times here, and three times again in the other file

jayasree-007: Thanks for the suggestion. I've refactored the comparison logic into a shared helper and updated both BuildOutputTable and StockItemTestResultTable to use it.

SchrodingersGat:
@jayasree-007 just style issues that need to be addressed here

jayasree-007:
sure, will check on it

jayasree-007:
I investigated the Read the Docs failure. The referenced directories (plugin/samples, report/templates/report, and contrib/container) are present and tracked in the PR commit a64b2f16. However, docs/main.py constructs the GitHub source URLs using the upstream inventree/InvenTree repository and the PR commit hash. Since this commit is on my fork, those generated upstream URLs return 404. I reproduced the 404 locally with both HEAD and GET requests. This appears to be unrelated to the changes in this PR. Could you please confirm how you'd like me to proceed?

SchrodingersGat:
You are missing a required migration file - run `invoke migrate` and commit the generated file

SchrodingersGat:
@jayasree-007 thanks for the contribution!

jayasree-007:
@SchrodingersGat Thank you so much, it was great making my first contribution here.

PR #12550, opened 2026-08-04
reset errors on update

### LINKED ISSUE #12549: Unable to manually select a component on BOM import

### Describe the bug*

When importing a BOM that has unresolvable components, the error persists, even though a component was manually selected.

Image

### Steps to Reproduce

Use eg. this BOM with the demo dataset and import it to any assembly:

[BOM_short.csv](https://github.com/user-attachments/files/30709500/BOM_short.csv)

Mapped columns:
- Component -> Name
- Reference -> Designator
- Amount -> Quantity

### Expected behaviour

The error should be resolved, after selecting an existing component from the database and the row can be imported afterwards.

### Deployment Method

Docker

### Version Information

InvenTree 1.4.3

I tried to reproduce

Reproducible

### Relevant log output

```shell

```

ribsey:
In order to allow the user to manually select a component during BOM import, the validation should be redone (by clearing the errors) after a new component was selected

- fixes #12549

_I did not look at the code closely, so I'm not sure if it is a valid fix. But in my case, it worked._


SchrodingersGat:
@ribsey please fix style issues here

You can setup git hooks to automatically format the code for you on commit:

https://docs.inventree.org/en/stable/develop/contributing/#setup-devtools

SchrodingersGat:
@ribsey thanks for the contribution!

PR #12585, opened 2026-08-08
feat(frontend): Add persistant filtered detail navigation in the url

### LINKED ISSUE #12397: [FR] Add Next and Previous Icon to switch easily between 2 objects

Jr-kenny:
The detail navigation now follows the review feedback.

The helper is a `.tsx` file, and the navigation actions use the shared `navigateToLink` handler. Previous and Next use Tabler icons with accessible labels and tooltips. A subtle `x of y` indicator shows the current position in the filtered result set, using the API count for the full total.

I added a Playwright regression that opens a detail page from the table and moves from the first record to the second.

I checked this with `yarn tsc --noEmit`, `yarn build`, Biome on the changed files, and a local Playwright browser flow against the demo dataset.

Closes #12397


src/frontend/src/functions/DetailNavigation.tsx
SchrodingersGat, line 1: this should be a `.tsx` file, not `.ts`

src/frontend/src/hooks/UseDetailNavigation.tsx
```diff
@@ -0,0 +1,122 @@
...
+          event.button !== 0 ||
+          event.ctrlKey ||
+          event.metaKey ||
+          event.shiftKey
+        ) {
+          return;
+        }
+
+        event.preventDefault();
+        navigate(href);
```
SchrodingersGat: Look at existing `navigateToLink` function - we already implement these checks generically

SchrodingersGat:
### User Interface Buttons

I would like to see the navigation buttons being icons (tabler) - not text. Also, a subtle `x of y` text might be helpful. See example image below

image

SchrodingersGat:
### Unit Tests 

This will require regression testing via playwright.

SchrodingersGat:
@Jr-kenny this is starting to come together nicely

I think that the placement is a bit wonky here - perhaps it makes more sense in the "breadcrumb" bar - and please adjust the layout to vertically center the items, larger icons, etc

image

SchrodingersGat:
An implementation request - putting all of the attributes into the URL gets pretty messy. Perhaps we can instead use a hash of the values, and then store the actual data in sessionStorage - keyed to the hash? So this way all we add is `?nav=1A2BC3`

matmair:
@SchrodingersGat that makes it not shareable, invalidating most of the usefulness of links

SchrodingersGat:
Ok, I was not thinking about sharing, for me I think the main value is the quick navigation within the user's own table context.

Jr-kenny:
Yeah, I see what you mean. The current implementation puts the API URL, filters, index, PK, and field directly in the URL. I’ll switch that to a short `nav` hash like `?nav=1A2BC3`, with the navigation data stored in `sessionStorage` under that hash. I’ll update the link builder, reader, and Playwright coverage as part of that.

Jr-kenny:
> Ok, I was not thinking about sharing, for me I think the main value is the quick navigation within the user's own table context.

@SchrodingersGat 
Just to confirm before I change it umm should I go ahead with the sessionStorage + short ?nav= hash approach for local navigation, even though those links won’t be shareable? Or do we want to keep the navigation state in the URL so links remain portable?

matmair:
The bounty is scoped to be generally useful, not just for one user https://github.com/inventree/InvenTree/issues/12397#issuecomment-5140111349

SchrodingersGat:
To be clear, I am not saying that I do not want the links to be shareable - I simply had neglected to consider that in my previous review.

With regard to shareable links, there is a consideration we need to make - the API does not always return the same results for each user. For example, there are many API endpoints which filter against "assigned_to_me" (or similar) - which will obviously return different results for each user

That being said I am keen to see the links be shareable - but, perhaps there is a way to also make the URLs cleaner?

SchrodingersGat:
### URL Shortening

In terms of URL shortening, let's use this example: `http://localhost:5173/web/manufacturing/build-order/7/details?_nav_api=%2Fapi%2Fbuild%2F&_nav_index=17&_nav_pk=7&_nav_field=pk&_nav_query=part_detail%3Dtrue%26outstanding%3Dtrue`

#### Variable Names

The variable names can be shortened, as an example

| Long Name | Short Name |
| --- | --- |
| `_nav_api` | `_na` |
| `_nav_query` | `_nq` |
| `_nav_index` | `_ni` |

#### Nav Field

Most of the time, the nav field will be `pk` - you can omit that unless it is *not* `pk`

### Nav API

The `_nav_api` values are mostly going to point to hard-coded fixed text values - we could have a lookup table for these to shorten the URL even further? e.g. `_nav_api=XYZ`

SchrodingersGat:
There is some visual "instability" when clicking through the navigation items - see video below. It would be better if the buttons did not alter the height of the navigation bar.


https://github.com/user-attachments/assets/99cda0f1-96c2-4c37-8891-c27cb77a7401

SchrodingersGat:
Also, please put the text first, before the buttons, so that the position of the buttons does not move as the length of the text changes:

image

SchrodingersGat:
@Jr-kenny also, can you please add a documentation section into `docs/concepts/ui.md` detailing the new feature

src/frontend/src/functions/DetailNavigation.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+export const DETAIL_NAVIGATION_PARAMS = {
+  api: '_na',
+  query: '_nq',
+  index: '_ni',
+  pk: '_np',
+  field: '_nf'
+} as const;
+
+const LEGACY_DETAIL_NAVIGATION_PARAMS = {
```
SchrodingersGat: This is not necessary, there is no "legacy" outside the context of this PR

src/frontend/src/functions/DetailNavigation.tsx
```diff
@@ -0,0 +1,279 @@
...
+  field: '_nav_field'
+} as const;
+
+const PAGINATION_PARAMS = new Set(['limit', 'offset', 'page']);
+const DEFAULT_DETAIL_NAVIGATION_FIELD = 'pk';
+
+// Keep common built-in list endpoints compact while preserving the full URL
+// for plugin and otherwise unknown endpoints.
+const DETAIL_NAVIGATION_API_ALIASES: Record<string, string> = {
+  '/api/part/': 'p',
```
SchrodingersGat: Use the existing `ApiEndpoint` enum rather than re-defining hard-coded API URLs

src/frontend/src/functions/DetailNavigation.tsx
```diff
@@ -0,0 +1,279 @@
...
+  '/api/parameter/template/': 'pt',
+  '/api/user/': 'u',
+  '/api/user/group/': 'ug',
+  '/api/project-code/': 'pr',
+  '/api/tag/': 'tag',
+  '/api/attachment/': 'at',
+  '/api/machine/': 'm'
+};
+
+const DETAIL_NAVIGATION_API_URLS = new Map(
```
SchrodingersGat: With the suggested re-mapping as above I do not think this is needed anmore

SchrodingersGat (dismissed):

SchrodingersGat:
@Jr-kenny nice, getting very close I think. Review needed from @matmair

src/frontend/src/components/nav/PageDetail.tsx
```diff
@@ -184,11 +253,9 @@ export function PageDetail({
...
               )}
             </Group>
-            {computedActions && (
-              <Group gap={5} justify='right' wrap='nowrap' align='flex-start'>
-                {computedActions.map((action, idx) => (
-                  <Fragment key={idx}>{action}</Fragment>
-                ))}
+            {computedActions.length > 0 && (
+              <Group gap={5} justify='right' wrap='nowrap' align='center'>
+                {computedActions}
```
matmair: why is this changed? wont this cause key issues?

src/frontend/src/components/nav/PageDetail.tsx
```diff
@@ -130,14 +141,72 @@ export function PageDetail({
...
     <>
       <PageTitle title={pageTitleString} />
       <Stack gap='xs'>
-        {computedBreadcrumbs && computedBreadcrumbs.length > 0 && (
+        {((computedBreadcrumbs?.length ?? 0) > 0 || hasDetailNavigation) && (
           <BreadcrumbList
             navCallback={breadcrumbAction}
-            breadcrumbs={computedBreadcrumbs}
+            breadcrumbs={computedBreadcrumbs ?? []}
+            rightSection={
```
matmair: This should be in a seperate component, not the Breadcrumbs imo

src/frontend/src/functions/DetailNavigation.tsx
```diff
@@ -0,0 +1,279 @@
...
+}
+
+function removeDetailNavigationParams(params: URLSearchParams) {
+  Object.values(DETAIL_NAVIGATION_PARAMS).forEach((key) => {
+    params.delete(key);
+  });
+
+  Object.values(LEGACY_DETAIL_NAVIGATION_PARAMS).forEach((key) => {
+    params.delete(key);
+  });
```
matmair: this is not needed IMO

matmair (approved):
this is the most promising PR for sure; as this is a new function these legacy catches against previous iterations of the PR are unncessary

Jr-kenny:
@matmair I pulled the detail navigation out of `BreadcrumbList`. `PageDetailNavigationBar` now owns the shared bar and renders `BreadcrumbList` and `DetailNavigation` as siblings, so the controls stay in the same visual position without the breadcrumb component owning them. Is this the separation you had in mind?

matmair:
@Jr-kenny there is a test failing consistently; please address that

matmair:
also: disable actions that are no allowed, do not remove them. that leads to instability

https://github.com/user-attachments/assets/6f412d36-8ca5-487d-901f-5a668cbedfef

src/frontend/src/components/nav/DetailNavigation.tsx
```diff
@@ -37,34 +37,36 @@ export function DetailNavigation({
...
-      )}
+      <Tooltip label={t`Previous`} position='top'>
+        <ActionIcon
+          component='a'
+          href={navigation.previous?.href}
+          onClick={
+            navigation.previous?.onClick ?? ((event) => event.preventDefault())
+          }
+          data-disabled={!navigation.previous || undefined}
+          aria-disabled={!navigation.previous}
```
matmair: just set `disabled`, do not reinvent it

matmair (approved):
I am happy with merging this and awarding the bounty
@SchrodingersGat agreed?

SchrodingersGat:
@matmair yes, good to go on this. Thanks for the contribution @Jr-kenny

matmair:
@Jr-kenny please claim your bounty on bounty hub https://www.bountyhub.dev/en/bounty/view/a157ff08-20cf-466a-9ab3-a321f515133f

Jr-kenny:
@matmair thanks again for merging #12585 and pointing me to the bounty.

I tried to claim it on BountyHub, but the claim page says: **“This bounty is exclusive, you need to be assigned to work on it in order to claim it.”**

The issue itself currently has no assignee, and BountyHub doesn’t seem to recognize me as assigned. Since you mentioned earlier that whoever’s solution was merged would be assigned, does'nt the bounty need to be assigned to `Jr-kenny` first before I can import and claim PR #12585?

Screenshot 2026-08-19 at 10 42 34 AM

Jr-kenny:
One more question on the bounty payment side — I’m new to BountyHub, so I didn’t realize beforehand that the payout setup requires SWIFT/BIC details.

My current USD account only supports ACH and gives me an account number + routing number, but no SWIFT/BIC.

Is there any alternative way you normally handle bounty payments in cases like this, or does it strictly have to go through BountyHub’s SWIFT-supported payout setup?

matmair:
I will write their support. All bounties go through bountyhub; they have stripe in the backend so there might be other options

matmair:
@Jr-kenny try again

Jr-kenny:
Just did  I’ve sent the assignment request again. Thanks!

matmair:
You should now be able to claim the bounty with this pr link

matmair:
@Jr-kenny have you send your claim? I see that it is now assigned but I can not submit a claim for you, that seems to be only possible from your end

Jr-kenny:
Not yet  , the assignment issue is fixed now, thanks.
i just even claimed 
I’m still having trouble with the payout setup on BountyHub. It requires an account number + SWIFT/BIC, while my current USD account only supports ACH/routing details.

I’ve already contacted BountyHub support to ask whether there’s another supported payout option, but no response

PR #12590, opened 2026-08-08
[bug] Plugin reload fix

SchrodingersGat:
Fixes bug where API endpoint could be operating on stale plugin instance


src/backend/InvenTree/plugin/api.py
```diff
@@ -358,7 +358,10 @@ def check_plugin(
     if not plugin_cfg.active:
         raise NotFound(detail=f"Plugin '{ref}' is not active")
 
-    plugin = plugin_cfg.plugin
```
matmair, line 361: should we remove this public propoerty complemetly to not incourage these kind if issues in the future?

matmair (approved):

PR #12596, opened 2026-08-09
Fix Install Stock Items shows only selected parts

### LINKED ISSUE #12232: Install Stock Items shows all variants of part in stock list when Allow Variants is off

### Describe the bug*

When using BOM enforcement with an item that has variants, even if "Allow Variants" is off in the BOM, if you select the "parent" part in the Part selector, all variants are shown in the Stock list. It fails on install stating it isn't in the BOM.

### Steps to Reproduce

1. Create a part that has variants
2. Add the main part to a BOM without allowing variants
3. Go to Install Stock Item on the stock with the BOM
4. Select the parent part in the part selector
5. All variants are shown in the stock list

### Expected behaviour

With variants off, only the selected part should be shown in the stock list.

In general, filter on show as well as on submit to declutter the experience.

### Deployment Method

Docker

### Version Information

InvenTree-Version: 1.3.5
Django Version: 5.2.15
Commit Hash: d8cac9b
Commit Date: 2026-06-11
Commit Branch: null
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-7.0.0-15-generic-x86_64-with-glibc2.41
Installer: DOC
Active plugins: [{"name":"InvenTreeBarcode","slug":"inventreebarcode","version":"2.2.0"},{"name":"BOM Exporter","slug":"bom-exporter","version":"1.1.0"},{"name":"InvenTree Exporter","slug":"inventree-exporter","version":"1.0.0"},{"name":"Parameter Exporter","slug":"parameter-exporter","version":"2.0.0"},{"name":"InvenTreeEmailNotifications","slug":"inventree-email-notification","version":"1.0.0"},{"name":"InvenTreeUINotifications","slug":"inventree-ui-notification","version":"1.0.0"},{"name":"InvenTreeCurrencyExchange","slug":"inventreecurrencyexchange","version":"1.0.0"},{"name":"InvenTreeMachines","slug":"inventree-machines","version":"1.0.0"},{"name":"InvenTreeLabel","slug":"inventreelabel","version":"1.1.0"},{"name":"InvenTreeLabelMachine","slug":"inventreelabelmachine","version":"1.0.0"},{"name":"InvenTreeLabelSheet","slug":"inventreelabelsheet","version":"1.0.1"},{"name":"DigiKeyBarcodePlugin","slug":"digikeyplugin","version":"1.0.1"},{"name":"LCSCBarcodePlugin","slug":"lcscplugin","version":"1.0.1"},{"name":"MouserBarcodePlugin","slug":"mouserplugin","version":"1.0.1"},{"name":"TMEBarcodePlugin","slug":"tmeplugin","version":"1.0.1"}]

I tried to reproduce

Reproducible

### Relevant log output

```shell

```

siddhartha928:
## What does this PR do?

Fixes the "Install Stock Item" dialog so the stock item selector only lists stock for the exact part chosen in the part selector, instead of also including stock for all of that part's variants. This closes a gap where the UI would let a user pick a variant's stock item that the backend would then reject at submit time.

## Why was this PR needed?

When installing a stock item into an assembly with a BOM, the part selector already correctly restricts choices based on the BOM (respecting each line's "Allow Variants" setting). However, the stock item selector is filtered only by the selected part, and the stock list API's include_variants filter defaults to true when not specified, so it silently re-expanded the results to include every variant's stock, even for BOM lines with variants disallowed. This meant a user could select a variant's stock item in the UI that the install serializer's BOM check would then reject with "Selected part is not in the Bill of Materials", a mismatch between what the UI showed and what the API would accept. The fix passes include_variants: false explicitly in the stock item selector's filters so it never shows options the backend won't allow.

## What are the relevant issue numbers?

Closes [#12232](https://github.com/inventree/InvenTree/issues/12232)

## Does this PR meet the acceptance criteria?

- [x] Tests added for new/changed behavior
- [x] Follows project style guide
- [x] No breaking changes introduced


siddhartha928:
Hi @SchrodingersGat , this is my first contribution to this project. I've fixed [#12232](https://github.com/inventree/InvenTree/issues/12232) by making the "Install Stock Item" dialog's stock selector respect the same BOM/variant restrictions the part selector already enforces, so it no longer offers variant stock items that the backend would reject at submit time. Would appreciate a review when you have time!

SchrodingersGat:
@siddhartha928 thanks for the contribution

PR #12597, opened 2026-08-09
Add shareable previous/next detail navigation

### LINKED ISSUE #12397: [FR] Add Next and Previous Icon to switch easily between 2 objects

FurinaDog:
## Summary
- carry the source table endpoint, filters, ordering, and absolute offset in the detail URL
- resolve adjacent records against the same filtered and ordered API list, so copied URLs work in a new browser and across pagination
- preserve the URL context for preview-to-detail links and panel changes
- keep generic table users, including plugins, free of page-specific state

## Safety and edge cases
- accepts only relative `/api/` list endpoints
- hides controls if the current item can no longer be located after the list changes
- keeps navigation-only query parameters out of table API filters

## Testing
- `yarn build`
- `npx playwright test tests/pages/pui_part.spec.ts --list`
- added a Playwright regression test for filtered navigation, URL persistence, and reload

Closes #12397


FurinaDog:
Closing this PR because the latest CI run for b69ccdfe8 still fails in unrelated existing frontend tests (sales/build pages timing out while table-select-filters remains disabled). The failures do not exercise the detail-navigation changes, and I am not expanding this bounty PR into unrelated test-infrastructure changes.

PR #12621, opened 2026-08-11
[bug] Handle duplicate email addresses for magic links

SchrodingersGat:
If multiple users exist with the same email, do not generate magic link emails

### DIFF
--- a/src/backend/InvenTree/InvenTree/magic_login.py
+++ b/src/backend/InvenTree/InvenTree/magic_login.py
@@ -69,6 +69,9 @@ class GetSimpleLoginView(GenericAPIView):
             return User.objects.get(email=email)
         except User.DoesNotExist:
             return None
+        except User.MultipleObjectsReturned:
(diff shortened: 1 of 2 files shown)


src/backend/InvenTree/InvenTree/magic_login.py
```diff
@@ -69,6 +69,9 @@ def get_user(self, email):
             return User.objects.get(email=email)
         except User.DoesNotExist:
             return None
+        except User.MultipleObjectsReturned:
+            logger.warning('Multiple users found with email: %s', email)
```
matmair, line 73: Imo this should be a raised error, not just a log

SchrodingersGat: Raise to the user, or log error to the DB?

matmair: i think it woul be prudent to raise it to the user

PR #12623, opened 2026-08-11
refactor(backend): change fail limit to match allauth defaults more closely

matmair:
hopefully this will be mreo successfull in fixin the flakeiness


matmair:
this does not flip in my local tests but we will see how it works on GH actions

PR #12631, opened 2026-08-12
(feat): currency exchange

Saman-Zand-H:


docs/docs/plugins/builtin/iranian_currency_exchange.md
```diff
@@ -0,0 +1,86 @@
...
+title: Iranian Currency Exchange
+---
+
+## Iranian Currency Exchange
+
+The **Iranian Currency Exchange** plugin provides a USD to IRR exchange rate.
+It uses a manually configured rate by default and can optionally fetch the
+free-market USD rate from [TGJU](https://www.tgju.org/currency).
+
+The plugin uses InvenTree's existing Django-Q2 scheduler. No additional Celery
```
matmair, line 11: Why is celery even mentioned? We do not use that anywhere

docs/docs/plugins/builtin/iranian_currency_exchange.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Manual rate
+
+Leave **Enable TGJU USD rate consumer** disabled and enter **Manual USD to IRR
+rate** as the number of Iranian rials per one US dollar. The plugin then makes
+no TGJU requests, so rate updates work without an Internet connection. After
+changing the manual rate, open Currency Management and select **Refresh
+exchange rates** to apply it immediately.
+
+### TGJU consumer
```
matmair, line 41: This is too much technical not; if someone is interested in this level they can just look at the source

src/backend/InvenTree/plugin/builtin/integration/iranian_currency_exchange.py
matmair, line 1: This should be shipped as a separate plugin; I do not think this has broad enough use case to be shipped to everyone

matmair (changes requested):
While this looks sound I am not too eager to ship single currency backends via core; this is a great candidate for a plugin IMO

docs/docs/plugins/builtin/iranian_currency_exchange.md
Saman-Zand-H: actually i just forked it and wanted to do these changes for  myself i didn't mean to add it to the main repo as a PR. i'll close it.

SchrodingersGat:
IRR is available from frankfurter which is our default exchange. Have you not been able to get this to work out of the box?

Ref: https://frankfurter.dev/currencies/irr/

PR #12648, opened 2026-08-17
Add bulk stock creation from the barcode scan page

mattstyles333:
## Description

The Scan page currently lists barcodes and has a stub action. This fills that in for parts: select scanned parts, then create stock items with a shared quantity and location.

The same part can be scanned more than once (each scan is a package). Other model types still de-duplicate.

## Checklist

- [x] Pull request is targeted against `master`
- [x] One feature (frontend only; no API / schema change)
- [x] User-facing strings go through Lingui (`t` / `Trans`)
- [x] Playwright coverage in existing `pui_scan.spec.ts`
- [x] User docs on the barcode scan page
- [x] No Spex / GitLab / deploy-specific changes

## Test plan

- [ ] Scan two different parts, select both, Create Stock Items, confirm quantity/location apply to both
- [ ] Scan the same part twice, confirm both rows appear (no "already scanned")
- [ ] Scan a stock item twice, confirm duplicate is still blocked
- [ ] User without stock add permission: create button disabled
- [ ] Playwright: `pui_scan.spec.ts` (duplicate parts + create-stock modal)

### DIFF
--- a/src/frontend/src/pages/Index/Scan.tsx
+++ b/src/frontend/src/pages/Index/Scan.tsx
@@ -1,7 +1,6 @@
 import { t } from '@lingui/core/macro';
 import { Trans } from '@lingui/react/macro';
 import {
-  ActionIcon,
   Alert,
   Divider,
   Grid,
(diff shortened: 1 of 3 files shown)


src/frontend/src/pages/Index/Scan.tsx
```diff
@@ -160,7 +245,9 @@ export default function Scan() {
           <Trans>Selected elements are not known</Trans>
         </Group>
       );
-    } else if (uniqueObjectTypes.size > 1) {
```
matmair, line 163: Why is this changed?

src/frontend/src/pages/Index/Scan.tsx
```diff
@@ -110,7 +119,6 @@ export default function Scan() {
             }
           }
 
-          // If no match is found, add an empty result
```
matmair, line 113: Why is this removed?

PR #12655, opened 2026-08-18
Add previous/next navigation to detail pages

### LINKED ISSUE #12397: [FR] Add Next and Previous Icon to switch easily between 2 objects

Lescure1004:
## What

Adds previous/next navigation to detail pages so a user can move between
the filtered records they came from, without having to return to the list.

Closes #12397

## How

- A small zustand store (`TableNavigationState`) records the ordered list of
  primary keys and the current record when a table row is clicked.
- `PageDetail` reads that context and renders previous/next buttons in the
  header, disabled at the first/last record, navigating via `getDetailUrl()`.
- When no list context exists (e.g. opening a detail URL directly), the
  buttons are not rendered.

## Checklist

- [x] Frontend change only
- [x] i18n keys added (`Previous`, `Next`)
- [x] Playwright test added

### DIFF
--- a/src/frontend/src/components/nav/PageDetail.tsx
+++ b/src/frontend/src/components/nav/PageDetail.tsx
@@ -1,13 +1,19 @@
-import { Group, Paper, Space, Stack, Text } from '@mantine/core';
+import { ActionIcon, Group, Paper, Space, Stack, Text } from '@mantine/core';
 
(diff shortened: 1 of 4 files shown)


matmair:

PR #12317, opened 2026-07-06
refactor PurchaseOrder APIs to viewset

matmair:
This is the first major functional refactor of a complex API tree to modelviewsets; In the long run it would make sense to generally switch to this pattern in most places:
- it requires less knowledge of api endpoint naming and building practices
- it is even more DRY
- if used in plugins we can move to more "kitchensink" generic modelviewsets that detect usage of features and append the correct api paths (where necessary; we are slowly moving to generic endpoints for things like attachments, notes, images anyway)

PR #12668, opened 2026-08-19
feat(frontend): Add filter navigation remove button

matmair:
Follow up to https://github.com/inventree/InvenTree/pull/12585/

This adds an addition button to the page navigation group to remove the filters - providing a fast (and error-safe) way to remove navigation parameters from the current view for cleaner URLs for sharing.
<details><summary>old design</summary>
<p>

grafik
</p>
</details> 
new design
grafik

Includes fix for po tag filtering bug in https://github.com/inventree/InvenTree/pull/12317

### DIFF
--- a/docs/docs/concepts/ui/index.md
+++ b/docs/docs/concepts/ui/index.md
@@ -92,6 +92,9 @@ The navigation context is stored in the detail URL. This makes links shareable a
 
 The navigation controls also work with detail pages provided by frontend plugins that use the standard table and page components.
 
+!!! tip
+    This behaviour can be controlled by users with the setting `USE_TABLE_NAVIGATION`.
+
 ### Navigation Tree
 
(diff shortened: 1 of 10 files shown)


SchrodingersGat:
This is a great idea. I think that the button should be on the right, so that the position does not shift when the test length changes.

matmair:
I am a bit worried that users might accidentally hit the clear button while trying to use the navigation buttons; should we add a separator or something to space things out or do you not see that as a problem?

SchrodingersGat:
A small spacer / separator would be sufficient I think

matmair:
@SchrodingersGat I have updated the design

SchrodingersGat:
@matmair frontend tests are proving to be super flaky here

matmair:
There is a bug introduced in the api refactor for POs; specifically the tags filter; working on a fix

matmair:
@SchrodingersGat I have found a fix; do you want to merge this with the fix and the new feature or should I split it out into 2?

SchrodingersGat:
Happy to merge these together

SchrodingersGat:
Thanks @matmair

PR #12077, opened 2026-06-03
[feature] tags support

SchrodingersGat:
Brings 'proper' support for tags

### Features

- Filter models by tags, via the API
- Exposes API endpoint with all available tags
- Edit tags directly via the UI
- Display tags on detail pages
- Filter tables by tags
- Add "tags" support for multiple new model types

### Tasks

- [x] Implementation
- [x] Back-end unit tests
- [x] Migrations
- [x] Add tags to demo dataset - https://github.com/inventree/demo-dataset/pull/118
- [x] Playwright testing
- [x] Documentation
- [x] CHANGELOG update
- [x] API version update

PR #12672, opened 2026-08-20
feat(backend): rename filter

matmair:
see https://github.com/inventree/InvenTree/pull/12077#issuecomment-5360661213
Option 1: rename the added filter

PR #12673, opened 2026-08-20
feat(backend): rename inclusion parameter

matmair:
see https://github.com/inventree/InvenTree/pull/12077#issuecomment-5360661213
Option 2: rename the inclusion parameter (that is much older)


matmair:
replaced by https://github.com/inventree/InvenTree/pull/12672

PR #12678, opened 2026-08-21
Fix 500 errors for transitions

SchrodingersGat:
Order transitions attempted with an invalid PK currently throw a `500` instead of `404`.

### DIFF
--- a/src/backend/InvenTree/order/api.py
+++ b/src/backend/InvenTree/order/api.py
@@ -444,16 +444,33 @@ class PurchaseOrderViewSet(
         queryset = serializers.PurchaseOrderSerializer.annotate_queryset(queryset)
         return queryset
 
+    def get_order(self):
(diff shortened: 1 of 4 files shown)


src/backend/InvenTree/order/api.py
```diff
@@ -534,6 +563,8 @@ def issue(self, request, pk=None):
     )
     def receive(self, request, pk=None):
         """API endpoint to receive stock items against a PurchaseOrder."""
+        self.get_order()
```
matmair, line 566: Imo this check should run in @action and not be implemented individually on each action as that adds an inapparent failure point; should I PR something for this?

SchrodingersGat: @matmair yes that seems like a pretty reasonable follow-up to this

matmair: included in https://github.com/inventree/InvenTree/pull/12693

PR #12682, opened 2026-08-22
[Feature] Non-Conformance

### LINKED ISSUE #4860: [Feature] Build Order Non Conformance
After adding the ability to scrap a build output I'd like to look into the possibility of tracking "non conformances" stemming from build orders. Somewhat like an RMA but created from failed items out of a build order. Open non conformances can inform production staff about potential issues and can be addressed as part of standard production processes

### References

- https://github.com/inventree/InvenTree/pull/4846
- https://github.com/inventree/InvenTree/pull/4800
- https://github.com/inventree/InvenTree/issues/4826

SchrodingersGat:
- Implements "non conformance" functionality
- Closes https://github.com/inventree/InvenTree/issues/4860


matmair:
Very interesting for bettering MES area of the platform.
As this is an advanced concept that might not be enabled on most sites, I think this would be a great candidate for a built-in plugin as there is not much interaction with the rest of the platform. This could be used to detect and discuss implementation patterns that are non ergonomic (the way FSM transitions are exposed through APIs for example).

src/backend/InvenTree/build/api.py
```diff
@@ -1046,6 +1058,233 @@ class BuildCancel(BuildOrderContextMixin, CreateAPI):
...
+
+
+class NCRListOutputOptions(OutputConfiguration):
+    """Output options for the NCRList endpoint."""
+
+    OPTIONS = [InvenTreeOutputOption('part_detail', default=True)]
+
+
+class NCRList(NCRMixin, OutputOptionsMixin, ListCreateAPI):
+    """API endpoint for accessing a list of NonConformance (NCR) objects.
```
matmair, line 1147: implementing this via a viewset might be more efficient (see https://github.com/inventree/InvenTree/pull/12317)

matmair: see also https://github.com/inventree/InvenTree/pull/12693

PR #12689, opened 2026-08-23
User permissions check for Attachment API

SchrodingersGat:
- Check user role permissions when retrieving Attachment objects via the API
- Hide entries that the user does not have appropriate `view` role for
- Return `PermissionDenied` if the user tries GET access


matmair (approved):
At this point we should probably make a custom viweset for these generic classes to ensure we do not have to patch out the same security converters with every single one

SchrodingersGat:
@matmair a common viewset for these common models would be a great idea

PR #12693, opened 2026-08-23
feat(backend): Add generic API helper for transitions

matmair:
This PR adds a generic viewset mixin that enables writing much more concise APIs when using FSM transitions - follow up to https://github.com/inventree/InvenTree/pull/12317


matmair:
@SchrodingersGat @wolflu05 what do you think about this approach? I use it in some of my plugins, it was partially implemented automatically by Claude due to the high level of duplication in a model with a lot of transitions

matmair:

src/backend/InvenTree/order/api.py
```diff
@@ -405,6 +406,17 @@ class PurchaseOrderViewSet(
         'supplier', 'created_by'
     )
     serializer_class = serializers.PurchaseOrderSerializer
+    # TODO @matmair remove legacy return codes
```
SchrodingersGat, line 409: @matmair are you going to update this as part of this PR?

SchrodingersGat (commented):
Can you also apply these helpers to the build model API endpoints?

src/backend/InvenTree/order/api.py
matmair: No, this would be breaking and should only be done with 2.0

matmair:
> Can you also apply these helpers to the build model API endpoints?

this new generic approach only applies to DRF viewsets, not GenericViews and the build APIs are not on viewsets yet so that would expand the scope of the PR quite a bit

matmair:
i just found a potential information leak in this approach; must review options

matmair:
@SchrodingersGat the info leak in the transitions table is resolved now; this is ready for review

src/backend/InvenTree/generic/states/api_helpers.py
```diff
@@ -0,0 +1,411 @@
...
+            viewset actually mixes in ``OutputOptionsMixin`` - passing it to a plain ``as_view()``
+            that has no such attribute raises ``TypeError``.
+
+    Returns:
+        A DRF ``@action``-decorated method, ready to assign as a viewset class attribute.
+    """
+    attr_name = name or method_name
+    segment = url_path or attr_name.replace('_', '-')
+    # TODO @matmair move all transition actions under the common prefix
+    # path = f'{TRANSITION_URL_PREFIX}/{segment}'
```
SchrodingersGat, line 130: can this be removed now?

src/backend/InvenTree/generic/states/api_helpers.py
```diff
@@ -0,0 +1,411 @@
...
+        serializer_class: Serializer to use for the response body. Defaults to an empty serializer
+        reset_output_options: Force ``output_options=None`` for this action. Only valid when the
+            viewset actually mixes in ``OutputOptionsMixin`` - passing it to a plain ``as_view()``
+            that has no such attribute raises ``TypeError``.
+
+    Returns:
+        A DRF ``@action``-decorated method, ready to assign as a viewset class attribute.
+    """
+    attr_name = name or method_name
+    segment = url_path or attr_name.replace('_', '-')
```
SchrodingersGat, line 128: Are you going to implement this `TODO`?

src/backend/InvenTree/generic/states/api_helpers.py
```diff
@@ -0,0 +1,411 @@
...
+        """
+        instance = self.get_object()
+        # permission check
+        model = instance._meta.model
+        if not (
+            check_user_permission(request.user, model, 'view')
+            or check_user_permission(request.user, model, 'change')
+        ):
+            raise PermissionDenied(
+                _('User does not have permission to edit this attachment')
```
SchrodingersGat, line 382: "attachment" seems incorrect here - should be generic "item" or similar?

PR #12698, opened 2026-08-24
[FR] add well-known mixin to let plugins define well-known urls

### LINKED ISSUE #11741: [FR] add well-known mixin to let plugins define well-known data entry points

### Problem statement

[well-known URIs](https://en.wikipedia.org/wiki/Well-known_URI) are used as discovery mechanisms in various contexts. These often depend on being on the root path of an application which makes them harder to use as users must not enter the url of their instance but of the url-root of the plugin they are using - which they might not be aware of.

### Suggested solution

Add a plugin mixin that allows the definition of well-known names and to what URL they should redirect or which json data to provide

adding entries to individual plugins also works in practice but i like to have shared logic in a shared place

### Examples of other systems

_No response_

- [ ] I want to develop this.

matmair:
Fixes https://github.com/inventree/InvenTree/issues/11741
Adds a way for plugins to add .well-known entry points to the instance; helpful for A&A, system integrations and auto-discovery. Full docs provided with more insights.

### DIFF
--- a/docs/docs/plugins/develop.md
+++ b/docs/docs/plugins/develop.md
@@ -145,6 +145,7 @@ Supported mixin classes are:
 | [UserInterfaceMixin](./mixins/ui.md) | Add custom user interface features |
 | [UrlsMixin](./mixins/urls.md) | Respond to custom URL endpoints |
 | [ValidationMixin](./mixins/validation.md) | Provide custom validation of database models |
+| [WellKnownMixin](./mixins/wellknown.md) | Provide well-known endpoints on the root of an instance |
 
 ## Plugin Concepts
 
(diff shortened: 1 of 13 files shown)


matmair:

SchrodingersGat:
As the well-known endpoints are not going to be changing frequently - i.e. only when the plugin registry hash changes, can we cache the `wellknownindexview` endpoint - rather than regenerating the list of endpoints on each request?

We could build this cache whenever the plugin registry is reloaded, or first time the API endpoints is requested, and this would reduce the DB hits on subsequent requests.

matmair:
No, for some discovery use cases the response needs to change depending on data in the request (most often headers)

matmair:
We could cache the plugin list but each needs to be called on each index request

SchrodingersGat (approved):

SchrodingersGat:
@matmair ok that's fair enough. Nice implementation overall

PR #12732, opened 2026-08-28
Fix media file access for single-machine deployment without reverse proxy

chinaglowing:
## Problem

In single-machine deployments without a reverse proxy (e.g., Windows standalone with SQLite), media files (exported reports, labels, etc.) return 404 errors. Additionally, media file requests are blocked by the security middleware.

### `middleware.py` (+3 lines)
Add `MEDIA_URL` to `paths_own_security` - media files are served with DRF token / session auth in single-machine mode, so they need to bypass the default auth middleware redirect.

### `urls.py` (+14 lines, -4 lines)
Serve media files via `django.views.static.serve` **outside** `DEBUG` mode. Previously, media file routing was only configured when `DEBUG = True`, which means production/single-machine deployments without a reverse proxy had no way to access media files.

The new `re_path` serves media files at `/media/<path>` using Django's built-in static serve view, with authentication handled by `AuthRequiredMiddleware` + DRF token/session auth.

## Testing
- Verified on Windows single-machine deployment (SQLite, port 17456, no reverse proxy)
- Report/label export downloads work correctly (no more 404)
- Auth still enforced: unauthenticated requests are redirected to login

## Notes
- Does not affect deployments with reverse proxy (nginx/caddy) - existing config takes precedence
- No translation changes (translations are handled via Crowdin per project guidelines)


src/backend/InvenTree/InvenTree/urls.py
matmair, line 1: I don't know where you think you added authorisation checks for media files. I don't see it and there are no checks so this is a hard pass without proper docs and tests

matmair (changes requested):

SchrodingersGat:
This approach conflicts with a standard django requirement - static and media files are not served via django in production.

PR #12738, opened 2026-08-29
Fix order allocation variant validation logic and serializer read_only_fields

bhumin18:
### Summary
This PR fixes a data validation logic bug in order allocation models and cleans up serializer `read_only_fields` definitions:

1. **`order/models.py` (`SalesOrderAllocation.clean()` & `TransferOrderAllocation.clean()`):**
   - Fixed the variant check from `self.line.part not in variants` to `self.item.part not in variants`.
   - **Why this matters:** Because `variants = self.line.part.get_descendants(include_self=True)`, checking `self.line.part not in variants` was always `False`, which allowed allocating completely unrelated stock items to order line items without triggering the validation error. Now, only the required part or its valid descendant variants can be allocated.

2. **`order/serializers.py` (`SalesOrderAllocationSerializer` & `TransferOrderAllocationSerializer`):**
   - Removed the erroneous empty string `''` from `read_only_fields = ['line']`.

3. **`order/test_api.py` (`OrderAllocationValidationTest`):**
   - Added unit tests verifying that allocating unrelated stock items raises `ValidationError`, while allocating valid variant stock items succeeds.


matmair:
There are a lot of style changed here that make no sense; make sure to read the contributions guide - especially regarding formatting https://docs.inventree.org/en/stable/develop/contributing/

SchrodingersGat (approved):

SchrodingersGat:
@bhumin18 thank you for cleaning up the PR, much easier to review it now.

This is a good catch, thanks for submitting a fix for this.

PR #12747, opened 2026-08-30
chore: add types to templates; add AI agent catch

matmair:
This adds:
- types (better for API filtering than labels) by default to the FR/bug templates
- clarification to the template that lead at least Codex and Claude to stop automatically filing length bug reports that do not follow the template

Next step would be an AI text detection bot that labels fully AI reports automatically but lets see first if this helps ppl fill out the full template.


SchrodingersGat (approved):

SchrodingersGat:
Thanks @matmair

PR #12760, opened 2026-09-01
Run docker container as non-root user

### LINKED ISSUE #6589: [Docker] Container permissions

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/6589


matmair:
Looks good in general; have you tested this starting up on an existing setup?

SchrodingersGat:
> Looks good in general; have you tested this starting up on an existing setup?

Not yet. Still have to perform testing

PR #12763, opened 2026-09-02
Fix permissions for SelectionList

SchrodingersGat:
SelectionList access should be gated on staff role permissions.

### DIFF
--- a/src/backend/InvenTree/common/test_api.py
+++ b/src/backend/InvenTree/common/test_api.py
@@ -185,6 +185,38 @@ class ParameterAPITests(InvenTreeAPITestCase):
         self.assertIsNone(response.data['model_type'])
         self.assertFalse(response.data['enabled'])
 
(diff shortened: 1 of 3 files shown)


src/backend/InvenTree/common/test_api.py
```diff
@@ -2333,6 +2365,103 @@ def test_patch_list_without_choices_preserves_entries(self):
         self.assertTrue(SelectionListEntry.objects.filter(pk=self.entry.pk).exists())
 
 
+class SelectionListStaffPermissionAPITests(InvenTreeAPITestCase):
```
matmair, line 2368: Do we need a whole test class to test a widely used permission class?

matmair (approved):

matmair:
I would consider this breaking as it changes current api behavior. A backport would be uncommon

PR #12766, opened 2026-09-02
[bug] API permission fixes:

SchrodingersGat:
- User can only change their own profile
- Users can only access notes against models they are scoped to
- Users can only access parameters against models they are scoped to


SchrodingersGat:
@matmair another one for review

matmair (approved):

PR #12770, opened 2026-09-02
Prevent delete operation on /api/user/me/

SchrodingersGat:
`UserMe` endpoint inherits from the `UserDetail` endpoint - this fix explicitly removes support for the 'delete' operation


matmair:
This will need an api version bump and thus should not be backported

matmair (approved):

Issue #12769, opened 2026-09-02
collect_plugins_static_files() has no locking, so concurrent django-q processes corrupt plugin static output


Kamaar:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

## Related, and why this is not a duplicate

- **#12130** (closed as *not planned*) reports the same symptom. Its diagnosis, in a comment,
  is that a race occurs because `invoke plugins` was being run *externally* at the same time as
  the server started. That is one way to trigger it, but the concurrency is internal and needs
  no external trigger: see below. Nothing was changed, so the behaviour is still present.
- **#7709** (closed as *completed*, via #8502) is the older umbrella issue for plugin static
  files. It fixed the "files never collected" symptom; it did not make the collector safe to run
  from several processes at once, which is what this is about.

## Description

`install_plugins_file()` calls `plugin.staticfiles.collect_plugins_static_files()` with no
synchronisation, and the guard that is supposed to stop it running repeatedly is process-local:

```python
# plugin/registry.py
def install_plugin_file(self):
    file_hash = plugins_file_hash()
    if file_hash != settings.PLUGIN_FILE_HASH:
        install_plugins_file()
        settings.PLUGIN_FILE_HASH = file_hash
```

`settings.PLUGIN_FILE_HASH` starts as `''` (set in `settings.py`) and is only ever assigned in
memory, so **every process** sees a mismatch on **every start** and runs the installer. With a
server plus django-q workers, that is N concurrent `pip install` runs and N concurrent static
collections per restart — not once per plugins-file change, as the guard implies.

`registry.set_ready()` reaches the same call for any process that is either the main thread or a
worker thread:

```python
if InvenTreeSetting.get_setting('PLUGIN_ON_STARTUP', create=False, cache=False):
    ...
    elif InvenTree.ready.isInWorkerThread() or InvenTree.ready.isInMainThread():
        registry.install_plugin_file()
```

So no external `invoke plugins` is needed to collide with startup — the processes collide with
each other. The tracebacks below come from `django_q/worker.py` in some cases and
`django_q/monitor.py` in others, in a single restart.

`copy_plugin_static_files()` clears the destination directory and *then* copies into it, so two
processes interleaving those phases interfere with each other in three distinct ways:

1. `clear_static_dir()` → `staticfiles_storage.delete(path)` → `os.rmdir` fails, because another
   process has already begun writing files into the directory being removed.
[...]

PR #12776, opened 2026-09-03
Plugin load fix

SchrodingersGat:
Addresses concurrency issues related to installing plugins from file, as identified in https://github.com/inventree/InvenTree/issues/12769

Additionally ensures that static file installation (via plugins) is now concurrency safe - preventing partial file updates or file corruption.


Kamaar:
Reviewed the diff — this looks right, and it addresses the root cause rather than the symptom. Confirmed there are no `settings.PLUGIN_FILE_HASH` references left on the branch.

One thing I would raise: `_claim_hash_setting()` writes the new hash *before* the work is done, and `_release_hash_claim()` reverts it on failure. The `finally` covers exceptions, but not the process being killed outright — a container stopped mid-install, or the OOM killer. In that case `_PLUGIN_FILE_HASH` is left at the new value with the install never having run, and once the lease expires every other process sees `current == new_value` and skips. Plugins would then stay uninstalled until the plugins file changes again.

That is a narrow regression against today's behaviour, where every process retries on every start. Writing the hash only *after* a successful install would avoid it: the lease alone is enough to stop concurrent runs, since a process that cannot acquire it skips anyway, and a kill would simply leave the hash unchanged so the next start retries.

Minor: `staticfiles.py` now imports `_acquire_lease_blocking` and `_release_lease` from `plugin.registry`. Since they have become cross-module API, they might be worth promoting out of the underscore namespace, or into a small shared module.

I have an instance that reproduces the original bug reliably (Docker, server plus worker, a plugin shipping 29 static files). Happy to run this branch there and report back — it would take re-enabling `PLUGIN_ON_STARTUP`, which is currently off as the workaround.

Kamaar:
Tested this branch on the instance where the original bug reproduces.

**Method.** I could not run master wholesale on a 1.5.2 deployment — master's `registry.py` imports `get_wellknown_urls` from `plugin/urls.py`, which does not exist in 1.5.2 — so I applied only this PR's hunks for `plugin/registry.py` and `plugin/staticfiles.py` onto the 1.5.2 files. They applied cleanly, with line offsets only, and `staticfiles.py` is byte-identical between 1.5.2 and master before this PR. The patched files went into **both** the server and the worker container, since the collision between those two processes is the thing being tested. `PLUGIN_ON_STARTUP` was re-enabled, so the code path under test is live.

**Result.** Five consecutive `docker compose restart inventree-server inventree-worker` cycles:

```
run 1: http=200  files=15  staticfiles errors=0
run 2: http=200  files=15  staticfiles errors=0
run 3: http=200  files=15  staticfiles errors=0
run 4: http=200  files=15  staticfiles errors=0
run 5: http=200  files=15  staticfiles errors=0
```

Verified independently through the API afterwards: all 12 files listed in the plugin's `.vite/manifest.json` are served with 200, the panel bundle imports and exports the expected symbol, and the plugin's own endpoints respond. Before this branch, the same instance lost the static directory reliably across restarts.

Caveat worth stating: a race is probabilistic, so five clean restarts is a good signal rather than a proof. But the previous behaviour was consistent enough that five failures would have been the expected outcome without the fix.

**Unrelated finding, while getting the instance into a testable state.** My `plugins.txt` had accumulated three entries for the same package — `@v2.0.0`, `@v2.0.1` and `@main` — which made `pip install -r plugins.txt` fail with `ResolutionImpossible`, so no plugin was installed at all. The cause is in `update_plugins_file()`:

```python
def compare_line(line: str):
    return re.match(rf'^{package_reference}[\s=@]', line.strip())
```

For a VCS install `package_reference` is `name@git+url@tag`, so installing the same plugin at a different tag does not match the existing line and gets appended instead of replacing it. `package_reference` is also interpolated into the regex unescaped, and it contains `+`, `.` and `/` — in `git+https` the `+` becomes a quantifier — so the comparison is unreliable even for an identical reference. Happy to open that as its own issue if useful; it seems distinct from this one.

matmair:
@Kamaar stop posting verbatim AI chat slop into discussion with other humans; these ultra long texts that wonder of in unrelated issues waste everyone’s time

matmair (approved):

matmair:
@SchrodingersGat once this is merged I would love to have a point release (I have prepped https://github.com/inventree/InvenTree/pull/12795) as this could explain some hard-to-reproduce deployment issues

matmair:
question: how does one auqire a lock if there is no database (during initial setup)

PR #12784, opened 2026-09-04
Improve barcode scanning with native BarcodeDetector

BI7KHI:
Use the browser-native BarcodeDetector (Shape Detection API) for camera barcode scanning when available, falling back to html5-qrcode. Native decoding is hardware-accelerated and far more robust to low light, blur and small codes than the pure-JS software decoder.

Also:
- Prefer the rear camera and higher resolution for better recognition
- Add a flashlight (torch) toggle for low-light scanning
- Full-frame detection (no aiming box) with multi-code support
- Relax camera constraints (ideal instead of exact)
- Fix a bug where the native <video> element was not rendered before scan

demo


src/frontend/src/components/barcodes/BarcodeCameraInput.tsx
```diff
@@ -57,45 +155,150 @@ export default function BarcodeCameraInput({
...
+    const target = device ?? camId;
+    const scanner = legacyScannerRef.current;
+    if (target && scanner && !scanningEnabled) {
+      scanner
         .start(
-          camId.id,
-          { fps: 10, qrbox: { width: 250, height: 250 } },
+          target.id,
+          {
+            fps: 15,
```
matmair: Why is this changed?

src/frontend/src/components/barcodes/BarcodeCameraInput.tsx
```diff
@@ -57,45 +155,150 @@ export default function BarcodeCameraInput({
...
+            qrbox: { width: 300, height: 300 },
+            aspectRatio: 1.0,
+            videoConstraints: { facingMode: 'environment' }
```
matmair: Why is this added? Won't this disable using the selfie camera on a large pickstation?

src/frontend/src/components/barcodes/BarcodeCameraInput.tsx
BI7KHI: I reverted this change in commit `0f93dee`. Higher FPS gives better scan performance at the cost of increased device CPU overhead, so the reviewer's concern is valid.

src/frontend/src/components/barcodes/BarcodeCameraInput.tsx
BI7KHI: Good catch — the `videoConstraints.facingMode` override was too aggressive.

I removed the facingMode override from both the native and the html5-qrcode fallback paths `f0ba1aa`. The camera selector now always drives which device is used, while the auto-selection logic still prefers the rear camera by device label when no camera has been saved yet.

SchrodingersGat (approved):

PR #12830, opened 2026-09-09
Squash migrations

### LINKED ISSUE #5015: Squash migrations for better performance

SchrodingersGat:
- Closes https://github.com/inventree/InvenTree/issues/5015
- Squashes migrations prior to 1.0.0

### Tasks

- [x] Get CI passing
- [x] Create a 1.0.0 database artifact to migrate *from* during CI
- [x] Add documentation around migration procedure across 1.0.0 release boundary


SchrodingersGat:
@matmair lmk what you think of this approach. Everything after 1.0.0 is left in place

src/backend/InvenTree/company/migrations/0003_remove_supplierpart_minimum_squashed_0047_supplierpart_pack_size.py
```diff
@@ -0,0 +1,407 @@
...
+from django.db import migrations, models
+
+
+# Functions from the following migrations need manual copying.
+# Move them and any dependencies into this file, then update the
+# RunPython operations to refer to the local versions:
+# company.migrations.0019_auto_20200413_0642
+# company.migrations.0024_unique_name_email_constraint
+# company.migrations.0026_auto_20201110_1011
+# company.migrations.0036_supplierpart_update_2
```
matmair: was this addressed?

src/backend/InvenTree/part/migrations/0003_auto_20190525_2226_squashed_0060_merge_20201112_1722.py
```diff
@@ -0,0 +1,454 @@
...
+import stdimage.models
+from django.conf import settings
+from django.db import migrations, models
+
+
+# Functions from the following migrations need manual copying.
+# Move them and any dependencies into this file, then update the
+# RunPython operations to refer to the local versions:
+# part.migrations.0034_auto_20200404_1238
+# part.migrations.0056_auto_20201110_1125
```
matmair: was this addressed?

matmair (approved):
looks good in general; great to get this addressed;
I might have overlooked it; is there a specific error message / communication that ensures the user is aware of this restriction?

Would be great to get a dedicated Error Code too.

src/backend/InvenTree/company/migrations/0003_remove_supplierpart_minimum_squashed_0047_supplierpart_pack_size.py
SchrodingersGat: Fixed now

src/backend/InvenTree/part/migrations/0003_auto_20190525_2226_squashed_0060_merge_20201112_1722.py
SchrodingersGat: This has now been addressed

SchrodingersGat:
@matmair a specific error message has now been implemented, and is explicitly checked for in CI

SchrodingersGat:
image

Compared to master:

image

matmair:
Looks great! It is a large change but I think we have the major things that could go wrong covered.
Thank you for taking the time to address this, this is much cleaner then my attempts.

SchrodingersGat:
It's good to finally get this working. We can let it run on master for a while as we prepare for `1.6.0`

PR #12851, opened 2026-09-14
Suggestion: support typed note content (HTML, JSON, plain text)

ledvinap:
This proposes adding an immutable content type to notes and note templates: HTML, JSON, or plain text.
 
Notes have previously been used to store scripts results and JSON metadata, where preserving the original source matters. HTML only may break this workflow; this PR allows more note use-cases, with easy future extension

HTML remains the default rich-text format. JSON is validated without reformatting; JSON and plain
text are stored unchanged and displayed literally in the interface and reports. Existing notes remain HTML (it is possible to set type before migration)
 
Built-in content handlers provide a possible basis for additional formats later. Discussion is welcome on the approach, supported formats, and whether this belongs upstream.
 
The implementation was written by GPT-6-Astra, with the important parts verified by me. If this approach is considered suitable for merging, I’ll perform another verification pass.


matmair:
This reads like you are using a wrench as a hammer.

Structured data for usage other than human consumption should be put into metadata, custom plugin model data or test results if you need it stock specific. Notes are for user-rendered, human readable information.

src/backend/InvenTree/common/notes.py
```diff
@@ -0,0 +1,159 @@
...
+
+class RawNoteContent:
+    """Raw note content with literal display and no embedded-image support."""
+
+    supports_images = False
+
+    @staticmethod
+    def clean(content: str) -> str:
+        """Return the source text unchanged."""
+        return content
```
matmair, line 31: This is what we call a "stored XSS" https://community.owasp.org/attacks/xss/

src/backend/InvenTree/common/notes.py
```diff
@@ -0,0 +1,159 @@
...
+
+        try:
+            json.loads(content, parse_constant=reject_constant)
+        except RecursionError as exc:
+            raise ValidationError({'content': _('JSON is nested too deeply.')}) from exc
+        except ValueError as exc:
+            raise ValidationError({
+                'content': _('Invalid JSON: %(error)s') % {'error': str(exc)}
+            }) from exc
+        return content
```
matmair, line 60: Possible stored xss

src/frontend/src/locales/en/messages.po
matmair, line 1: This does not belong into PRs as per contribution guidelines

src/frontend/src/components/editors/NotesEditor.tsx
```diff
@@ -538,165 +566,199 @@ export default function NotesEditor({
...
+                      autosize
+                      styles={{ input: { fontFamily: 'monospace' } }}
+                      onChange={(event) => {
+                        setRawContent(event.currentTarget.value);
+                        setIsDirty(true);
+                        setSaveError(null);
+                      }}
+                    />
+                  ) : (
+                    <Text
```
matmair, line 590: And here is the sink for the xss

matmair (changes requested):

src/backend/InvenTree/common/notes.py
ledvinap: I'm not sure how to handle this better - `RawNoteContent` contains raw bytes by design - and consumers shall handle it accordingly. `.raw` of content handler selects rendering. Maybe it can be inverted, with 'html' / 'is_html' as property meaning that content needs rendering (and XSS validation).

Also there is nothing to validate for raw content, it is valid by  design. The inclusion was mainly to demonstrate new content_type + basis of json 

(fixed typos)

src/frontend/src/components/editors/NotesEditor.tsx
ledvinap: `{rawContext}` is string child here, React will escape it (`dangerouslySetInnerHTML` is necessary for xss here)

ledvinap:
> This reads like you are using a wrench as a hammer.

Yes, you are right, and in this case it a bit of overreaction, I can migrate to HTML notes quite easily. 
My main concern were test-result notes - I use test notes to attach stock-item configuration that is operator-configurable.

Metadata are  an option for my case, but plugin is necessary, which looks like overkill.

> Structured data for usage other than human consumption should be put into metadata, custom plugin model data or test results if you need it stock specific. Notes are for user-rendered, human readable information.

I need some middle-ground - human readable and human-editable, but consumable by scripts/plugins

src/backend/InvenTree/common/notes.py
ledvinap: `json` is meant to be raw; it may be prettified in future, but never included directly in HTML.

src/backend/InvenTree/common/notes.py
matmair: Before it goes into the database it hast to be validated and cleaned; we have multiple templating systems - you don't know where in the system this content will be consumed by an attacker

matmair:
This can not be merged in this state and with this data concept.
If you want to have a fundamental discussion about adding a new way of handling data a feature request is the correct path. We have been bitten before by merging stuff too fast and then having to deal with length vulnerability disclosures.

ledvinap:
> This can not be merged in this state and with this data concept. If you want to have a fundamental discussion about adding a new way of handling data a feature request is the correct path. We have been bitten before by merging stuff too fast and then having to deal with length vulnerability disclosures.

Yes, thanks for commenting PR problems. 
The idea was more about getting necessary code changes together / getting reference implementation so that there is solid base for discussion and full change footprint is visible.
These notes are not my priority now, I can easily apply workaround.

If it is decided that this PR is useful, I will finish it/improve it as necessary.

matmair:
@SchrodingersGat what is you opinion on this? I don't think we need another arbitrary data store system in addtion to metadata

SchrodingersGat:
Agreed, there is no need to pursue this further.

PR #12864, opened 2026-09-16
docs: fix local dev server port reference (8080 -> 8000)

### LINKED ISSUE #12501: Reference URL For Docs Server
Hi,

I've been getting Inventree set up for the past week or so and I've identified a few places for some improvements to the setup docs. While I'll handle all of that in other issues, I did see while compiling the docs for myself that the documentation setup guide has a reference to the docs being hosted at 'localhost:8080', while my mkdocs install automatically served the documents at '127.0.0.1:8000'.

I've had some issues with the install so it is possible that is an error from my installation, but I figured I'd check if other users saw this behaviour or if this is now intended behaviour. If this behaviour is seen by others, I'm happy to generate a pull request with an updated readme for the documentation setup guide.

Cheers.

520ceg520:
The 'mkdocs serve' command uses MkDocs' default port 8000, not 8080. The 8080 port is only used by the 'invoke dev.docs-server' task.

Fixes #12501


matmair:
8000 is the default port InvenTree's server uses so not sure this is great

SchrodingersGat:
The docs should just reference the `invoke` task I think

PR #12867, opened 2026-09-17
Preserve note image files when transactions roll back

ledvinap:
This is bugfix from #12851. AI-written as usual, verified.

**Description:**

When a transaction rolls back after removing an image from a note, deleting the note, or deleting its parent part, the database rows and original content are restored, but the image file has already been deleted.

Defer file deletion using `transaction.on_commit()` on the database connection supplied by the deletion signal. Committed deletions still remove files; rolled-back deletions preserve them.

Adds regression coverage for all three rollback paths and updates existing cleanup tests to execute commit callbacks. Independent of #12851.

**Validation:**

- All three rollback cases fail without the fix.
- All 18 focused image, cleanup-task, API, and permission tests pass in isolated SQLite.
- Ruff lint and format checks pass with `--preview`.
- `git diff --check` passes.

Reproduction uses model transactions; no specific UI trigger has been established.


SchrodingersGat (approved):

SchrodingersGat:
@ledvinap LGTM - was this discovered via usage?

ledvinap:
No. When working on #12851, gpt-6-astra review spotted it. 

Rationale behind it seems solid, here is possible trigger:

  1. Set STOCK_ALLOW_DELETE_SERIALIZED=False.
  2. Select an ordinary stock item with a note image and a serialized stock item.
  3. Send DELETE /api/stock/ with {"items": [ordinary_id, serialized_id]}, with the ordinary item processed first.
  4. The serialized item rejects deletion with HTTP 400: “Serialized stock items cannot be deleted”.

 The transaction restores both stock items, the note content, and the image record. Without the fix, the image file disappears; with the fix, it survives.

PR #12868, opened 2026-09-17
[bug] Fix lookup error in API

SchrodingersGat:
Fixes order dependency bug on the PurchaseOrder API viewset. This caused the `/api/po/status/` endpoint to return 404 - because of a duplicate "match" against the pk endpoint

By ensuring that the `pk` field is always integer based we avoid this bug completely. Also added regression test


SchrodingersGat:
Replaced by https://github.com/inventree/InvenTree/pull/12885

PR #12870, opened 2026-09-17
Ensure generated label PDFs have correct extension

ledvinap:
Label templates may intentionally use extensionless filename patterns — for example, templates shared with physical-printer plugins. 
When the built-in PDF plugin generates a downloadable file, append .pdf if missing  without changing the template pattern.
Preserve existing .pdf/.PDF suffixes and use labels.pdf when the filename is empty, None, or omitted. HTML debug output remains unchanged.

Adds regression coverage for suffixless, existing-suffix, empty, None, and omitted filenames.


SchrodingersGat (approved):

SchrodingersGat:
Does this fix need to be applied to reports also?

ledvinap:
> Does this fix need to be applied to reports also?

It is already handed for reports, but it can be slightly improved (no .PDF.pdf, report.pdf instead of .pdf for empty report_)name) (https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/report/models.py#L662)

PR #12871, opened 2026-09-17
Fix validation of dependent serializer fields

ledvinap:
AI generated, as usual:

`DependentField` validates its wrapper but calls only the child’s `to_internal_value()`. That skips the child’s own validators and serializer-level `validate()`. Calling `child.run_validation()` restores the complete validation process.

A concrete use is [machine printer options](https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/plugin/builtin/labels/inventree_machine.py#L180): the selected printer driver supplies a serializer dynamically. For example, a driver could reject an incompatible combination:

```python
def validate(self, attrs):
    if attrs['auto_cut'] and attrs['media'] == 'continuous':
        raise serializers.ValidationError(
            'Automatic cutting is unavailable for continuous media.'
        )
    return attrs
```

Both individual fields are valid, but their combination is rejected only by `validate()`. Before this fix, that check never runs and the API can dispatch the print task.

That combination is illustrative; the [actual regression test](https://github.com/ledvinap/InvenTree/blob/e4d6c8d99d987a607c3c81cc8be404cc1e3a2c98/src/backend/InvenTree/plugin/base/label/test_label_mixin.py#L275) supplies a rejecting driver serializer and demonstrates **HTTP 201 before the fix, HTTP 400 with no task dispatch afterward**.


SchrodingersGat (approved):

SchrodingersGat:
@ledvinap thanks for the fix

PR #12872, opened 2026-09-17
Preserve label printing failures and skip failed job redelivery

ledvinap:
AI, as usual.

Label printing exceptions currently delete the monitored DataOutput, losing the error details. 
An unavailable plugin leaves the job unfinished without an error, leaving the “Printing” popup stuck.

This change records both failures using mark_failure() while preserving exception propagation. Workers skip outputs with recorded errors, preventing redelivery from reprinting partially completed batches. Existing missing/completed-output guards remain intact.

The problem more pronounced with my  (unpublished yet) changes, where multiple checks may fail a print.
 
Adds regressions for printing exceptions, unavailable plugins, and redelivery. No frontend changes or migrations are required.

PR #12874, opened 2026-09-17
Preserve validation details when rendering labels

ledvinap:
Label rendering wrappers currently replace all exceptions with generic “Error rendering label to PDF/HTML” messages, losing useful validation details raised by upstream template tags or plugins.

Handle `ValidationError` separately in both wrappers and re-raise it unchanged. Preserve existing logging, including for validation failures, so diagnostics remain available after failed `DataOutput` records are deleted. Unexpected exceptions retain their existing generic messages.

This fix applies independently to master. Together with #12872, the specific validation details are also retained in failed results and logging can be removed.

Mocked tests cover validation-detail preservation, logging, and unchanged handling of unexpected exceptions in both wrappers. Focused tests, Ruff checks, and `git diff --check` pass; the full suite was not run.


matmair:
Have you tested and ensured that no internal information or secrets are leaked to unprivileged user through all possible error messages?

ledvinap:
> Have you tested and ensured that no internal information or secrets are leaked to unprivileged user through all possible error messages?

Fair point.

I did not consider ValidationError as dangerous in this respect. It is normally user-facing problem.
Also report code does return validation errors (https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/report/models.py#L584-L586, https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/report/models.py#L623-L625), this PR extends that behavior to label wrappers; but that does not prove it to be safe.

My personal/developer-centric opinion is that its value outweighs negatives.


AI analysis:

• Current built-in validation messages can include internal details and raw template values:
 
  - Arithmetic tags include both operands and their Python types through debug_vars() ([report/templatetags/report.py](https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/report/templatetags/report.py#L750)).
  - Currency tags include rejected amounts, multipliers, and currency codes.
  - File-path validation includes the supplied path.
  - Locale validation includes the locale and underlying exception text.
 
Other messages are fixed descriptions, such as missing QR-code data or an invalid model argument.
 
I haven’t identified an explicit credential disclosure in these inspected messages, but operands or paths could contain information the printing user should not see. ValidationError does not guarantee a sanitized message.

matmair:
1) follow the contributions document; if you setup your repo correctly with prek it is not possible to cause this CI failure
2) i am not really a fan of the added test runtimes, especially seeing that this is a pattern in your PRs (see https://github.com/inventree/InvenTree/pull/12872) but I will let @SchrodingersGat decide

ledvinap:
> 1. follow the contributions document; if you setup your repo correctly with prek it is not possible to cause this CI failure

Sorry, fixed now, using extra commit; should I add new commits or force-push in future?

> 2. i am not really a fan of the added test runtimes, especially seeing that this is a pattern in your PRs (see [Preserve label printing failures and skip failed job redelivery #12872](https://github.com/inventree/InvenTree/pull/12872)) but I will let @SchrodingersGat decide

I'll adapt as requested. GPT is generating the tests, I just verify that they make sense and do not test that 1+1 is still 2.  With AI tools, it is usually better to have more tests to avoid missing something.

In this concrete case the tests are pretty cheap, so more code to review is probably bigger problem. But I did not test other PRS yet, it may be worse.

---
AI:

Measured both added tests locally, including discovery, class/test setup and teardown, plugin construction, mocks, and assertions:

| Mode, 100 repetitions | Median | 95th percentile |
|---|---:|---:|
| Normal | **6.1 ms** | 6.8 ms |
| With project coverage settings | **15.3 ms** | 16.4 ms |

Running them as a standalone Django command took **4.57 seconds median** across five fresh processes, including startup and shutdown. That startup cost is already paid when they run inside the existing suite.

They require **no database setup, fixtures, or actual rendering**. Estimated incremental suite cost is therefore approximately **6–15 ms locally**.

SchrodingersGat (approved):

SchrodingersGat:
Happy with this overall, thanks for the implementation @ledvinap

PR #12878, opened 2026-09-18
feat: support password change well-known

matmair:
as per https://adamj.eu/tech/2026/09/16/django-change-password-url/

this enables password managers to directly direct users to the password change form


SchrodingersGat (approved):

SchrodingersGat:
Nice addition @matmair

PR #12880, opened 2026-09-18
Fix port mapping in docker-compose.yml

isosloth:
Fixed a port mapping issue in docker-compose.yml

Inside the container, the INVENTREE_HTTP_PORT and INVENTREE_HTTPS_PORT are used, so they're binding to the port configured in the .env, which means that the previous ${INVENTREE_HTTP_PORT:-80}:80 tried to expose port :80 to the configured one, even though the Port is different.


SchrodingersGat:
@isosloth can you please provide some information on how the existing setup is insufficient in a production environment?

isosloth:
Thanks for asking. The current port mapping breaks as soon as `INVENTREE_HTTP_PORT` is set to anything other than 80.

The compose file currently has:

```yaml
ports:
  - ${INVENTREE_HTTP_PORT:-80}:80
```

This publishes the configured port on the host but always forwards it to port 80 **inside** the container. Caddy does not listen on 80 in that case. It binds to the port from its site address (`INVENTREE_SITE_URL`), which uses the same port as configured in `.env`. So nothing is listening on container port 80, and the proxy is unreachable.

Example with `INVENTREE_HTTP_PORT=8001`, output of `sudo docker compose exec inventree-proxy netstat -tlnp`:

```
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 127.0.0.11:39693        0.0.0.0:*               LISTEN      -
tcp        0      0 127.0.0.1:2019          0.0.0.0:*               LISTEN      1/caddy
tcp        0      0 :::8001                 :::*                    LISTEN      1/caddy
```

and the output of `sudo docker compose ps`:
```
79c1a8a35dfd   caddy:alpine "caddy run --config …"   14 minutes ago   Up 14 minutes (healthy)   443/udp, 2019/tcp, 0.0.0.0:8001->80/tcp, [::]:8001->80/tcp, 0.0.0.0:8002->443/tcp, [::]:8002->443/tcp   inventree-proxy
```

Caddy is listening on 8001 inside the container, but the mapping forwards `8001:80`, so requests never reach it. The listening ports are identical with the original and my modified compose file. Only the published mapping differs.

The fix is to use the same variable on both sides:

```yaml
ports:
  - ${INVENTREE_HTTP_PORT:-80}:${INVENTREE_HTTP_PORT:-80}
```

With `INVENTREE_HTTP_PORT=8001` this becomes `8001:8001`, and the proxy is reachable again. With the default (80) the behavior is unchanged, so existing setups are not affected.

This matters in production because a non-default HTTP port is a common setup, for example when port 80 is already used by another service on the host (which it is in my case).

SchrodingersGat:
But the caddy process (internally) only listens on ports `80` and `443` - `INVENTREE_HTTP_PORT` is not used anywhere else...

PR #12890, opened 2026-09-19
refactor: evaluate enviroment once in ready

matmair:
in my environments these paths are pretty hot and the dict lookup is a bit cheaper on modern arm/x64.


SchrodingersGat (approved):

SchrodingersGat:
@matmair thanks for the fix!

PR #12900, opened 2026-09-21
feat: add plugin support for custom page integration

psxjt5:
This PR adds functionality for plugins to add SPA routes, enabling plugins to define new frontend pages as required.

Code changes made:

- Added an extra `route` UI feature type (backend) (`mixins.py`)
- Added the `get_ui_routes` method to enable plugins to define custom routes (`mixins.py`)
- Added the new 'route' UI feature type as a `PluginUIFeatureType` (frontend) (`PluginUIFeature.tsx`)
- Added `RouteUIFeature` (`PluginUIFeatureTypes.ts`)
- Added the plugin route handler (`PluginRoutes.tsx`)
    - Uses the existing `RemoteComponent` mechanism (used with panels, dashboard widgets etc.) to load the source component specified by the plugin for the selected route.
- Added a new `plugin/*` route which directs /web/plugin/* requests to the `PluginRoutes` handler (`router.tsx`)
- Added an example of this into `user_interface_sample.py` and `sample_route.js`.
- Added a test to `ui/tests.py`
- Updated the documentation with the new `route` UI feature (`ui.md`)

## Defining a new page

New pages are therefore defined like this within a plugin:

```
def get_ui_routes(self, request, context, **kwargs):        
        return [{
           'key': 'my-plugin-page',
            'title': 'My Plugin Page',
            'source': self.plugin_static_file('my_page.js:getFeature'),
            'options': {
                'path': 'page/:pk',
            },
        }]
```

Routes are exposed under the plugin-specific URL namespace. The resulting URL in this case is: `/web/plugin/my-plugin/page/<pk>`.

## Utilised API Calls

The existing `usePluginUIFeature` endpoint has been extended to support the `route` feature type and is now called (with `featureType: 'route'`) when someone navigates to a URL under the `plugin/*` prefix, so existing pages are unaffected.

## Component Rendering

Components are rendered using the existing `RemoteComponent` mechanism.


src/backend/InvenTree/plugin/base/ui/mixins.py
```diff
@@ -22,6 +22,7 @@
     'template_preview',  # Custom template preview
     'navigation',  # Custom navigation items
     'primary_action',  # Custom primary action buttons
+    'route',  # Custom react route
```
matmair:
```suggestion
    'route',  # Custom react router path
```

src/backend/InvenTree/plugin/samples/integration/user_interface_sample.py
```diff
@@ -222,7 +222,7 @@ def get_ui_navigation_items(self, request, context, **kwargs):
                 'key': 'sample-nav-item',
                 'title': 'Sample Nav Item',
                 'icon': 'ti:menu',
-                'options': {'url': '/sample/page/'},
+                'options': {'url': 'plugin/sampleui/test'},
```
matmair, line 225: Why is this changed?

src/backend/InvenTree/plugin/samples/static/plugins/sampleui/sample_route.js
```diff
@@ -0,0 +1,20 @@
+export function getBasicPage(_context) {
+    return React.createElement(
+        'div',
+        { style: { padding: 24 } },
+        React.createElement('h1', null, 'Sample Plugin Route'),
+        React.createElement('p', null, 'Plugin route is working.')
+    );
```
matmair, line 7: is this the intended syntax? Seems a bit verbose

matmair (commented):
Looks good in general; great implementation.

2 things that I noticed bahviour wise on testing:
- the navigation item is not highlighted as active when selected
- no item is added to the side nav tree; not sure if we need to address that in this patch

Image vs 

Image

src/backend/InvenTree/plugin/samples/integration/user_interface_sample.py
psxjt5: This navigation item was pointing at a non-existent page anyway, so I thought it could also be used to demo the route feature. Happy to change it back and create a new button instead?

src/backend/InvenTree/plugin/samples/static/plugins/sampleui/sample_route.js
psxjt5: I've changed the text here to something a bit more relevant to the sampleUI plugin, and in line with what some of the other examples say in ead264e307dca564ff4bcb5e212a3d7b427309a3.

psxjt5:
> Looks good in general; great implementation.
> 
> 2 things that I noticed bahviour wise on testing:
> 
> * the navigation item is not highlighted as active when selected
> * no item is added to the side nav tree; not sure if we need to address that in this patch
> 
> Image vs
> 
> Image

In terms of the navigation link highlighting, the built-in links (Parts, Manufacturing etc.) match the tab name to the beginning of the URL (so the Manufacturing link is highlighted when the current URL begins with `/manufacturing/` etc.) to derive which tab is actively highlighted.

The dynamic tabs (added by plugins) use the title of the navigation item as their tab value instead. As a result, the comparison against the first part of the URL breaks (i.e., for URL `/plugin/sampleui/test`, `plugin` is being compared with the button title, `Sample Nav Item`,). So this probably needs addressing in a separate issue.

On top of this, with plugin routes being exposed under `/web/plugin/<plugin-name>/...` the current mechanism would not allow pages from different plugins to be distinguishable, and so for dynamic items it may be better to match the entire URL anyway.

src/backend/InvenTree/plugin/samples/integration/user_interface_sample.py
matmair: no that makes sense; keep it

matmair:
@psxjt5 would you be able to dress the navigation highlighting detail? I think everything else looks good

src/backend/InvenTree/plugin/samples/static/plugins/sampleui/sample_route.js
matmair: the question is more about the broad usage of React.createElement; more complex UI elements will be pretty hard to read

psxjt5:
> @psxjt5 would you be able to dress the navigation highlighting detail? I think everything else looks good

@matmair I can give it a go, do you want me to do that as part of this PR or should I create a new issue and work on that separately?

matmair:
I would prefer if we could get this done here but only if this does not prolong this PR more than a week

src/backend/InvenTree/plugin/samples/static/plugins/sampleui/sample_route.js
psxjt5: Ah, I see. In this case the `React.createElement` syntax is being used because these examples for the SampleUI plugin are .js files (presumably as they're just small files used as examples). I've also tried using `.tsx` files with regular JSX in my own plugin and those work too.

psxjt5:
> I would prefer if we could get this done here but only if this does not prolong this PR more than a week

@matmair Ok, I've implemented the necessary fixes for the navigation hlghlighting in a12d03b28c34bb149aa325db7668d1e0fd825354.

matmair (approved):
this looks great; thank you!

matmair:
@SchrodingersGat is this changelog worthy?

psxjt5:
> this looks great; thank you!

No problem. Thanks for all your help!

SchrodingersGat:
@psxjt5 great implementation here, seems to work very well. Can you please add a `CHANGELOG.md` entry for this addition, then OK to merge from my end

psxjt5:
> @psxjt5 great implementation here, seems to work very well. Can you please add a `CHANGELOG.md` entry for this addition, then OK to merge from my end

@SchrodingersGat Changelog entry added here: https://github.com/inventree/InvenTree/pull/12900/commits/fa8bf4501bc65dc11413236a9745dcfebcdce9b4

matmair:
@psxjt5 looking into the frontend issues, seemingly not related

PR #11961, opened 2026-05-18
[refactor] Attachment images

SchrodingersGat:
This PR adds / refactors some functionality related to attachments

### References

- Closes https://github.com/inventree/InvenTree/issues/5703

### Changes

- Adds "thumbnail" for attachments which are images
- Display thumbnail in attachment table
- Prevent upload of files above a user-configurable size
- Data migration for existing uploaded attachments

### Rationale

This forms the first part of a re-think of how we might approach generic images against any model instance - see https://github.com/inventree/InvenTree/pull/10484

However, none of that will be implemented here - this is a low-scope change to allow discrimination of attachments which are also images, and display them as such in the UI

### Tasks

- [x] Offload thumbnail generation to the background worker
- [x] Database migration unit tests
- [ ] Demo dataset migration
- [x] Unit tests for new features

### Future Work (Potentially, TBD)

- Replace part / company "image" fields with standard attachments - https://github.com/inventree/InvenTree/issues/9788
- The "primary" image associated with an instance is just an uploaded attachment
- Remove django-std-image entirely - https://github.com/inventree/InvenTree/issues/11647
- Remove django_cleanup entirely - handled internally
- Remove the "select part image from existing" functionality - in favor of generic attached images

PR #12902, opened 2026-09-21
[bug] Fix for Attachment thumbnail generation

SchrodingersGat:
- Fix issue introduced in https://github.com/inventree/InvenTree/pull/11961
- Attachment thumbnails were previously generated in the top-level media dir
- Now, they are created alongside the attachment they are generated for


matmair:
@SchrodingersGat does this need a backport?

Issue #12873, opened 2026-09-17
Harden label printing against concurrent delivery and interrupted jobs


ledvinap:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

Related to the #12872 that preserves label-print failures in `DataOutput` and skips redelivery of failed jobs. The problem is a bit theoretical and ignoring it as 'too unlikely' is reasonable; it is likely that other issues have higher priority.

The #12872  fixes failure reporting and prevents printing again when a later delivery observes a recorded failure. However, label printing still has concurrency and interruption cases that can produce duplicate labels or misleading status.

### Failure scenarios

| Scenario | Consequence | Did it exist with delete-on-error? |
|---|---|---|
| Two workers read the same unfinished output before either finishes | Both pass the guards and can print the same labels. | Yes. Deleting the output did not stop an execution already holding it. |
| A worker saves a stale output after another worker records failure | Progress or completion saves can clear the recorded error and replace it with apparent success. | The stale-write race existed. Previously, a stale save could recreate the deleted output; retaining outputs makes overwriting their error another manifestation. |
| Printing partially succeeds, then the worker times out or terminates | The output can remain unfinished, allowing redelivery to print the batch again. | Yes. Django-Q’s `TimeoutException` inherits from `SystemExit`, bypassing `except Exception`. Abrupt termination also bypasses deletion. |
| A plugin marks the output complete before `after_printing()` fails | The frontend can observe success and stop polling before the failure becomes visible. | Yes. Previously the later failure deleted the output instead of recording an error. |
| An output is deleted while a worker still holds it | A subsequent model save can recreate the deleted row. | Yes, through progress/completion saves, but not through the delete-on-error handler. The related PR extends this risk to failure recording via mark_failure(). |

### How overlapping execution can occur

InvenTree supports multiple worker processes: the default is four, restricted to one per cluster for SQLite or without a global cache. Queue leasing normally prevents duplicate delivery, but its timer starts at dequeue, before waiting in the local worker queue.

For example:

| Time | Event |
|---|---|
| 0 s | The job is leased and placed in a local worker queue. |
| 280 s | Worker A starts printing. |
| 300 s | The queue lease expires. |
| 305 s | Worker B receives the same job and starts executing it while A is still printing. |

Worker A’s 90-second execution timeout has not expired. If neither execution has recorded completion or failure, both can pass the unfinished-output guard.

[...]

Issue #11650, opened 2026-03-31
label printing background task timeouts - labels get printed multiple times


simonkuehling:
### Please verify that this bug has NOT been raised before.

- [x] I checked and didn't find a similar issue

### Describe the bug*

When printing multiple stock item labels (inventree-brother-plugin with brother ethernet label printers) at once I often run into an issue that not all labels get printed, but the printjob is randomly repeated 2-3x.

I found a traceback regarding a timeout from the backgroud worker process that seems to match the time of printing - but I cannot wrap my head around the real cause of the error yet. any idea?

### Steps to Reproduce

1. open a list of multiple stock items
2. select all of them (more than ~5 in my case)
3. click "print labels" icon on top of the list

### Expected behaviour

background job gets properly scheduled, only one label printed each, no labels missing

### Deployment Method

Docker

### Version Information

InvenTree-Version: 1.2.0 dev
Django Version: 5.2.8
Commit Hash: f72efb8
Commit Date: 2025-11-20
Commit Branch: null
Database: django.db.backends.postgresql
Debug-Mode: False
Deployed using Docker: True
Platform: Linux-6.1.0-rpi7-rpi-v8-aarch64-with-glibc2.41
Installer: DOC
[...]

PR #12913, opened 2026-09-22
Background Task Features

SchrodingersGat:
- Ref: https://github.com/inventree/InvenTree/issues/12873
- Ref: https://github.com/inventree/InvenTree/issues/11650

This PR introduces two new features for background tasks:

### Optional Single-Shot Mode

`offload_task` can now be called with `retry=False` to disable retry on failure. This can be used for tasks which should only run once.

Utilizes the [ack_failure](https://django-q2.readthedocs.io/en/master/tasks.html#ack-failure) task attribute from django-q

### Per-Task Timeout

Introduces per-task timeout configuration. Utilizes the [timeout](https://django-q2.readthedocs.io/en/master/tasks.html#timeout) task attribute from django-q

----------------------------

Once this is merged, we could apply this to label and report generation - potentially closing https://github.com/inventree/InvenTree/issues/11650


ledvinap:
@SchrodingersGat : 
Based on gpt-6 review, edited:


1. **[P2] Per-task timeout can exceed the redelivery interval.** 

IMO refusing timeout longer than retry interval is OK for InvenTree; long print job is now primary candidate for extended timeout, but few minutes shall be enough.

It is also possible to subclass ORM broker and override `timeout()` [orm.py](https://github.com/django-q2/django-q2/blob/master/django_q/brokers/orm.py#L27,) but that is probably overkill for problem that nobody complained about.

2. **[P2] Deduplication silently ignores the requested options.** [tasks.py](https://github.com/inventree/InvenTree/blob/master/src/backend/InvenTree/InvenTree/tasks.py#L345-L353)

This is mainly policy choice + `check_duplicates` doc shall mention it.

3. **[P2] A terminal `retry=False` failure misses the staff notification.** 

Moving `after_failed_task` to `post_execute` signal and fetching `Task`  using `task` dict seems reasonable fix:

```python
@receiver(post_execute, sender='django_q', dispatch_uid='failure_post_execute_notification')
def after_failed_task(sender, task, **kwargs):
    """Notify staff when a saved task failure will not be retried."""
    from InvenTree.exceptions import log_error

    if task['success']:          # this execution attempt succeeded
        return

    instance = Task.objects.filter(pk=task['id'], success=False).first()
    if instance is None:        # Task succeeded (in other execution)
        return

    n = instance.attempt_count
    limit_reached = Conf.MAX_ATTEMPTS > 0 and n >= Conf.MAX_ATTEMPTS

    if not (task.get('ack_failure', False) or limit_reached):
        return

    # Create a new Error object associated with this failed task
    # This will, in turn, trigger a notification to staff users via the Error post_save signal
    message = f"Task '{instance.func} ({instance.pk})' failed after {n} attempts"
    logger.error(message)

    log_error(
        'task_failure',
        scope='worker',
        error_name='Task Failure',
        error_info=message,
        error_data=str(instance.result) if instance.result else '',
    )

File: src/backend/InvenTree/InvenTree/tasks.py [hunk 1]
```diff
@@ -302,8 +326,21 @@ def offload_task(
...
         check_duplicates: If True, check for existing identical tasks before offloading
+        retry: If False, the task is attempted exactly once and is never retried if it
+            fails (see note below)
+        timeout: Optional per-task override (in seconds) of the worker's task timeout
         **kwargs: Keyword arguments to be passed to the task function
 
+    Note:
+        django-q2 has no concept of a per-task retry limit: the ORM broker (which
+        InvenTree always uses) simply leaves a failed task's queue entry in place, so it
+        gets redelivered indefinitely (governed by the cluster-wide 'retry' timeout)
```
ledvinap, line 337: (indefinitely -> global cluster limit still applies)

src/backend/InvenTree/InvenTree/tasks.py
```diff
@@ -354,7 +391,16 @@ def offload_task(
...
+            task_kwargs = dict(kwargs)
+            if not retry:
+                # Bandaid for django-q2 having no per-task retry limit: 'ack_failure'
+                # is its one native per-task option that acknowledges (and so drops)
+                # a task as soon as it fails, rather than leaving it to be redelivered
+                task_kwargs['ack_failure'] = True
+            if timeout is not None:
+                task_kwargs['timeout'] = timeout
+
+            task = AsyncTask(taskname, *args, group=group, **task_kwargs)
```
ledvinap, line 403: (q_options  will reduce task argument collision a bit, but offload_task signature / kwargs handling will consume args sooner)
```suggestion
            task_kwargs = dict(kwargs)
            q_options = dict(task_kwargs.pop('q_options', {}))
            q_options['group'] = group

            if not retry:
                # Bandaid for django-q2 having no per-task retry limit: 'ack_failure'
                # is its one native per-task option that acknowledges (and so drops)
                # a task as soon as it fails, rather than leaving it to be redelivered
                q_options['ack_failure'] = True
            if timeout is not None:
                q_options['timeout'] = timeout

            task = AsyncTask(taskname, *args, q_options=q_options, **task_kwargs)
            
```

src/backend/InvenTree/InvenTree/tasks.py
```diff
@@ -490,6 +544,14 @@ def bulk_offload_task(
             'started': timezone.now(),
         }
 
+        if not retry:
+            # See offload_task() - 'ack_failure' drops the task the moment it fails,
+            # instead of leaving its queue entry to be redelivered
+            task['ack_failure'] = True
+
+        if timeout is not None:
+            task['timeout'] = timeout
```
ledvinap, line 553: This is probably good spot to fail when timeout is longer than configured retry interval.

ledvinap (approved):

src/backend/InvenTree/InvenTree/tasks.py
```diff
@@ -328,22 +390,27 @@ def offload_task(
         check_duplicates: If True, check for existing identical tasks before offloading
         retry: If False, the task is attempted exactly once and is never retried if it
             fails (see note below)
-        timeout: Optional per-task override (in seconds) of the worker's task timeout
+        timeout: Optional per-task override (in seconds) of the worker's task timeout.
+            Clamped down (with a warning) if it would leave less than 120s of headroom
```
ledvinap: (`HEADROOM = 30`)

ledvinap:
@SchrodingersGat : I updated https://github.com/inventree/InvenTree/pull/12913#issuecomment-5781643478; most cases are handed; look at `instance = Task.objects.filter(pk=task['id'], success=False).first()`, it may be slightly more consistent

PR #12914, opened 2026-09-22
Allocate mixin

### LINKED ISSUE #11678: [FR] auto allocate stock with priority

### Problem statement

We currently issue all build orders automatically from our machines or from bar codes, and stock is automatically allocated. I would like to add a priority to a stock location for the automatic allocation. For example, we always use reel components before tray parts even if both are on the machine as reels feed faster than trays for us.

### Suggested solution

I currently work around it, by changing the code for exclude to actually make it a lower priority rather than excluded. It's a bit hacky though. 

just using exclude location and manually fixing things if that one runs out.

### Examples of other systems

_No response_

- [ ] I want to develop this.

SchrodingersGat:
Closes https://github.com/inventree/InvenTree/issues/11678

Adds a new plugin mixin class which supports customization of auto-allocation against orders. Plugins can choose to adjust the auto-allocated / auto-selected stock, which allows low level control of how auto-allocation functions, without imposing limitations or complex configuration of default InvenTree behaviour


matmair (approved):
Looks good to me;
Maybe it makes sense in a separate PR to create a allocation state machine that tracks how these decisions are made and move the whole thing out of the model?

